PR #412, opened 2014-12-20
if skipping phase init in adsynt2, then skip amplitude init also

mskala:
The adsynt2 opcode allows skipping _phase_ initialization, but that is not very useful unless we also skip _amplitude interpolation_ initialization.  Amplitudes are interpolated at k-rate.  In a tied note, if adsynt2 is allowed to initialize the amplitudes to zero when it is already playing, it puts an annoying click into the audio, most noticeable when ksmps is large.  The proposed change makes "initialization skip" apply to amplitude interpolation as well as phase - which was probably the intended behaviour all along.

By the way, although it's not really necessary to credit me on minor patches like this, if you want to do so, my name is correctly spelled with a "k".


kunstmusik:
Closing this pull request as John has noted the code has been applied.

PR #453, opened 2015-04-05
added static building for opcodes to resolve csound tilde issue on windo...

rorywalsh:
...ws

This doesn't seem to be necessary on any other platforms and leads to larger plugin library sizes, but so far it's the only thing I could do to get csound~ to load. 

csound~ needs be modified so that it enforces -+rtaudio=null


vlazzarini:
The thing to do is to protect it so that it only affects Windows. Whoever accrpts this merge request should do that.

vlazzarini:
Note that the linux build is failing too.

rorywalsh:
I have had to comment out the building of padsynth to get it to build for me on Windows.

kunstmusik:
I don't think we can use this patch; a change like that should be globally controlled I think. This could maybe be added to make_plugin in the main CMakeList.txt and isolated to windows.

rorywalsh:
That sounds reasonable. Or rather than isolate this particular build setup to windows, perhaps we could simply enable a CMAKE build option for static linking to pthread and liibgcc?

PR #462, opened 2015-05-03
Unset DOT_PATH in doxyfiles

fsateler:
The paths set are probably where @gogins has it installed, but it is very unlikely anyone else has such a path. Clear them so that we simply rely on the PATH environment variable.

I am submitting this as a pull request because I do not want to surreptitiously disrupt anyone's workflow, but I think this should be merged. Merging this would allow debian to drop a patch.


fsateler:
Thanks!

PR #467, opened 2015-05-14
Add uniform default ladspa path

fsateler:
Replace the different checks in different files by a single fallback

Submitting as a PR since I don't know if it was intentional that csladspa had a mac specific path but dssi didn't


vlazzarini:
The only issue with this, I think is that it will change how things currently work on Linux for csladspa. My reading of the code is that if there is no LADSPA_PATH set, then it will look for CSDs in the current directory. If this is thought to be a minor issue, I am happy to accept the patch.

fsateler:
Ah interesting point. And that makes sense, given that it would be unusual to ship CSDs in /usr/lib/ladspa. But I guess a similar concern couldbe raised in mac?

vlazzarini:
it could, however, it has been always the case since the first release that plugin CSDs go in the default place (the only choice really in that platform).

kunstmusik:
@fsateler Is this pull request still in the works or can we close this?

fsateler:
On 31 May 2015 at 22:32, Steven Yi <notifications@github.com> wrote:

> @fsateler <https://github.com/fsateler> Is this pull request still in the
> works or can we close this?
>

The thing remaining is more of a policy one. I certainly cannot advocate to
change csound's behavior just to make it easier for me as a downstream
packager.

Defaulting csladspa to read from /usr/lib/ladspa instead of . is weird but
not totally wrong IME: after all, most LADSPA hosts are GUIs and therefore
either need LADSPA_PATH set or will have . == ~ , which is not quite usable
either. This makes me think it is not likely people expect csladspa to read
from the current directory. But then I may just be lacking imagination.


-- 

Saludos,
Felipe Sateler

fsateler:
I'm not pushing this further. I'd appreciate if a change like this is made, but its not the end of the world either, I can patch csound downstream.

PR #509, opened 2015-07-30
Avoid unsafe and potentially overlapping string copies.

megastep:
There were some problems in the way the option strings were handled for MIDI options, which could lead to calls with `strcpy()` with identical strings, which is undefined and led to some early crashes in certain conditions.


vlazzarini:
thanks

PR #538, opened 2015-10-21
added static linking to many plugin opcodes that were causing crashes…

rorywalsh:
… on Windows when loaded in VST plugins


kunstmusik:
As the CI build shows, I don't think this change works for non-Windows builds.  Also, there has got to be a simpler way to do this.  Could you try adding these flags to make_plugin in the main CMakeLists.txt, and conditionally if(WINDOWS)?

rorywalsh:
The best would be an option for static linking for plugin opcodes, but I
don't know enough about cmake to implement such a thing. I can take a look.

jpffitch:
Is this resolved with the new 6.07 build?

rorywalsh:
It seems to be. At least I've yet to hit any issues with it.

jpffitch:
believed fixed other way

Oeyvind:
I tested here too and it looks good afaics!

2016-05-21 18:27 GMT+02:00 John ffitch <notifications@github.com>:

> believed fixed other way
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/csound/csound/pull/538#issuecomment-220786988>
>



-- 

Oeyvind Brandtsegg
Professor of Music Technology
NTNU
7491 Trondheim
Norway
Cell: +47 92 203 205

http://www.partikkelaudio.com/
http://soundcloud.com/brandtsegg
http://flyndresang.no/
http://soundcloud.com/t-emp

PR #584, opened 2016-01-16
Set install path to Homebrew's default and set the rpath

ryutaroikeda:
This fixes the "could not fix CsoundLib64" and the "library not loaded" issues when installing csound via homebrew on OSX (see https://github.com/csound/homebrew-csound/issues/12). CsoundLib64.framework is now installed in /usr/local/Frameworks, and this path is added to the rpath.


kunstmusik:
@ryutaroikeda Thanks for submitting this!  I do have concerns though about having Homebrew information within the Csound build and how rpath might interfere with existing installer build scripts.  I think there may be other issues as I think the root cause is that the expansion of ~ is no longer to the user's home directory (and this might affect the java extensions install).  I'll research a bit further.

kunstmusik:
Closing as this has been addressed in https://github.com/csound/homebrew-csound/issues/14

PR #635, opened 2016-04-30
Paulstretch Opcode

PaulBatchelor:
Here is an opcode implementation of the paulstretch algorithm:
http://hypermammut.sourceforge.net/paulstretch/
It is largely based off of the Numpy/Scipy python implementation, found here:
https://github.com/paulnasca/paulstretch_python

For arguments, it takes in a stretch factor, window size (in seconds), and an ftable containing an audio buffer.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -755,6 +755,7 @@ set(stdopcod_SRCS
     Opcodes/nlfilt.c
     Opcodes/oscbnk.c
     Opcodes/pluck.c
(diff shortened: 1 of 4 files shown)


vlazzarini:
merged, thanks. We need a manual page now

PaulBatchelor:
I can do this. It seems like all the manual pages are in another repo. Do I just make a PR there, or am I missing something?

vlazzarini:
yes pull request will work

PR #731, opened 2016-12-16
Fixed websocket opcode

eddyc:
Fixed bug where opcode wouldn't send data to browser


vlazzarini:
could you resolve the conflicts before we merge this?

eddyc:
Don't now why there were so many, but that's fixed now

Issue #720, opened 2016-11-30
readme.txt file in minimal Windows archives should be encoded using UTF-8


nwhetsell:
This file is in Csound6.08_Win32_minimal.zip and Csound6.08_Win64_minimal.zip on the [releases page](https://github.com/csound/csound/releases), and it appears to be encoded using [Windows‑1252](https://en.wikipedia.org/wiki/Windows-1252). Many text editors will assume an encoding of UTF‑8 and display this file as

```
… you will need to add the ëbiní directory to your system path, and create an OPCODE6DIR64 system variable that points to this ëlibí folder. 
```

instead of the expected

```
… you will need to add the ‘bin’ directory to your system path, and create an OPCODE6DIR64 system variable that points to this ‘lib’ folder. 
```

As an alternative, if the left and right quotes are replaced with straight quotes (or omitted entirely), the file can be encoded using ASCII.

(As a side-note, it may be worth adding instructions to the readme.txt file describing how to “add the ‘bin’ directory to your system path, and create an OPCODE6DIR64 system variable that points to this ‘lib’ folder.”)

rorywalsh:
That's a fair point about the encoding. Feel free to prepare a version of
that file with the appropriate instructions. I'll happily add it to future
zip files.

nwhetsell:
Is the readme.txt file included in the Windows archives in the Csound repository?

rorywalsh:
No, but if you write one we can add it somewhere.

nwhetsell:
[Here’s a new README.txt](https://github.com/csound/csound/files/629223/README.txt), containing:

```
This folder contains a minimal build of Csound for 64-bit Windows. If you're
trying to install Csound along with its documentation,
[CsoundQt](https://csoundqt.github.io), and other extras, download and run one
of the setup executables at https://github.com/csound/csound/releases. The names
of the setup executables begin with Setup_Csound6_x64_.

It's a good idea, but not essential, to move the bin, include, and lib folders
to roughly where Csound's setup executables install them. To do this, open an
administrator Command Prompt in this folder and enter

    set CsoundPath=%ProgramFiles%\Csound6_x64
    if not exist "%CsoundPath%\NUL" mkdir "%CsoundPath%"
    move bin "%CsoundPath%"
    move include "%CsoundPath%"
    move lib "%CsoundPath%"

To easily run the executables in the bin folder, you must add the bin folder to
your Windows path and create an OPCODE6DIR64 environment variable. To do this on
Windows 10:

[...]

PR #738, opened 2016-12-29
Remove duplicate Windows csound64.lib creation script

nwhetsell:
This pull request removes mingw64/make_import_library_32.cmd (mentioned [here](https://github.com/csound/csound/issues/720#issuecomment-265945302)), and it lets mingw64/make_import_library.cmd also create a 32-bit csound64.lib by running

```batch
make_import_library /machine:x86
```

This pull request also fixes various issues in mingw64/README.md so that, for example, numbered lists increment properly when displayed on GitHub.

I should note that there’s still another csound64.lib creation script at https://github.com/csound/csound/blob/develop/make_import_library.cmd. I’m not sure why it’s there, or whether it can be removed.


nwhetsell:
By the way, passing `/machine:x86` to this script makes use of [the fact that](https://msdn.microsoft.com/en-us/library/h34w59b3.aspx):

> LIB processes options in the order specified on the command line and in command files. If an option is repeated with different arguments, the last one to be processed takes precedence.

kunstmusik:
@rorywalsh @gogins  This looks like something for you all to review?

gogins:
Looks good to me.

Thanks,
Mike


Michael Gogins
Irreducible Productions

rorywalsh:
Me too.

PR #777, opened 2017-04-08
Pvs bin setter methods should be void on return

eddyc:
Control reaches end of non-void function for pvs bin setters


kunstmusik:
@vlazzarini Is this okay to merge?

vlazzarini:
instead, I've made another change to fix it.

PR #787, opened 2017-05-02
webaudio playback improvement

hfmanson:
The webaudio buffer is decoupled for ksmps. Every call to _performKsmps yields ksmps samples. This function is called repeatedly until enough samples are gathered to entirely fill the webaudio buffer. The last csound samples are preserved and used to fill the next first samples of the next webaudio buffer


kunstmusik:
@hfmanson Thanks for looking into this. Could I ask you to rebase this to the develop branch?  We merged no-pthread into develop a while back and no-pthread should have been deleted from GitHub then; there are also a number of changes that might not be necessary in the PR with the changes in develop.

PR #788, opened 2017-05-02
webaudio improvement on develop branch

hfmanson:
The webaudio buffer is decoupled for ksmps. Every call to _performKsmps yields ksmps samples. This function is called repeatedly until enough samples are gathered to entirely fill the webaudio buffer. The last csound samples are preserved and used to fill the next first samples of the next webaudio buffer

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,7 +478,7 @@ if(NOT MSVC)
   find_library(PTHREAD_LIBRARY pthread)
 
   if(NOT PTHREAD_LIBRARY)
-      message(FATAL_ERROR "Csound requires the pthread library")
+      message(STATUS "Csound requires the pthread library")
   endif()
 
   set(CMAKE_REQUIRED_INCLUDES pthread.h)
--- a/H/csGblMtx.h
+++ b/H/csGblMtx.h
@@ -53,7 +53,7 @@ void csoundUnLock() {
 extern "C" {
   #endif
 
-static INIT_ONCE g_InitOnce = INIT_ONCE_STATIC_INIT;
+static INIT_ONCE g_InitOnce = INIT_ONCE_STATIC_INIT; 
 static CRITICAL_SECTION* csound_global_lock;
 
 static BOOL CALLBACK InitHandleFunction ( PINIT_ONCE InitOnce, PVOID Parameter,
@@ -62,13 +62,13 @@ static BOOL CALLBACK InitHandleFunction ( PINIT_ONCE InitOnce, PVOID Parameter,
     CRITICAL_SECTION* cs = (CRITICAL_SECTION*) malloc(sizeof(CRITICAL_SECTION));
     InitializeCriticalSection(cs);
     *lpContext = cs;
-    return 1;
+	return 1;
 }
 
 
 
 void csoundLock() {
-    BOOL status;
+    BOOL status; 
     CRITICAL_SECTION* cs;
 
     status = InitOnceExecuteOnce(&g_InitOnce, InitHandleFunction, NULL, &cs);
@@ -79,7 +79,7 @@ void csoundLock() {
 
 void csoundUnLock() {
 
-    BOOL status;
+    BOOL status; 
     CRITICAL_SECTION* cs;
 
     status = InitOnceExecuteOnce(&g_InitOnce, InitHandleFunction, NULL, &cs);
@@ -93,8 +93,23 @@ void csoundUnLock() {
 }
 #endif
 
+#else /* END WIN32 */
+#ifdef __cplusplus
+extern "C" {
+#endif
 
-#endif /* END WIN32 */
+void csoundLock() {
+}
+
+void csoundUnLock() {
+}
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif
 
 
 #endif      /* CSOUND_CSGBLMTX_H */
+
--- a/OOps/midiops.c
+++ b/OOps/midiops.c
@@ -25,6 +25,7 @@
 #include "csoundCore.h"                                 /*      MIDIOPS.C   */
 #include "midiops.h"
 #include <math.h>
+#include <time.h>
 #include "namedins.h"           /* IV - Oct 31 2002 */
 
 #define dv127   (FL(1.0)/FL(127.0))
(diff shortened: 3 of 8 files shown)


vlazzarini:
let's go for it. This was in my todo list for a couple of years. Thanks a lot @hfmanson !

kunstmusik:
Thanks for the reviews @eddyc and @vlazzarini , and thanks for the contribution @hfmanson!

kunstmusik:
@hfmanson I'm reviewing this again for release and just realized that you had pulled out CsoundObj.js and FileList.js from linking into the generated libcsound.js and included them separately. Is there a critical reason for this?  I'd rather keep it compiled into libcsound.js so that it doesn't change the files to source in and require end users to modify their applications too much, but if there's a technical reason why this is necessary then that's fine.

hfmanson:
@kunstmusik The reason to leave CsoundObj.js and FileList.js out of libcsounds.js is to enable emscripten 
 to create webassembly instead of asm.js. Thanks to this [stackoverflow question](http://stackoverflow.com/questions/44047603/webkit-wasm-stacktrace-debugging) I got this working yesterday and have a  [demo](http://mansoft.nl/csound/). I'll create a new pull request shortly.

vlazzarini:
To me it looks like the correct thing to do, but we could two builds (one targeting webassembly, separating the js files, and the other targeting asm.js). Do think it's possible @hfmanson ? Also, ideally we should wait for the webassembly pull request before we close the release. It would be a great thing to have it as part of 6.09. Actually a major development. Thanks again @hfmanson , we owe you a few beers.

kunstmusik:
@hfmanson is it the case then that same two JS files get loaded whether you first load in asm.js or wasm first? That'd be useful or runtime determination of which Csound backend or not. If it works that way, we'll just have to communicate that change to users in Release Notes.

(And thanks for researching WASM implementation!)

hfmanson:
@kunstmusik on [this commit](https://github.com/csound/csound/commit/7f9100fac72cbee65f732da1ffee3ac5b3da90e9) I added CsoundObj.js and FileList.js to require.js. Later I improved [JS loading](https://github.com/csound/csound/commit/2a77b80ed84563c500e7122585873177d94ea2a6)
 by first requiring nothing but running the 'main' function. In 'main' first the global Module object is prepared for csound initialization and then "jquery", "bootstrap", "libcsound" and "CsoundObj" are required. When csound is initialized the Module['_main'] is executed which after requiring the remaining .js modules initialize the ace widgets.

kunstmusik:
I saw all of this but I'm also not using require.js in my own project so this has changed the API to force loading of additional js files beyond libcsoun d.js. It's not a big deal though, just a breaking change, but it seems warranted if it enables wasm. It complicates dynamic script loading a little bit but it's manageable. I'll plan to go ahead and release as-is and put a note about it in the release notes.

hfmanson:
I think it's better to make that breaking change now, so later on people
can use the asm.js and webassembly version with the same API.

kunstmusik:
Yup, sounds good to me, just needed to understand what that change was for.  Thanks again!

gogins:
Yes, let's preserve the API.

Regards,
Mike


Michael Gogins
Irreducible Productions

vlazzarini:
Ok, should we consider a versioning way of flagging the API change?

kunstmusik:
Well, I should clarify as I perhaps misspoke: the CsoundObj API in Javascript hasn't changed, just that one now has to include the CsoundObj.js and FileList.js files separately. (Previously, they were compiled into libcsound.js, so one got the functionality available in those files automatically.) The change to separate out the files makes sense in light of WASM, as one would use the same JS files but one could load either the asm.js or wasm backends.  

The alternative is to return the asm.js compile to embedding those two files, but leave them separate for WASM. I'm not sure that gains us much rather than just making the break now, as users will have to change to the new usage for WASM anyways.  

I think a note in the Change log should be enough with a note to users on the mailing list.

hfmanson:
Probably it's possible to use the old way of prepending the CsoundObj.js and FileList.js files to the final csound.js when using wasm. The only difference between asm.js and wasm is that a .wasm file is generated and libcsound.js will be a 'launcher' for the wasm code. However for development purposes it is handier to have some javascript merger do the job of merging several javascript files into one. When i was fixing  webaudio problems it would be time consuming to relink libcsound.js all the time for just a few changes in CsoundObj.js. I would like to hear your comments on this.

vlazzarini:
On a separate note, is there a build script or makefile for the wasm build?

hfmanson:
I created an additional build script build-wasm.sh

vlazzarini:
Thanks @hfmanson !

vlazzarini:
Looking at the sources now, I can't see the build-wasm.sh, could you contribute it to the repo?

vlazzarini:
@hfmanson I picked up your build script and I am attempting to build it, but I had a couple of issues: 

1. there was a multiply defined conflict for strlcat. I've added -DHAVE_STRLCAT=1 to the cmake options that fixed it.

2. Now the final step to create the wasm file fails. After getting a flurry of stuff printed  to the terminal, I get this error

```
Assertion failed: (WasmValidator().validate(wasm)), function processAsm, file /Users/clb/emslave/buildslave/osx-tagbuilder/emsdk/binaryen/tag-1.37.9/src/asm2wasm.h, line 1476.
Traceback (most recent call last):
  File "/Users/victor/src/emsdk_portable/emscripten/1.37.9/emcc", line 13, in <module>
    emcc.run()
  File "/Users/victor/src/emsdk_portable/emscripten/1.37.9/emcc.py", line 2170, in run
    subprocess.check_call(cmd)
  File "/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 541, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '[u'/Users/victor/src/emsdk_portable/clang/e1.37.9_64bit/binaryen/bin/asm2wasm', 'libcsound.temp.asm.js', '--total-memory=16777216', '--emit-jsified-potential-traps', '-O3', '--mem-init=libcsound.js.mem', '--mem-base=1024', '--table-max=-1', '--wasm-only', '-g', '-o', 'libcsound.wast', '-S']' returned non-zero exit status -6
```

do you know how to fix it? Am I missing any tools? I have the latest emscripten installed.

hfmanson:
I just create a new pull request containing WASM support

vlazzarini:
builds perfectly here now. thanks again.

gogins:
Thanks, I'll check this out.

Regards,
Mike

PR #789, opened 2017-05-04
main.js correction and webaudio input audio

hfmanson:
- In main.js first csound must be 'required' and setup, then Module['_main'] is called which should then require the ace widgets. Now csound also loads correctly when -O0 is used in emscripten
- Finished audio input implementation in CsoundObj.js


kunstmusik:
Looks good to me, going to merge.

PR #790, opened 2017-05-07
Bug in hdf5read

eddyc:
Editing the input argument names directly resulted in a bug where a second instance of the hd5read opcode would have inaccurate input argument names.


jpffitch:
This undoes recent canges to Victo;s code so he should comment
==John

jpffitch:
This undoes a recent change from victor so he should comment

eddyc:
If it does it's a mistake on my part, but I think the commit where I changed it is overwritten with Victor's other solution, if you look at the files changed for this pull request there is only HDF5IO.h and HDF5IO.c.

jpffitch:
I would merge if I knew how as long as plugins.h is not changed

vlazzarini:
I think you should close this pull request and try another one with no change to plugins.h

jpffitch:
Merged code in HDF

PR #795, opened 2017-05-19
Added warnings for index out of range in setrow and setcol

eddyc:
Added warnings for index out of range in setrow and setcol


kunstmusik:
Should this be returning OK if there's the index out of bounds error?

vlazzarini:
No, it should return the result of PerfError(). I'll fix this once I merged.

PR #797, opened 2017-05-23
Numerous CMake changes for MSVC build

stekyne:
Hi guys,

I've been working off this branch for a while and it's accumulated a lot of changes at this stage. I'd like to merge it back to develop now that it's building correctly on travis and appveyor.  

I originally was going to do some big changes but I had to avoid that in the end and now it's just a collection of small changes. It does now build the swig interfaces, python, portmidi and liblo. I've put all the build settings into the custom-msvc.cmake file as well. There are some miscellaneous changes here and there in general. 

If you guys think it looks ok, I'll merge it in.


msvc/README.md
```diff
@@ -69,81 +71,34 @@ You can pass in default arguments to the csound instance by right-clicking on "c
...
-|USE_OPEN_MP                        |                   |          |
-|USE_PORTAUDIO                      |                   | Working |
-|USE_PORTMIDI                       | Manual build      | Enabled/Needs work |
-|USE_PULSEAUDIO		                  |                   | Disabled |
-|USE_SYSTEM_PORTSMF                 |                   |          |
-
 ## Work in progress / work to do
-1. Python integration into build. Most likely user will need to install it via msi
-2. PortMIDI/SMF manual build script needs to be created
+1. [Fixed] Python integration into build. Most likely user will need to install it via msi
```
fsateler: If it's fixed it should be removed from the to do list.

InOut/CMakeLists.txt
```diff
@@ -169,24 +170,22 @@ else()
...
-    endif()
-    if(WIN32)
-      if(NEED_PORTTIME)
-        include_directories(${PORTTIME_INCLUDE_PATH})
-      endif()
-        list(APPEND pmidi_LIBS ${CSOUND_WINDOWS_LIBRARIES})
-    endif()
-    if(LINUX AND ALSA_LIBRARY)
-        list(APPEND pmidi_LIBS ${ALSA_LIBRARY})
+			include_directories(${PORTTIME_INCLUDE_PATH})
```
fsateler: Indentation is off (using tabs instead of spaces) in some (but not all) cmake changes.

CMakeLists.txt
```diff
@@ -436,7 +435,7 @@ find_package(OpenMP)
 find_library(LIBSNDFILE_LIBRARY sndfile)
 
 if(NOT LIBSNDFILE_LIBRARY AND WIN32)
-find_library(LIBSNDFILE_LIBRARY sndfile-1)
+find_library(LIBSNDFILE_LIBRARY libsndfile-1)
```
fsateler: This could be `find_library(LIBSNDFILE_LIBRARY sndfile libsndfile-1)` above instead of this if?

stekyne: Agreed, I have it like this elsewhere. I'll change it now.

InOut/CMakeLists.txt
stekyne: Not sure whats an easy fix here outside of being disciplined and always using spaces. Maybe we need a pre-commit git hook to remove tabs. I'll investigate a solution

fsateler: Editorconfig is an option, there are plugins for it for most editors.

msvc/README.md
stekyne: Yeah agreed. I was mainly keeping it there for historic reasons but probably not point. I'll remove it

InOut/CMakeLists.txt
stekyne: It seems there is support in VS for it. I'll have a look

msvc/Custom-vs.cmake
```diff
@@ -1,28 +1,93 @@
...
+set(BUILD_WEBSOCKET_OPCODE ON)
+set(BUILD_WIIMOTE_OPCODES OFF)
+set(BUILD_WINDSOUND ON)
+set(NEED_PORTTIME OFF)
+
+# Csound use features
+set(USE_ALSA 0) # N/A
+set(USE_ATOMIC_BUILTIN 0) # Needs code changes for MSVC
+set(USE_AUDIOUNIT 0) # N/A
+set(USE_CURL 0) # Not working, necessary even?
```
kunstmusik: I saw this and other uses of USE_CURL that were introduced into code. I think if we're going to have an option to explicitly turn of use of CURL, we can do it by check USE_CURL in CMake and optionally not set -DHAVE_CURL rather than adding USE_CURL everywhere in source.  I'd rather see this removed or updated before any merge.

InOut/CMakeLists.txt
```diff
@@ -169,24 +170,22 @@ else()
...
+    elseif(WIN32)
+		if(NEED_PORTTIME)
+			include_directories(${PORTTIME_INCLUDE_PATH})
+		endif()
+		target_link_libraries(pmidi ${CSOUND_WINDOWS_LIBRARIES})
+    elseif(LINUX AND ALSA_LIBRARY)
+		target_link_libraries(pmidi ${ALSA_LIBRARY})
     endif()
-    make_plugin(pmidi pmidi.c "${pmidi_LIBS}")
+    target_link_libraries(pmidi ${PORTMIDI_LIBRARY})
```
kunstmusik: I think we may have talked about this on Slack, but we should review our use of target_link_libraries and make sure it's consistent: if we're going to do it in some places, might as well do it everywhere.

Opcodes/py/pythonopcodes.h
```diff
@@ -21,7 +21,13 @@
...
 #define _pycsound_pythonopcodes_h_
 
-#include <Python.h>
+#ifdef _DEBUG
+# undef _DEBUG
+#  include <Python.h>
+# define _DEBUG
+#else
+# include <Python.h>
+#endif
```
kunstmusik: Thanks for adding this, it should address linking issues I've had for python interfaces/opcodes.

msvc/Custom-vs.cmake
stekyne: ok yeah that makes sense. I'll make that change.

InOut/CMakeLists.txt
stekyne: I'll make the changes to use target link libraries everywhere so. I'd like to phase out any custom functions if possible, we shouldn't really need them.

Opcodes/py/pythonopcodes.h
stekyne: No problem. It's required because in debug builds it looks for python*_d.lib which won't exist.

kunstmusik: Exactly the issue I was having but had not spent the time to fix here.  Thanks again!

InOut/CMakeLists.txt
kunstmusik: Sounds great, very much appreciated!

msvc/Custom-vs.cmake
stekyne: It seems I added this without looking at the cause. HAVE_CURL should be disabled if USE_CURL is 0 but the default value in the option() function was overriding it. I guess it's mostly used on linux so disabling it wasn't noticed. I've moved the user options to be above the custom cmake inclusion and it works properly now.

kunstmusik:
Thanks for the CURL changes, the target_link_libraries issue should be done eventually but the PR is fine to merge at this point IMO.

stekyne:
Great, I'll do that now so that seeing as the release has gone out.

PR #813, opened 2017-06-20
Editorconfig use for consistent text style across platforms

stekyne:
http://editorconfig.org/

Felipe mentioned this in a previous pull request so I had a look. Maybe we should try and use this seeing as VS can really interfere with the style (auto-formatting) of the code. It's supported in a lot of editors natively and lots of plugin support for other editors too.

I just put in some default settings so I guess we need to agree on the style and what the format should be. Things like line ending, number of spaces, tabs/spaces etc.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -2,7 +2,11 @@ root = true
 
 [*]
 insert_final_newline = true
-
-#[*.{py,h,hh,hpp,c,cc,cpp}]
+trim_trailing_whitespace = true
+end_of_line = lf
 indent_style = space
 indent_size = 4
+
+#[*.{py,h,hh,hpp,c,cc,cpp}]
+#indent_style = space
+#indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+root = true
+
+[*]
+insert_final_newline = true
+
+#[*.{py,h,hh,hpp,c,cc,cpp}]
+indent_style = space
+indent_size = 4


jpffitch:
No tabs
No trailing spaces
Indent by 2 except top level of function
Lines less than 85 where possible
No space before semicolon
No space between fn name and args
Space in syntax not function

Probably more but those were the format from Csound 3

I expect no agreement

⁣Sent from TypeApp ​

vlazzarini:
I have no problem with the proposed style. I am not quite sure how to use editorconfig but I guess I will learn.

stekyne:
I don't think there is any input from the developers once it's in place. It will basically advise the editor how the formatting should be. If you use an editor without the plugin or no support, it takes no effect and you're free to mess up the formatting as you wish!

Realistically it's probably just for VS users so we don't inject tabs everywhere or screw up line endings etc.

gogins:
There is a sed script (indent.sh) for reformatting Csound code, I suggest
its settings be used for the plugin.

Regards,
Mike



Michael Gogins
Irreducible Productions

kunstmusik:
One thing I dislike about the indent.sh is that it gets run once in a while
and then makes merges between branches very difficult as well as viewing
git history for files. I'd love to figure out something that will apply
style as part of the git commit process.

stekyne:
Yeah steven that is a problem with scripts like that. It's happened with VS already, you accidentally hit auto format and the whole file is marked as a change in git. That's partially the reason I wanted to fix this now, to reduce the git diff noise in commits.

We could create a git pre commit hook to format the code as it's being committed. It could strip out tabs and replace with spaces, add end of file lines, remove white space etc. I'm not entirely familiar with how to do it but I'll investigate now. 

With  editorconfig, we can at least avoid some of this problem by aligning our editors. The git hook could then catch any formatting errors that slip by.

jpffitch:
Never eard of editrcofig.  Does it work i the oe true editor?  Witiut 
emacs I cannot work.

gogins:
Something like this in csound/.git/hooks/pre-commit:

https://gist.github.com/kblomqvist/bb59e781ce3e0006b644

One thing for certain -- the SAME tool should be used by ALL developers on
ALL platforms. I think astyle would do. Any editor worth its salt can be
jiggered to call astyle from a hotkey or a hook.

Best,
Mike


Michael Gogins
Irreducible Productions

stekyne:
There is an emacs plugin: https://github.com/editorconfig/editorconfig-emacs#readme

stekyne:
Just going to merge this in and see how it works out. We can modify as needed.

PR #819, opened 2017-06-26
Added Csound iOS Swift Examples

nikhilsinghmus:
Added Csound iOS Swift Examples
Modified CsoundMotion class to make CMMotionManager instance public
Renamed Csound iOS Examples to Csound iOS Obj-C Examples
Updated manual and scripts


kunstmusik:
@nikhilsinghmus: I tried to build the fork here and I get an error about a Csound-iOS-SwiftExamples-Bridging-header.h does not exist. Is this file missing from the PR or is this something that is supposed to be generated?

PR #829, opened 2017-07-23
Update README.md

tiagmoraismorgado:


jpffitch:
Why change my name?

⁣Sent from TypeApp ​

jpffitch:
I have applied the only real change

⁣Sent from TypeApp ​

tiagmoraismorgado:
sorry, i thought that there was a spelling error. i will put it back as it should be

PR #839, opened 2017-08-22
Fix build failure with GMM 5.2+

fsateler:
Check if GMM defines the operator<< for std::vector<>, and if not,
define our own version copied from older GMM.

Since now the cmake code is a bit more involved, move it into a FindGMM
module.

I have tested against GMM 5.1 and 5.2


fsateler:
Hmm. For some reason, the appveyor build is unable to detect that gmm 5.1 does have the `operator <<` . Help from the windows devs appreciated.

gogins:
I do not know, but I can say that the current AppVeyor build exclusively
uses Visual Studio 15 2017 for amd64 (same as x64). The compiler behavior
may have changed a bit from Visual Studio 14 2015. Also, I had to change
some operator definitions in CsoundAC for Visual Studio 14 2015 (as
compared with mingw64). Perhaps you need to specialize the types for gmm's
operator << in our code.


Michael Gogins
Irreducible Productions

fsateler:
I don't have a windows machine to test so I went for an alternate solution: define a different overload so the function is always available. Lets see if appveyor likes this better...

PR #845, opened 2017-08-31
Some Travis updates

fsateler:
Unfortunately cmake does not seem to pick up eigen. Maybe it's too old version?


fsateler:
This makes tests a lot faster as they use the container-based images.

Re eigen, this would not be a regression, as it was not being installed previously

kunstmusik:
@fsateler libeigen3-dev is listed twice in the change, is this intentional?

fsateler:
Ehm, no its not, I somehow missed that it was already there. Force pushed a fixed commit.

kunstmusik:
@fsateler Is this commit supposed to show up in this PR?  (I'm not seeing an update.)

fsateler:
Sorry, now it is.

kunstmusik:
Looks good to me and the previous test passed Travis, so I'll merge now. Thanks!

PR #847, opened 2017-09-01
starting over PR in an ad-hoc branch

gesellkammer:
# beosc

Beosc is a set of opcodes for bandwidth enhanced additive synthesis. It is based on the supercollider port of loris. It does not depend on any external library and only implements resynthesis. This is done through two opcodes

* `beosc`: bandwidth-enhanced oscillator, a linearly interpolated oscillator amplitude modulated by noise.
* `beadsynt`: bandwidth-enhanced additive synthesis. Based on csound's adsynt2, adds a bandwidth parameter per oscillator, the noise modulation can have either a uniform or gaussian distribution, with optional frequency interpolation and wavetable lookup interpolation. 
* `tabrowlin`: assuming a 2d (flattened) function table containing multiple rows of sampled streams (for instance, the amplitudes of a set of oscilators), extract one row of that data
with a possible linear interpolation between adjacent rows. 

# emugens

## New opcodes

* `arrayreshape`: given a 1D or 2D array, change the shape of the array. Useful to convert a 1D flat array (possibly copyied from a table) to a 2D array
* `arrayslice`: copy a slice of an array or a table, with optional start, end, and step values. Overlaps with slicearray and could be disposed of if slicearray would add a `step` parameter and the ability to copy from a table also.
* `ftslice`: copy a portion of a table to another table, with optional start, end, and step values
* `arrayprint`: print a (possibly multidimensional) array to stdout
* `arrayview`, `arrayviewend`: create a 'view' on a ftable or another array, possibly indicating start and end indices. This is just a pointer to the original memory, so nothing is copyied.

## Additions

* `linlin`: accepts arrays in the form of `kOut[] linlin kIn[], klow, khigh`, or `kOut[] linlin kx, kLow[], kHigh[]` 
* `cmp`: accepts arrays and 'in between' comparison: 
`kOut[] cmp klow, "<", kIn[], "<=", khigh`
* `bpf`: accepts any number of pairs


kunstmusik:
Thanks this PR is much easier to read.  There is an issue with beosc.c in that it has a license header of LGPL but the two licenses cited are GPL, which is not compatible.  Being a separate plugin, it might be possible to make it entirely GPL, but it might be possible to make it LGPL if the code in beosc.c does not use the GPL code directly and is its own implementation.

vlazzarini:
Thanks for this contribution. There are a few things that we need you to review:
1. The beosc license seems problematic. Maybe you could ask permission to the original authors to license your code as LGPL.  I think making it GPL might make it impossible for us to include it in the source distribution.

2. emugens: arrayslice should be subsumed within slicearray. We'll do it, or you can modify the
original slicearray code if you wish.

3. array**: can you rename these so that they don't start with an "a". Currently the parser is not able to distinguish between variables starting with "a" and opcodes starting with "a". If someone has a variable with these exact names, there will be problems. Note that arrayreshape can be called just reshape.

4. we're not sure about the safety of arrayview*, could you remove it for the moment and open a ticket describing the functionality you'd like implemented and we'll discuss it there. 

5. Make sure all a-rate opcodes include support for sample-accurate mode.

6. can you supply man pages for the opcodes submitted?

We're just like to add that we're very grateful for the contributions, it's great work. We're just making sure there are no issues that might cause us problems in the future.

gesellkammer:
Regarding licenses:

All code here is original, it reimplements the algorithm of Loris as 
described in the original papers and follows the code of Loris regarding 
the filter applyed to the noise used for amplitude modulation. The idea of 
implementing it this way, porting the resynthesis side of it, was inspired 
in the similar work done in the BEASTmulch ugens for supercollider. There 
is also no code taken from them either. Because I don't know anything about 
licenses and didn't want to take credit for the algorithm, I included both 
licenses. Could these licenses be left out?

Regarding arrayslice / slicearray, I would rather NOT touch files I don't 
"own", particularly not so "big" files like arrays.c. I'd rather see 
someone more experiences implement the changes to slicearray. In concrete, 
what's asked is the inclusion of a "step" parameter for slicing. A similar 
change should be done to copyf2array, to take start, end and step 
parameters, all possibly as k-rate params.

About arrayview, I need this functionality for two reasons. One is to be 
able to pass arrays to user opcodes, which currently when passed as arrays 
results in a measurable increase in CPU. The second one is to avoid copying 
to tables for other opcodes which are implemented to use tables. Before 
changing all these to use arrays, having a way to move efficiently from one 
to the other is crutial when dealing with large tables / arrays. I will 
leave it out but I would like to see a solution for this in some way.

Regarding naming: a would try to avoid names like "reshape", which could 
mean anything, in favor of longer but more specific names, like 
"rehapearray". The idea of beginning all opcodes with a common prefix, like 
array... is nice when using autocompletion and to reduce the complexity of 
the already quite chaotic global namespace in csound.

vlazzarini:
Ok, if code is yours, just remove any mention of other licences, that should be suficient. 

We'll do the slicearray etc additions.

Thanks for clarifying your use cases. We could provide overloads to table-using opcodes to take in arrays, which will cover one of your cases (oscillators can do this already). The other seems to me to do with pass by reference, rather than pass by copy, and I can see you might want some optimisation. @kunstmusik, can we plan to address this in 7? 

It's up to you to name your opcodes, but try to avoid starting them with i, a, k, etc. This issue should be addressed in 7, but it is still a source of potential clash.

Thanks

gesellkammer:
I fixed all issues mentioned by @vlazzarini and pushed them to my branch. My knowledge of git being very limited, do I need to do a new PR or will these be pulled automatically?

gesellkammer:
New commits fix issues with sc_phasor and sample-accurate loops

gesellkammer:
I see the fixes in sc_phasor. Can we then leave out my changes in sc_phasor and merge this PR?

kunstmusik:
@gesellkammer: We were looking at this PR and I think it is not merge-able in its current state. It looks like commit 22728e8 added a file  CMakeLists.txt.user, and it is also not possible to leave out parts of a PR (the sc_phasor changes you mention to leave out). 

I know some projects require commits for PRs be squashed and rebased to the head of the branch they are merging to, but I'm not even sure how to do that exactly.  Could you file yet again another PR with just the valid changes?

gesellkammer:
I think that at this point it is better not to continue filing any PRs. Thanks for your time.

kunstmusik:
@gesellkammer Well as mentioned, the PR can not be merged with the state it is in, but it is not to say the parts can not be merged some other way.  If you would like the other parts manually merged, it is more work but doable. 

However, we would still need new manual pages for the new/revised opcodes. So:

1. Would you like us to not do the PR but merge the changes manually?
2. If so, do you have manual text available? (If you can have it even in just text format, someone can help to add those to the manual project)

vlazzarini:
If you send us your files, we can try adding them to the sources, should be no problem.

PR #863, opened 2017-10-21
Fix python test case runner for Windows

stekyne:
@kunstmusik Could you take a look and see if this is suitable or not? 

This changes worked for me locally and all tests were passing. Would be nice to have appveyor run these tests after each build if possible.

### DIFF
--- a/tests/commandline/test.py
+++ b/tests/commandline/test.py
@@ -161,7 +161,10 @@ def runTest():
     tests += udoTests
 
     output = ""
-    tempfile = "/tmp/csound_test_output.txt"
+    if(os.name == 'nt'):
+    	tempfile = "/csound_test_output.txt"
+    else:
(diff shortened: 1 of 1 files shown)


kunstmusik:
I don't think this is the right fix and that what was probably happening is that the csound on your path was being used and not the one that gets built.  The part about /tmp was good though.  I did a change based on this code and pushed it to develop:

https://github.com/csound/csound/commit/57a17fafdc3fbcdc374be0f919d509cb7523ab1c

I'm pretty sure that is correct and it at least gets the tests to run here for both MinGW and VS.  For VS, I ran "cmake --build . --target csdtests" within my csound/msvc/csound-vs folder.  

@stekyne - could you review that commit, and would you know how we could hook this into the AppVeyor build?

PR #880, opened 2017-11-19
Fix format string attributes and usages to allow -Werror=format

fsateler:
Disabling format errors is bad, as introducing such errors is easy. Lets allow using `-Wformat` and `-Werror=format`.

Maybe even enable `-Wformat` by default?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,12 +456,6 @@ else()
     add_definitions("-DBETA")
 endif()
 
-if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
-    add_definitions("-Wno-format")
-endif()
-
-
-
 #if(USE_DOUBLE)
 #    set(BUILD_PLUGINS_DIR ${BUILD_DIR}/plugins64)
 #else()
@@ -598,7 +592,7 @@ set(HEADERS_TO_CHECK
     unistd.h io.h fcntl.h stdint.h
     sys/time.h sys/types.h termios.h
     values.h winsock.h sys/socket.h
-    dirent.h )
+    dirent.h inttypes.h)
 
 foreach(header ${HEADERS_TO_CHECK})
     # Convert to uppercase and replace [./] with _
@@ -607,6 +601,11 @@ foreach(header ${HEADERS_TO_CHECK})
     string(REGEX REPLACE [./] "_" upper_header ${tmp})
     check_include_file(${header} HAVE_${upper_header})
 endforeach()
+
+if (NOT HAVE_INTTYPES_H)
+    message(FATAL_ERROR "Csound requires inttypes.h")
+endif()
+
 
 check_deps(USE_LRINT HAVE_LRINT)
 if(USE_LRINT)
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -29,6 +29,7 @@
 #include <math.h>
 #include <ctype.h>
 #include <string.h>
+#include <inttypes.h>
 
 #include "oload.h"
 #include "insert.h"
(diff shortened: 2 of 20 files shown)


fsateler:
CI has passed here too.

kunstmusik:
I should have commented on this a week or two ago when I quickly looked at it, but I think it's a really good idea.  I'm not familiar with using inttypes.h for formatting like this but if this is what is the norm these days then that seems fine to me.  I'd just want to make sure it build alright on other platforms (WebAssembly would be easy to build/test; I have no access to a Raspberry PI though which would be one where I imagine tooling might be a bit older.)

jpffitch:
I am not so sure.  It makes printer unreadable

⁣Sent from TypeApp ​

vlazzarini:
Can we resolve this PR?

jpffitch:
I remain sceptical.  I cannot see what problem this attacks, no indication 
of how it interacts with Str calls.  Have there been any problems with 
current scheme?  I am not aware of them

fsateler:
The problem is that string formatting errors are a common source of vulnerabilities/bugs. Debian builds most software with `-Wformat=error` so that packages fail to build if the format string is wrong, to avoid the entire class of vulnerabilities/bugs. Unfortunately, because csound has so many such errors, the flag cannot be used, and I have to disable it explicitly. I'd like to restore the formatting errors, as they can uncover true bugs (like [this one](https://github.com/csound/csound/pull/880/files#diff-c441289cc5049b6ef8505e7ac5c73b7eL71)), and they help prevent bugs when running in other architectures (due to different sizes of types).

In general, format strings are easy to verify for the compiler, so it seems a waste to not let it help.

If the objections are to the use of the stdint macros (which I agree are a bit ugly), then the solution would be to change those usages to casts to `long` or `long long` and use the regular formatters.

jpffitch:
I repeat I have not seen a bug related to formats for years.  You say
"as they can uncover true bugs (like this one),"
but the link does not show a bug.

How does it interact with internationalisation?

> If the objections are to the use of the stdint macros (which I agree are a bit
> ugly), then the solution would be to change those usages to casts to long or
> long long and use the regular formatters.

i thought that was what we do.  Clearly we are at cross purposes

fsateler:
> I repeat I have not seen a bug related to formats for years.  You say "as they can uncover true bugs (like this one),"  but the link does not show a bug.

Well, a `%.100s` format is invalid. So if this works it is merely by luck, and could stop working with any libc upgrade.

> How does it interact with internationalisation?

I don't think this makes much change. Possibly it will generate some amount of fuzzed strings due to changed specifiers. I'll take a look.

> i thought that was what we do.  Clearly we are at cross purposes

I'm not sure why you say we are at cross purposes. I can change the PR to use casts instead of the stdint specifiers, would that solve your concerns?

jpffitch:
Anoter question; do the inttype.h macros exist in MSVC?

fsateler:
>  do the inttype.h macros exist in MSVC?

Appveyor build succeeded so I think yes. I added a hard dependency on inttypes.h in this patch.

fsateler:
I have resolved the conflicts. The question of wether to use inttypes.h or not is still out there. I can change to not use that if you want to.

vlazzarini:
inttypes is a C99 header, is it not? I am happy with it. It seems to pass the windows build anyway

vlazzarini:
Will wait until travis is finished, then merge

vlazzarini:
travis was taking too long but it was only to do with the macos queue

vlazzarini:
@fsateler I will need to revert this PR. It somehow broke our tests, eg:
```
MacBook-Pro:debug victor$ ./csound -n ../tests/commandline/test12.csd 
time resolution is 1000.000 ns
0dBFS level = 32768.0
--Csound version 6.11 beta (double samples) Jan 20 2018 
[commit: 57485442b455fbc89f1f3b8895b361d919c63d3b]
libsndfile-1.0.25
UnifiedCSD:  ../tests/commandline/test12.csd
STARTING FILE
Creating orchestra
closing tag
Creating score
rtaudio: PortAudio module enabled ... 
using callback interface
error:  Variable '__synthetic_300' used before defined
Line -1
error:  Variable type for __synthetic_300 could not be determined.
Stopping on parser failure
cannot compile orchestra
end of score.		   overall amps:      0.0
1 errors in performance
Elapsed time at end of performance: real: 0.003s, CPU: 0.002s
no sound written to disk

csound command: Segmentation fault
```

Could you review it again? I am sorry about this.

kunstmusik:
I think we're okay now with commit df5658c .

vlazzarini:
yes we are.

fsateler:
Sorry for the delay, I was out for the weekend.

I'm confused now: do I need to do anything? It appears this PR was not reverted after all.

kunstmusik:
Nothing to do,the problem was found and fixed. Thanks again!

PR #889, opened 2017-12-11
Allow aftertouch to be init to 0

alcomposer:
Addition of `--aft-zero` flag to Csound to allow aftertouch to be init to 0. Without `--aft-zero` flag set the value of aftertouch is default to 127 (current behaviour).


jpffitch:
I am still not convinced this flag belongs in OPARM.  I would put in CSOUND with value 127 or 0.

Also someone has to update the manual.

And I repeat is this for 6.10 or later?

PR #912, opened 2018-01-12
make an attempt to find Faust2 with cmake

hlolli:
Some cmake functions to determine if Faust2 is installed, borrowed from https://github.com/OSSIA/score/blob/master/CMake/modules/FindFaust.cmake with a clue from @stekyne 

I have faust installed and it was able to locate it on my machine. Hope this doesn't cause problems with those who don't have Faust installed.


kunstmusik:
Shouldn't we just copy FindFaust.cmake into our cmake/ folder and use it with a find_package(), rather than integrate directly into that CMakeLists.txt file?  Seems like it'd be a bit tidier that way.

vlazzarini:
yep, I think it is better as a separate file/package. Can you try that?

kunstmusik:
The FindFaust.cmake in this PR is reduced compared to the one for OSSIA score.  It's also a bit odd as the finder script looks to be used just to see if Faust is there, but then the older code is still there. Seems like the PR isn't really complete somehow?  (i.e., the finder code for faust and accompanying flag setting should probably be all in the finder script, then referenced from our CMake file)

hlolli:
That's right, I only took the part where it's looking for if Faust is installed, and then continue as before, I guess I can clean some of the code that's already inside the Opcodes/CMakeLists.txt file. https://github.com/csound/csound/blob/develop/Opcodes/CMakeLists.txt#L47-L51 either delete it and use the code from OSSIA or just move it. My initial intention was to only find if Faust is present and install the opcodes, so I didn't have the confidence in changeing the code I'm guessing Victor wrote.

vlazzarini:
@hlolli please go ahead and change it. The worst case is that you break the faustcsound build, but then I will just ask you to fix it ;)

Opcodes/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ find_library(WIIUSE_LIB wiiuse)
 check_include_file(p5glove.h P5GLOVE_H)
 find_library(P5GLOVE_LIB p5glove)
 
-if(BUILD_FAUST_OPCODES)
+if(BUILD_FAUST_OPCODES AND FAUST_FOUND)
```
fsateler: Use `check_deps(BUILD_FAUST_OPCODES FAUST_FOUND)` please. This gives a message to users when it is disabled, and can be made to fail with `-DFAIL_MISSING=ON` for distribution builds.

cmake/Modules/FindFAUST.cmake
```diff
@@ -0,0 +1,19 @@
...
+  /usr/local/lib/faust/architecture/
+  "${FAUST_INCLUDE_DIR_HINT}"
+  )
+
+set(FAUST_NAMES ${FAUST_NAMES} libfaust.so libfaust.dylib faust.dll faust libfaust)
+find_library(FAUST_LIBRARY
+  NAMES ${FAUST_NAMES}
+  HINTS "${FAUST_LIB_DIR_HINT}")
+
+if(FAUST_INCLUDE_DIR AND FAUST_LIBRARY)
```
fsateler: You can use `find_package_handle_standard_args(FAUST FAUST_INCLUDE_DIR FAUST_LIBRARY)` to set this. This macro also handles REQUIRED and QUIETLY keywords. You need to `include(FindPackageHandleStandardArgs)` to use it though.

Opcodes/CMakeLists.txt
hlolli: Will add!

cmake/Modules/FindFAUST.cmake
vlazzarini: @hlolli did you make these changes?

hlolli: No not these, I'll add them!

hlolli: Done!

vlazzarini: all ready to merge?

PR #918, opened 2018-01-23
Updating installer..

rorywalsh:


kunstmusik:
Keeping track, two things needs fixing:  

* The generated float-version.h header is not put into the installer.  This needs a fix to copy *.h, another copy of *.hpp, and then a copy of the individual float-version.h.  The current installer copies float-version.h.in into the installer which is wrong.

* CsoundQt requires libsndfile-1.dll because it dynamically links with it.  (Waiting on CsoundQt build with that provided.)

PR #919, opened 2018-01-24
wasm compilation minor changes

hlolli:
So I have 3 modifications to propose, one being the .gitignore which is self-evident.
One is adding `-s NO_EXIT_RUNTIME=0` to emcc step of the wasm compilation, not includeing this causes a warning being thrown useing the newest versions of the emscripten compiler as is described here https://kripken.github.io/emscripten-site/docs/getting_started/FAQ.html#what-does-exiting-the-runtime-mean-why-don-t-atexit-s-run

the warning was printed like this
```
atexit() called, but NO_EXIT_RUNTIME is set, so atexits() will not be called. set NO_EXIT_RUNTIME to 0 (see the FAQ)
0dBFS level = 32768.0
--Csound version 6.11 (float samples) Jan 23 2018 
[commit: d567b0d2226f75a1d97c394a54df6881eed27e14]
libsndfile-1.0.25
0dBFS level = 1.0
orch now loaded
SECTION 1:
new alloc for instr 1:
```
adding NO_EXIT_RUNTIME=0 removes it.

The last change is seemingly redundant use of `CFLAGS="-s WASM=1"` into the configure step of libsndfile (emconfigure). By removeing this I was able to compile, because for some strange reason this caused emconfigure not to find my c-compiler. And if this really is redundant, then best to remove it.

Best would be if someone can confirm that these canges do not break the wasm build in any way


kunstmusik:
I tested locally and it seems to have compiled libsndfile and wasm csound just fine.  Examples project ran in FF.  Merging now.

PR #924, opened 2018-02-02
Fix to cmake Xcode Generator for recent xcode projects. 

benmca:
Currently, builds for multiple targets in the generated XCode project fail because of misconfiguration in the generated targets. The generated targets all have libstdc++ as the default c++ lib for llvm, and we want libc++. 

One workaround - and I'd assume this won't introduce any issues - is to up the the target version as suggested by XCode's warnings. I upped target version to 10.9 for generated Xcode project. Targeting a lower version causes llvm to default to using libstdc++ instead of libc++, and not have C++11 features on by default, which breaks some subprojects using inline namespaces - nicely outlined here:

https://stackoverflow.com/questions/12542971/using-libstdc-compiled-libraries-with-clang-stdlib-libc

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -409,8 +409,8 @@ message(STATUS "Building for OSX")
   endif()
 
   if(CMAKE_OSX_SYSROOT AND NOT CMAKE_OSX_SYSROOT MATCHES ".*OSX10.6.*")
-    set(CMAKE_OSX_DEPLOYMENT_TARGET 10.7)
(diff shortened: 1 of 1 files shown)


vlazzarini:
I don't think we should change this in CMakeLists.txt. A better solution is to allow this variable to be set in Custom.cmake. I've fixed the code so that this can be done in all cases (previously it could only be done if the default sysroot was not found.

benmca:
Worked! Thanks Victor.

PR #925, opened 2018-02-02
Libcsound exposed fn

hlolli:
I'm guessing newer emscripten compilers don't automatically add `'FS'` and `'cwrap'` to `Module`. And providing this flag solves it. Maybe I'm forgetting some other global variables. We could also consider wrapping libcsound in IIFE as not to contaminate the global namespace (window/process).


hlolli:
More on this here:
https://kripken.github.io/emscripten-site/docs/api_reference/preamble.js.html

```
 you want to call preamble methods from somewhere the compiler can’t see, like another script tag on the HTML, you need to export them. To do so, add them to EXTRA_EXPORTED_RUNTIME_METHODS (for example, -s 'EXTRA_EXPORTED_RUNTIME_METHODS=["ccall", "cwrap"]' will export call and cwrap). Once exported, you can access them on the Module object (as Module.ccall, for example).
```

PR #938, opened 2018-03-18
Updated usage of the ableton API

stekyne:
I found the changes required to make this build again but I have no means of testing it. I'm fairly sure it's correct but there is a similar API call that it could be as well. Maybe @gogins  can confirm?


kunstmusik:
Merging this as we are trying to finish up for 6.11.0.  I was able to build the branch locally and test using the manual/examples/ableton_link.csd, linking with Reaktor 6 and have Csound in sync with Reaktor's metronome. (Win10)

gogins:
Great!

PR #958, opened 2018-05-11
Swift 4 update, added Accelerate to linked frameworks in examples projects

nikhilsinghmus:
Minor update to Swift 4 and Xcode's newest recommended settings, and adding Accelerate to the linked frameworks in the example projects so that users don't have to.


kunstmusik:
@vlazzarini Perhaps an updated release?  Thinking we may get further questions about Accelerate otherwise.

vlazzarini:
Ok, no need for a new build, just re-packaging is that right?

kunstmusik:
Yes, I think that's right, libcsound.a and all that should be just fine.

vlazzarini:
Ok done.

PR #967, opened 2018-06-05
audioWorklet === undefined if unimplemented instead of null

hfmanson:


kunstmusik:
I checked and indeed it shows up undefined rather than null in both Edge and Firefox on Win 10.  Merged, thanks!

PR #971, opened 2018-07-06
AudioWorklet output channel fix + MIDI support

hfmanson:
In AudioWorkletNode each input and output can have multiple channels. The number of channels in an output is specified in options.outputChannelCount per output. In Csound we only have one output, which can contain multiple channels. Now my stereo tobias.csd works. 

Midi support fix

### DIFF
--- a/Emscripten/src/CsoundNode.js
+++ b/Emscripten/src/CsoundNode.js
@@ -52,9 +52,6 @@ class CsoundNode extends AudioWorkletNode {
      */
     constructor(context, options) {
         options = options || {};
-        options.numberOfInputs  = 1;
-        options.numberOfOutputs = 2;
-        options.channelCount = 2;
         
         super(context, 'Csound', options);
 
@@ -309,7 +306,8 @@ class CsoundNodeFactory {
     static createNode(inputChannelCount=1, outputChannelCount=2) {
         var options = {};
         options.numberOfInputs  = inputChannelCount;
-        options.numberOfOutputs  = outputChannelCount;
+        options.numberOfOutputs = 1;
+        options.outputChannelCount = [ outputChannelCount ];
         return new CsoundNode(CSOUND_AUDIO_CONTEXT, options);
     }
 }
--- a/Emscripten/src/CsoundObj.js
+++ b/Emscripten/src/CsoundObj.js
@@ -307,6 +307,43 @@ class CsoundObj {
         }
     }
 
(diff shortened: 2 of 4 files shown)


kunstmusik:
@hfmanson The output channel fix makes sense to me, but could you explain the MIDI support change?  Just want to make sure I understand the change and that it is there to restore older built-in WebMIDI support to CsoundObj.js, yes?

hfmanson:
I copied enableMidiInput(midiInputCallback) from a previous version of CsoundObj.js and modified it so the midiInputCallback parameter can be optional. When a MIDI event occurs on the main thread it is posted to the audioprocessor (ScriptProcessorNode or AudioWorklet) through the already present midiMessage function

kunstmusik:
I tested a build with the changes and found that:

1. Stereo was indeed not working and is now fixed
2. MIDI input worked. I do not use csound.js so that change would best be reviewed by @vlazzarini. The addition of the enableMidiInput() method to CsoundObj looks good to me.  One can decide to use it if they want to, or they can do the WebMIDI code themselves and have finer control over how to handle MIDI devices.  

@vlazzarini - I think these changes are good.  If you're okay with the csound.js change, I will go ahead and merge and close.

kunstmusik:
One note: we could use an arrow function so that we wouldn't have to use the "that", something like:

```javascript
        const handleMidiInput = (evt) => {
            this.midiMessage(evt.data[0], evt.data[1], evt.data[2]);
            //console.log("midievent: " + event.data[0] + ", " + event.data[1] + ", " + event.data[2]);
        };
```

hfmanson:
@kunstmusik Just tested the arrow function. 'this' is undefined, it seems handleMidiInput is called on another MIDI thread

kunstmusik:
@hfmanson Check out the code in context here:

https://github.com/kunstmusik/csound-audioworklet/blob/master/csound/CsoundObj.js#L311-L315

I get console output when driving this test page:

https://kunstmusik.github.io/csound-audioworklet/midiInputTest.html

with MIDI output.  The console.log shows:

CsoundObj.js:313 ƒ midiMessage(byte1, byte2, byte3) {
        this.node.midiMessage(byte1, byte2, byte3);
    }

hfmanson:
I am afraid I got fooled by the debugger, now the arrow function works correctly. I'm ok with that change

kunstmusik:
Merged. I will add the arrow function shortly. Thanks all!

PR #974, opened 2018-07-10
that -> this

hfmanson:


kunstmusik:
@hfmanson thanks for catching that!

PR #976, opened 2018-07-11
Module['print'] no longer available

hfmanson:
Module['print'] is no longer available in CsoundObj.js and all references should be removed

### DIFF
--- a/Emscripten/src/CsoundObj.js
+++ b/Emscripten/src/CsoundObj.js
@@ -287,6 +287,7 @@ class CsoundObj {
         let that = this;
 
         if (navigator.getUserMedia === null) {
+            console.log("Audio Input not supported in this browser");
             audioInputCallback(false);
         } else {
             let onSuccess = function(stream) {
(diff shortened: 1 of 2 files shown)


kunstmusik:
I'm wondering about having messages with console.log(), but I don't know if that's useful or not since for these two scenarios both have success/fail callbacks. @vlazzarini any thoughts?

vlazzarini:
Not sure. I thought 'print' was still available for the SPN case.

hfmanson:
I think it would be best to export a function CsoundObj_message in CsoundObj,c which encapsulates csoundMessage and add a wrapper to CsoundProcessor.js and CsoundScriptProcessorNode.js called 'message'

hfmanson:
You can replace Module['print'] with console.log. console.log will display in the csound console window when used after moduleDidLoad
<pre>
function moduleDidLoad() {
   csound.Csound.enableMidiInput(undefined);
   ...
}
</pre>

kunstmusik:
@hfmanson Using console.log seems good to me, but not sure how removing enableMidiInput() is necessary?  Could you explain what moving that to moduleDidLoad does?

hfmanson:
Now the webpage using csound should enable MIDI itself through csound.Csound.enableMidiInput(undefined); in moduleDidLoad. The advantage is, when MIDI is not supported by the browser, an error message will display in the csound console window

kunstmusik:
Going to merge as the change makes sense to me. (@vlazzarini may want to review and modify csound.js though.)

PR #981, opened 2018-07-17
XHTML awareness

hfmanson:
This PR makes csound XHTML aware. included is minimal.xml which is the XML version of minimal.html

### DIFF
--- a/Emscripten/.gitignore
+++ b/Emscripten/.gitignore
@@ -6,7 +6,6 @@ dist
 dist-wasm
 dist-web-csound
 ace
-examples
 examples-wasm
 csound-emscripten-*
 csound-wasm-*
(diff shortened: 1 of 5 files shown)


kunstmusik:
@hfmanson just reviewing this now.  Looks good but could you verify that these changes are compatible with HTML5?  (I haven't looked at or thought about XHTML in ages so I don't feel confident in my assessment.) Thanks!

hfmanson:
@kunstmusik I already did. Now elements are explicitly created in the HTML namespace http://www.w3.org/1999/xhtml . createElementNS is always available because the DOM is namespace aware

kunstmusik:
@hfmanson thanks just wanted confirmation. Will merge now.

PR #1008, opened 2018-09-07
Update PKGBUILD

drlazz:
Dead links replacement.


vlazzarini:
Sorry @drlazz, we've just completed some housekeeping that involved replacing the old develop branch, and for this reason your pull request was closed by the system.

If you want to submit it again, we'll be happy to look at it. You will need a fresh clone to do this, otherwise it won't work.

PR #1020, opened 2018-09-14
Update emugens,  band-enhanced additive synthesis

gesellkammer:
This PR updates emugens, fixing many bugs and adding new opcodes. It adds a new set of opcodes to implement band-enhanced additive synthesis, as implemented in loris, in csound.

## emugens

* `printarray`: print 1D and 2D arrays
* `tab2array`: copy a slice of a table (start, end, step) to an array
* `ftslice`: copy a slice of a table to another table
* `bpf`: breakpoint linear interpolation, now accepts unlimited number of breakpoints and accepts arrays
* `bpfcos`: similar to bpf, but with cosine interpolation
* `linlin`: accepts arrays, allows to make linear interpolation between arrays
* `lincos`: similar to linlin, with cosine interpolation
* or and and bitwise operation between two arrays

## beosc

* `beosc`: band enhanced oscillator, a port of BEOsc from supercollider, with added gaussian noise and linear interpolation
* `beadsynt`: a bank of band-enhanced oscillators, makes resynthesis of files analyzed via loris possible and efficient
* `getrowlin`: get a row from a 2D array, possibly a slice (giving start, end and step for the row), with linear interpolation between adjacent rows
* `tabrowlin`: the same but for a table

The band-enhanced opcodes are developed together with a python library (https://github.com/gesellkammer/loristrck) and set of scripts to analyze, filter and resynthesize soundfiles via partial tracking.

For an example of beadsynt, see https://github.com/gesellkammer/ugens/blob/master/beosc/examples/beadsynt3.csd

### DIFF
--- a/Opcodes/emugens/.gitignore
+++ b/Opcodes/emugens/.gitignore
@@ -1,1 +1,0 @@
-compile_commands.json
--- a/Opcodes/emugens/.gitignore
+++ b/Opcodes/emugens/.gitignore
@@ -1,0 +1,1 @@
+compile_commands.json
--- a/Opcodes/emugens/emugens.c
+++ b/Opcodes/emugens/emugens.c
@@ -18,122 +18,307 @@
 
     You should have received a copy of the GNU Lesser General Public
     License along with Csound; if not, write to the Free Software
+
     Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA
     02110-1301 USA
 */
 
 #include <csdl.h>
+#include <math.h>
+
+#define SAMPLE_ACCURATE \
+    uint32_t n, nsmps = CS_KSMPS;                                    \
+    MYFLT *out = p->out;                                             \
+    uint32_t offset = p->h.insdshead->ksmps_offset;                  \
+    uint32_t early = p->h.insdshead->ksmps_no_end;                   \
+    if (UNLIKELY(offset)) memset(out, '\0', offset*sizeof(MYFLT));   \
+    if (UNLIKELY(early)) {                                           \
+        nsmps -= early;                                              \
+        memset(&out[nsmps], '\0', early*sizeof(MYFLT));              \
+    }                                                                \
+
+#define INITERR(m) (csound->InitError(csound, m))
+#define MSG(m) (csound->Message(csound, Str(m)))
+#define MSGF(m, ...) (csound->Message(csound, Str(m), __VA_ARGS__))
+#define PERFERROR(m) (csound->PerfError(csound, &(p->h), "%s", m))
+
+/* from Opcodes/arrays.c, original name: tabensure */
+static inline void
+arrayensure(CSOUND *csound, ARRAYDAT *p, size_t size) {
+    if (p->data==NULL || p->dimensions == 0 ||
+        (p->dimensions==1 && p->sizes[0] < size)) {
+        size_t ss;
+        if (p->data == NULL) {
+            CS_VARIABLE *var = p->arrayType->createVariable(csound, NULL);
+            p->arrayMemberSize = var->memBlockSize;
+        }
+        ss = p->arrayMemberSize*size;
+        if (p->data==NULL)
+            p->data = (MYFLT*)csound->Calloc(csound, ss);
+        else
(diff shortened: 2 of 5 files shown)


vlazzarini:
1) have you used a fresh clone to do this PR or have you rebased an existing fork? We have to be careful not to bring tainted commits back into develop.

2) Appveyor is failing; it might be unrelated but we need to check.

kunstmusik:
@vlazzarini I think we can do rebasing ourselves in the github UI or perform the merge locally with rebase and double check any problems with VST-tainted branches.  I too a look at the commits locally and I think it's post-update of develop so I think merging should work out without much hassle. 

In terms of new opcodes, we should double-check with names and overlap with existing opcodes.  beosc and beadsynt seem like no problems, but bpf seems similar to linseg and bpfcos with cosseg, for example.  And considering things like would it be better to have a separate getrowlin opcode or just modify the existing getrow to allow start:end:step? I remember talking with some people confused about the introduction of mtof after it was added as it's pretty much the same as cpsmidinn and would like to be careful here. (Probably best to get check in with rest of community to see if they have any preferences/thoughts on opcode names and overlaps, merge as-is, etc.)

gesellkammer:
@vlazzarini: the errors in appveyor are related to M_PI not being defined. I see that the build is done in windows, so the errors are probably related to some difference relating to what is guaranteed to be defined in math.h

vlazzarini:
ok, maybe @kunstmusik or @skyne can help with the Windows issue.

Opcodes/emugens/.gitignore
```diff
@@ -0,0 +1 @@
+compile_commands.json
```
fsateler: I don't think this file belongs in the repo. This may be added to the base gitignore to ignore any compile-commands

Opcodes/emugens/emugens.c
```diff
@@ -18,255 +18,527 @@
...
+    if (x>=*data[datalen-2]) {
+        *p->r = *data[datalen-1];
+        return OK;
+    }
+    for(i=2; i<datalen; i+=2) {
+        x1 = *data[i];
+        y1 = *data[i+1];
+        if( x <= x1 ) {
+            dx = ((x-x0) / (x1-x0)) * M_PI + M_PI;
+            *p->r = y0 + ((y1 - y0) * (1 + cos(dx)) / 2.0);
```
fsateler: You need to `#include <math.h>` for `M_PI`, `cos` and other math functions. I think it;s better to be explicit about the dependencies rather than rely on it being included indirectly (this might fix the appveyor issue).

gesellkammer: ok, fixed both issues

jpffitch:
M_PI definition seems to have vanished evertwhere!  There are definitions 
in InOut/libmpadec/mpadec_config.h and Opcodes/doppler.cpp but uses in 
other files in OOps and Opcodes seemnaked.  It should be in csounfCore.h I 
thought

Seemsmidssing from base macros as well

Opcodes/emugens/emugens.c
jpffitch: Surely people should be using PI defined in csoundCore.h with the PI_F variant

skyne:
Hi, sorry to disturb the conversation, but i think @vlazzarini , i'm not the man who you wanted to tag here :)

vlazzarini:
Still failing on Windows, it appears.

vlazzarini:
Looks you need another token defined on MSVC (that is why I asked you @skyne)
https://msdn.microsoft.com/en-us/library/4hwaceh6.aspx

gesellkammer:
> Looks you need another token defined on MSVC 
@vlazzarini: I assume this comment was previous to the last commit, where PI from csoundCore.h is used instead, as indicated by @jpffitch.

vlazzarini:
yes it was. Let's wait for appveyor now

kunstmusik:
Checking in, bpf having a time input makes sense to have it separate.  I'm at the Web Audio Conference in Berlin and won't be able to participate on this again until next week; otherwise looks like Appveyor is building this branch now okay.

vlazzarini:
I am happy to merge this. Do we have man pages forthcoming?

gesellkammer:
Yes, man pages will follow

PR #1025, opened 2018-09-23
use -lncurses instead of -lcurses

hlolli:
On NixOs, -lcurses isn't available and looking at Faust Makefiles, there is only mention of -lncurses but not -lcurses. Ncurses seems to be a library commonly used and easy to find.


vlazzarini:
Did you get a fresh clone to make the PR, or rebased your repo?

hlolli:
What I did, was that I cloned this (original) repo, changed the remote origin url, and force pushed to my fork. I can ofc delete my fork and re-fork if that makes things safer?

vlazzarini:
I just wanted to make sure this will not bring any tainted commits that will bring the duplicate history back into the repo

kunstmusik:
@vlazzarini There's a couple of options, but right now I think best for us to manually test locally first. Another option is to use the rebase option in the PR interface.  For manual, the PR interface has:

"You can also open this in GitHub Desktop or view command line instructions."

I've been following the command line instructions to test locally (following Step 1), primarily just checking the "git rev-list --count HEAD" as well as checking "git merge-base HEAD develop" to see the commit hash for the branching point.  

In this case, this PR looks clean.  If the contents of the commit are good to you, we can merge it.  

For the future, we should probably make it a policy to test all PRs locally and then choose from there if it's good what to do (use Github PR mechanism for merge or rebase, or apply changes locally and push via Step 2 in the command line instructions).  Probably best to do this kind of process moving forward so that we always have at least one level of verification and testing before merging in.

vlazzarini:
ok. I'll have to read on how to merge the PR locally, but that method seems ok to me.

PR #1033, opened 2018-09-28
Emugens fix

gesellkammer:
Fixes misc. bugs of previous pull request


gesellkammer:
Can you point me to an example in the code, so that I know how to fix this?

vlazzarini:
I suspect @jpffitch  might be referring to the global arrays  such as

```
int32_t _pcs[] = {9, 11, 0, 2, 4, 5, 7};
```

it looks like they're ready-only so should be OK, but may be best marked const or
placed inside the functions where they are used

jpffitch:
For example look at Opcodes/OSC.c where there is a global structure.

static CS_NOINLINE OSC_GLOBALS *alloc_globals(CSOUND *csound)
{
     OSC_GLOBALS *pp;

     pp = (OSC_GLOBALS*) csound->QueryGlobalVariable(csound, 
"_OSC_globals");
     if (pp != NULL)
       return pp;
     if (UNLIKELY(csound->CreateGlobalVariable(csound, "_OSC_globals",
                                               sizeof(OSC_GLOBALS)) != 0)){
       csound->ErrorMsg(csound, "%s", Str("OSC: failed to allocate globals"));
       return NULL;
     }
     pp = (OSC_GLOBALS*) csound->QueryGlobalVariable(csound, "_OSC_globals");
     pp->csound = csound;
     pp->mutex_ = csound->Create_Mutex(0);
     csound->RegisterResetCallback(csound, (void*) pp,
                                   (int32_t (*)(CSOUND *, void *)) OSC_reset);
     return pp;
}

and other places.  look for QueryGlobalVariable

Another simple place is in Opcodes/zak.c

jpffitch:
No  I meant the printarray_fmt opcide keeps a string betwen opcodes in the 
new PR; was not in develop branch yet

Readonly statics are OK

vlazzarini:
yes, that's not allowed alright. Needs to be placed in a globally allocated object as you described.

gesellkammer:
Is there still anything that needs to be changed with the current PR?

gesellkammer:
Honestly, the fix proposed was too complex for my own knowledge of csound's internals and my programming skills in general. I would be happy to reconsider it, with some help.

vlazzarini:
Happy to help. What do you need?

gesellkammer:
The idea behind the now removed opcode "printarray_setfmt" was a global way to control the format and most of all the precision of numbers when being printed, and at the same time be able to customize other options, such as line width (when to introduce a line break), etc. But instead of introducing an opcode to control another opcode, I would rather have a global way of setting options which can be used by the opcodes, something like was introduced with the A4 variable, but for general purpose. Something like

`setoption "printfmt" "%.4f"`

This would be documented in each opcode which makes use of such variables, so that they can be shared across opcodes (in this case, for instance, the newly introduced `ftprint` could default to this format whenever the user does not use a format explicitly)

vlazzarini:
It's not clear to me which opcodes this would apply to and how. 

A4 is great that, like 0dbfs, has a clear
application scope.

So we need to see how your idea can be similarly developed

gesellkammer:
At this point I would opt to leave the default format as fixed (it is an optional parameter, so the default is only used when the parameter is left unset or an empty string "" is used, all this is documented in the manual) and make any further changes based on usage.

jpffitch:
beosc.c does not comple
/home/jpff/csound6/Opcodes/emugens/beosc.c:650:20: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
         for (c=0; c<p->count; c++) {
                    ^
/home/jpff/csound6/Opcodes/emugens/beosc.c: In function ‘beadsynt_init’:
/home/jpff/csound6/Opcodes/emugens/beosc.c:674:19: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if (ftp->flen < count) {
                   ^
/home/jpff/csound6/Opcodes/emugens/beosc.c:683:19: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if (ftp->flen < count) {
                   ^
/home/jpff/csound6/Opcodes/emugens/beosc.c:692:19: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if (ftp->flen < count) {
                   ^
/home/jpff/csound6/Opcodes/emugens/beosc.c: In function ‘tabrowcopyarr_k’:
/home/jpff/csound6/Opcodes/emugens/beosc.c:1216:14: error: unused variable ‘numitems’ [-Werror=unused-variable]
     uint32_t numitems = (uint32_t)ceil((end - start) / (float)step);
              ^~~~~~~~
/home/jpff/csound6/Opcodes/emugens/beosc.c: In function ‘beadsynt_init_common’:
/home/jpff/csound6/Opcodes/emugens/beosc.c:602:61: error: ‘count’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     if (p->lphs.auxp==NULL || p->lphs.size < sizeof(int32_t)*count)
                                              ~~~~~~~~~~~~~~~^~~~~~
/home/jpff/csound6/Opcodes/emugens/beosc.c: In function ‘beosc_akiii’:
/home/jpff/csound6/Opcodes/emugens/beosc.c:113:25: error: ‘seed’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     tmp1  = (uint64_t) ((int32_t) (*seedptr) * (int64_t) 742938285);
                         ^~~~~~~~~~~~~~~~~~~~
/home/jpff/csound6/Opcodes/emugens/beosc.c:443:14: note: ‘seed’ was declared here
     uint32_t seed;
              ^~~~
/home/jpff/csound6/Opcodes/emugens/beosc.c: In function ‘beosc_kkiii’:
/home/jpff/csound6/Opcodes/emugens/beosc.c:113:25: error: ‘seed’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     tmp1  = (uint64_t) ((int32_t) (*seedptr) * (int64_t) 742938285);
                         ^~~~~~~~~~~~~~~~~~~~
/home/jpff/csound6/Opcodes/emugens/beosc.c:346:14: note: ‘seed’ was declared here
     uint32_t seed;
              ^~~~
cc1: all warnings being treated as errors

jpffitch:
emugen.c does not compile either.  I have fixed the trivial problems in beos but do not understand enough to fix unitialised variables.  Not being compiled until code is fixed.

emugens.c: (will look to see if I can fix)

/home/jpff/csound6/Opcodes/emugens/emugens.c: In function ‘ftprint_perf’:
/home/jpff/csound6/Opcodes/emugens/emugens.c:1447:23: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     if(end < 1 || end > ftplen)
                       ^
/home/jpff/csound6/Opcodes/emugens/emugens.c:1454:19: error: comparison between signed and unsigned integer expressions [-Werror=sign-compare]
     for(i=start; i<end; i+=step) {
                   ^
/home/jpff/csound6/Opcodes/emugens/emugens.c:1455:45: error: value computed is not used [-Werror=unused-value]
         printf(default_printfmt, ftable[i]) + 1;
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~
cc1: all warnings being treated as errors

jpffitch:
Fixed emugens.c but beos needs real work

vlazzarini:
how come this passed Travis? We need to fix this otherwise CI is not really doing anything for us.

jpffitch:
I was going to ask the same.

The code is wrong wrt initialisation of seed at least.  Fixed a number of 
them

jpffitch:
Another issue.  This code defines an opcode called tabslice which is also defined in arrays.c
Have not managed to determine what the differences are but this does need to be resolved.  Also the array forms of bitwise or and bitwise and really belong in arrays.c together with the array|constant etc.

jpffitch:
The array forms of bitwise or and bitwise and really belong in arrays.c together with the array|constant etc.

vlazzarini:
is tabslice an overload?

gesellkammer:
Yes, basically that is correct. I just pushed a fix to the last PR with 
this and some extra bound checking: 
https://github.com/csound/csound/pull/1037

PR #1041, opened 2018-10-06
Update to Bela API 1.4

giuliomoro:
Blind patch to match new Bela API https://github.com/BelaPlatform/Bela/commit/f057db414bc191328c48be4ff04af0947ece5253

Untested.


vlazzarini:
looks good, will merge.

PR #1049, opened 2018-10-14
added strtoarr and strstrip opcode

rorywalsh:
Added a native version of one of Joachim's UDOs and a strstrip opcode that will remove instances of a substring from a larger string. I had to add "make_plugin(strtoarr Opcodes/strtoarr.cpp)" to CMakeLists.txt, not sure this is the preferred way of adding plugin opcodes to the build.  

Comments welcome.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1463,3 +1463,5 @@ endif(DOXYGEN_FOUND)
 # install CMake module
 install(FILES ${CMAKE_SOURCE_DIR}/cmake/Modules/FindCsound.cmake
         DESTINATION "${CMAKE_INSTALL_PREFIX}/share/cmake/Csound")
+
+make_plugin(strtoarr Opcodes/strtoarr.cpp)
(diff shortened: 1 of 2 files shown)


kunstmusik:
I would rather see this as a separate plugin from Csound main repo and we
put a freeze on any new opcode libs for Csound 6.  I also wonder about the
names: there are common names for "splitting" strings and replacing text
(replace with empty string) and it'd be a bit rushed to commit to these
opcode names and features.

rorywalsh:
I'm open to suggestions regarding the names. strtoarr could be strtoarray and strstrip could be strremove. I don't mind putting them into a separate repo, but these are light weight additions and don't add any  dependencies.

hlolli:
strstrip, simply substr with indicies, to try to be as parallel to other programming language paradigms?

hlolli:
sorry nevermind, this string remove does something else, as there's already such mechanism for substring implemented.

hlolli:
Unrelated: An opcode for file2string would be a welcome addition (as opposed to read just 1 line at a time which is all there is afaik).

rorywalsh:
My attempt to slip these into core Csound has obviously failed :sob: So I'll set up a new repo for them with CL so other can push to it and have their opcodes built too. I agree that a file2string would be cool. I can add one.

PR #1091, opened 2018-12-16
FLTK widget colour updates during performance

1bpm:
At the moment FLsetColor, FLsetColor2 and FLsetTextColor only work at initialisation. These additions allow widget colours to be changed after the panel has been created/during performance by calling the relevant redraw function.


1bpm:
Great thank you

PR #1092, opened 2018-12-17
Fixed FLjoy x/y handle issue with FLsetVal

1bpm:
Using FLsetVal with FLjoy, setting the y value with ihandley doesn't appear to have any effect. Setting x value with ihandlex affects both axes of the FLjoy simultaneously. This let the handles work independently and properly with FLsetVal.

### DIFF
--- a/InOut/widgets.cpp
+++ b/InOut/widgets.cpp
@@ -3022,15 +3022,11 @@ static void fl_setWidgetValue_(CSOUND *csound,
       break;
     case 3:                                     // FLjoy
       {
-        static int  flag = 0;
-        // FLsetVal always requires two adjacent calls when setting FLjoy
-        if (!flag) {
+        if (v.joy == JOY_X) {
           ((Fl_Positioner *) o)->xvalue(val);
-          flag = 1;
         }
         else {
           ((Fl_Positioner *) o)->yvalue(val);
-          flag = 0;
         }
       }
       break;


jpffitch:
Should it not be possible to set both at once?

⁣Sent from TypeApp ​

1bpm:
Yes, at the moment though flsetval on the x handle changes both x and y to the same value. I have just found this commit will probably break flsetval_i on fljoys though, so will close and review in more detail.

1bpm:
OK this doesn't actually break flsetval_i. I think either documentation or code needs changing - the proper usage of setval with fljoy is only documented on this page: http://www.csounds.com/maldonado/Manual/WidgetOpcodes.html - does Gabriel still maintain/contribute?

Relevant bits from that page that aren't in the main docs:
- when using FLsetVal_i to set the values of FLjoy, two adjacent instances of FLsetVal_i are always required, the first matching the X value and the second the Y value of FLjoy.
- it is not possible to use FLsetVal on FLjoy.

This commit does change both of those - FLsetVal_i can be used in any order, and FLsetVal can be used on FLjoy. It wouldn't affect any old documents using FLsetVal_i as long as the guidance on that page was followed.

Here are some tests I did with the current and altered approaches, note comments in score:

```<CsoundSynthesizer>
<CsOptions>
-odac
</CsOptions>
<CsInstruments>
sr = 44100
kr = 4410
nchnls = 2

FLpanel "test", 500, 500
	gkjx, gkjy, gijx, gijy FLjoy "Joy", 0, 1, 0, 1, 0, 0, -1, -1, 400, 400, 50, 50
FLpanelEnd
FLrun


instr 1 ; k rate set one axis with line
	k1 line 0, p3, 0.5
	FLsetVal 1, abs(k1), gijx
endin


instr 2 ; init set val
	FLsetVal_i p4, gijx
	FLsetVal_i p5, gijy
endin


instr 3	; as instr 2 but order of calls inverted
	FLsetVal_i p5, gijy
	FLsetVal_i p4, gijx
endin


instr 4 ; init set on y
	FLsetVal_i p4, gijy
endin


</CsInstruments>
<CsScore>
f1 0 16384 10 1


/* test 1: FLsetVal ; OK - both function the same */
; original: X increments from 0.1 to 0.5; Y decrements from 0.5 to 0.1
; new: X increments from 0.1 to 0.5; Y decrements from 0.5 to 0.1
i2 0 1 0.1 0.5
i2 1 1 0.2 0.4
i2 2 1 0.3 0.3
i2 3 1 0.4 0.2
i2 4 1 0.5 0.1

/* test 2: FLsetVal ; Not OK - original has axes muddled */
; original: X decrements from 0.5 to 0.1; Y increments from 0.1 to 0.5
; new: X increments from 0.1 to 0.5; Y decrements from 0.5 to 0.1
i3 5 1 0.1 0.5
i3 6 1 0.2 0.4
i3 7 1 0.3 0.3
i3 8 1 0.4 0.2
i3 9 1 0.5 0.1


/* test 3: FLsetVal_i ; Not OK - only sets Y handle but original takes repeated calls to FLsetVal_i as alternate X/Y */
i4 10 1 0.9 ; original: X set to 0.9 ; new: Y set to 0.9
i4 11 1 0.1 ; new: Y set to 0.1 ; new: Y set to 0.1


/* test 4: FLsetVal ; Not OK - original moves both axes when FLsetVal is only used on one axis */
; original: X and Y move from 0 to 0.5
; new: X only moves from 0 to 0.5
i1 15 15 


</CsScore>
</CsoundSynthesizer>

```

PR #1106, opened 2019-01-26
Fixes for #1046 (fluidsynth 2 integration)

dvzrv:
Opcodes/fluidOpcodes/fluidOpcodes.cpp: Removing extra { introduced in #1046, when extending for fluidsynth >= v2. Minor style fixes.


kunstmusik:
Not sure why Appveyor isn't shown here but it did succeed over there. Will merge now.

PR #1110, opened 2019-02-03
perfThread: wait the recording thread before waiting for the performance thread

### LINKED ISSUE #1103: testPerfThread hangs on mips
Build failure was observed on debian buildds: https://buildd.debian.org/status/fetch.php?pkg=csound&arch=mips&ver=1%3A6.12.2~dfsg-2&stamp=1545519725&raw=0

There is apparently a deadlock in which the main thread is waiting on the perf thread, and the perf thread is waiting on the recording thread, but apparently nobody told the recording thread to stop.

The following diff (which was applied as a workaround in the debian package) appears to fix the issue, but I'm not sure if this is the correct fix. It looks a bit better to me anyway: the recording thread should not survive the perf thread, or should it?

```diff
diff --git a/interfaces/csPerfThread.cpp b/interfaces/csPerfThread.cpp
index 842201a02..f9e854583 100644
--- a/interfaces/csPerfThread.cpp
+++ b/interfaces/csPerfThread.cpp
@@ -676,15 +676,15 @@ int CsoundPerformanceThread::Join()
     int retval;
     retval = status;
 
-    if (perfThread) {
-      retval = csoundJoinThread(perfThread);
-      perfThread = (void*) 0;
-    }
     if (recordData.running) {
         recordData.running = false;
         csoundCondSignal(recordData.condvar);
         csoundJoinThread(recordData.thread);
     }
+    if (perfThread) {
+      retval = csoundJoinThread(perfThread);
+      perfThread = (void*) 0;
+    }
 
     // delete any pending messages
     {
```

fsateler:
If we don't tell the recording thread to stop, we might enter a deadlock as the perf thread waits for the record
thread but it has not been stopped yet.

Fixes #1103

### DIFF
--- a/interfaces/csPerfThread.cpp
+++ b/interfaces/csPerfThread.cpp
@@ -676,14 +676,14 @@ int CsoundPerformanceThread::Join()
     int retval;
     retval = status;
 
-    if (perfThread) {
-      retval = csoundJoinThread(perfThread);
-      perfThread = (void*) 0;
-    }
     if (recordData.running) {
         recordData.running = false;
         csoundCondSignal(recordData.condvar);
         csoundJoinThread(recordData.thread);
+    }
+    if (perfThread) {
+      retval = csoundJoinThread(perfThread);
+      perfThread = (void*) 0;
     }
 
     // delete any pending messages


vlazzarini:
thx

PR #1118, opened 2019-03-06
Fix build with Altivec

pkubaj:


vlazzarini:
just a question, when did you clone the repo for this PR?
Just checking that it will not contain garbage commits which we had to deal last year before I merge.

I can verify it myself, but it might be quicker if you let me know.

Thanks for the PR

pkubaj:
I tested it against the version that is in FreeBSD ports tree (commit df2edd269).

PR #1123, opened 2019-03-26
Add Haiku compatibility

PeteCA:
These additions allow Csound to run under Haiku, with  realtime audio out and MIDI in.
(No audio in or MIDI out currently)

All additions have been protected with ifdefs, so other platforms should not be affected.


Frontends/CMakeLists.txt
```diff
@@ -51,6 +51,8 @@ if(BUILD_CSBEATS)
         set_source_files_properties(${FLEX_csbeats_SCANNER_OUTPUTS} PROPERTIES
         COMPILE_FLAGS -DYY_NO_UNISTD_H)
       make_executable(csbeats "${csbeats_SRCS}" "")
+    elseif(HAIKU)
+      make_executable(csbeats "${csbeats_SRCS}" "")
```
fsateler: We've seen this pattern a few times. I wonder if we shouldn't instead in the main CMakeLists.txt do this:

```cmake
# Some platforms have folded libm into libc so no need to link them
if (HAIKU OR MSVC) 
  set(MATH_LIBRARY "")
else()
  set(MATH_LIBRARY "m")
endif()
```

And then later use that instead of adding a large amount of ifs for each target that requires math functions. We even have a special check in `make_plugin` (ugh)

jpffitch: I agree with Felipe; keeping cmake simple is a worthy end.

kunstmusik:
+1

PeteCA:
Yes, please, if you can simplify things.  I did it that way because I have no experience with cmake, and that was the most obvious way of making the necessary fix.

PeteCA:
I could fix it here if someone explicitly lays out the preferred code. (I have the same unfamiliarlty with cmake.)

I also haven't ever had to  update a PR before.  If I push the revision. does the PR get updated automatically?

fsateler:
> Should we take the PR as it is and try to fix it in the various CMakeLists.txt files afterward?
I am not ure I know cmake well enough

I'll try to prep a patch.

fsateler:
Patch at #1126 . Feel free to merge and I'll update that.

PeteCA:
As #1126 has apparently passed, I suppose I should just wait until that is merged and add it to this?
Trouble is, I'm not exactly sure how to do that... Do I just pull from master to my local repo, then push
again with the changes?  (Sorry -- never did this before.)

fsateler:
I have rebased #1126  on top of this branch. @PeteCA please check if my branch also builds on Haiku so that we don't regress.

This branch can be merged from my POV

PeteCA:
Sorry for my ignorance (haven't been using Git(Hub) all that much yet), but how exactly do I get the code from your branch?  This sounds like what I wanted to do but couldn't find out how!
(You seem to have picked up my additions just fine!)

PeteCA:
OK, sorry.  Never mind...  I eventually [after about half-an-hour's trial and error] figured out the correct incantations to get your code. Everything seems to be just as it should.  So, as I understand it, merging your PR is all that's needed now?

kunstmusik:
@fsateler and @PeteCA I went ahead and merged this, but then realized there were two file changes that I didn't see until after the merge. One removed iOS/libsndfile.a (which, I'm not sure why that was in the repo). The other modified the iOS/release.sh script version number.  

@vlazzarini : I haven't looked at iOS in a while. Do we need to restore both of those changes?  Looks like the libsndfile.a was added by you in on March 18th in commit d7754a9e.

PeteCA:
Oops.. I think my source was 'develop' of March 17th, so
I guess the addition was after that.  Do what you need to.

vlazzarini:
yes, I put the libsndfile.a there because that's the new one with the extra support that was compiled for us. The other change is just the package version. I'd say we can easily put these back.

PR #1126, opened 2019-03-27
cmake: Generalize usage of math library

fsateler:
Instead of filling all targets with `if(MSVC)` directives, use
a variable to store the needed math library.

I don't have access to a windows developer environment, so let's see if appveyor passes.


fsateler:
Hmm, why is there no appveyor?

stekyne:
Not sure why it didn't report that but it did fail.
https://ci.appveyor.com/project/csound/csound/builds/23411773

Looks like there is a cmake error from this:
CMake Error at Frontends/CMakeLists.txt:51 (make_executable):
  make_executable Function invoked with incorrect arguments for function
  named: make_executable
CMake Error at Frontends/CMakeLists.txt:53 (target_include_directories):
  Cannot specify include directories for target "csbeats" which is not built
  by this project.
CMake Error at Frontends/CMakeLists.txt:54 (target_include_directories):
  Cannot specify include directories for target "csbeats" which is not built
  by this project.

fsateler:
Thanks. looks like a missing quote. Lets see if new version passes.

PR #1129, opened 2019-03-31
Tabensure fix

gesellkammer:
This PR fixes the wrong usage of tabensure at init time. It only initializes an array if it has not been initialized before, and defers any resizing to perf time. WIth this scheme, any init action on arrays does not collide with the expected size at perf-time.

The logic behind it is: 
* at init time, if out array has not been initialized / created, create it and make ensure the given size
* at perf time, check size and ensure it is desired size. Don't check if it has been initialized, since this is checked at i-time. 

## TODO 

* We should probably check dimensions also

## Questions 

* How to extend this easily to multidimensional arrays?


gesellkammer:
Should I remove this PR based on the discussion around csound's de facto semantics of not supporting perf time array resizing?

kunstmusik:
This change is limited to the opcodes in emugens.c and beosc.c; doesn't this change to need to be system-wide to be viable?

gesellkammer:
> doesn't this change to need to be system-wide to be viable?

These changes make the opcodes I introduced to behave correctly. Of course it would be of great benefit to extend such a scheme to the rest. At least it should be documented that any reuse of an array as an l-value results in surprising behaviour.

PR #1134, opened 2019-04-07
Modified Hash Table implementation to expand on load

kunstmusik:
Modified hash table to expand hash table buckets based on load and load factor. Performance compared to prior implementation should be about equal on small loads and much improved on higher loads.


stekyne:
Looks ok to me!

jpffitch:
Any reason not to merge?

PR #1139, opened 2019-04-15
Fix plugin search path for multiple csound instances

PeteCA:
Unlike other OSs, Haiku has to allow for several possible default plugin directories, so for that csmodule.c has a static string of the possibilities, separated as usual by colon characters.  This string is split to extract the individual directories, replacing the colon with a null. This meant that after the first search,
only the first name would be found.  The simple fix is to restore the separator after the string has been copied.  This should not affect any other systems that never have more than one default.

(I was also going to fix the tabs and long lines from my last patch, but they seem to have magically disappeared!  I guess that was John?  Many thanks for that.)


kunstmusik:
@PeteCA Is this PR still valid?
@vlazzarini I haven't tested, but this change is small and looks okay to me. Shall we merge?

PeteCA:
On Fri, May 17, 2019 at 06:20:14AM -0700, Steven Yi wrote:
> @PeteCA Is this PR still valid?
Yes.  It's needed for Haiku, because I found it necessary to provide several
possible directories to search for plugins (because the Haiku package system
could put them in different places).  Because the original code replaces the
separators with nulls, this only works once.  If you start another CSOUND
instance, only the first directory is looked at.

kunstmusik:
Thanks @PeteCA, I'm pulling the trigger to merge now.

PeteCA:
Thanks, Steven.  Now everything's up-to-date (at my end).

vlazzarini:
ok from me (late). I'll fix anything that needs fixing.

PR #1149, opened 2019-06-05
increase the maximum name length for the jack-client name

hlolli:
32 characters is bit short amount for my use-case (I'm creating jack clients based on java classpath names). 256 should be a sensible maximum.

### DIFF
--- a/H/cs_jack.h
+++ b/H/cs_jack.h
@@ -21,7 +21,7 @@
     02110-1301 USA
 */
 
-#define MAX_NAME_LEN    32      /* for client and port name */
+#define MAX_NAME_LEN    256      /* for client and port name */
 
 typedef struct RtJackBuffer_ {
 #ifdef LINUX


hlolli:
Maybe this is nonesense to put it all the way up to 256, on linux I'm defenitely able to exceed 32 characters limit when compiling with a greater value. But I'm unable to do the same with mac, I compiled directly from git and I'm unable to exceed 32 char limit (so it's a jack2 side) But reading this:
 https://github.com/jackaudio/jack2/blob/4828349f3dac9dd84d743011489b24999bb0705f/common/JackConstants.h#L35 

seems that jack2 has hard 64 char limit on client_name. So I need to research a bit...

hlolli:
meh... ok I'll close this PR, even if jack2 would allow longer names, it would be released sometime in the future. I guess I'll need to trim my names down...

Crossing fingers that they'll extend it https://github.com/jackaudio/jack2/issues/474

PR #1154, opened 2019-07-05
Bugfixes to emugens, adopts changes made to arrays

gesellkammer:
This PR fixes some bugs in emugens:

* ntom was not reading to the end of the string
* beadsynt now works with i arrays
* tabrowlin and getrowlin would calculate wrong size under certain conditions

All array operations now use tabinit and tabcheck as defined in include/arrays.h, following the new array semantics where no new allocations are performed on perf time.


jpffitch:
Can you suggest the change needed in the manual for 
"beadsynt now works with i arrays" please?  Not sure what it means.

gesellkammer:
beadsynt takes three arrays as inputs and these were hardcoded to be 
k-arrays, now they can be either i or k. The manual is correct, the 
implementation was wrong.

for release notes, if you want to include something, it could be 
"beadsynt corrected to work with both k- and i- arrays, as indicated in 
the manual"

jpffitch:
Thanks

PR #1156, opened 2019-07-11
Use the overloaded Scope::log(float*)

giuliomoro:
as this is a) more efficient b) less unlikely to be discontinued in the future.

This assume that `schannel.samples[frmcount]` is a `float`, and so we take its address instead.


vlazzarini:
can't assume MYFLT is a flot

PR #1160, opened 2019-07-23
support string pfields with pindex opcode

1bpm:
Adds an additional pindex opcode for string p fields. Changes existing pindex to raise InitError if the p field is invalid.

### DIFF
--- a/Opcodes/nlfilt.c
+++ b/Opcodes/nlfilt.c
@@ -41,6 +41,12 @@ typedef struct {
 
 typedef struct {
         OPDS    h;
+        STRINGDAT   *ians;
+        MYFLT   *index;
+} PFIELDSTR;
+
+typedef struct {
+        OPDS    h;
         MYFLT   *inits[24];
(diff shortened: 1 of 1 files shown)


jpffitch:
Can you convince me that the assignment of the cs_strdup does not leak memory from a previous value?
Not sure myself but I suspect a Free call is needed.

1bpm:
I do not think I would be able to convince you at the moment - if you have a suspicion would be inclined to agree. Just to clarify something like the following before the cs_strdup ? 
if (p->ians->data!=NULL) csound->Free(csound, p->ians->data);

jpffitch:
That is what I was thinking was needed.

⁣Sent from TypeApp ​

1bpm:
Just got round to adding the line. The tests I have tried seemed OK.

1bpm:
Another push as I noticed I did not wrap the error strings in Str().

jpffitch:
Thank you

PR #1166, opened 2019-08-05
Fix build of libstdutil an Android

sandroandrade:
libstdutil doesn't build for Android because of a wrong (moved?) place for sdif.c.


kunstmusik:
Thanks for the contribution!

PR #1188, opened 2019-09-20
cmake fix so it is aware of pthreads and atomic-builtins in Haiku

PeteCA:
I found that the UDP server didn't work in Haiku, simply because cmake didn't find pthreads (and atomic-builtins, which were later found to be needed).  Haiku has no pthreads library because they are incorporated in the root library, so CMakeLists.txt has been amended for the Haiku build.

Atomic-builtins were not being seen because the test compile failed.  The test source uses 'int32_t' or 'int64_t' -- defined in 'stdint.h'.  For some reason this include is not required in Linux, so the test compiles, but a Haiku compile fails without it.  I've added it to the source in 'cmake/CheckAtomic.cmake'.  This should not affect builds under other OSs, as that file should always be present.  I've checked that it's fine under Linux, but have no way of testing other systems.


jpffitch:
Looks OK

PeteCA:
Thanks for merging!

PR #1191, opened 2019-10-03
adjustment for change in Haiku headers

PeteCA:
Haiku have modified their chain of network-related headers,  which prevented Csound from compiling in recent Haiku versions.  This patch corrects for that.  It won't affect other systems. (Not sure why it won't merge.)

BTW, there is confusion in other modules between e.g. "int32" and "int32_t" which prevents a simpler version of this patch from compiling.  I'll try to get an issue raised.


PeteCA:
I thought the conflict was my mismatched sfont.c, but I think I fixed that, and it still seems to have failed.  Not sure how to resolve it.

Begasus:
> I thought the conflict was my mismatched sfont.c, but I think I fixed that, and it still seems to have failed. Not sure how to resolve it.

sfont.c doesn't give a conflict, not sure if the change is needed, but it doesn't brake the build

PeteCA:
I apologize for that.  What happened is that I found a slight bug in sfont.c, but wasn't sure of the proper fix.  I made a private temporary fix in the source for myself, but left it to John to do a proper one, which he did.  Unfortunately I forgot about my private change that was still in my PR, and it conflicted with the proper one.  I tried to update my code to match, but it seems I did something wrong.  Please ignore my version of that file.

PeteCA:
I've added the needed update to OSC.c, which I missed before because liblo was not in my test system. I think that's the lot!

PeteCA:
Sorry it took me so long to fix the conflict!  Seemed to get distracted a lot...

Begasus:
Same here works ok on Haiku (also made some local changes for the installation part in CMakeList.txt) :)

PR #1151, opened 2019-06-10
small fixes for rtjack compilation on win32

hlolli:
After long battle with jack2 on windows, I was able to make it run, and to be able to compile it, I needed to make minor changes to cmake and rtjack.c. But it wont run from the official jack2 installation, since it hasn't been updated since 2016. If someone really wants to use jack2 with windows and csound, I would recommend extracting the pre-built jack2 from http://users.notam02.no/~kjetism/radium/demos/windows64/?C=M;O=D there they distribute newer binaries with the same ABI. It's just a question of time when the jack developers make a new windows release and the errors from their current outdated release are fixed.

Issue #926, opened 2018-02-08
VCPKG-MSVC: Replace manual builds with more VCPKGs


stekyne:
Portaudio - 
This package wasn't originally used because it does not contain ASIO as a feature. However, we can manually extract the ASIO SDK into VCPKG and it will build with ASIO support.

Portmidi - 
This package has been added in the last 2 months so we can remove the manual build.

Fluidsynth - 
Currently in the processing of adding it myself to VCPKG. Should be done soon.

Liblo - 
Needs to be added to VCPKG but should be straightforward. 

HDF5 - 
This has been fixed in VCPKG. It (VCPKG) now has feature-packages. This remove the MSPI dependency so we can install the vanilla version without issue.

This will remove a lot of manual builds from the downloadDependency.ps1 script. It also could lead to us adding libcsound as a VCPKG itself. It will automatically pull down the dependencies within VCPKG then. 

There are a few other files that we download so I have to figure out what to do with them. Maybe add them as submodules or something like that.

kunstmusik:
Awesome, thanks so much for looking into this!

stekyne:
@kunstmusik I've nearly got this going. Just with libsndfile, you switched the build to static to get around some issue. Can you remember what it was and what was the test case? Just to ensure I've not re-introduced it.

kunstmusik:
Yes, see these two threads:

http://csound.1045644.n5.nabble.com/Csnd-dev-Windows-Problem-again-with-libsndfile-critical-td5759488.html
http://csound.1045644.n5.nabble.com/Csnd-dev-libsndfile-and-MSVC-build-td5757593.html

stekyne:
Just to update. This is done from Csound's end but there is one issue left in VCPKG. Should reduce the build time and clean it up in general. There will be 13 packages in VCPKG that we make of us of then. It can also be used for other platforms (Mac/Linux) if a full build from source is required.

stekyne:
Will be fixed by #1208. Closing now

PR #1192, opened 2019-10-15
dont try to find libsndfile includes on msvc

hlolli:
With this change, cmake works directly with the source-tree using MSVC. This is my command which I used to test #1151 there was also some doubt about this written in the comments, so I hope the new comments clarify this TODO/question.

```
cmake ..\ -G "Visual Studio 15 2017 Win64" -DUSE_DOUBLE=1 -DUSE_JACK=1 -DBUILD_JACK_OPCODES=0 -DCMAKE_BUILD_TYPE="Rel" -DCMAKE_INSTALL_PREFIX="dist" -DJACK_LIB="C:\Users\User\Downloads\radium_64bit_windows-5.9.65-demo\radium_64bit_windows-5.9.65-demo\bin\jack_local\libjack64.lib" -DJACK_LIBRARY="C:\Users\User\Downloads\radium_64bit_windows-5.9.65-demo\radium_64bit_windows-5.9.65-demo\bin\jack_local\libjack64.lib" -DJACK_INCLUDE_PATH="C:\Users\User\Documents\jack2\common" -DLIBSNDFILE_LIBRARY="C:\Program Files\Mega-Nerd\libsndfile\lib\libsndfile-1.lib" -DSNDFILE_H_PATH="C:\Program Files\Mega-Nerd\libsndfile\include" -DBISON_EXECUTABLE="..\..\..\Downloads\win_flex_bison-latest\win_bison.exe" -DFLEX_EXECUTABLE="C:\Users\User\Downloads\win_flex_bison-latest\win_flex.exe"
cmake --build .
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -549,8 +549,8 @@ find_package(SWIG)
 find_package(VSTSDK2X)
 
 
-# TODO should this work for MSVC also?
-if(WIN32)
+# LIBSNDFILE_LIBRARY and SNDFILE_H_PATH flags are enough for MSVC
+if(WIN32 AND NOT MSVC)
   find_library(FLAC_LIB NAMES FLAC flac)
   find_library(OGG_LIB ogg)
   find_library(SPEEX_LIB speex)


stekyne:
I don't really understand your first statement there. Works directly with source tree? 

I wrote that comment when doing the MSVC initial port. At the time I wasn't sure if libsndfile needed those other dependent libraries. I think it probably does need them to be consistent across each platform. 

For #926, I'm going to be using libsndfile's VCPKG which does seem to include some of these dependencies (https://github.com/microsoft/vcpkg/blob/3cf69f964d1fd374acdb31fb30070059e3f8954e/ports/libsndfile/CONTROL#L9). 

Also we seem to download these dependencies manually with VCPKG as well (https://github.com/csound/csound/blob/6b7da4b7127b6273a787cf05037818d13d2488d7/msvc/downloadDependencies.ps1#L96). I'm not sure if they're needed by anything else but libsndfile, if they're not, then that's probably a mistake that can be removed.

If this helps you in someway then go ahead with merging it and I can change it later if needs be. It seems to be failing the build though.

hlolli:
The ogg and flac sources are all in libsndfile https://github.com/erikd/libsndfile/tree/master/src. We could rethink this part but that would be more up to you since I don't use windows from day to day :). I was under the impression that Csound was built with cygwin, and that it was done from a third-party Cmake files.

No stress on this PR. It could well be that some package managers provide stripped down version of libsndfile that doesn't include for example libogg and libFLAC, that would mean a breaking change. But from my personal experience, I couldn't get libogg and libflac to behave well with Windows.

stekyne:
Those files seem to be just interfaces to those libraries for libsndfile. Their main readme still says to download the dependencies via apt-get etc. Csound is built on Appveyor with MSVC these days. 

Libsndfile can be built without those dependencies anyway. It's usually best to search for them with Cmake and if that fails, then either let it build without them or fail if they're required. 

Tbh I think a lot of the Cmake files need a bit of refactoring and cleaning up anyway.

hlolli:
I think what explains it here in my case, is that I install libsndfile and
link against the pre-compiled library files. I'll maybe have a look and try
to get these dependencies to fail more gracefully. But I'd go for
recommending users wanting to install Csound to download the official
installer here http://www.mega-nerd.com/libsndfile/ and it should save some
time and headache.

kunstmusik:
@hlolli Csound on Windows is built with MSVC. Also, libsndfile sources have interfaces to connect with FLAC and ogg but does not include sources. 

I remember that we used to do static linking with libsndfile with MSYS2 build  and therefore needed the dependent libraries. I don't remember what happened with linking with MSVC but I suppose this is related.

hlolli:
I guess for static linking we need all the source files.  There's defenitely pro and contra here. Installing libsndfile from installer means less dependencies but we loose the ability to statically link. Maybe we should set the cmake file to only need the flac/oggvorbis sources when we are statically linking?

PR #1208, opened 2019-10-30
MSVC build changes

stekyne:
Further attempts at cleaning up the MSVC build and improving build times.

- [x] Remove libsndfile manual build and add vcpkg
- [x] Remove liblo manual build and add vcpkg
- [x] Remove portaudio manual build and add vcpkg
- [x] Remove portmidi manual build and add vcpkg
- [x] Remove HDF5 manual build and add vcpkg
- [x] Remove dirent file from Csound and add vcpkg version
- [x] Use chocolatey for other binary executables. Swig, flex, bison.
- [x] Update project to Visual Studio 2019 as well Appveyor image
- [x] Update VC runtime redist details on appveyor
- [x] Need to manually download and extract ASIOSDK for portaudio vcpkg
- [x] Update documentation for VS build
- [x] Remove STK manual build and add VCPKG
- [x] Updated Windows installer for changes

Lot of stuff is done and working. There were some problems making it all work cross platform but seems to have settled now. Main thing is to abstract out every dependency in Cmake to a "find_package()" request. I've added a few Find_*.cmake modules to the repo and they're being used to find these dependencies cross platform and regardless of how they are installed on the system (apt-get, homebrew or vcpkg or manually installed to default dirs). 

That's the model I'd like to continue and it will be less of an issue when we remove some opcodes from the repo.

The PR is not quite ready yet but I'll open it up for people to have a look and point out any issues.


kunstmusik:
This all looks like a very nice step forward to me. Only issue that stuck out to me is the change in checking for if the build option was on first, then checking for deps.  I think in the past we've checked for deps, then checked for the build option.  I think it makes sense to switch to the former. 

Also, reporting of build options results (i.e. "BUILDING xxx" or "NOT BUILDING xxx") is something we could standardize and have a single report at end of configuration (similiar to autotools).  But that'd be for another time I think. 

I'd say whenever you're ready, we can double check on macos and linux locally then merge.

stekyne:
I was thinking about the first issue you mentioned while doing this work but couldn't decide the right behaviour. If the user doesn't want a feature enabled, I think we  should avoid looking for the dependencies to speed up the process. Likewise we shouldn't be looking for dependencies that don't exist on a particular platform. Which is mostly the audio interface stuff. 

I'd also like the custom macros to check dependencies to be removed and use the idiomatic CMake ways of doing this.  There is also the problem of whether the build should fail if a dependency isn't found but the user has selected that feature to be enabled. Right now, it will just not build that feature but I think that's too loose. I know there is a custom flag to fail if not found but it doesn't always work as expected I've found.

The building report sounds good to me. 

I will start a new ticket after this with a list of CMake changes to implement and we can discuss it further.

stekyne:
This is ready for OSX testing if anyone could help out. As far as I can tell, Linux/TravisCI is working fine with the changes.

kunstmusik:
I pulled the branch and did a quick ```mkdir build; cd build; cmake ..; make -j6" and--after disabling websocket opcodes, a known issue on mac--it built. I don't have a number of 3rd party libs installed here for a number of the opcode libraries, so I can't comment on if everything is building correctly, but I'd be in favor of merging and debugging any issues back in develop (assuming the linux and windows builds complete).

stekyne:
Yeah no problem. It rebased fairly easy so I should be able to keep it up to date if we want to take some more time. I'll wait for the CI to finish at least and go from there.

stekyne:
Merged it in. I'll open a ticket for any issues found from now.

PR #1209, opened 2019-10-31
Feature/git actions workflow

ShantanuJamble:


stekyne:
Looks good. Maybe there doesn't need to be so much spaces between commands? Might make it a bit easier to read if it's more compact.

ShantanuJamble:
> Looks good. Maybe there doesn't need to be so much spaces between commands? Might make it a bit easier to read if it's more compact.

@stekyne I apologize this was created by Github actions automatically when I was testing it. That's the reason I closed it.

stekyne:
Ah ok sorry, I jumped the gun!

PR #1221, opened 2019-11-15
Added more array versions to bpfcos

gesellkammer:
This PR adds more array versions to bpfcos

In the second case, it assumes that there is a curve defined by a series of points multidimensional points (kx, ky, kz) and interpolates ky and kz based on a given kx.

All versions work with both k- and i- arguments.

```csound

ky  bpfcos  kx, kxs[], kys[]
ky, kx  bpfcos  kx, kxs[], kys[], kzs[]

```


jpffitch:
Manual page changes?

PR #1228, opened 2019-11-19
Use Google Test as the unit test framework

stekyne:
- [x] Added vcpkg for Windows
- [x] Update Cmake files
- [x] Update unit tests to use GTest API
- [x] Update GHActions to include gtest for linux
- [x] Update GHActions to include gtest for osx

Initial PR to see changes and any comments. There are a few failing tests but I think it's down to the tests themselves or differences between Linux/OSX and MSVC. I'll comment on that below.


tests/c/csound_orc_semantics_test.cpp
```diff
@@ -0,0 +1,242 @@
...
+    
+    entries = find_opcode2(csound, "pcauchy");
+    opc = resolve_opcode(csound, entries, "i", "k");
+    ASSERT_TRUE (opc->iopadr != NULL);
+    
+    opc = resolve_opcode(csound, entries, "k", "k");
+    ASSERT_TRUE (opc->kopadr != NULL);
+    
+    opc = resolve_opcode(csound, entries, "a", "k");
+    ASSERT_TRUE (opc->aopadr != NULL);
```
stekyne: @kunstmusik This test fails on windows at this line 95. The other i and k value ones work. Not clear to me why aopadr isn't populated. Any ideas?
```
G:\Programming\GitHub\csound\tests\c\csound_orc_semantics_test.cpp(95): error: Value of: opc->aopadr != NULL
  Actual: false
Expected: true
```

tests/c/csound_orc_semantics_test.cpp
```diff
@@ -0,0 +1,242 @@
...
+    
+    csound->Free(csound, entries);
+}
+
+TEST_F (OrcSemanticsTest, FindOpcodeNewTest)
+{
+    ASSERT_TRUE (find_opcode_new(csound, "##error", "i", "i") != NULL);
+    ASSERT_TRUE (find_opcode_new(csound, "##error", NULL, "i") == NULL);
+    ASSERT_TRUE (find_opcode_new(csound, "##xin64", "i", NULL) != NULL);
```
stekyne: This assertion fails as well. Can't find xin64
```
G:\Programming\GitHub\csound\tests\c\csound_orc_semantics_test.cpp(104): error: Value of: find_opcode_new(csound, "##xin64", "i", NULL) != NULL
  Actual: false
Expected: true
```

tests/c/csound_orc_compile_test.cpp
```diff
@@ -0,0 +1,170 @@
...
+{
+    char* instrument =
+        "instr 1 \n"
+        "k1 expon p4, p3, p4*0.001 \n"
+        "a1 randi  k1, p5   \n"
+        "out  a1   \n"
+        "endin \n";
+
+    TREE *tree = csoundParseOrc(csound, instrument);
+    ASSERT_EQ (tree->next->line, 0);
```
stekyne: It's returning 1 instead of 0. Maybe the test can be updated or is it expected to be 0?
```
G:\Programming\GitHub\csound\tests\c\csound_orc_compile_test.cpp(169): error: Expected equality of these values:
  tree->next->line
    Which is: 1
  0
```

tests/c/io_test.cpp
```diff
@@ -125,199 +124,117 @@ void test_keyboard_io(void)
...
     ret = csoundStart(csound);
-    CU_ASSERT(ret == CSOUND_SUCCESS);
+    ASSERT_TRUE (ret == CSOUND_SUCCESS);
     ret = csoundPerformKsmps(csound);
-    CU_ASSERT(ret == CSOUND_SUCCESS);
+    ASSERT_TRUE (ret == CSOUND_SUCCESS);
     MYFLT val = csoundGetControlChannel(csound, "key", &err);
-//    CU_ASSERT(err == CSOUND_SUCCESS);
-    CU_ASSERT_DOUBLE_EQUAL(val, 100.0, 0.001);
+    ASSERT_DOUBLE_EQ (val, 100.0);
```
stekyne: Csound method returns -1 instead of 100. Seems like the API is wrong here.
```
G:\Programming\GitHub\csound\tests\c\io_test.cpp(135): error: Expected equality of these values:
  val
    Which is: -1
  100.0
    Which is: 100
```

tests/c/csound_orc_semantics_test.cpp
kunstmusik: check out commit 194ad034519296a1472ec4ad773ccc8f208c9d78 looks like xin64 was removed.

tests/c/csound_orc_semantics_test.cpp
kunstmusik: I think this test needs updating as I believe @jpffitch had done work to remove use of kopadr and aopadr and there is just init-time and perf-time functions to think about.

tests/c/csound_orc_compile_test.cpp
kunstmusik: Unsure, as line number reporting has changed over the years.  @jpffitch are line numbers starting at 0 or 1 with the parser now?

tests/c/csound_orc_semantics_test.cpp
stekyne: Is it removed or renamed xin.64? Can just delete that assertion I guess.

stekyne:
@kunstmusik you mentioned someone might take a look at this? did that ever happen? If not, I can maybe fix it up and deliver the framework + commented out failing tests. At least we can just fix the tests later.

kunstmusik:
My student had looked but we stopped as the tests that needed fixing required domain knowledge of Csound and I didn't have time to go through it. I'd say please go ahead with it and I can review what he did and try to get into this soon.

stekyne:
@kunstmusik Ok sounds good. Did they get it running on linux/osx CI? That's all I'd have to add to get the PR into the codebase I'd say and from there we can just fix up tests as normal.

stekyne:
@kunstmusik this is working in CI now for linux/osx/windows but some tests are failing. What do you think? Using gtest in vcpkg works pretty well. It's not such a big dependency. I think it's best to maybe use this work instead of switching to another test framework again. Not sure why the wasm build is failing though.

CMakeLists.txt
```diff
@@ -1169,6 +1169,7 @@ if(EMSCRIPTEN)
   include_directories(emscripten/deps/libsndfile-1.0.25/src/)
 endif()
 
+# TODO include target rather than libs to also bring in transitive dependencies
```
kunstmusik, line 1172: Is there something required to address this comment?

tests/c/CMakeLists.txt
```diff
@@ -1,88 +1,47 @@
...
-add_test(NAME testServer
-        COMMAND $<TARGET_FILE:testServer> ${CMAKE_SOURCE_DIR}/tests/c/ -arg2 ${TEST_ARGS})
-
-
-endif(BUILD_TESTS)
-
-
-
-
+# FIXME currently require static lib, this should include build static csound condition
```
kunstmusik, line 3: Seems this FIXME is addressed?  If so, could you remove the comment?

kunstmusik (changes requested):
Just need responses on the two comments regarding the FIXME comments.

kunstmusik:
Tested to work on macOS arm64; assume from CI builds that this works on other platforms.  @jpffitch and @vlazzarini : You'll need to install googletest to run the unit tests, which are now in C++.  Should be fairly easy to get setup and should be well worth it for the more modern testing framework.

PR #1231, opened 2019-12-04
Fixes outvalue

gesellkammer:
This PR, if applied, fixes `outvalue` wrongly picking `outvalue.Si` over `outvalue.k`. It also removes a second, spurious call to `koutval` during the init phase.

Unrelated to this, it adds a new opcode "ftexists" which checks if a given ftable exists. This still needs support from the API to be able to skip error messages when the table is not found.


vlazzarini:
Thanks, maybe we should wait until @jpff modifies the function to allow skipping printouts, then you can submit your new opcode.

PR #1232, opened 2019-12-13
New array based channel opcodes..

rorywalsh:
Included in this PR are array based variants of the chnset/chnget opcodes. Based on Victor and Steven's suggestion, I named them `chnget/k/i/a/S` and `chnset/k/i/a/S`. I'm also adding an OSX job to the newly commit azure pipelines script, along with an azure build script for installers/macos. Also included here are two test .csd files which I've put into tests/soak.

I will update the manual entries for the channel opcodes if this PR is accepted.


vlazzarini:
Looks appveyor failed, can you check?

jpffitch:
outArray is not a field in the p structure so fails to build

rorywalsh:
Thanks guys, I'll fix this.

rorywalsh:
I can't test this on Windows right now, so perhaps it's best to let this PR
sit here a bit until I can.

rorywalsh:
I just tested this on Windows and it's working as expected. Would someone
mind taking a look at the new chnget_array_opcode... and
chnset_aray_opcode... functions to make sure they are thread safe? I
followed the convention used in the other channel opcodes, but I would
appreciate it if someone would give them the once over.

jpffitch:
Does this need any changes in the manual?

rorywalsh:
I made a PR for the manual some weeks ago..

PR #1233, opened 2019-12-16
modify zacl to omit the second argument if not needed

gesellkammer:
This PR modifies the `zacl` opcode to be able to omit the second argument if only one channel needs to be cleared.

With this applied, the following lines would be the same:

```csound
    zacl 3, 3
    zacl 3
```


jpffitch:
Do you have a manual edit?  Otherwise it is ok.

⁣Sent from TypeApp ​

PR #1234, opened 2019-12-18
add offset index to copya2ftab

gesellkammer:
With this PR it is possible to determine an offset into the ftable when copying from an array (the offset defaults to 0). 

`copya2ftab` will copy the elements of the array into the ftable starting at the given offset.

I can send a PR for the manual documenting this change.


gesellkammer:
New opcodes added:
* lastcycle: returns 1 if this is the last performance-cycle of the event
* strstrip: strip whitespace from string, either from left, right or both sides

jpffitch:
Do you have manual pages for the new opcodes?

Would it not be easier to make the index optional defaulting to zero?

jpffitch:
If a negative offset is specified your code reads outside the array i think.  Surely this should be policed.

gesellkammer:
* copya2ftab checks for negative index
* now index optional (default 0) in copya2ftab

jpffitch:
Thank you.  Will deal with it soon.

⁣Sent from TypeApp ​

PR #1253, opened 2020-02-09
Bela: updated include folders

giuliomoro:
fixed include folders for latest Bela code


jpffitch:
Do you know what the build problems are withh enscripten  and android?

giuliomoro:
No idea really, I changed something that was Bela-specific only.

PR #1263, opened 2020-02-19
Fix for leaked file pointer

iskunk:
`cpupercent_init()` calls `fopen()` on `/proc/stat`, but never closes it.
Create a Csound file handle so it is closed on program exit.
```
 552 bytes in 1 blocks are still reachable in loss record 4 of 8
    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
    by 0x5B73E49: __fopen_internal (iofopen.c:65)
    by 0x5B73E49: fopen@@GLIBC_2.2.5 (iofopen.c:89)
    by 0x30F617: cpupercent_init (cpumeter.c:78)
    by 0x167507: init_pass (insert.c:116)
    by 0x168E5D: insert_event (insert.c:472)
    by 0x1680B0: insert (insert.c:298)
    by 0x17A098: process_score_event (musmon.c:829)
    by 0x17ABD2: sensevents (musmon.c:1038)
    by 0x14DC5E: csoundPerform (csound.c:2243)
    by 0x14A548: main (csound_main.c:328)
```

### DIFF
--- a/Opcodes/cpumeter.c
+++ b/Opcodes/cpumeter.c
@@ -95,6 +95,7 @@ int32_t cpupercent_init(CSOUND *csound, CPUMETER* p)
     p->cpus = (CPU_t *) p->cpu_a.auxp;
     k = cpupercent_renew(csound, p);
     p->cnt = (p->trig = (int32_t)(*p->itrig * csound->GetSr(csound)));
+    csoundCreateFileHandle(csound, &p->fp, CSFILE_STD, "/proc/stat");
     return k;
 }
 


iskunk:
This issue was provoked by the "cpumeter" soak test:
```
csound -d -W -m0 -i /dev/null /path/to/csound/tests/soak/cpumeter.csd -o ./cpumeter.wav
```

[Here](https://github.com/csound/csound/files/4226788/csound-vg-log.txt) is the full Valgrind log.

jpffitch:
Should use csound->CreateFileHandle but I can patch that up later

jpffitch:
Would it be better to add a deinit process so the closing happens in perf?

iskunk:
Ah, missed that. Please straighten out any issues you see with my changes, of course; a few of them involved some guessing as to the right approach.

I'm not in a good position to say what's the best approach :-]  I'm not clear on how explicit opcode deinit is meant to work, since it seems like the normal approach is to allocate everything in a tracked manner such that it can be automatically freed later. Adding a Csound file-handle record for the `/proc/stat` file pointer was intended to extend that same MO to the leaked element, but there might be a better way in light of reinit, etc.

jpffitch:
I think the potential problem is if a user uses cpumeter many times for short duration, as a live coder might.  Then there would be lots of FILE* things until the end of csound.  The alternative is to register a deinit callback so as soon as the instrument ends the FILE* is recovered.   I admit to finding this area rather opaque....  I will check it.

Thanks for all your efforts; a little embarasing that we did not catch some of these with the static analyser.

iskunk:
Yeah, I've noticed everyone's quite concerned about perf-time things that could slow down a live performance. I've tried to keep that in mind, but admit that additional rigmarole may be needed.

Happy to help :-)  After all these fixes are in, I have a few changes to the test-suite framework that facilitated this work.

PR #1268, opened 2020-02-19
Fix for various Valgrind-reported memory leaks

iskunk:
`csoundJoinThread()` does not `free()` the small bit of memory `malloc()`'ed in `csoundCreateThread()` to hold the `pthread_t` object.

The `CsoundPerformanceThread` destructor does not free the mutex and condition variable created toward the end of `CsoundPerformanceThread::csPerfThread_constructor()`.

A new library function is needed to destroy a condition variable.

### DIFF
--- a/Top/threads.c
+++ b/Top/threads.c
@@ -205,6 +205,7 @@ PUBLIC uintptr_t csoundJoinThread(void *thread)
     if(thread == NULL) return 0;
     pthreadReturnValue = pthread_join(*pthread,
                                       &threadRoutineReturnValue);
+    free(pthread);
     if (pthreadReturnValue) {
         return (uintptr_t) ((intptr_t) pthreadReturnValue);
     } else {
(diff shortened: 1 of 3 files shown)


iskunk:
This issue was provoked by the "testPerfThread" C test:
```
testPerfThread /path/to/csound/tests/c/ -arg2 -+env:OPCODE6DIR64=/tmp/csound-build
```

[Here](https://github.com/csound/csound/files/4227662/csound-vg-log.txt) is the full Valgrind log. (Lines may be a bit inconsistent in `csPerfThread.cpp`, as I have a second change to submit in that file)

jpffitch:
After a quick look it seems that the patch was not applied in Windows case so the build failed.  Unfortunately I have no Windows system or expertise and little c++ so it will take a little time to make this PR acceptable

iskunk:
Whoops, missed the Windows and stub implementations. Those are in there now, though I have no means of testing the Windows build myself.

PR #1270, opened 2020-02-20
Make test suite compatible with Python 3

iskunk:
* Invoke Python as `python3` instead of `python` (the latter is usually Python2)

* Use `#!/usr/bin/env python3` shebang line consistently

* Tkinter is [now named](https://docs.python.org/2/library/tkinter.html) `tkinter` (lower case) in Python 3

* Convert all `print foo` statements to `print(foo)`

### DIFF
--- a/tests/commandline/CMakeLists.txt
+++ b/tests/commandline/CMakeLists.txt
@@ -1,10 +1,10 @@
 cmake_minimum_required(VERSION 2.8)
 
 if(MSVC) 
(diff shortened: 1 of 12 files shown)


iskunk:
The AppVeyor build environment appears to be missing a Python 3 installation.

The Travis CI build environment appears to be missing the `python3-tk` package.

stekyne:
Python3 is installed on the Appveyor image but it's not on the path by default. I think we need to either add something to the yaml file or else manually alter the Path.

PR #1274, opened 2020-02-25
Deleted old and unused files

laserbat:
Couldn't find any evidence that anybody needs these files nor any reason to keep them in the repo. In my opinion it would be great to reduce amount of clutter, so that's what I'm trying to do.


jpffitch:
Some of those filesare needed/used.  nsliders.tk for example.

Rather tha a blanket PR could we have a list of the files yu think need deleting so they can e hecke.  Also best vis mailing list i think

laserbat:
Alright, I'll ask about it on mailing list. Thank you

PR #1277, opened 2020-02-25
Eliminate "//*" sequences interpreted as comment-block starts

iskunk:
This change addresses a relatively minor issue I've encountered: The use of the C++-style `//` comment marker, in conjunction with comment text that begins with an asterisk, can result in a `//*` sequence which at least two text editors ([GNU nano](https://www.nano-editor.org/), [NEdit](https://en.wikipedia.org/wiki/NEdit)) interpret as the start of a long C-style `/* ... */` comment block. This causes long sections of active code to appear commented-out.

### DIFF
--- a/Engine/csound_prs.lex
+++ b/Engine/csound_prs.lex
@@ -1693,7 +1693,7 @@ static int bodmas(CSOUND *csound, yyscan_t yyscanner, int* term)
           goto parseNumber;
         //        case '^':
         //type = 0;
-        //*++op = c; c = getscochar(csound, 1); break;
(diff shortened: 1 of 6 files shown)


jpffitch:
comes from not using the one true editor!

hlolli:
All Hail St. iGNUcius, Our GNU/Lord and GNU/Savior!
~ Emacs

iskunk:
Heh, these days, a Web browser serves as my secondary operating system of choice :-)

(FYI, I did leave in the few `/* ... *//* ... */` sequences I found, as those were interpreted correctly.)

PR #1281, opened 2020-02-26
Fix for Valgrind-reported memory leak

iskunk:
`widget_reset()` fails to clear `widgetGlobals->AddrSetValue`.
```
 224 bytes in 1 blocks are definitely lost in loss record 211 of 333
    at 0x4C3041F: operator new(unsigned long) (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
    by 0x168D10FF: __gnu_cxx::new_allocator<ADDR_SET_VALUE>::allocate(unsigned long, void const*) (new_allocator.h:111)
    by 0x168CF8F5: std::allocator_traits<std::allocator<ADDR_SET_VALUE> >::allocate(std::allocator<ADDR_SET_VALUE>&, unsigned long) (alloc_traits.h:436)
    by 0x168CD4AF: std::_Vector_base<ADDR_SET_VALUE, std::allocator<ADDR_SET_VALUE> >::_M_allocate(unsigned long) (stl_vector.h:172)
    by 0x168CF07D: void std::vector<ADDR_SET_VALUE, std::allocator<ADDR_SET_VALUE> >::_M_realloc_insert<ADDR_SET_VALUE>(__gnu_cxx::__normal_iterator<ADDR_SET_VALUE*, std::vector<ADDR_SET_VALUE, std::allocator<ADDR_SET_VALUE> > >, ADDR_SET_VALUE&&) (vector.tcc:406)
    by 0x168CC8A7: void std::vector<ADDR_SET_VALUE, std::allocator<ADDR_SET_VALUE> >::emplace_back<ADDR_SET_VALUE>(ADDR_SET_VALUE&&) (vector.tcc:105)
    by 0x168C95A5: std::vector<ADDR_SET_VALUE, std::allocator<ADDR_SET_VALUE> >::push_back(ADDR_SET_VALUE&&) (stl_vector.h:954)
    by 0x168BD424: fl_joystick (widgets.cpp:3927)
    by 0x4E9DC05: init0 (insert.c:249)
    by 0x4EADD54: musmon (musmon.c:313)
    by 0x5033D63: csoundStart (main.c:549)
    by 0x5033DA7: csoundCompile (main.c:556)
    by 0x109913: main (csound_main.c:326)
```

PR #1282, opened 2020-02-26
Ensure that widget_reset() is called on reset

iskunk:
One thing I noticed in my testing is that some Csound invocations that use FLTK widgets were leaking memory, because the `widget_reset()` function (in which said memory is freed) was not being called.

I tracked this down to the `csoundModuleInit()` function in `InOut/winFLTK.c`. The problem appears to be that the `csound->RegisterResetCallback()` call that hooks in `widget_reset()` was not being called, even though the `widget_init()` call was.

Looking at the code, it is notable that the first call occurs only if a number of conditions are satisfied, whereas the second one occurs in the (likely) absence of errors. This appears incorrect---either both calls should happen (ideally together), or neither. This change moves the first call to the same place as the second. It compiles cleanly, and has a neutral-to-positive effect on Valgrind results for all existing tests.

The tests specifically affected by this are following two:
```
csound -d -W -m0 -i /dev/null /path/to/csound/tests/soak/crossfm.csd -o ./crossfm.wav
csound -d -W -m0 -i /dev/null /path/to/csound/tests/soak/noise-2.csd -o ./noise-2.wav
```

Note that this change is a minimal fix, but perhaps more lines (comment? "graph init" message? other callbacks?) should be moved down in the same way.

### DIFF
--- a/InOut/winFLTK.c
+++ b/InOut/winFLTK.c
@@ -100,7 +100,6 @@ PUBLIC int csoundModuleInit(CSOUND *csound)
           csound->SetKillGraphCallback(csound, KillGraph_FLTK);
           csound->SetExitGraphCallback(csound, ExitGraph_FLTK);
            /* seemed to crash, but not anymore... */
-          csound->RegisterResetCallback(csound, NULL, widget_reset);
           csound->Message(csound, "graph init...\n");
 
         }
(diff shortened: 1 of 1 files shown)


iskunk:
One detail I failed to note earlier: My fix in #1281 actually depends on this one (or else `widget_reset()` is not called in any event). This fix was the harder one to articulate, so it came second :-]

PR #1290, opened 2020-03-06
chn_k: add possibility to set the input/output mode as string

gesellkammer:
This PR adds the possibility to set the imode of a channel via a string, like

```csound

chn_k "foo", "r"
chn_k "bar", "w"
chn_k "baz", "rw"
```

This is the same as `chn_k "foo", 1` but more readable


gesellkammer:
@kunstmusik: did you get a chance to review the corrections?

kunstmusik:
@gesellkammer Just looked, haven't tested but from reading code it looks good to me. @jpffitch @vlazzarini any issues with this change? Guess we'd need documentation change and example too.

jpffitch:
I am agnostcabout this.

gesellkammer:
Fix to reshapearray, as pointed out by @jpffitch 

> There are (at least) 3 problems with this opcode
> 
> 1:  if the sizes of the array and the size of the reshaped array do not match
> it needs an error message. Currently it is rather silent.
> 
> 2: in calculating the sizes in numcols is zero (the default value)
> multiplying by zero always leads to an error.
> 
> 3:  in numrows is 0 you end up with a 0 x 0 array.
>

kunstmusik:
Ugh, was going to merge this but now there's unrelated commits. Guess let's make sure these other changes are good, then merge, then move on to the coreaudio one which looked like it needed this one to merge first.

jpffitch:
The arrayreshape code looks OK to me.  I am happy fotr that to be merge 
in.  Steven to decide on the channel component I think.

kunstmusik:
Merged. @gesellkammer please submit updated docs/example for chn_k if not done so already. Thanks!

PR #1299, opened 2020-03-14
CoreAudio fix for nchnls mismatch

gesellkammer:
In macOS, when using coreaudio (auhal) as the backend, a mismatch in nchnls and the actual number of channels of the device will lead to csound hanging if nchnls > device channels. This PR fixes this issue by checking that device channels >= nchnls, and fails if this is not the case. This behaviour is similar to what portaudio does under the same circumstances. 
This fixes similar issues reported to csoundqt, where csoundqt would hang waiting for csound's audio thread to join when stopping a performance.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -218,3 +218,10 @@ CMakeLists.txt.user
 
 # emacs autosaved files
 \#*\#
+
+
+# cmake / doxygen 
+CMakeDoxy*
+CPackConfig.cmake
+CPackSourceConfig.cmake
+Doxyfile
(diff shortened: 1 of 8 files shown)


kunstmusik:
@gesellkammer this branch has changes not related to the subject and there is a conflict now. Could you rebase the changes so that we can see just the coreaudio change?

gesellkammer:
Is there something still preventing this PR from being merged?

jpffitch:
It needs amac user to confirm

⁣Get TypeApp for Android ​

kunstmusik:
This pull request is hard to evaluate since it has many commits showing that don't have anything to do with the original PR.  Separate branches and rebasing per PR would make this a lot simpler. I'll try to look at it now.

kunstmusik:
I built and lightly tested. Code changes that I could find relevant to the PR looked good. Merged.

PR #1308, opened 2020-04-04
Add an opcode "errormsg" 

gesellkammer:
This PR adds the opcode `errormsg`, which allows the user to throw an error from its own code. By default it throws a PerfError, which stops the current event. Two other error kinds are defined: a "warning", which prints a warning in the console only once, and "info", which prints a warning every cycle.

```csound

if kcondition == 1 then
    errormsg "An error has occurred!"
    ; this is the same as errormsg, "error", ...
elseif kcondition == 2 then
    errormsg "warning", "This is just a warning"
endif
```

`errormsg` works only at k-time. If needed, an i-variant can be added (maybe as `errormsg_i`)

NB: another name for this opcode might be `perferror`, and an init variant could then be "initerror", following the csound API.

### DIFF
--- a/Opcodes/emugens/emugens.c
+++ b/Opcodes/emugens/emugens.c
@@ -2271,6 +2271,80 @@ strstrip(CSOUND *csound, STR1_1 *p) {
 }
 
 
+/* errormsg
+ *
+ * errormsg "mymsg", Skind="error"       ; k-rate
+ *
+ * Skind can be "error", "warning" or "info"
+ * If "error", (the default) a performance error is thrown, which will delete the current event
+ * If "warning", a warning will be thrown but the event goes on (a warning is only shown once)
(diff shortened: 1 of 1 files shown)


jpffitch:
You can unify these if you look at csound->mode to see if you are ininit or perf.

⁣Sent from TypeApp ​

gesellkammer:
How would the dispatching work with csound->mode?

jpffitch:
In the dispatch loops for opcodes either init or perf the variables 
csound->mode is set to 1 (init) or 2 (perf).  Outside the loops it s
is zero, and csound->op is set to a string of the opcode name.

It may be the case that it is only visible in base code but that could be 
fixed easily.

gesellkammer:
I don't think that this is accessible to plugin opcodes. But maybe this could be moved into the core?

PR #1312, opened 2020-04-11
lisp cffi bindings: fix dylib name

azimut:
I tested only to the point that I can load the dylib now.


kunstmusik:
csnd6 is a different library than csound64 (the former is a C++ API).  The Csound library on macOS is CsoundLib64 (no extension) and is provided the CsoundLib64.framework.  Please try that name instead. Closing out this PR as invalid.

PR #1315, opened 2020-04-12
optional parameter for GEN20, 9

joachimheintz:
the sinc function is currently set to a fixed range from -pi to +pi.
the proposed change adds an optional multiplier for the x-range (1 by default).
so  `f1   0   4096   -20   9   1` would result in the current behaviour, whereas `f1   0   4096   -20   9   1 3` would generate a sinc function from -3pi to +3pi.

i tested the change and it seems to work.  i don't understand why there is no check in the code for x=0.  perhaps someone can review.

once it is merged, i can add it to the manual.  i also have new figures for https://csound.com/docs/manual/MiscWindows.html.

### DIFF
--- a/Engine/fgens.c
+++ b/Engine/fgens.c
@@ -1313,8 +1313,8 @@ static int gen20(FGDATA *ff, FUNC *ftp)
           ft[i] = FL(1.0);
         return OK;
     case 9:                     /* Sinc */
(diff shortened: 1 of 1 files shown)


jpffitch:
I will deal with x==0 case soon

joachimheintz:
thanks -- i will go for the manual page now.

PR #1316, opened 2020-04-12
fix bug in GEN20,8

joachimheintz:
gen20, 8 always returned 1 as value for the rectangle window, for instance in
`i0 ftgen 8, 0, 8192, -20, 8, .1`
i think this fixes it.


jpffitch:
The manual is a little vague on this but I think your fix is what was intended.  Just a slight worry if anyone has been using the broken code.

PR #1332, opened 2020-05-28
Scugens rename

gesellkammer:
Following a request from @joachimheintz, I propose a road to renaming the opcodes in scugens. The main point of the request was that the name of the opcodes, which included the sc_ prefix was not descriptive of their function. 

Since these opcodes are quite new, the oldest being introduced in 2019, I propose a roadmap for renaming them as follows:

For the next release (6.15), an alias for each is proposed, and both names are valid. The usage of the old name is marked as deprecated in the manual and a warning is printed in the console. 

The following release (6.16?) would have the original names removed

* `sc_lag` → `lag`
* `sc_lagud` → `lagud`
* `sc_trig` → `trighold`
* `sc_phasor` → `phasortrig`

If approved, I would add an entry for all those new names in the manual and modify the old manual names to mark them as deprecated with a note to their new name.


jpffitch:
Missing } means it fails to compile

Remember to adjust the manual

PR #1335, opened 2020-06-12
Do not ignore return value of setuid

jkrshnmenon:
Ignoring the return value of setuid might lead to a security issue.

Since the csound_main.c creates log files after invoking set_rt_priority,
if setuid fails, these files would be owned by root.


vlazzarini:
Seems ok, but the appveyor build failed for some reason that might be unrelated, needs to be checked

jpffitch:
Looks OK to me too  I merged by hand and srems OK on Linux.

PR #1342, opened 2020-06-26
Let syntax-check-only return an error if syntax failed

gesellkammer:
When csound is called with --syntax-check-only it always returns CSOUND_EXITJMP_SUCCESS. With this PR the function csoundCompile will return an error code (CSOUND_ERROR) if called with that flag and parsing failed. 
This allows frontends like CsoundQt to check syntax of a csd without running it and prevent a series of crashes were a failed compilation should not be carried further.

### DIFF
--- a/Top/main.c
+++ b/Top/main.c
@@ -164,6 +164,7 @@ PUBLIC int csoundCompileArgs(CSOUND *csound, int argc, const char **argv)
     volatile int     csdFound = 0;
     volatile int ac = argc;
     char    *fileDir;
+    int     compiledOk = 0;
 
 
     if ((n = setjmp(csound->exitjmp)) != 0) {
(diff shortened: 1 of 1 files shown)


rorywalsh:
Thanks @gesellkammer. This is going to be really useful.

jpffitch:
I had to revert this PR as the source in Top/main.c no longer compiles 


The problem is something to do wih the variablr compilesOk clashing wit the setjmp/longjmp
Sorry lost the actual wording as I needed to build a new system.

`If/when I have time I will look again but  no`t sure when that will be

jpffitch:
/home/jpff/csound6/Top/bust_main.c: In function ‘csoundCompileArgs’:
/home/jpff/csound6/Top/bust_main.c:162:13: error: variable ‘compiledOk’ might be clobbered by ‘longjmp’ or ‘vfork’ [-Werror=clobbered]
     int     compiledOk = 0;
             ^~~~~~~~~~
cc1: all warnings being treated as errors

PR #1345, opened 2020-06-27
Let syntax check return an error when failed

gesellkammer:
Trying this again, as it was prematurely closed. This PR is necessary in order to avoid csound crashing when in --daemon mode and the first compile fails. This crash brings down also any frontend (CsoundQt, ctcsound) with it. With this in place it is possible to preform a syntax check and see if it succeeded or not.


vlazzarini:
Thanks. It seems to be bringing in a backup file into the repo. Can you check?

jpffitch:
committed

jpffitch:
istalled from a local copy to avoif d the half-editted fie

vlazzarini:
Is this incorporated? Can it be closed.

jpffitch:
Certainlt; I thought it was closed.

PR #1356, opened 2020-07-13
Fix ftset (init version) and make lastcycle more robust

gesellkammer:
* ftset had a wrong OENTRY entry and the order of appearance was wrong. This PR fixes that. 
* lastcycle, contrary to what was stated in the manual, would not fire if no time extending opcode is used and the instr is turned off. This is now solved by adding an extra cycle if necessary. The manual has been updated with this information


jpffitch:
The fset opcode is incorrectly defined set multicore

⁣Get TypeApp for Android ​

gesellkammer:
I fixed this locally but since the PR is closed it will not propagate, so I would need to open a new PR. @jpffitch: could you please fix this? One just needs to add TW to the oentry

PR #1362, opened 2020-07-24
Added opcode to support new Trill sensors for Bela

berntisak:
Added the opcode `trill` which can be used to read data from Trill sensors

usage: `kactiveTouches, ksizes[], kvertLocations[], khoriLocations[] trill inumTouches, itrillID`

### DIFF
--- a/Bela/main.cpp
+++ b/Bela/main.cpp
@@ -23,23 +23,27 @@
 
 #include <getopt.h>
 #include <Bela.h>
-#include <Midi.h>
-#include <Scope.h>
-#include <csound.hpp>
-#include <plugin.h>
+#include <libraries/Midi/Midi.h>
+#include <libraries/Scope/Scope.h>
+#include <libraries/Trill/Trill.h>
+#include <csound/csound.hpp>
+#include <csound/plugin.h>
 #include <vector>
 #include <sstream>
 #include <iostream>
 #include <atomic>
 
+
 static int OpenMidiInDevice(CSOUND *csound, void **userData, const char *dev);
 static int CloseMidiInDevice(CSOUND *csound, void *userData);
 static int ReadMidiData(CSOUND *csound, void *userData, unsigned char *mbuf,
-			int nbytes);
+      int nbytes);
 static int OpenMidiOutDevice(CSOUND *csound, void **userData, const char *dev);
 static int CloseMidiOutDevice(CSOUND *csound, void *userData);
 static int WriteMidiData(CSOUND *csound, void *userData, const unsigned char *mbuf,
-			 int nbytes);
+       int nbytes);
+
+std::vector<Trill*> gTouchSensors;
 
 /** DigiIn opcode 
     ksig digiInBela ipin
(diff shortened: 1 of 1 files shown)


Bela/main.cpp
```diff
@@ -199,6 +203,71 @@ struct DigiIO : csnd::InPlug<3> {
...
+void trillReadLoop(void*)
+{
+  while(!Bela_stopRequested())
+  {
+    for(unsigned int n = 0; n < gTouchSensors.size(); ++n)
+    {
+      Trill* t = gTouchSensors[n];
+      t->readI2C();
+    }
+    usleep(50000);
```
giuliomoro, line 267: This could be shorter, 10 to 15 millisecond would be ideal, but it should become smaller when many sensors are connected . Could you maybe expose this as a control parameter somehow?

Bela/main.cpp
```diff
@@ -223,8 +292,8 @@ bool csound_setup(BelaContext *context, void *p)
   CsData *csData = (CsData *) p;
   Csound *csound;
   const char *args[] = { "csound", csData->csdfile.c_str(), "-iadc",
-			 "-odac", "-+rtaudio=null",
-			 "--realtime", "--daemon" };
+       "-odac", "-+rtaudio=null",
```
giuliomoro, line 295: are these (and other) whitespace changes expected? they make the diff excessively large. If these are intended (e.g.: to conform to code standards), I would recommend to have them all in a dedicated commit.

Bela/main.cpp
```diff
@@ -245,24 +314,40 @@ bool csound_setup(BelaContext *context, void *p)
...
   if(csnd::plugin<DigiIO>((csnd::Csound *) csound->GetCsound(), "digiIOBela" ,
-			   "", "aaa", csnd::thread::ia) != 0)
+         "", "aaa", csnd::thread::ia) != 0)
     printf("Warning: could not add digiIOBela a-rate opcode\n");
-  
+
+  /* set up trill opcode */
+  if(csnd::plugin<TrillIn>((csnd::Csound *) csound->GetCsound(), "trill" ,
+         "kk[]k[]k[]", "ii", csnd::thread::ik) != 0)
+    printf("Warning: could not add trillBar a-rate opcode\n");
```
giuliomoro, line 338: why `trillBar`? just `Trill` should be good

Bela/main.cpp
```diff
@@ -245,24 +314,40 @@ bool csound_setup(BelaContext *context, void *p)
...
+  /* set up trill opcode */
+  if(csnd::plugin<TrillIn>((csnd::Csound *) csound->GetCsound(), "trill" ,
+         "kk[]k[]k[]", "ii", csnd::thread::ik) != 0)
+    printf("Warning: could not add trillBar a-rate opcode\n");  
+
+  /* set up trill sensors */
+  unsigned int i2cBus = 1;
+  for(uint8_t addr = 0x20; addr <= 0x50; ++addr) {
+    Trill::Device device = Trill::probe(i2cBus, addr);
+    if(Trill::NONE != device && Trill::CRAFT != device) {
```
giuliomoro, line 344: remove ` && Trill::CRAFT != device`, otherwise it will only work with Trill Craft

Bela/main.cpp
```diff
@@ -199,6 +203,71 @@ struct DigiIO : csnd::InPlug<3> {
...
+    touchVertLocation = new float[numTouches];
+    touchHoriLocation = new float[numTouches];    
+    return OK;
+  }
+  
+  int kperf () {
+    activeTouches = gTouchSensors[trillID]->getNumTouches();
+    outargs[0] = MYFLT (activeTouches);
+    // Read locations from Trill sensor
+    for(unsigned int i = 0; i < activeTouches; i++) {
```
giuliomoro, line 235: this only works when the devices are in `Centroid` mode, which is the default mode for all devices except Trill Craft. For Trill Craft you should probably read `gTouchSensors[trillId].rawData[n]` instead. Then below in the loop you could differentiate between 1D and 2D devices. See [here](https://github.com/BelaPlatform/Bela/blob/dev/core/default_libpd_render.cpp#L900-L927) for an example

Bela/main.cpp
giuliomoro: as long as it's not `NONE`, you are good to go!

Bela/main.cpp
berntisak: Not really - must have happened when copying the original main.cpp code from Sublime to the Bela IDE and back. I'll try to revert them before updating or making a new PR

Bela/main.cpp
berntisak: Old name that I forgot to update. Fixed now

vlazzarini:
I'll merge this and improvements can be submitted as a new PR

Bela/main.cpp
berntisak: Done

Bela/main.cpp
berntisak: Maybe I should make a trillCraft opcode for the Trill Craft? I've only had a chance to test Ring, Square and Bar. I also realised even the Square might have its specific opcode to be able to vary between separate touches and compoundtouch. Or I guess that could be a general argument. I'll have a look at it for the next PR

Bela/main.cpp
berntisak: Hmm, it would be best to have that as a separate opcode then I think? Since you can have multiple instances of the trill opcodes, you dont want all of them to adjust this speed. For now, I will lower it

PR #1376, opened 2020-09-09
LFSR opcode

DaveSeidel:
```
/**
 * Linear Feedback Shift Register (LFSR) opcode.
 * 
 * Description
 * 
 *      Output is a series of pseudo-random positive integers. This is the technique
 *      used in so-called "Turing machine" synth modules and is usually used to
 *      generate melodic sequences. This implementation is adapted from the firmware
 *      for the Ornament & Crime module, as used in the Quantermain and Meta-Q apps.
 * 
 * Syntax
 *
 *      knum lfsr ilen, iprob [, iseed]
 *
 *      knum = lfsr(ilen, iprob [, iseed])
 *
 * Initialization
 * 
 *      ilen -- length of shift register, valid values are 1-31 (inclusive). The
 *      larger the length, the larger the resulting integers in the output. You
 *      can use this to constrain the output to a suitable range.
 * 
 *      iprob -- probability, valid values 1-255 (inclusive). Controls the spread
 *      of the output; larger values result in a wider spread of values.
 * 
 *      iseed (optional, default -1) -- initial state of the shift register, as a
 *      pattern of bits. The value is treated as an unsigned integer, so the default
 *      of -1 is effectively all bits on (0b11111111...).
 *  
 * Performance
 * 
 *      knum -- Integer output.
 */
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -225,3 +225,6 @@ CMakeDoxy*
 CPackConfig.cmake
 CPackSourceConfig.cmake
 Doxyfile
+
+# vscode
+.vscode/
--- a/Opcodes/CMakeLists.txt
+++ b/Opcodes/CMakeLists.txt
@@ -57,6 +57,7 @@ make_plugin(getftargs getftargs.c)
 make_plugin(liveconv liveconv.c)
 ##make_plugin(zak zak.c)
 make_plugin(gtf gammatone.c)
+make_plugin(lfsr lfsr.cpp)
 
 make_plugin(fareygen fareygen.c)
 make_plugin(quadbezier quadbezier.c)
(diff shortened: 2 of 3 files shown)


DaveSeidel:
See source code for copyright info.

DaveSeidel:
See https://github.com/csound/manual/pull/331

DaveSeidel:
Previous builds failed because MSVC is fussy about the time.h header, but the most recent failure is in the installer build and doesn't seem to have anything to do with my change:
```
Error on line 183 in C:\projects\csound\installer\windows\csound6_x64_appveyor.iss: Source file "C:\projects\csound\installer\windows\..\..\msvc\csound-vs\Release\pmidi.dll" does not exist.
```

Issue #1388, opened 2020-10-06
slicearray on array of strings doesn't work


hlolli:
With web-ide and linux build of https://github.com/csound/csound/commit/53c103ca4b9013c8046813d18eb2b843b86af5d2

fails to slice, it prints
```
new alloc for instr SUCCESS:
iArr2[0] is 1
iArr2[1] is 2
iArr2[2] is 3
new alloc for instr FAIL:
incorrect value for SArr2[0]
incorrect value for SArr2[1]
incorrect value for SArr2[2]
```

If you were to print the values of SArr2 after the slice operation, they'd be empty.

```Csound
<CsoundSynthesizer>
<CsOptions>
</CsOptions>
<CsInstruments>

sr = 48000
ksmps = 1
nchnls = 2
0dbfs = 1.0


instr SUCCESS
  iArr1[] fillarray 1,2,3,4,5,6,7
  iArr2[] init 3
 ;; EDITED: was slicearray_i (same result with slicearray)
  iArr2 slicearray iArr1, 0, 2

  if iArr2[0] == 1 then
    prints "iArr2[0] is 1\n"
  else
    prints "incorrect value for iArr2[0]\n"
  endif

  if iArr2[1] == 2 then
    prints "iArr2[1] is 2\n"
  else
    prints "incorrect value for iArr2[1]\n"
  endif

  if iArr2[2] == 3 then
    prints "iArr2[2] is 3\n"
  else
    prints "incorrect value for iArr2[2]\n"
  endif
endin

instr FAIL
  SArr1[] fillarray "1","2","3","4","5","6","7"
  SArr2[] init 3
  SArr2 slicearray SArr1, 0, 2

  if strcmp( SArr2[0], "1" ) == 0 then
    prints "SArr2[0] is \"1\"\n"
  else
    prints "incorrect value for SArr2[0]\n"
  endif

  if strcmp( SArr2[1], "2" ) == 0 then
    prints "SArr2[1] is \"2\"\n"
  else
    prints "incorrect value for SArr2[1]\n"
  endif

  if strcmp( SArr2[2], "3" ) == 0 then
    prints "SArr2[2] is \"3\"\n"
  else
    prints "incorrect value for SArr2[2]\n"
  endif
endin

</CsInstruments>
<CsScore>
i "SUCCESS" 0 0
i "FAIL" 0 0
</CsScore>
</CsoundSynthesizer>
```

hlolli:
[...]

PR #1393, opened 2020-10-14
fix slicearray for StringArrays

hlolli:
Running the csd I posted in the ticket I wrote in #1388 I get

```
new alloc for instr SUCCESS:
iArr2[0] is 1
iArr2[1] is 2
iArr2[2] is 3
new alloc for instr FAIL:
incorrect value for SArr2[0]
incorrect value for SArr2[1]
incorrect value for SArr2[2]
```

applying this change I get

```
new alloc for instr SUCCESS:
iArr2[0] is 1
iArr2[1] is 2
iArr2[2] is 3
new alloc for instr FAIL:
SArr2[0] is "1"
SArr2[1] is "2"
SArr2[2] is "3"
```

### DIFF
--- a/Opcodes/arrays.c
+++ b/Opcodes/arrays.c
@@ -4262,7 +4262,7 @@ static OENTRY arrayvars_localops[] =
       (SUBR) tabslice, (SUBR) tabslice, NULL },
     { "slicearray.a", sizeof(TABSLICE), 0, 2, "a[]", "a[]iip",
       (SUBR) tabslice, (SUBR) tabslice, NULL },
-    { "slicearray.S", sizeof(TABSLICE), 0, 2, "S[]", "S[]iip",
+    { "slicearray.S", sizeof(TABSLICE), 0, 1, "S[]", "S[]iip",
       (SUBR) tabslice, (SUBR) tabslice, NULL },
     { "slicearray_i", sizeof(TABSLICE), 0, 1, ".[]", "i[]iip",
       (SUBR) tabslice, NULL },


kunstmusik:
I don't think this is a good solution. slicearray is set to run at perf-time and this is a backwards incompatible change. slicearray_i is designed to run at init-time; why not add a slicearray_i implementation that works with S[]?

hlolli:
This is backwards compatible. This is delivering expected behavior from what is already present in i[] arrays.

```csound
<CsoundSynthesizer>
<CsOptions>
</CsOptions>
<CsInstruments>

sr = 48000
ksmps = 1
nchnls = 2
0dbfs = 1.0

giPrinted = 0

instr PERFRATE
  iArr1[] fillarray 1,2,3,4,5,6,7
  iArr2[] init 3
  iArr2 slicearray iArr1, 0, 2

  if iArr2[0] == 1 then
    printks "PERFRATE iArr2[0] is 1\n", p3
  else
    printks "PERFRATE incorrect value for iArr2[0]\n", p3
  endif

  if iArr2[1] == 2 then
    printks "PERFRATE iArr2[1] is 2\n", p3
  else
    printks "PERFRATE incorrect value for iArr2[1]\n", p3
  endif

  if iArr2[2] == 3 then
    printks "PERFRATE iArr2[2] is 3\n", p3
  else
    printks "PERFRATE incorrect value for iArr2[2]\n", p3
  endif
endin

instr INITRATE
  iArr1[] fillarray 1,2,3,4,5,6,7
  iArr2[] init 3
  iArr2 slicearray iArr1, 0, 2

  if iArr2[0] == 1 then
    prints "INITRATE iArr2[0] is 1\n"
  else
    prints "INITRATE incorrect value for iArr2[0]\n"
  endif

  if iArr2[1] == 2 then
    prints "INITRATE iArr2[1] is 2\n"
  else
    prints "INITRATE incorrect value for iArr2[1]\n"
  endif

  if iArr2[2] == 3 then
    prints "INITRATE iArr2[2] is 3\n"
  else
    prints "INITRATE incorrect value for iArr2[2]\n"
  endif
endin

instr STRIRATE
  SArr1[] fillarray "1","2","3","4","5","6","7"
  SArr2[] init 3
  SArr2 slicearray SArr1, 0, 2

  if strcmp( SArr2[0], "1" ) == 0 then
    prints "STRIRATE SArr2[0] is \"1\"\n"
  else
    prints "STRIRATE incorrect value for SArr2[0]\n"
  endif

  if strcmp( SArr2[1], "2" ) == 0 then
    prints "STRIRATE SArr2[1] is \"2\"\n"
  else
    prints "STRIRATE incorrect value for SArr2[1]\n"
  endif

  if strcmp( SArr2[2], "3" ) == 0 then
    prints "STRIRATE SArr2[2] is \"3\"\n"
  else
    prints "STRIRATE incorrect value for SArr2[2]\n"
  endif
endin

instr STRKRATE
  SArr1[] fillarray "1","2","3","4","5","6","7"
  SArr2[] init 3
  SArr2 slicearray SArr1, 0, 2

  if strcmp( SArr2[0], "1" ) == 0 then
    printks "STRKRATE SArr2[0] is \"1\"\n", p3
  else
    printks "STRKRATE incorrect value for SArr2[0]\n", p3
  endif

  if strcmp( SArr2[1], "2" ) == 0 then
    printks "STRKRATE SArr2[1] is \"2\"\n", p3
  else
    printks "STRKRATE incorrect value for SArr2[1]\n", p3
  endif

  if strcmp( SArr2[2], "3" ) == 0 then
    printks "STRKRATE SArr2[2] is \"3\"\n", p3
  else
    printks "STRKRATE incorrect value for SArr2[2]\n", p3
  endif
endin


</CsInstruments>
<CsScore>
i "PERFRATE" 0 1
i "INITRATE" 0 1
i "STRIRATE" 1 1
i "STRKRATE" 1 1
</CsScore>
</CsoundSynthesizer>
```

prints
```
new alloc for instr PERFRATE:
new alloc for instr INITRATE:
INITRATE iArr2[0] is 1
INITRATE iArr2[1] is 2
INITRATE iArr2[2] is 3
PERFRATE iArr2[0] is 1
PERFRATE iArr2[1] is 2
PERFRATE iArr2[2] is 3
B  0.000 ..  1.000 T  1.000 TT  1.000 M:  0.00000  0.00000
new alloc for instr STRIRATE:
STRIRATE SArr2[0] is "1"
STRIRATE SArr2[1] is "2"
STRIRATE SArr2[2] is "3"
new alloc for instr STRKRATE:
STRKRATE SArr2[0] is "1"
STRKRATE SArr2[1] is "2"
STRKRATE SArr2[2] is "3"
```

if it was up to me, wo shouldn't be making opcode underscore rate, it's very confusing, especially when the behviors cross over. If the opcode is internally looping over performance rate, but the contents within it can't update any faster, it doesn't make sense, but there may be a reason for this I'm missing. I'm totally aware of for example event vs event_i, how they differ in behavior, but I also don't see why event:k couldn't be introduced too and event_i be deprecated. That's just me, but I know this inconsistency is relied upon already.

kunstmusik:
Again, this is **NOT** backwards compatible. The reason there is an event vs event_i, as well as slicearray vs. slicearray_i, is that there is no way to differentiate the rate of operation based upon just the types of input and output arguments. Because of this, two differently named opcodes are necessary. For example, with event, you may wish to use only i-rate params but also want it to run at perf-time, or you may wish to have event_i run only with i-rate params and only at init-time once. 

This PR changes when slicearray operates. S[] can be mutated at perf-time, including array length, and the current perf-time slicearray accounts for this.  You want the behavior of slicearray_i, which explicitly runs at init-time, but for S[]. Again, the solution for this is to modify slicearray_i, not break slicearray for S[].

hlolli:
```
you may wish to use only i-rate params but also want it to run at perf-time
```

this I don't understand, you use k-rate for per-time and i-rate for initialization time. If you update i-rate var on k-rate, you'll only be able to read the value at i-rate from i-rate opcodes. Something went wrong in this design process imo. But I must have missed it when it took place in the past and will accept this.

hlolli:
Also why not make string readable both on i and k rate. Another issue imo.

kunstmusik:
You're being very dismissive while failing to understand that there are real issues here. 

"you use k-rate for perf-time and i-rate for initialization time."

is not a good way to think about this here for two reasons. First, look at this:

```csound
;; perf code
instr Runner
  if (changed2(gksomeVal) == 1) then
    event("i", 1, 0, 4)
  endif
endin

;; init code
instr ScoreLoop1
  event_i("i", 1, 0, 4)
  event_i("i", 2, 0, 4)
  event_i("i", 3, 0, 4)
  event_i("i", 4, 0, 4)
endin
```

The latter can not use event() or it will generate every block and person might set a duration when scheduling ScoreLoop1. The former you want to generate a note at perf-time whenever a signal changes. Both use the exact same rates of arguments and therefore *MUST* be differentiated by opcode name. In this case event has not output and is used completely for side-effects. 

Your argument about "if you update i-rate var on k-rate, you'll only be able to read the value at i-rate from i-rate opcodes." is correct but does not apply to this situation as it only applies to i- vs. k-.  S-vars do not have rates associated with them.  (Yes, this is a design issue; I wrote a lot about it in my PhD thesis, that rates ultimately come from opcodes, not datatypes). This is because the concept of rates associated with datatypes was perfectly sensible for Csound's original design, but things like strings were added over time that do not have rates.  You can have a string that mutates at init-time or at perf-time, but it you have no concept of this from the S-type itself. 

So again, if you're going to create a set of opcodes that has the possibility of processing input and output arguments of the same type and different rates, you MUST have different opcode names. 

Also, Strings are readable at both at init and perf-time. That's why there are string processing opcodes like strsub and strsubk to differentiate when you want strings processing to happen.

PR #1396, opened 2020-10-16
Add slicearray_i for string arrays

hlolli:
The command-line test for test_array_function_call.csd was failing and wasn't called when called with the other csdtests. I'm still unuser if the output: is comared during the test?
Otherwise this seems an easy case to add, given that slicearray_i for .[] types is calling the same function.


kunstmusik:
I added a k[] form for the case of working with gk[] at init-time. tabslice should probably do a type check of the input and output array types, but that can be tracked in a separate issue.

PR #1399, opened 2020-10-17
Implementing declare using new udo decleration mechanism

hlolli:
A draft PR targeting the new parser3. The test test_declare.csd runs successfully.


kunstmusik:
@hlolli I've merged this and made a couple of small changes (one to code style, other to add test_declare.csd into the test.py runner).  We would benefit from an additional test with expected failures or issues (i.e., calling declare a 2nd time, malformed declare statements, etc.)

PR #1405, opened 2020-11-06
Enable support for text-file comments in readk*

iskunk:
This is a proposed commit to enable the `readk*` opcodes to skip past comments in the ASCII-text input formats (`iformat` = 7 or 8), same as what GEN23 currently does.

Being able to use comments in the input text files is useful for documenting the contents/purpose of the file, adding column headers, etc.


vlazzarini:
sorry, this seems to have fallen through the cracks. Looks fine to me.

PR #1409, opened 2020-12-02
add version of pow for the case kout[] = kx ^ ivalues[] 

gesellkammer:


kunstmusik:
This looks good to me, are there updated manual pages already?

gesellkammer:
I don't think that the manual page needs updating in this regard (although a good candidate for an update, since it does not say anything about the rates supported, arrays, etc.)

gesellkammer:
Can this be merged?

kunstmusik:
Yes, sorry, I have been a bit out of it the past few days with work. I'll merge now and add an issue to the manual repo that the page for ^ needs updating to reflect this change.

PR #1431, opened 2021-02-02
 better rendering of console, XHTML aware

hfmanson:
replace innerHTML with createTextNode, createElementNS and textNode. 
UpdateStatus: 'keep' parameter defaults to true
minimal demo's in html/xml using div console

### DIFF
--- a/Emscripten/examples/minimal-div.html
+++ b/Emscripten/examples/minimal-div.html
@@ -63,11 +63,12 @@
 
     function handleMessage(message) {
       var element = document.getElementById('console');
-      element.value += message;
+      element.appendChild(document.createTextNode(message));
+      element.appendChild(document.createElementNS("http://www.w3.org/1999/xhtml", "br"));
       element.scrollTop = 99999; // focus on bottom
       count += 1;
       if (count == 1000) {
-        element.value = ' ';
+        element.textNode = ' ';
         count = 0;
       }
     }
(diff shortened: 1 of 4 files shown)


kunstmusik:
@hfmanson I think the changes look good, but at the same time the new version of WebAudio Csound will likely move away from csound.js (or at least, require a rewritten version to work with the revised core WebAudio Csound API) and encourage direct use of the core API.  

@vlazzarini I think we should simplify to supporting just the core API now and phase out use of csound.js as I think almost no one has used the PNaCl version of Csound at this point.  I think the examples should be easy to migrate from csound.js to the new API.  Thoughts?

hfmanson:
is there a branch where the new version of WebAudio Csound is developed on?

kunstmusik:
Yes it is the feature/webaudio-csound branch:

https://github.com/csound/csound/tree/feature/webaudio-csound

The work for this version is done in the wasm folder and has significant changes using a different compiler, updated API, etc.  There's some documentation there but @hlolli and I are going to put together a separate website (https://github.com/csound/webaudio/) that will eventually be served at csound.com/webaudio that will have installation information, API documentation, examples, tutorials, etc.

vlazzarini:
happy to phase out csound.js in this upcoming review. Just removing it and the examples using should be ok.

kunstmusik:
Thanks @vlazzarini!

FYI: I went ahead and merged; we'll see about updating and incorporating the old examples into the new webaudio sub-site for csound.com. I'll also make a note in the ChangeLog that csound.js is going away.

PR #1434, opened 2021-02-06
Fix spelling errors

snd1:
This fixes several spelling errors.

Btw i also found "timout" but i think you should change it, if this wasnt on purpose ;-)

Cheers,
Dennis


jpffitch:
The .po files are automatically generated from source files in. C or c++ so they only confuse
Will look attheothers

jpffitch:
I have incorporated te speling errors

PR #1436, opened 2021-02-12
generating types.d.ts out of jsdoc

hfmanson:
using tsd-jsdoc it is possible to create .d.ts out of csound JSDoc in src/modules.

TODO: upgrade tsd-jsdoc to generate ES6 style functions `inputMessage: (scoreEvent: string) => Promise<number>;` instead of `function inputMessage(scoreEvent: string): Promise<number>;`

### DIFF
--- a/wasm/browser/package.json
+++ b/wasm/browser/package.json
@@ -60,6 +60,7 @@
     "build:release:prod": "cross-env BUILD_TARGET=release_prod rollup -c --environment production",
     "build:release": "yarn build:release:dev && yarn build:release:prod",
     "build:readme": "node --experimental-modules script/generate-readme.mjs > README.md",
+    "build:index": "jsdoc -t node_modules/tsd-jsdoc/dist -r src -d .",
     "test:ci": "node tests/selenium_runner.js",
     "test:open": "http-server ./tests -c-1 --no-dotfiles"
   },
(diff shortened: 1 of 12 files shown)


hlolli:
Did you forget to check in the resulting typings file?

hfmanson:
types.d.ts is generated by running `yarn build:index`

hlolli:
Quite impressive. My critique about the generated file, is that it's a namespace. From my point of view it means that the type and data are seperated, whereas module would be the data and the type, making things automatic. And there are also more reasons to use module instead of namespace as listed here https://michelenasti.com/2019/01/23/is-typescript-namespace-feature-deprecated.html

Also for this to work I guess the external types for events and filesystem need to be also imported, I don't see them imported. Also declare namespace CsoundObj appears at least three times. Not sure if it would render them unusable or not, but two of them are seems to be duplicates and third is empty.

How it's setup I see how if can be difficult to automate this. But you could make it be a module and fix the imports and duplication I think we should use this.

vlazzarini:
what's the status of this PR? @hlolli @kunstmusik ?

vlazzarini:
Bump. Can we have a decision on this? Is it OK to merge?

hlolli:
since this time, for more agressive dead code elimination, the codebase is now using google-closure compiler, and the jsdocs will conflict with the closure annotations, which are often identical but more often not identical. And the concept of using jsdocs to export typings is sadly something we will trade off for getting the best out of google-closure compiler.

Issue #1402, opened 2020-10-19
text formatting characters printed to non-ttys


giuliomoro:
Issue:

`print_messages()` in `Engine/insert.c` and  `csoundDefaultMessageCallback()` in `Top/csound.c` print some text formatting escape sequences to `stdout` and/or `stderr`. These escape sequences are always printed unconditionally unless we are on `WIN32`. However, when sending the output to something other than a terminal (e.g.: the console of the Bela IDE, but also the `journalctl` output for a `systemd` service), the sequences are printed as they are and no text is formatted.

Solution:

make the use of the escape sequences conditional on the return value of `isatty()`. If `0 == isatty(1)` and/or `0 == isatty(2)` then escape sequences should not be sent to `stdout` and/or `stderr`, respectively.

Question:

It's probably inefficient to call `isatty()` for every call to the print functions, so where is the best place to cache the return value for `isatty(1)` and `isatty(2)`?

giuliomoro:
bump!

vlazzarini:
Sorry I did not see this ticket before.

You could add a member to CSOUND (at the end of the structure) and cache it there. Probably the best solution. 

The alternative would be to create a global engine var and store it there, but I think not the optimal solution in this case.

jpffitch:
I too did not see this.

Yes a field in CSOUND structure is the obvious way, assuming isatty is nor a simple lookup

giuliomoro:
thanks, will submit a patch.

jpffitch:
Did you send the patch?

giuliomoro:
I did right now, thanks for the reminder. https://github.com/csound/csound/pull/1443

kunstmusik:
I haven't looked into this, but this seems related to the -+msg_color=boolean flag? Also wondering if there are situations where one wants the explicitly have the formatting characters (i.e., csound embedded in an IDE).

giuliomoro:
> this seems related to the -+msg_color=boolean flag?

It may well be ? I am not familiar with the code base or with using Csound. I tried to undo my change and add 

```
<CsOptions>
-+msg_color=0
</CsOptions>
```

and I still get printed such as:

```
[mUnifiedCSD:  /root/Bela/projects/exampleTempProject/_main.csd
``` 

Happy to hear about a better way for achieving what I am aiming for (or the correct way of using the `msg_color` option.

PR #1443, opened 2021-03-07
Prevent writing formatting characters to non-ttys. 

giuliomoro:
Tested on Linux.

Closes https://github.com/csound/csound/issues/1402


giuliomoro:
Maybe a more elegant solution is to have an `else` instead of an early return.

PR #1448, opened 2021-03-09
better formatting when listing devices in alsa; proper search in bpf …

### LINKED ISSUE #1447: ftprint does not print at krate
i assumed that the following code would print every second the provided table. (i made this example as small as possible)

```
<CsoundSynthesizer>

<CsOptions>
-odac
</CsOptions>

<CsInstruments>
sr        = 48000
ksmps     = 32
nchnls    = 2
0dbfs = 1
instr 1
  itab = p4  
  ktrig metro 1
  ftprint itab,ktrig
endin

</CsInstruments>

<CsScore>
f10 0 8 -2  0 1 2 3  4 5 6 7
i1 0 20 10
e

</CsScore>
</CsoundSynthesizer>
```
 but it is only printed once.

system: linux ubuntu

0dBFS level = 32768.0
--Csound version 6.15 (double samples) Mar  9 2021
[commit: e6c9ce1d1f20cb2a688402be089961f3583d2b37]
libsndfile-1.0.28
UnifiedCSD:  ftprint.csd
STARTING FILE
Creating options
Creating orchestra
closing tag
Creating score
rtaudio: ALSA module enabled
rtmidi: ALSA Raw MIDI module enabled
Elapsed time at end of orchestra compile: real: 0.002s, CPU: 0.002s
sorting score ...
	... done
Elapsed time at end of score sort: real: 0.002s, CPU: 0.002s
0dBFS level = 1.0
orch now loaded
ALSA output: total buffer size: 1024, period size: 256
writing 512 sample blks of 64-bit floats to dac
SECTION 1:
ftable 10:
ftable 10:	8 points, scalemax 7.000
       _

      -
     _

    -
   .

  '
 .
________
new alloc for instr 1:
ftable 10:
   0: 0.0000 1.0000 2.0000 3.0000 4.0000 5.0000 6.0000 7.0000 

may be there is a trick but the manual says:
ktrig - The table will be printed whenever this value changes from 0 to possitive. Can be used together with metro to print at a given time interval. A value of -1 indicates to print at each k-cycle (default = 1)

gesellkammer:
…(bisect); fix bug in ftprint where trigger was not working as advertised (fixes #1447)

### DIFF
--- a/InOut/rtalsa.c
+++ b/InOut/rtalsa.c
@@ -603,6 +603,16 @@ static void list_devices(CSOUND *csound)
     csound->Free(csound, line_);
 }
 
+static void trim_trailing_whitespace(char *s) {
+    int i = 0, index = -1;
+    while(s[i] != '\0') {
+        if(s[i] != ' ' && s[i] != '\t' && s[i] != '\n')
+            index = i;
+        i++;
+    }
+    s[index+1] = '\0';
+}
+
 int listDevices(CSOUND *csound, CS_AUDIODEVICE *list, int isOutput){
 
   IGN(csound);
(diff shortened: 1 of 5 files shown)


kunstmusik:
I loaded up a VM with Linux Mint and compiled and tested this. As far as I can see it looks good to me. I don't know if there are any API clients making any assumptions about the device name length but, if so, I'd say they should fix to follow this change.  Can't speak to other changes for bpf but will assume the author knows best and that this code is backwards compatible. Would prefer either smaller feature-specific commits rather than multiple feature changes in one PR or adding @gesellkammer for direct commit access to Csound to simplify contributions if that's of interest. 

Merging now.

PR #1455, opened 2021-03-28
fix bug in named_instr_assign_numbers

gesellkammer:
This fixes a very nasty bug where adding a new named instr after a high numbered instr would result in a crash.


gesellkammer:
The fix is actually not completely correct, since both indexes are not really merged. I will correct this PR later today

kunstmusik:
Did a quick test on macOS (commandline tests, some live coding) and I didn't see any issues. Code looks good to me, merging now, thanks!

Issue #1456, opened 2021-04-06
ftgenonce string keying returns the wrong table.


jackxbritton:
I am loading samples by listing the files in a directory and lazily loading them with ftgenonce in my instrument. Unfortunately, ftgenonce doesn't seem to be keying the arguments correctly, so I get the wrong table back much of the time (the sample filenames all have different names).

The following program demonstrates the error with csound 6.15:

```xml
<CsoundSynthesizer>

<CsOptions>
-odac
</CsOptions>

<CsInstruments>

nchnls = 1
0dbfs = 1

gSamples[] directory "samples/cr-78"
instr sampler

inot init p4
ifn ftgenonce 0, 0, 0, 1, gSamples[inot], 0, 0, 0

prints "%s -> %d\n", gSamples[inot], ifn

isamplelen init nsamp(ifn) / ftsr(ifn)
xtratim isamplelen

asamp loscil 1, 1, ifn, 1, 0
out asamp

endin

</CsInstruments>

<CsScore>

r 30 INOT
i "sampler"  0 0.25 $INOT
s
e

</CsScore>
</CsoundSynthesizer>
```

I then run the following (bash):

```bash
csound -d 1456.csd 2> >(grep '\->')
```
..and get the following output. It shows that different filenames are resulting in the same function tables.

```
samples/cr-78/bongo_high.wav -> 101
samples/cr-78/bongo_low.wav -> 102
samples/cr-78/conga_low.wav -> 103
samples/cr-78/cowbell.wav -> 103
samples/cr-78/cymbal.wav -> 104
samples/cr-78/guiro_long.wav -> 104
samples/cr-78/guiro_long2.wav -> 105
samples/cr-78/guiro_short.wav -> 105
samples/cr-78/hat_bongo.wav -> 106
samples/cr-78/hat_bongo2.wav -> 106
samples/cr-78/hat_closed.wav -> 107
samples/cr-78/hat_rim.wav -> 107
samples/cr-78/hi_hat.wav -> 108
samples/cr-78/hi_hat2.wav -> 108
samples/cr-78/hi_hat_metal.wav -> 109
samples/cr-78/kick1.wav -> 110
samples/cr-78/kick2.wav -> 111
samples/cr-78/kick3.wav -> 111
samples/cr-78/kick_click.wav -> 109
samples/cr-78/kick_cowbell.wav -> 112
samples/cr-78/kick_cowbell2.wav -> 112
samples/cr-78/kick_cymbal.wav -> 113
[...]

PR #1461, opened 2021-04-09
Fix djb2 hash.

jackxbritton:
#1456 

Open to criticism, I can't reproduce any warnings. Looks like standard C to me.


vlazzarini:
thanks. It is standard C but some compilers are overzealous and think single = in a conditional may be an error. Parentheses keep them quiet.

PR #1467, opened 2021-05-01
Update of the testing facility

iskunk:
I've been sitting on these changes for a while, making use of them in my own testing. The main goals are (1) allow the test suite to run on Python 3 without losing support for Python 2, and (2) run the tests in the build tree, so that the source tree doesn't get polluted with test outputs.

Commit message body follows:

* Make scripts compatible with both Python 2 and 3

* Detect at configuration time the available Python interpreter

* Make tests run in the build tree, not in the source tree

* Don't write files directly into `/tmp`

* Use proper exit status value from `os.system()`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -441,6 +441,7 @@ find_package(FAUST)
 find_package(Java)
 find_package(JNI)
 find_package(MUSICXML)
+find_package(Python)
 find_package(VSTSDK2X)
 
 #if(MUSICXML_FOUND)
--- a/tests/c/CMakeLists.txt
+++ b/tests/c/CMakeLists.txt
@@ -56,26 +56,22 @@ add_test(NAME testCircularBuffer
 #add_executable(testCscore cscore_tests.c)
 #target_link_libraries(testCscore ${CSOUNDLIB} ${CUNIT_LIBRARY} pthread)
 #add_test(NAME testCscore
-#        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests/c/
 #        COMMAND $<TARGET_FILE:testCscore> ${CMAKE_SOURCE_DIR}/tests/c/ -arg2 ${TEST_ARGS})
 
 add_executable(testPerfThread perfthread_test.cpp)
 target_link_libraries(testPerfThread ${CSOUNDLIB} ${CUNIT_LIBRARY} pthread libcsnd6)
 add_test(NAME testPerfThread
-        WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/tests/c/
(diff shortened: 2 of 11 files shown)


vlazzarini:
@jpffitch could you review this?

jpffitch:
Why the change to not use /tmp for a temporary file?
Not knowng python I am not really able to judge the rest, and I do not use a buikd tree separate from the sorces.

It is probably OK

jpffitch:
Also why does it fail the test on `Debian?`  We need an explanaton and/or a fix

iskunk:
Hi @vlazzarini, @jpffitch,

Writing files with fixed filenames into `/tmp` is a risky practice best avoided. There is the security concern of a [symlink race](https://en.wikipedia.org/wiki/Symlink_race), and the possibility of a denial of service if a different user creates the same-named file. Generally, files in `/tmp` should be created using `mkstemp(3)` or `tmpfile(3)`, or a utility program like `mktemp(1)`, as these ensure that the filename used is unique (thus sidestepping the pitfalls).

That boilerplate aside, there isn't even a need to write into `/tmp` to begin with, since the build tree provides a more appropriate location anyway. The test output file is yet another output of a particular build, after all.

With regards to the failed build, are you talking about Travis CI build 6155.1? That appears to be Ubuntu Trusty. Is there a Debian build I'm not aware of?

With regards to the Ubuntu Trusty build that failed, I see two issues:

* CMake 3.9.2 does not have a `FindPython.cmake` module. It has an older, now-deprecated module named `FindPythonInterp.cmake`. (Note that CMake 3.9 is fairly old; the last release was in late 2017.)

* The `testPerfThread` test failed. I'm not sure what's happening here, as this is one of a handful of tests that inexplicably hangs for me (will be reported in a future issue). Given that the same test passed on a newer version of Ubuntu, could this be unrelated to my change?

vlazzarini:
I have seen one of the builds fail on testPerThread this week, but I resubmitted the job and it passed. I never have that test fail to me locally on MacOS.

jpffitch:
OK; I tink it is time to take this PR` and if the test fails we fix it from there.

I will try to accept but githu doe not like my envoronment

iskunk:
Thanks gentlemen.

@vlazzarini, I looked through some of your follow-up commits. I was surprised that Tk was giving you trouble, as it worked for both Python 2 and Python 3 for me. Could it have been failing because the tests were suddenly using Python 3, and a `tkinter` package for Python 3 was not installed? (I encountered a similar issue when I tested my changes on Python 2.)

Also, is there still a CMake option for building the Python (3) interface? My change should not have affected the interface-build portion at all; I did not want to mess with that.

(Please let me know if I can answer any further questions)

vlazzarini:
That's all fixed now, Steven removed tkinter. 

The Python2 interface is a separate issue, it needed to go, and in responding to another open issue, I removed it. It is replaced by ctcsound.

PR #1471, opened 2021-05-04
Copy tests/soak/nsliders.tk to builddir to fix three tests

iskunk:
The `checkbox`, `flashtxt`, and `setctrl` tests need this file in the current directory in order to complete successfully, and now that tests can run in a separate builddir (PR #1467), this will ensure the file continues to be accessible in that scenario.

----

(Now that the big test PR is done, I'm focusing on individual tests.)

As a completely separate aside, by the way, the following tests fail due to `fox.wav` not being present in `tests/soak/`:
```
poscil3-file.csd
pvsftr.csd
pvslock.csd
pvstanal.csd
vpow_i-2.csd
```


vlazzarini:
ok, thx for these.

Issue #1249, opened 2020-02-04
Search paths for plugins


gesellkammer:
As discussed in the dev list, csound should search for plugins at multiple paths by default. In macos, if csound is installed via installer, plugins are searched in `/Library/Frameworks/CsoundLib64.framework/Versions/6.0/Resources/Opcodes64`, whereas if installed from source plugins are searched at `~/Library/...`. Both paths, and probably a third one, something like `~/Library/Csound/Opcodes64` should be searched by default. In linux, both `/usr/local/lib/csound/plugins64-6.0` and `/usr/lib/csound/plugins64-6.0` and maybe also `~/.local/lib/csound/plugins64-6.0`. There should be at least one location where a user can install plugins without super-user permissions.

vlazzarini:
There are new API functions and flags in Csound to allow for this. For users, they can now load all plugins from a given directory via

```
--opcode-dir=DIR        load all plugins from DIR
```

gesellkammer:
I would strongly suggest doing the opposite: there should be a default for each platform, and maybe a flag to prevent using that default, for the niche case where a pro user needs to test multiple versions/installations. Supercollider works this way, the pd application has settings to do the same. Without this any package manager for plugins would be very brittle.

vlazzarini:
There is a default for each platform already; on OSX it is the framework; elsewhere it is defined by CMAKE_INSTALL_PREFIX.
These can be overridden by OPCODE6DIR64. The code search currently does not look into more than one directory,  but
it can be modified. Last time I looked it appeared more involved than expected, but it can be done.

The flag I added is useful for anyone to load their custom plugins without having to have a separate option for each library.
It is a different use case altogether.

gesellkammer:
The request is explicitly to have a default path for plugins at the user level (ie, without admin rights) and without having to touch OPCODE6DIR64. Notice that setting OPCODE6DIR64 to a wrong value will prevent csound from finding its own plugins and rt modules. I can take a look and do a PR, but I need to reassurance that this is a desired feature.

vlazzarini:
But OPCODE6DIR64 is user level. Not sure what you are asking? It's a shell variable, no need for admin rights whatsoever.
Users can set it and unset it as they wish. We do not need another feature.

gesellkammer:
If that variable points to something other that the default or does not include the default, csound fails to find its own plugins. This is very error prone for inexperienced users and I can't imagine building a package manager which relies on inexperienced users doing complex things.

[...]

PR #1473, opened 2021-05-09
Add a user folder to the paths searched for plugins

gesellkammer:
This is a draft implementing #1249. It includes a new define in the cmake file, "CS_DEFAULT_USER_PLUGINDIR", which is a relative path to a folder within the users home directory. In linux and mac the prefix is $HOME, in windows it is %LOCALAPPDATA%. This is defined in the cmake file because we need access there to APIVERSION. 

In csmodule.c we add the user dir to the search path if this macro is defined. If the user defines OPCODE6DIR64, the path
given there overrides everything, including the user plugins dir.

### DIFF
--- a/Top/csmodule.c
+++ b/Top/csmodule.c
@@ -401,6 +401,7 @@ static int csoundCheckOpcodeDeny(CSOUND * csound, const char *fname)
     return 0;
 }
 
+
 /**
  * Load plugin libraries for Csound instance 'csound', and call
  * pre-initialisation functions.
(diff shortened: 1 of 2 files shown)


gesellkammer:
With the given implementation and current api version, the user plugin dir would be:

* Linux: `~/.local/lib/csound/6.0/plugins64`
* Mac: `~/Library/csound/6.0/plugins64`
* Windoes: `C:\Users\<User>\AppData\Local\csound\6.0\plugins64`

Top/csmodule.c
```diff
@@ -451,10 +451,45 @@ int csoundLoadModules(CSOUND *csound)
...
+#if defined(LINUX) || defined(__MACH__)
+    char *prefix = getenv("HOME");
+#elif defined(WIN32)
+    // C:\Users\<User>\AppData\Local\csound\6.0\plugins64
+    char *prefix = getenv("LOCALAPPDATA");
+#endif
+    int pos = strlen(dname);
+    int prefixlen = strlen(prefix);
+    if(pos + prefixlen + 2 > _buflen - 1) {
+      csound->Message(csound, "plugins search path two long\n");
```
kunstmusik, line 472: typo: "two" should be "too"

kunstmusik:
@gesellkammer Are there any further changes expected? If not, seems good to merge to me.

gesellkammer:
There are no further changes from me, but it was very superficially tested in other platforms other than linux

kunstmusik:
No worries, I think it'll be easier for everyone to test and find any issues once it's merged.  Will do so now, thanks!

PR #1494, opened 2021-06-24
Add an optional prefix to soundfont instruments printed via sfilist

gesellkammer:
This adds an optional prefix to the instruments listed by sfilist (to list instruments in a soundfont). The use case is to make the printed lines easier to parse.

```
gisf	sfload	"sf_GMbank.sf2"
	sfilist	gisf, ">>> "
```


jpffitch:
Does this need a manual change?

PR #1509, opened 2021-08-04
VCPKG/CI improvements

stekyne:
VCPKG has come on quite a bit since we initially adopted it. It now has manifest files (like package.json for NPM), versioning, binary caching, support for other platforms than Windows.

This PR is to clean up the whole VCPKG integration and use the best practices advised. This is also important as we had a number of VCPKG related build breakages last year which highlighted how unstable it can be in it's current form.

I'd like to remove appveyor as the Windows CI and just use Github Actions for all platforms. It's much easier to use one system to orchestrate releases etc. Github Actions previously didn't have caching folders but I think this has been added. The VCPKG binary caching feature should remove this need anyways or possibly used along side it.

With the manifest file, the call to build the project will automatically ensure the dependencies listed will be installed. This removes the need for a Windows specific build script (downloadDependencies.ps1) and can be used in a crossplatform manner. Regardless of updating VCPKG, the dependencies should stay at the same level due to the versioning requirements in the manifest. This should make the builds more consistent over time and reduce/eliminate the breakages we saw last year.

Binary caching means vcpkg will check with a cached location (local folder or Nuget repo) to see if it can pull down prebuilt packages or rebuild in the CI if needed and then update the cache. We can specifiy a Github repo for nuget which will be the official source of the dependencies. This also means that devs on their local machines can pull down these cached versions and avoid building locally. In essense, the dependencies will be built once and then reused across CI and local builds which should be much quicker for everyone.

VCPKG is supported on Linux and MacOS now. This means we could build releases which have the exact same sources for dependent libraries across all platforms. All built and packaged in the one CI and further automate our releases in general. This should also minimise potential variable behaviour seen by using different versioned dependent libs on different platforms. It's unclear if this is desired yet or to stick with the status quo but it should be easy to add at least and experiment.

Tasks: 
 - [x] Add manifest file for packages
 - [x] Add versioning to manifest file to lock dependencies
 - [x] Add VCPKG as a submodule (advised for CMake integration) + CMakeList file changes to integrate
 - [x] Added Github Action build for Windows
 - [x] Add binary caching via Nuget (for Windows at least)
 - [ ] Add optional build support for Linux and MacOS
 - [x] Remove appveyor build
 - [x] Add custom portaudio port for asiosdk


vcpkg.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "csound",
+    "version-string": "6.16.2",
+    "dependencies": [
+        "eigen3",
```
kunstmusik, line 5: Are these libaries being installed as static libraries for linking?

CMakeLists.txt
```diff
@@ -1,14 +1,22 @@
 cmake_minimum_required(VERSION 3.5)
+
 # silence RPATH cmake warning
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
+# Use vcpkg submodule toolchain by default
+set(VCPKG_OVERLAY_TRIPLETS ${CMAKE_CURRENT_SOURCE_DIR}/msvc)
```
kunstmusik, line 8: Will this work okay on non-msvc systems?

.github/workflows/csound_builds.yml
```diff
@@ -1,65 +1,84 @@
...
 
-on: 
-  push:
-    branches:
-      - develop
-      - master
-    pull_request:
-      branches:
-        - develop
+on:
```
kunstmusik, line 3: Any particular reason to reformat this to 4-spaces?

.github/workflows/csound_builds.yml
```diff
@@ -1,65 +1,84 @@
...
+            - name: Checkout Source Code
+              uses: actions/checkout@v1
+              with:
+                  fetch-depth: 1
+            - name: Install system dependencies
+              run: |
+                  choco install swig -y --version=4.0.1 --allow-downgrade
+                  choco install winflexbison -y
+            - name: Configure build
+              run: cmake -B build -S . -DVCPKG_TARGET_TRIPLET=x64-windows-csound
```
kunstmusik, line 80: Will the TRIPLET flag be required for anyone building on Windows?

kunstmusik (commented):
To note, I had experimented with vcpkg on MacOS a couple months ago and found that I had issues with building universal binaries (i.e., x86_64 and arm64). It seemed there might have been a way to get around it by building the app twice and having separate vcpkg caches, as it seemed to delete the cache when I switched between arm64 and x86_64, then manually creating the universal binary via lipo (similar to what we do with iOS builds).  I did not take it further than that but wanted to mention it.

vcpkg.json
stekyne: Whether they're static or dynamic is based on the triplet. We have a custom triplet file "x64-windows-csound" that installs some as static and some as dynamic. We can/must have a custom triplet for each platform.

CMakeLists.txt
stekyne: This is temporary. We'll probably have the triplet files in the root dir or else a custom folder.

I was thinking of adding a "platform" dir and we could push in all the build stuff into that. It's quite busy atm with existing ones e.g Android, Bela, Emscripten, mingw64, mingw64-linux, msvc.

.github/workflows/csound_builds.yml
stekyne: Sorry my autoformatter went nuts. I'll correct it now.

.github/workflows/csound_builds.yml
stekyne: Yeah it will be required. This is also temporary. I will try to deduce in CMake what it should be but it was difficult as some env vars aren't set when this needs to be configured (before first project() cmake call). 

There should be a programmatic solution to pick a sane default and be allowed to be overriden via command line or env var.

stekyne:
> To note, I had experimented with vcpkg on MacOS a couple months ago and found that I had issues with building universal binaries (i.e., x86_64 and arm64). It seemed there might have been a way to get around it by building the app twice and having separate vcpkg caches, as it seemed to delete the cache when I switched between arm64 and x86_64, then manually creating the universal binary via lipo (similar to what we do with iOS builds). I did not take it further than that but wanted to mention it.

It should cache them separately as they are using different triplets. This should work by default now but it's possible you were using vcpkg before this was implemented. I will try and take a look but currently don't have access to a mac. I'll see about it. That part of this PR may get split off to another piece of work. Maybe it's best to focus on Windows for now.

stekyne:
@kunstmusik Maybe there is enough work done to merge this particular PR. I've got the Windows build working same as it does for Appveyor currently. The installer and zip output should be the same and local testing suggests that is the case. Once this is merged into develop, we can switch off and remove appveyor completely. I'll remove the associated files in another PR then.

I've left the Linux and Mac builds for now due to some issues with ports in VCPKG not currently working. I have a PR in VCPKG to fix liblo at the minute. https://github.com/microsoft/vcpkg/pull/19660 

The binary caching for VCPKG is working well though and it's a good improvement by itself. The build time might be cut down even more with some more GH action caching and configuration changes.

kunstmusik:
I checked out the pr branch and tested here on Win10. Using the same cmake commands to configure and build, I was able to create a Csound build.  Perhaps documenting the git submodule and cmake commands to use for general dev on Windows would be useful. Couple things:

* VCPKG tends to get really slow while installing dependencies (seems to slow down as it goes).  Tooks hours here to install. I have only 8gb RAM which I wonder is a factor?  Also, perhaps windows defender?

* With these changes, seems we can remove the msvc folder. I want to double check with you though, are all of the libraries which were statically linked still statically linked in this version?

* Going to go ahead with merging as it seems safe to do so and new PR can address any outstanding issues. Thanks!

vlazzarini:
The only issue is that Appveyor is failing. Perhaps this should have been addressed before the merge.

kunstmusik:
> The only issue is that Appveyor is failing. Perhaps this should have been addressed before the merge.

The github action builds an installer and should be available to anyone who is logged into Github. You can see it on this page:

https://github.com/csound/csound/actions/runs/1165292998

and it will be a part of future runs too. I haven't tested the installer, but assuming it's got everything, I think we can dump appveyor.

vlazzarini:
Ok, so I presume we will use this for the next release then.
If that's all good, can we stop requesting builds from appveyor and remove checks/badges? 
It is showing as failing in the github site.

vlazzarini:
I've commented out the README.md badge. Not sure how to stop triggering builds and displaying CI checks.

vlazzarini:
I went ahead and deactivated the webhook and removed the appveyor integration for csound/csound repo. It's still there for the manual. Now the checks are only coming from the github actions.

stekyne:
> I checked out the pr branch and tested here on Win10. Using the same cmake commands to configure and build, I was able to create a Csound build. Perhaps documenting the git submodule and cmake commands to use for general dev on Windows would be useful. Couple things:
> 
> * VCPKG tends to get really slow while installing dependencies (seems to slow down as it goes).  Tooks hours here to install. I have only 8gb RAM which I wonder is a factor?  Also, perhaps windows defender?
> * With these changes, seems we can remove the msvc folder. I want to double check with you though, are all of the libraries which were statically linked still statically linked in this version?
> * Going to go ahead with merging as it seems safe to do so and new PR can address any outstanding issues. Thanks!

That's strange alright for the slowness. It should take like 40 mins to build all the dependencies give or take. 8gb of RAM might be a factor but for sure I've had to add exclusions for Windows defender as it does end up using a lot of CPU time. The good thing is that this is only done once even locally. Each package version is cached locally in the user directory so it's external to the VCPKG repo. If you create a new VCPKG dir and install the same packages, they'll just get copied across. This is way beter than what we had before and should prevent all these rebuilds that were required before.

The static/dynamic linking of the dependencies is the exact same. We're using the same "x64-windows-csound" custom triplet that we used in the old appveyor build.

I'll remove the MSVC folder now and update the readme.md file.

One last thing, I've adopted this "platform" folder. I think it would be worth pushing in all the platform stuff into here to reduce the number of folders in the root directory. It also makes sense to group them anyway.

@vlazzarini We definitely should remove Appveyor before continuing. It's redundent at this stage and causing build failures. I can test out the binaries a bit more but I checked the number of files and they match up. Also running a few examples worked fine.

vlazzarini:
@stekyne that's done now (I think 😀)

PR #1514, opened 2021-08-16
Fix jack autoconnect

gesellkammer:
This PR fixes the following issues when using jack as the rt audio backend:

* csound could crash if given an invalid dac/adc value. Now it prints a warning and does not autoconnect
* when given a dac pattern (-odac:pattern) it would try to fill that pattern with numbers, from 1 to the number of channels in the .csd. This would make it impossible to autoconnect to ports starting at 0 or ports using other naming mechanisms. Now this pattern is used as a regular expression passed to jack_get_ports (https://jackaudio.org/api/group__PortSearching.html) and the connection is tried with the ports matching the given regex. Ports are sorted via a natural sort routine (so that a port_in10 comes after port_in2) replicating the previous behaviour for client presenting numbered ports. This fixes an issue under pipewire where the ports order can be modified during the lifetime of a client.


vlazzarini:
There is a problem with natsort.c licence, it's GPL by the looks of it. We use LGPL and that will not work (all our code will become GPL if this goes in).

gesellkammer:
I found a replacement released under the MIT license (http://www.davekoelle.com/files/alphanum.hpp), would that work with csound's lgpl?

vlazzarini:
I think so, since the MIT license is more permissive than the LGPL.

jpffitch:
a quick question: how does alphanum_isdigit differ from the required C function isdigit ?

⁣Get TypeApp for Android ​

gesellkammer:
Honestly I have not dived in the depths of locales across all platforms, but as far as I could follow, isdigit is locale dependent in windows. Since this routine is used to sort port numbers, which are not expected to follow any locales, I removed any locale dependent code from the original implementation  (original here: http://www.davekoelle.com/files/alphanum.hpp)

jpffitch:
Ok.  Never had much to do with locales

⁣Sent from TypeApp ​

PR #1518, opened 2021-08-30
Increase stack size to prevent heap corruption

docEdub:
I ran into an issue when using Csound WASM in the Oculus Quest 2 browser with large .csd files. The stack appears to be expanding into the memory used as the heap. This causes strange error messages to be reported by the browser and/or Csound depending on how much of the heap gets corrupted.

This change fixes the issue by reserving 4mb of memory for the stack. It is maybe not the most correct fix, but it works.


kunstmusik:
Seems like we should go with it and can revisit in the future. Thanks for the contribution!

PR #1519, opened 2021-09-03
Implement stack size more correctly and change it from 32mb to 128mb

docEdub:
After investigating the issue PR https://github.com/csound/csound/pull/1518 tried to fix, I learned that the WASM stack grows down into the code section of memory, not up into the heap! This change adjusts the memory setup accordingly, increasing the stack from 32mb (512 * 64k for each page) to 128mb, to accommodate large .csd files.

This change also sets the `WebAssembly.Memory` `maximum` argument to 1gb so low-powered devices use less CPU when dynamically growing the heap. This has a noticeable effect in the Oculus Quest 2 browser.

### DIFF
--- a/wasm/browser/src/module.js
+++ b/wasm/browser/src/module.js
@@ -7,8 +7,8 @@ import { dlinit } from "@root/dlinit";
 import { csoundWasiJsMessageCallback, initFS } from "@root/filesystem/worker-fs";
 import { logWasmModule as log } from "@root/logger";
 
-const BYTES_PER_MB = 1048576;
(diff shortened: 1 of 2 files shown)


docEdub:
We should probably expose a stack size argument in the Csound constructor so users can make it larger or smaller as needed instead of being stuck with whatever is hard-coded in `module.js`.

docEdub:
We might also want to write a checkerboard pattern at the end of the stack and check it periodically to see if it has been overwritten. That would let us know a stack overflow occurred.

PR #1522, opened 2021-09-07
env variable to avoid extra printing

gesellkammer:
Csound prints some output (0dbfs value, version, performance stats) even if all messaging has been disabled.  This is particularly undesired when using csound as a library, where the end user might not even be aware that csound is in use and is confronted with this spurious output. This PR adds an env variable "CSNOMESSAGES" which, when defined, disables these messages. If the variable is not defined, csound's behaviour is unaffected.


vlazzarini:
@jpffitch  are you happy with this solution for the problem? This printing seems to be before options come into effect. 
I have no major issues but the solution feels a little awkward.
There's only a few messages covered by this, maybe they should come later and be subject to options?

rorywalsh:
This a great, an options flag would be even better. It would mean we could also disable these messages in the web build.

gesellkammer:
The problem is that some of these messages are printed before options are read. If the printing can be moved then a flag already exists and the only thing to add would be a check before each of such messages that msglevel > 0

vlazzarini:
yes, I think moving is a better option. I have done this now and also silenced most typical messages with -m0. There are others to be caught now (like error messages etc), but we can suppress them easily now with `if (O->msglevel || O->odebug)`. I have committed that and I'll close this PR. Thanks for starting this discussion.

gesellkammer:
thanks!

PR #1523, opened 2021-09-08
Some fixes regarding comman line options and conditional  printing

gesellkammer:


gogins:
Thanks for this work,
Mike

Michael Gogins
Irreducible Productions

PR #1524, opened 2021-09-10
Stop writing out of bounds of buffer allocated in `allocFloatArray`

docEdub:
The `allocFloatArray` function currently allocates the bytes needed to hold the given number of doubles, plus 1 byte for a null terminator. When setting the null terminator byte, the index into the array is calculated incorrectly and ends up zeroing out a byte outside of the allocated buffer. This causes WASM errors to be displayed in the console and causes occasional pops and clicks in the audio output.

This change fixes the issue by not adding a null terminator in `allocFloatArray` since it is not needed.


docEdub:
Closed this PR because it was accidentally included in PR https://github.com/csound/csound/pull/1519, which has been merged.

PR #1532, opened 2021-09-20
VCPKG build for linux in CI

stekyne:
Testing out using VCPKG for linux in the CI.


stekyne:
@vlazzarini @kunstmusik Hey guys, I have VCPKG working for Linux builds whilst still keeping the existing package manager build. I think it might be a good way to get a consistent build. Using the same version of dependencies but also uniform access to the same version that some distros might not have. I'll get it working for mac next and this will probably cover a lot of that.

Issue #1542, opened 2021-11-08
Ensure greater consistency of csound.hpp with csound.h


gogins:
- [x]  There is no SetAudioChannel. _Added it to Csound.hpp._
- [x]  There is no SystemSr. _Added it to Csound.hpp._
- [x]  SetDebug is supposed to return a value. _**Not in csound.h**!_ _Removed the return value from Csound.hpp._

gogins:
These changes will be inherited by `CsoundThreaded` and other classes that inherit from `Csound`.

The only potential breaking change is if some C++ client expects `Csound::SetDebug` to return a value. I am not aware of any such usage in my own code and I probably use this stuff more than anybody.

PR #1543, opened 2021-11-08
#1542.

gogins:
Noticed a few inconsistencies between Csound.hpp and csound.h, which are fixed in this PR. There may be more inconsistencies.

Two changes add missing virtual methods at the end of the csound.hpp declaration. These should not cause any breakage.

One change removes an (unused) return value from Csound::SetDebug. This would be a breaking change if anybody was using this return value, but I'm not aware of anybody actually using this return value.

### DIFF
--- a/include/csound.hpp
+++ b/include/csound.hpp
@@ -928,6 +928,12 @@ public:
   {
     return csoundGetInputName(csound);
   }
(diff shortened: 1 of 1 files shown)


gogins:
All GitHub Actions builds are now succeeding, I will merge this PR and see if that is still the case.

PR #1575, opened 2022-01-16
Remove packages for code that has been moved out

stekyne:
Since we moved some opcodes out into the external plugins repo, we were still installing some VCPKGs which slows the build down a fair bit. Can safely remove them now.


stekyne:
Passing so I'll merge this in.

PR #1582, opened 2022-02-07
Update BUILD.md

joachimheintz:
Make sure to checkout master.


vlazzarini:
This is now done in #1940. It will be a work in progress. We have a fresh set of instructions for all the basic platforms, to which others can be added.

PR #1583, opened 2022-02-07
Updated installer to pick up VS2022 redist files

stekyne:
The GH agent for Windows has updated to the latest Windows and VS (2022). The paths for the redist files have therefore changed. All that is required is to update the GH workflow. The build appears to be working with no other changes for 2022.

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -63,7 +63,7 @@ jobs:
 
   windows_build:
     name: Windows build
-    runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


stekyne:
@vlazzarini This has fixed the windows build but the wasm one seems broken. Someone should maybe take a look at it.

vlazzarini:
yes thx @stekyne I'm aware of it, I have actually disabled that for the develop branch until someone takes a look.
Thx for taking care of this.

PR #1593, opened 2022-02-25
Fix midi devices listing and typo in csound

gesellkammer:


vlazzarini:
This should be fine but we cannot do a PR into master. If this is for Csound6 only, it should be done in the csound6 branch.
Master should only change  when we release.

kunstmusik:
I don't think this needs to be closed, just the target branch modified so I am reopening it. I will switch it to csound6 branch.

vlazzarini:
there's another PR

PR #1594, opened 2022-02-25
Fix midi list printing and typo in ctcsound

gesellkammer:


vlazzarini:
thanks, Eduardo. Sorry for the trouble.

PR #1608, opened 2022-05-24
fix qnan even when compiled with fast-math

gesellkammer:
Fix qnan even when compiler with -ffast-math (fast-math enables the assumption that math is always finite, so isnan is optimized to false).


jpffitch:
All systems could benefit from this. Is it is patch againstv6 or v7?

gesellkammer:
The file has not changed in v7 so it could be applied to that branch also

PR #1617, opened 2022-08-06
Csound6 branch: Fix for windows build

stekyne:
Some symbols were clashing in the build as we are using static libs. Renamed some of the CS ones to be different from internal libmp3lame (or whatever it is).


vlazzarini:
@stekyne thanks, we'll need these done to v7 in the develop branch too.

PR #1620, opened 2022-08-17
Bela changes and improvements

giuliomoro:


vlazzarini:
Hi Giulio,

Thanks for this. The develop branch now is the long-term new-version (7.0) development code (hopefully by the end of the year), it's perhaps what you'd call bleeding edge.

Would you do a similar PR to the csound6 branch as we're about to release 6.18? We're keeping 6.x as bugfix releases but I think they would benefit from these changes.

giuliomoro:
done thanks

PR #1621, opened 2022-08-17
Setting GIT_HASH_VALUE at build time

giuliomoro:
I find it more useful to get the git hash value at build time than at configure time. This way you ensure the commit is correct (though it would also be nice to check for "dirty" (i.e.: untracked changes) in the repo).


giuliomoro:
it may needs some quotes, let me give it a try (don't have a windows build machine so I'll just push here and see if it works)

giuliomoro:
yeah no didn't work

vlazzarini:
This PR has became stale. I'm closing it; if the changes are still needed, a new PR against current code is the best way forward.

PR #1640, opened 2022-10-11
Add Chocolatey package

### LINKED ISSUE #1364: Submit csound to Chocolately to have available for installation on Windows
We should look at submitting Csound to https://chocolatey.org/ so that users can install easily using choco as well as developers who might want to build against Csound.

stekyne:
Should resolve #1364

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -234,3 +234,5 @@ html
 manual.zip
 dist
 xcode
+*.nupkg
+*.exe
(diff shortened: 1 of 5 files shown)


kunstmusik:
@stekyne : double checking that this should this be merged to develop and not csound6 branch?

stekyne:
@kunstmusik I guess it could be either/both. It's really just to get the code in the repo. You will still need to upload manually so it doesn't make much of a difference for now. I can rebase to csound6 if they get back soon and accept the package but it could be another few weeks.

PR #1653, opened 2022-11-08
Merge csound6 6.18.0 into latest from develop

kunstmusik:
This merges changes for 6.18.0 to develop (csound7). There were a number of conflicts I looked and resolved, would like others to check out before merging into develop. Some concerns:

* 3 unit tests related to structs are failing. These are also failing in latest from develop so it seems unrelated to the merged code from csound6.
* I'm unsure I did the merges correctly for vcpkg and CI builds correctly. @stekyne could you take a look and check if I messed up anything there?
* @hlolli There's a unit test failing for browser tests related to TableCopyIn. Could you assist in looking at this one?


hlolli:
looks a bit like a regression in `tableCopyIn`, I need to compare with the c-api

hlolli:
I pushed a fix for the channel tests

hlolli:
it now fails on `error Couldn't find any versions for "@csound/wasm-bin" that matches "7.0.0-alpha1"`
need to check the yarn linking

hlolli:
@kunstmusik I'm fixing this in a hacky way, by not having yarn lookup non existing dependency, since we didn't push 7.0.0-alpha1 yet. But we should push it as soon as this is merged to prevent this becoming issue.

hlolli:
we may want to follow this tutorial to be extra sure that we are resolving csound6 by default

https://kevinkreuzer.medium.com/publishing-a-beta-or-alpha-version-to-npm-46035b630dd7

hlolli:
wasm errors gone now !

kunstmusik:
> wasm errors gone now !

Awesome, thanks for the quick fix!  I think we just need to fix up the Windows CI build and get a double check from @jpffitch or @vlazzarini and we can merge this to develop.

stekyne:
@kunstmusik it's fixed now. Updated vcpkg to latest version. The git has has to align with the vcpkg baseline otherwise it complains. Also the installer name is now different so those are both updated now.

stekyne (approved):

kunstmusik:
> @kunstmusik it's fixed now. Updated vcpkg to latest version. The git hash of the vcpkg submodule has has to align with the vcpkg manifest baseline otherwise it complains. Also the installer name is now different so those are both updated now.

That's fantastic, thanks very much for fixing this!  

@vlazzarini or @jpffitch : If you would review this PR and are okay with it, would appreciate it. Seems like it gets what we need from 6.18.0. If you're too busy, I feel good enough about the PR to go ahead with merging.

vlazzarini:
Sorry, I started looking at thos when it came in but never finished. I think it looks ok from my perspective.

kunstmusik:
> Sorry, I started looking at thos when it came in but never finished. I think it looks ok from my perspective.

Thanks @vlazzarini, I've gone ahead with merging the PR. We should be good to go now for CS7 dev on develop.

PR #1657, opened 2022-11-19
Add renderer to worklet+singlethread setup

hlolli:
Currently, rendering goes in sync with the realtime rendering. Maybe we'll need this as well for scriptProcessor setup. And maybe it was me that removed it, forgetting to re-add this mechanism.


kunstmusik (approved):
We'll need to do a larger audit/refactor of all this soon, but PR looks good to fix the issue.

PR #1660, opened 2022-12-11
allow inline struct-array property getters

hlolli:


hlolli:
these diffs have been merged to another PR https://github.com/csound/csound/pull/1682

PR #1663, opened 2022-12-13
feat(test): add opcode assert for testing

hlolli:
Currently the csdtests rely on exit code of the performance, this makes writing unit tests bit bad. So this is a suggestion that we can expand on for writing assertions in tests, and more accurately test and assert multiple times in a single csd.


vlazzarini:
@hlolli if you can resolve these conflicts then I can go ahead and finish reviewing and merging.

PR #1666, opened 2022-12-14
[WIP] chnget/chnset variants for structs

hlolli:


hlolli:
good idea, but given recent rework of channels I'l revisit this later

Issue #192, opened 2014-03-24
for loop [sf feature-requests #24]


kunstmusik:
**Originally reported by Sourceforge User: joachimheintz**

i suggest to have a c-like for loop (i- and k-rate), with this syntax (here i-rate as example):
for iIndx=0, iIndx<iMax, iIndx++
...
endfor

thanks for considering!

hlolli:
for-each in csound7?

hlolli:
I'm going to attempt to make for-of and for-in as they appear in javascript, which would be similar to @vlazzarini's suggestion

for-of

```csound
for k1 of karray do
  printk2 k1
od

for i1 of iarray do
  print i1
od

for iarr of itwodimensional do
  for i1 of iarr do
    print i1
  od
od
```

and for-in

```csound
for Smember in mytype do
  prints "member-name of struct %s\n", Smember
od
```

kunstmusik:
Closing as completed in https://github.com/csound/csound/pull/1670

PR #1670, opened 2022-12-21
feat: for-in array iterator

hlolli:
From a feature request from 2014
https://github.com/csound/csound/issues/192

EDIT: changed from for-of to for-in after discussion on the mailing list

### DIFF
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -101,6 +101,7 @@ SYMBOL          [\[\]+\-*/%\^\?:.,!]
 %x xstr
 %x udodef
 %x udoarg
+%x forloop
 
 %%
 <*>"\r"            { } /* EATUP THIS PART OF WINDOWS NEWLINE */
@@ -177,12 +178,6 @@ SYMBOL          [\[\]+\-*/%\^\?:.,!]
 "enduntil"      { *lvalp = make_token(csound, yytext);
                   (*lvalp)->type = OD_TOKEN;
                   return OD_TOKEN; }
-"for"           { *lvalp = make_token(csound, yytext);
-                  (*lvalp)->type = FOR_TOKEN;
-                  return FOR_TOKEN; }
-"in"            { *lvalp = make_token(csound, yytext);
-                  (*lvalp)->type = IN_TOKEN;
-                  return IN_TOKEN; }
 
 "goto"          { *lvalp = make_token(csound, yytext);
                   (*lvalp)->type = GOTO_TOKEN;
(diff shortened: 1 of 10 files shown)


Engine/csound_orc.lex
```diff
@@ -130,7 +130,7 @@ SYMBOL          [\[\]+\-*/%\^\?:.,!]
 "|"             { return '|'; }
 "&"             { return '&'; }
 "#"             { return '#'; }
-"�"             { return '~'; } /* \xC2?\xAC */
+"¬"            { return '~'; } /* \xC2?\xAC */
```
stekyne, line 133: does that character look right?

hlolli: it's so strange, I've tried to fix it few times, I just can't understand what's up with the encoding, just for sanity check, could you test locally if you can replace it with the original char, because every editor I use create the same diff, and more strange even, is that when I look at your comment, I see this

when I look at the diff I see this strange thing there

when I look at the file alone, it again, looks correct

stekyne: image
It looks like the old version is wrong but it didn't look like that first time 😓 I don't really understand. Locally in vscode it looks fine to me. I guess ignore it for now

stekyne: image
Ah ok, it doesn't render correctly the previous version locally. So the diff looks correct in my screenshot above.

hlolli: could you maybe if you have time, make a PR (or just push to my branch) for the correct character? I wonder if something magical happens when it's correct from your text-editor, afterwards, I will for science, open the file and resave just to see if my editors are still being annoying on this, if that's the case, I'll just not touch this file or dig deeper.

kunstmusik: I'm unsure of the history here, seems like "~" became an alias for "¬" Oddly, we have a manual entry for "¬" but the example uses ~.

https://csound.com/docs/manual/opbitnot.html

@jpffitch or @vlazzarini : any insight into this? I assume the code should be restored to "¬".

Engine/csound_orc.lex
```diff
@@ -177,6 +177,12 @@ SYMBOL          [\[\]+\-*/%\^\?:.,!]
 "enduntil"      { *lvalp = make_token(csound, yytext);
                   (*lvalp)->type = OD_TOKEN;
                   return OD_TOKEN; }
+"for"           { *lvalp = make_token(csound, yytext);
+                  (*lvalp)->type = FOR_TOKEN;
+                  return FOR_TOKEN; }
+"in"            { *lvalp = make_token(csound, yytext);
```
kunstmusik, line 183: Does this break using the "in" opcode?

kunstmusik (commented):
Couple of issues regarding altered character in lexer and use of in keyword when there's already an in opcode

Engine/csound_orc.lex
hlolli: that was a concern I raised on the mailing list, but I let my trust on the tests go ahead of me, so I'll double check locally that all's well

hlolli: you were correct that it was a problem, but I have it fixed now with start-condition on "for"

joachimheintz:
=)

Issue #380, opened 2014-10-15
add switch/case handling


tgrey1:
More syntax sugar a-la C and most other languages, helps trim down large if/elseif/else blocks.

PR #1671, opened 2022-12-28
feat: switch case operators

hlolli:
A feature request from: https://github.com/csound/csound/issues/380


vlazzarini (approved):
all good, thanks.

vlazzarini (approved):

Issue #1669, opened 2022-12-19
Using b type (bool) variable in until statement segfaults


hlolli:
The following code segfaults on develop branch. It seems until statements only handle expressions and not identifier for boolean types.

```csound
<CsoundSynthesizer>
<CsInstruments>
sr=44100
ksmps=1
nchnls=1


instr 1
  ix = 0
  bi = ix == 10
  until bi do
   ix += 1
  od
endin


</CsInstruments>
<CsScore>
i1 0 0
e
</CsScore>
</CsoundSynthesizer>
```

vlazzarini:
Looking at this example, I don't understand its semantics.

The expression `bi = ix == 10` would evaluate to False and the loop would repeat forever, since bi never changes. This is what I would have expected.

```
  until bi do
   ix += 1
   bi = ix == 10
  od
```

Anyway, it fails in create_bool_expression() at the moment.

vlazzarini:
I have a PR fixing the crash. It would be good to have the value of a boolean in place of an expression, but at the moment we don't. Maybe we can open a separate ticket on that?

PR #1672, opened 2022-12-28
fix: using boolean tokens in boolean expressions

hlolli:
this fixes the segfault reported in this ticket https://github.com/csound/csound/issues/1669 as well as segfaults which I discovered for if-else statements as well.


vlazzarini:
I've checked this and the boolean test is not passing correctly. It may need a good deal of work to get it right

vlazzarini (approved):
all looks good now, test is passing and we have booleans working in control flow.

hlolli:
thanks for getting it to work, I look forward making use of for the purposes of [less cognitive load](https://github.com/zakirullin/cognitive-load)

vlazzarini:
absolutely. Actually it was working when you placed the PR but it stopped working with the changes we had made later on. All I had to do was to fix these to accept booleans.

PR #1673, opened 2022-12-28
fix: don't use opmtimization flag sse for m1 apple silicon compiler

hlolli:
The background to this change is that I'm trying to use now only arm64 binaries, including libsndfile. I tell cmake to use arm64 over the rosetta2 x86_64 translator with `set(CMAKE_OSX_ARCHITECTURES "arm64")` in custom.cmake. But this causes an error `error: unknown FP unit 'sse'`. If I just make sure that `-mfpmath=sse` isn't set when apple silicon processor is detected, everything will be alright for me.

The reason I don't use the recently added `CMAKE_APPLE_SILICON_PROCESSOR` is because it doesn't seem to be set correctly. And I think m1 users also using rosetta2 will experience the same issue as of today. This check using sysw, I took from https://gitlab.kitware.com/cmake/cmake/-/blob/v3.22.2/Modules/CMakeDetermineSystem.cmake#L50-84

I believe we should also add this to the master branch.
Note that if we want to release fat binaries, supporting both x86_64 and arm64 `set(CMAKE_OSX_ARCHITECTURES "arm64;x86_64")`, we'd need this change I believe.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ cmake_minimum_required(VERSION 3.5)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
+include(cmake/Modules/DetectAppleSilicon.cmake)
 
 option(USE_VCPKG "Use VCPKG to download and manage dependencies" OFF)
 
@@ -1440,7 +1441,7 @@ set (CPACK_PACKAGE_VERSION_PATCH "0")
 include (CPack)
 
 # Documentation
-if(BUILD_DOCS) 
+if(BUILD_DOCS)
 find_package(Doxygen)
 if(DOXYGEN_FOUND)
     configure_file(${CMAKE_CURRENT_SOURCE_DIR}/config.doxygen
@@ -1469,4 +1470,3 @@ else()
  install(DIRECTORY ${CMAKE_SOURCE_DIR}/samples
           DESTINATION ${CMAKE_INSTALL_PREFIX}/share)
 endif()
-
--- a/cmake/CompilerOptimizations.cmake
+++ b/cmake/CompilerOptimizations.cmake
@@ -18,15 +18,13 @@ if (HAS_CXX_FAST_MATH AND NOT MINGW)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ffast-math")
 endif()
 
-
-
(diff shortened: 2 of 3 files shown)


vlazzarini:
it's failing on Linux.
In any case, we should not apply any changes to the master branch because that has to be in sync with the released code. In any case, my M1 builds did not fail over this. Csound 6.18 for mac is a universal build and I had no trouble building it.

hlolli:
I will look into why linux is failing!

This issue seems to be a thing tough
https://github.com/Z3Prover/z3/issues/5848#issuecomment-1044527883
https://github.com/HaxeFoundation/hashlink/issues/457

On my macbook-pro m1 2021 model, with rosetta2 installed, I do get

```
> uname -m
x86_64
```
this annoying and strange result, which seems to trigger CMake to think that I want to target x86_64 (the variables regarding processor type indicate it).

I wonder if maybe the variable `HAS_FPMATH_SSE` shouldn't even be set to begin with. I'll take another look, but as far as I understand the arm64 processor, this optimization is supposed to be redundant.

vlazzarini:
I just checked and in my universal build here the SSE flag is not added, I think CMake does not include it because it does not pass the HAS_FPMATH_SSE  test.

vlazzarini:
This is what I have from my CMake configuration:
```
-- Performing Test HAS_FPMATH_SSE
-- Performing Test HAS_FPMATH_SSE - Failed
-- Performing Test HAS_CXX_FPMATH_SSE
-- Performing Test HAS_CXX_FPMATH_SSE - Failed
```

although it finds SSE2:

```
-- Performing Test HAS_SSE2
-- Performing Test HAS_SSE2 - Success
-- Performing Test HAS_CXX_SSE2
-- Performing Test HAS_CXX_SSE2 - Success
```

hlolli:
what happens when you build with
`set(CMAKE_OSX_ARCHITECTURES "arm64")` ?

But I believe I found a proper solution!

What I had to do was to include both these variables in my cmake custom file

```cmake
set(CMAKE_OSX_ARCHITECTURES "arm64")
set(CMAKE_APPLE_SILICON_PROCESSOR "arm64")
```

then it worked, because without setting `CMAKE_APPLE_SILICON_PROCESSOR` I just got in the CMake logs

```
The system is: Darwin - 21.6.0 - x86_64
```

when I include `set(CMAKE_APPLE_SILICON_PROCESSOR "arm64")` I get instead

```
The system is: Darwin - 21.6.0 - arm64
```

I could add this to build.md eventually. But thanks for the review, much simpler not to complicate things :)

Issue #1615, opened 2022-07-30
Resolving argumnts to outch broken in cs7; ok in cs6


jpffitch:
Runing te mnual example outch-3.csd gives an error

UnifiedCSD:  manual6-git/examples/outch-3.csd
error:  error: opcode 'int' for expression with arg types iaia not found, line 26

 from file manual6-git/examples/outch-3.csd (1)
error:  Unable to find opcode entry for 'outch' with matching argument types:

Found:
  (null) outch @

Candidates:
   outch Z

Line: 26
 from file manual6-git/examples/outch-3.csd (1)
Parsing failed due to syntax errors
Stopping on parser failure
cannot compile orchestra
end of score.              overall amps:      0.0
           overall samples out of range:        0
1 errors in performance
Elapsed time at end of performance: real: 0.001s, CPU: 0.001s

It seems to work with only one output channel but` so far attempts tohave two channels in out outch call ave faied

hlolli:
I've compared the code, and I don't see any smoking gun, it seems in most parts to be the exact same code. But what I do notice on develop, is that the tree seems to be incorrectly ordered, with the int opcall being incorrectly in place of the parent, which is outch.

```xml
 <treeright addresses="(0x600001498870 : 0x0)" type="284" name="T_OPCALL" varname="int" loc="16:./bug.csd">
  <treeleft addresses="(0x600001498410 : 0x600001c997e0)" type="284" name="T_OPCALL" varname="outch" loc="15:./bug.csd">
   <treeleft addresses="(0x6000014983a0 : 0x600001c99790)" type="294" name="T_IDENT" varname="outch" loc="15:./bug.csd">
   </treeleft>
   <treeright addresses="(0x600001498480 : 0x600001c99830)" type="297" name="INTEGER_TOKEN" value="1" loc="15:./bug.csd">
   </treeright>
   <treenext addresses="(0x6000014984f0 : 0x600001c99880)" type="285" name="T_FUNCTION" fname="poscil" loc="15:./bug.csd">
    <treeright addresses="(0x600001498560 : 0x600001c998d0)" type="298" name="NUMBER_TOKEN" value="0.100000" loc="15:./bug.csd">
    </treeright>
    <treenext addresses="(0x6000014985d0 : 0x600001c99920)" type="297" name="INTEGER_TOKEN" value="440" loc="15:./bug.csd">
    </treenext>
   </treenext>
[...]

PR #1674, opened 2022-12-29
fix: parsing of tree where expression-list is followed by identifier

hlolli:
fixes https://github.com/csound/csound/issues/1615


vlazzarini:
This PR looks stale and the issue it refers is fixed.

Issue #1677, opened 2022-12-29
cmake: libsamplerate is incorrectly considered found from library files alone


hlolli:
defining `LIBSAMPLERATE_LIBRARY` causes cmake to consider libsamplerate found and builds the utility `new_srconv` which depends on it. But it also depends on samplerate.h which we should detect, otherwise an error is thrown like this

```
[636/665] Building C object util/CMakeFiles/src_conv.dir/new_srconv.c.o
FAILED: util/CMakeFiles/src_conv.dir/new_srconv.c.o
[...]

PR #1678, opened 2022-12-29
cmake: properly find libsamplerate

hlolli:
I think it makes sense to keep the cmake module pattern for finding the libraries?

Note that the file util/CMakeLists.txt was DOS formatted, so whenever I changed it I kept getting ^M line endings, so I converted it to unix if that's ok?

closes: https://github.com/csound/csound/issues/1677


CMakeLists.txt
```diff
@@ -205,6 +207,11 @@ else()
     message(STATUS "Not using Custom.cmake file.")
 endif()
 
+if(LIBSAMPLERATE_LIBRARY AND LIBSAMPLERATE_INCLUDE_DIRECTORY)
+  set(LIBSAMPLERATE_FOUND TRUE)
```
stekyne, line 211: This should be set in the find module: https://github.com/csound/csound/blob/a1580f9cdf331c35dceb486f4231871ce0b00266/cmake/Modules/FindEIGEN3.cmake#L66

So you import it along with the rest of the set or unset vars. You should avoid using `include_` family of functions and use `target_` only to avoid including these paths for all targets.

cmake/Modules/FindLIBSAMPLERATE.cmake
```diff
@@ -0,0 +1,11 @@
+if(NOT LIBSAMPLERATE_LIBRARY)
+  if(LIBSAMPLERATE_LIBRARY_DIRECTORY)
+    find_library(LIBSAMPLERATE_LIBRARY NAMES samplerate libsamplerate-0 HINTS "${LIBSAMPLERATE_LIBRARY_DIRECTORY}")
+  else()
+    find_library(LIBSAMPLERATE_LIBRARY NAMES samplerate libsamplerate-0)
+  endif()
```
stekyne, line 6: I think you can just add this to hints whether it's empty or not and simplify this.

CMakeLists.txt
hlolli: I did that initially, the problem I encountered, is that I am defining the location of `LIBSAMPLERATE_INCLUDE_DIRECTORY` in my custom.cmake file. Since it says in the comment where the custom.cmake file is imported https://github.com/csound/csound/blob/develop/CMakeLists.txt#L197-L199 that the idea of it is to override variables from cmake after find_/locate_ operations are done, then it means that the value of `LIBSAMPLERATE_INCLUDE_DIRECTORY` wont exist until later.
I am aware that this is a terrible excuse. I'm just worried about needing to resort to -DVar.. flags for this, which is bit annoying.
The correct solution, in my opinion, is that custom.cmake will be placed at the top (or very close to it) and that way, the developers don't get access to "cheat-mode", because if cmake doesn't resolve the locations correctly when the custom file is evaluated at the beginning, it will also translate to non reproduce-able situations when we are trying to debug issues from users.
With custom.cmake at the top, I'd be totally open for having a second custom cmake file, called something like, overrides.cmake, that gets called after all variables are set.

The second point you made, preferring `target_` over `include_`, totally, I will fix :)

cmake/Modules/FindLIBSAMPLERATE.cmake
hlolli: nice I didn't know, I will fix!

CMakeLists.txt
hlolli: I'm thinking to just not have a _FOUND variable and check for both vars existing, it willsuffice until we can rethink (or not) the order of custom cmake import.

hlolli: fixed

cmake/Modules/FindLIBSAMPLERATE.cmake
hlolli: fixed

CMakeLists.txt
stekyne: Afaik the idea of custom.cmake is mostly for setting feature flags for the app. I don't think you should be overriding find_ module vars this way. You can add a path hint maybe.

vlazzarini:
do we still need this?

kunstmusik:
Closing out this PR as out of date as we have a FindSamplerate.cmake in cmake/modules now.

PR #1679, opened 2022-12-29
cmake: better detection for if the compiler is Clang

hlolli:
On my macbook, the value of ${CMAKE_CXX_COMPILER_ID} is AppleClang, so this match doesn't catch it with string-equals operation, and as I wanted to test out some of the utilities, there are few in util1 that wont configure because of this issue `if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANG OR MSVC)`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ project(Csound)
 
 ENABLE_TESTING()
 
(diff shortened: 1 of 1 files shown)


stekyne (approved):
Good find 👍 Seems to be the correct way to handle it

PR #1680, opened 2022-12-29
Update arrays_for_loop.csd

joachimheintz:
illegal use of carry fixed


hlolli:
thanks for fixing the score in the test ! :)

PR #1682, opened 2022-12-30
feat: fix multiple bugs related to using array types inside structs

hlolli:
This should allow using array of any variable as struct member. This also fixes other issues as well related to the fact that the argument naming was inconsistently added to the varPool.


vlazzarini:
@hlolli this looks complicated to merge, maybe you could submit a new PR against develop with your fixes.

vlazzarini:
Can we close this PR now? I think it's been superseded by other work.

hlolli:
correct, closing

PR #1696, opened 2023-03-01
renaming _CR define due to conflicts with std::chrono

rorywalsh:


vlazzarini:
Not passing some of the checks. Could you take a look?

rorywalsh:
Will do. 👍

rorywalsh:
The build errors on MacOS seems to be related to dependencies rather than changes to the code base? 

```
==> Running `brew cleanup wiiuse`...
==> Caveats
==> stk
The header files have been put in a standard search path, it is possible to use an include statement in programs as follows:

  #include "stk/FileLoop.h"
  #include "stk/FileWvOut.h"

src/ projects/ and rawwaves/ have all been copied to /usr/local/opt/stk/share/stk
Error: Process completed with exit code 1.
```

The wasm stuff is failing on:

`  Error: Action failed with error: Error: Unable to locate executable file: cachix. Please verify either the file path exists or the file can be found within a directory specified by the PATH environment variable. Also check the file mode to verify the file is executable.`

@hlolli do you have a take on the wasm stuff?

vlazzarini:
you're right. Is it possible to try and re-run the MacOs build just to see if it passes now? I'm rerunning the last one from csound6 branch and it passed the dependencies stage

https://github.com/csound/csound/actions/runs/4224196000/jobs/7523867745

PR #1701, opened 2023-03-29
Added x86 and x86_64 architectures to Android make files

giovannibedetti:
This change is made to add support for a small subset of Android devices that use a different architecture, for example _Asus Zenfone 2_, _Genymotion / Android emulator_, and most of all _Chromebooks_.

Unfortunately, building Csound for Android then fails because of the _mp3in_ opcode.
Removing that specific opcode from compilation makes the build successful. 

To be able to build I had to:
- Remove line 263 from _Android/CsoundAndroid/jni/Android.mk_
- Remove line 1194 and 1256 from _Top/csmodule.c_

Of course it would be better to fix the issue in the opcode.

### DIFF
--- a/Android/CsoundAndroid/jni/Application.mk
+++ b/Android/CsoundAndroid/jni/Application.mk
@@ -1,4 +1,4 @@
-APP_ABI := armeabi-v7a arm64-v8a
+APP_ABI := armeabi-v7a arm64-v8a x86 x86_64
 APP_CPPFLAGS += -fexceptions -frtti
 APP_OPTIM := release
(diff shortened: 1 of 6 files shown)


vlazzarini:
that opcode can be removed from
the build, libsndfile is now supporting mp3 files. So potentially if this is built with that support, then we don't need any specific mp3 opcodes.

vlazzarini:
I just saw this PR is against the master branch. Can you submit one to the csound6 (6.19) or develop (7.00) or both?

We can't merge into the master branch.

PR #1706, opened 2023-04-21
Make hash table functions available to plugins

docEdub:
Exposes the internal hash table functions so I can use them in a new set of websocket opcodes I'm working on.

### DIFF
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -548,10 +548,17 @@ static const CSOUND cenviron_ = {
     csoundCepsLP,
     csoundLPrms,
     csoundCreateThread2,
+    cs_hash_table_create,
+    cs_hash_table_get,
+    cs_hash_table_put,
+    cs_hash_table_remove,
+    cs_hash_table_mfree_complete,
+    cs_hash_table_get_key,
+    cs_hash_table_keys,
+    cs_hash_table_values,
     {
       NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL,
-      NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL, NULL,
-      NULL, NULL
+      NULL, NULL, NULL, NULL
     },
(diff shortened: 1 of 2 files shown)


vlazzarini:
SUBR dummyfn_2[] needs to match the size of the initialisation in csound.c, which appears to have changed.
Since we do not need ABI backwards compatibility with 6.x we can even remove this padding for now.

docEdub:
> SUBR dummyfn_2[] needs to match the size of the initialisation in csound.c, which appears to have changed. Since we do not need ABI backwards compatibility with 6.x we can even remove this padding for now.

Done. I'll let someone else remove it later if they want.

How do you feel about the new function names? I wasn't sure which nomenclature is right:
Prefixing with `HashTable...` like `HashTableGet` and `HashTablePut`, or prefixing with the verb like `GetHashTableValue`, `GetHashTableKey`, `SetHashTableValue` (or maybe `PutHashTableValue`?).

vlazzarini:
I don't have much of a preference, just whatever looks more uniform with what's there already.

vlazzarini:
let me know when you are ready for me to merge

docEdub:
> let me know when you are ready for me to merge

Ok, thanks. I'm developing this in parallel with an improved set of websocket opcodes, so I'm not 100% sure this PR has everything I need, yet. I'd like to get further along with those opcodes before marking this PR as ready.

vlazzarini:
just to say, if the plugins have external dependencies, they need to go in the plugins repo, not here.

docEdub:
> just to say, if the plugins have external dependencies, they need to go in the plugins repo, not here.

Thanks, ya, I'm currently working from a fork of that repo, but I'm leaning towards it living in its own separate repo when it's done.

vlazzarini:
that's even better

docEdub:
I closed this PR because I'm having to do it on the Csound 6 branch instead of the Csound 7 branch, so it will work with Cabbage.

vlazzarini:
why not both? It should not be too different.

PR #1713, opened 2023-06-05
Make hash table functions available to plugins

docEdub:
Exposes the internal hash table functions so I can use them in a new set of websocket opcodes I'm working on. Also exposes `csoundPeekCircularBuffer`.


vlazzarini:
This seems to be have been done against the master branch, which is locked and only changes after releases.

Please use csound6 for 6.x. A similar PR for 7.x (develop) is appreciated since we want to make sure these improvements are carried over.

PR #1720, opened 2023-07-07
Add a clear events function

bramtayl:
This is useful in case you want to clear previous events and start fresh

### DIFF
--- a/interfaces/csPerfThread.cpp
+++ b/interfaces/csPerfThread.cpp
@@ -477,20 +477,7 @@ int CsoundPerformanceThread::Perform()
     status = retval;
     csoundCleanup(csound);
     // delete any pending messages
-    csoundLockMutex(queueLock);
-    {
-      CsoundPerformanceThreadMessage *msg;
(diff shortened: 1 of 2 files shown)


kunstmusik:
Not sure this is doing as advertised? Seems to only clear pending messages for Perf Threads and not score events that are csound's scheduler? Or am I missing something?

bramtayl:
I'm most likely missing something, this was a bit of a stab in the dark

bramtayl:
There isn't a way to do this already, is there?

Oeyvind:
I think you can use turnoff3 for this (?). To delete instr events  "in the
queue of scheduled events" (https://csound.com/docs/manual/turnoff3.html).

fre. 7. juli 2023 kl. 17:06 skrev bramtayl ***@***.***>:

bramtayl:
I don't think that does quite what I want it to do. I want basically a new performance thread without having to restart the whole csound session

bramtayl:
But it works as a workaround for now. Would be inconvenient if I had a ton of instruments but I only have a few

Issue #1727, opened 2023-07-22
Build with strict flags fails


bramtayl:
Trying to get the vcpkg port building, and I'm noticing buidling fails with some strict flags, e.g.

`-Werror=misleading-indentation`
`-Werror=use-after-free`
`-Werror=incompatible-pointer-types`

and probably more. Would be good to run a linter or something and clean up all the warnings.

bramtayl:
Ok, so, by default, the debug build errors on Linux if there are any warnings, and there are plenty, so would be good to fix

PR #1730, opened 2023-07-24
vcpkg fixes

bramtayl:
Here are there errors I'm getting without these patches:

```
/home/brandon/vcpkg/buildtrees/csound/src/6.18.1-c8636a11f8.clean/Engine/memalloc.c:214:15: error: pointer ‘pp’ may be used after ‘realloc’ [-Werror=use-after-free]
  214 |       pp->ptr = oldp;
      |       ~~~~~~~~^~~~~~
/home/brandon/vcpkg/buildtrees/csound/src/6.18.1-c8636a11f8.clean/Engine/memalloc.c:208:9: note: call to ‘realloc’ here
  208 |     p = realloc((void*) pp, ALLOC_BYTES(size));
      |         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

/home/brandon/vcpkg/buildtrees/csound/src/6.18.1-c8636a11f8.clean/Engine/musmon.c:66:1: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]
   66 | if(csound->oparms->realtime) \
      | ^~

/home/brandon/vcpkg/buildtrees/csound/src/6.18.1-2b900c6227.clean/./include/sysdep.h:527:5: error: argument 3 of ‘__atomic_compare_exchange’ discards ‘volatile’ qualifier [-Werror=incompatible-pointer-types]
  527 |   !(__atomic_compare_exchange(val, (long *) &oldVal, &newVal, 0,        \
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/Engine/memalloc.c
+++ b/Engine/memalloc.c
@@ -207,13 +207,13 @@ void *mrealloc(CSOUND *csound, void *oldp, size_t size)
     /* allocate memory */
     p = realloc((void*) pp, ALLOC_BYTES(size));
     if (UNLIKELY(p == NULL)) {
-#ifdef MEMDEBUG
-      CSOUND_MEM_SPINLOCK
-      /* alloc failed, restore original header */
-      pp->magic = MEMALLOC_MAGIC;
-      pp->ptr = oldp;
-      CSOUND_MEM_SPINUNLOCK
-#endif
+// #ifdef MEMDEBUG
+//       CSOUND_MEM_SPINLOCK
+//       /* alloc failed, restore original header */
+//       pp->magic = MEMALLOC_MAGIC;
+//       pp->ptr = oldp;
+//       CSOUND_MEM_SPINUNLOCK
+// #endif
       memdie(csound, size);
       return NULL;
     }
--- a/Engine/musmon.c
+++ b/Engine/musmon.c
@@ -63,9 +63,9 @@ void print_csound_version(CSOUND*);
 
 #ifdef HAVE_PTHREAD_SPIN_LOCK
 #define RT_SPIN_UNLOCK \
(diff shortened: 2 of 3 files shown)


bramtayl:
Closes #1727

Engine/memalloc.c
```diff
@@ -207,13 +207,13 @@ void *mrealloc(CSOUND *csound, void *oldp, size_t size)
...
-      pp->ptr = oldp;
-      CSOUND_MEM_SPINUNLOCK
-#endif
+// #ifdef MEMDEBUG
+//       CSOUND_MEM_SPINLOCK
+//       /* alloc failed, restore original header */
+//       pp->magic = MEMALLOC_MAGIC;
+//       pp->ptr = oldp;
+//       CSOUND_MEM_SPINUNLOCK
+// #endif
```
stekyne, line 216: Not sure this is a good enough solution

bramtayl: Me either. I don't really understand the code well enough to fix it though

jpffitch:
What problem is it trying g to solve? 

⁣Get TypeApp for Android ​

jpffitch:
It is better of the error than delete  debuggingy code.
I have adjusted memalloc.c to fix that bug. The musmon c problem was just confused indentation which is fixed here and I will commit to cs6 and cs7 as soon as I can.

The sysdep.h issue is only in MSVC which I do not have so I am at a bit of a loss. It depends on the definition of _:atomic_compare_exchange and the type of the third argument and a conflict involvingo volatile. Perhaps a windows person can help.

==John ff

bramtayl:
Close in favor of #1735

PR #1735, opened 2023-08-09
vcpkg fixes

bramtayl:
These are the changes I had to make to get csound6 to build on vcpkg, see https://github.com/microsoft/vcpkg/pull/32701. I've left some comments about reasons for the changes.


CMakeLists.txt
```diff
@@ -240,12 +254,19 @@ elseif(NOT DEFINED LIBRARY_INSTALL_DIR)
 endif()
 message(STATUS "LIBRARY INSTALL DIR: ${LIBRARY_INSTALL_DIR}")
 
+# on windows, since the plugins are dlls, they need to go into the executable folder
```
stekyne, line 257: I don't quite get this comment. Why would this be the case? The plugins can go to the plugin directory along with any dependent DLLs for them. As long as it's on the path and set as the plugin folder, it'll work no?

CMakeLists.txt
```diff
@@ -455,15 +469,8 @@ SET(BUILD_SHARED_LIBS ON)
...
-    find_package(SndFile CONFIG REQUIRED)
-else()
-    find_library(LIBSNDFILE_LIBRARY NAMES sndfile libsndfile-1 libsndfile)
-
-    if(NOT LIBSNDFILE_LIBRARY)
-        message(FATAL_ERROR "Csound requires the sndfile library")
-    endif()
-endif()
+# provided by vcpkg
+find_package(SndFile CONFIG REQUIRED)
```
stekyne, line 473: The else part is for non VCPKG builds. You can see the libraries have multiple potential names. This will break the current builds for most mac/linux users.

stekyne (commented):
It looks like you're merging this into the csound6 branch, is that correct? The develop/csound7 branch has got a working VCPKG build behind the `USE_VCPKG` flag that also allows the old build to work. I'm not sure why some of these other fixes are needed for Mac etc but they might very well be needed. Some of the changes here will break the existing build for some users so not sure this can be merged as is.

CMakeLists.txt
bramtayl: vcpkg forbids dlls to show up anywhere inside the lib folder. Stupid policy but

bramtayl:
Yeah, this is for csound6. For the PR, I wasn't really worried about not breaking things, but yeah, many of these changes should be behind a USE_VCPKG flag. If there is a csound7 version released soon we could just use that too.

CMakeLists.txt
```diff
@@ -4,6 +4,11 @@ cmake_minimum_required(VERSION 3.5)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 
+if (APPLE)
+    set(CMAKE_INSTALL_RPATH "@loader_path")
+elseif (UNIX)
+    set(CMAKE_INSTALL_RPATH "$ORIGIN")
+endif()
```
bramtayl, line 11: This is to make sure the relative linking works. Apple uses `@loader_path` for libraries and `@executable_path` for executables, while Linux uses `$ORIGIN`.

CMakeLists.txt
```diff
@@ -357,19 +378,11 @@ endif()
 
 # Set plugins install directory
 if(USE_DOUBLE)
-    if(APPLE)
-        set(CSOUNDLIB "CsoundLib64")
-        set(PLUGIN_INSTALL_DIR "${CS_FRAMEWORK_DEST}/${CSOUNDLIB}.framework/Versions/${APIVERSION}/Resources/Opcodes64")
-    else()
-        set(CSOUNDLIB "csound64")
-    endif()
+    # No mac framework
```
bramtayl, line 381: Would be good to add an option to disable mac frameworks

CMakeLists.txt
stekyne: Hmm how do you mean? Is that written down somewhere?

bramtayl: Not sure but you can see it in the CI logs e.g. https://dev.azure.com/vcpkg/public/_build/results?buildId=92172&view=logs&jobId=7922e5c4-0103-5f8f-ad17-45ce9bb98e80&j=7922e5c4-0103-5f8f-ad17-45ce9bb98e80&t=66eef212-08d6-5e6c-a560-7d01149d0188

stekyne: I imagine all the binaries produced should end up in the bin folder. The lib folder is probably intermediate or dependent libs for the build process.

bramtayl: The plugins all end up in lib\csound\plugins64-6.0

CMakeLists.txt
bramtayl: rpaths also confuse me so could be wrong here

bramtayl:
Ok, I've split this up into several separate commits. I also moved a lot of changes behind non-default options to reduce breakage.

CMakeLists.txt
```diff
@@ -184,6 +196,9 @@ option(BUILD_INSTALLER "Build installer" OFF)
 option(BUILD_TESTS "Build tests" ON)
 option(USE_GIT_COMMIT "Show the git commit in version information" ON)
 option(REQUIRE_PTHREADS "For non-Windows systems, set whether Csound will use threads or not" ON)
+option(DEBUG_ERROR_ON_WARNING "Error on compiler warning" ON)
+option(APPLE_FRAMEWORK "Build framework on Apple" ON)
+option(USE_VCPKG_DEPENDENCIES "Use dependencies from vcpkg" OFF)
```
stekyne, line 201: There is already a USE_VCPKG no? What is this flag supposed to do

bramtayl: This is for building for the vcpkg build bots, not for vcpkg users. I guess I could change the name? We might be able to merge the two flags at some point?

bramtayl:
Each commit should be fairly separable, and maybe could be separate PRs? The only real vcpkg specfic one is "add vcpkg dependencies option", the rest are pretty much bug fixes.

CMakeLists.txt
stekyne: I don't quite get the comment about vcpkg build bots and vcpkg users. What do you mean? They are the same flag but I can see now that the `USE_VCPKG` flag is not referenced much in this branch.

I think trying to add this work to the older Csound6 branch is going to be a bit of a mess. Why not try this with develop? That has a working build for VCPKG for Linux and Windows at least, I think Mac is also working.

bramtayl: Maybe that would make sense. I'm just a little tentative about using an unreleased version?

bramtayl: The difference is the difference between `./vcpkg install csound` (where vcpkg does the building for you) and `cmake ... -DCMAKE_TOOLCHAIN_FILE=[vcpkg root]/scripts/buildsystems/vcpkg.cmake` (where users manually call cmake). But there should be a significant amount of overlap in the changes needed for both for sure. But yeah, I'll try working with the develop branch next

stekyne: Right I see what you're referring to. I think that the vcpkg build for Csound will just be a certain build configuration for the CMake build i.e. probably without any unsupported flags that require 3rd party deps not in vcpkg rather than a separate build flow in CMake. As for the unreleased version, I'm sure we can work around that for now. It is possible to use branch/unreleased versions in VCPKG afaik but they just don't encourage it.

PR #1739, opened 2023-09-06
Add sndfile module

bramtayl:


bramtayl:
Ref #1735

CMakeLists.txt
```diff
@@ -465,49 +465,23 @@ SET(BUILD_SHARED_LIBS ON)
...
-SET(CMAKE_REQUIRED_LIBRARIES ${LIBSNDFILE_LIBRARY})
-SET(CMAKE_REQUIRED_INCLUDES ${SNDFILE_H_PATH})
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_LIBSNDFILE")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_LIBSNDFILE")
 
-check_c_source_compiles("#include <sndfile.h>
-int main(int argc, char** argv) {
-printf(\"SF_FORMAT_MPEG: %d\", SF_FORMAT_MPEG);
-return 0;
-}" HAVE_SF_FORMAT_MPEG)
```
stekyne, line 498: Not sure this should be deleted. This did actually catch an issue recently. It is possible to have libsndfile without this sub dependency but the code will break without that. This ensures it's obvious to the user why.

CMakeLists.txt
```diff
@@ -465,49 +465,23 @@ SET(BUILD_SHARED_LIBS ON)
...
+    if (USE_MP3)
+        set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DSNDFILE_MP3")
+    endif()
 
-if (USE_MP3)
-  message(STATUS  "Using libsndfile MP3 support")
-  set(CMAKE_C_FLAGS  "${CMAKE_C_FLAGS} -DSNDFILE_MP3")
-else()
-  message(STATUS  "NOT Using libsndfile MP3 support")
+    if(MSVC)
```
stekyne, line 476: This should reference USE_VCPKG no?

stekyne:
This does look like a good change but lets make sure we're not losing any checks. I think the cmake modules should work on MSVC or at least put in some hint locations for the files even if we mostly always use VCPKG. We should try to aim to avoid referencing platforms directly and either have USE_VCPKG or just the cmake module which should abstract any platform info away.

CMakeLists.txt
bramtayl: Yeah, that makes more sense

bramtayl:
Ok, marking this as a work in progress then because it will be tricky to support windows without vcpkg too. If we want to support non-vcpkg windows builds it would probably be good to add CI for it too.

stekyne:
@bramtayl we don't have to support non-vcpkg builds for Windows explicitly. I haven't been super careful about keeping the mingw build working but I have tried to at least keep changes for VCPKG behind the USE_VCPKG flag. I'd say maybe just add some directory hints for the cmake find files for Windows and it would be enough.

bramtayl:
@vlazzarini I had to add #1748 to this; otherwise the vcpkg and windows builds both fail

bramtayl:
Ok, I've basically decided not to support non-vcpkg windows. This is already the status quo. To do so, we'd need to

- Add a find module for all of sndfile's dependencies and sub-dependencies
- Figuring out where to look for dll's for each

which I don't think is really maintainable without CI?

CMakeLists.txt
```diff
@@ -470,47 +473,27 @@ SET(BUILD_SHARED_LIBS ON)
...
 if(USE_LIBSNDFILE)
- set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_LIBSNDFILE")
- set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_LIBSNDFILE")
-
- if(USE_VCPKG)
-    find_package(SndFile CONFIG REQUIRED)
- else()
-    find_library(LIBSNDFILE_LIBRARY NAMES sndfile libsndfile-1 libsndfile)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DUSE_LIBSNDFILE")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DUSE_LIBSNDFILE")
```
stekyne, line 477: This shouldn't really be necessary if we're using targets correctly.

bramtayl: The only think I changed here was the indentation...but if I'm understanding correctly, it would definitely make more sense to do:

```
target_compile_definitions(${CSOUNDLIB} PRIVATE "USE_LIBSNDFILE)
```

CMakeLists.txt
```diff
@@ -83,6 +83,9 @@ include(CheckCXXCompilerFlag)
 include(CMakePushCheckState)
 
 list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/Modules/)
+if (NOT USE_VCPKG)
+    list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake/non-vcpkg-extras/)
+endif()
```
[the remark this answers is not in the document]
bramtayl: The advantage of this is we can use `find_package(SndFile)` instead of 

```
if (USE_VCPKG)
    find_package(SndFile CONFIG)
else()
    find_package(SndFile)
endif()
```

But I'm happy doing either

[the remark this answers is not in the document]
stekyne: I think we can abstract all of that out to a findX.cmake file though. So the main cmake files should just do find_package(x) and it will just work.

CMakeLists.txt
bramtayl: Ok, changed it to that

Opcodes/CMakeLists.txt
```diff
@@ -114,25 +114,24 @@ if(BUILD_DSSI_OPCODES)
 endif()
 
 ## OPCODES WITH EXTERNAL DEPENDENCIES ##
-find_package(LIBLO)
+find_package(LIBLO MODULE)
```
bramtayl, line 117: This isn't set up to use a config module yet so force the find module

CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ cmake_policy(SET CMP0077 NEW)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG 1)
```
stekyne, line 8: This should be tied to `USE_VCPKG` though. Otherwise it will be broken. Why not just use the `USE_VCPKG` flag in the fine.cmake files? Maybe it needs to be made global or something. I've never really tried it.

bramtayl: > Otherwise it will be broken

Hmm, what do you mean by that?

stekyne: Becaues if you specify `USE_VCPKG` as true and this flag as false, then it won't work correctly no? Whatever `USE_VCPKG` is set as should be what this flag also is set to.

bramtayl: Hm, still don't understand. As I understand it, vcpkg tends to provide "config" modules, while csound provides "find" modules. So we have three options:

1) Only load the csound "find" modules if vcpkg isn't available
2) Specify whether we want a "config" or a "find" module depending on whether or not we are using vcpkg
3) Choose to always use a config module if it's available

stekyne: USE_VCPKG should be the main flag to decide if the build is using it or not. If you introduce another flag, then it's possible to have USE_VCPKG=1 and CMAKE_FIND_PACKAGE_PREFER_CONFIG as 0, which wouldn't work. Just use the USE_VCPKG flag

bramtayl: Hmm, no sorry I still don't understand. `CMAKE_FIND_PACKAGE_PREFER_CONFIG` is an internal cmake setting, not a new option I'm adding. It just tells cmake that if there is both a `find` and a `config` package available, use the `config` one.

stekyne: Ah ok. I wasn't paying attention to that. Then this make sense. I wasn't aware of that option. Fair play

stekyne (approved):
Think it looks good. If there are no build breakages, should be good to go

bramtayl:
Ok, I decided to add back SndFile's dependencies, for completeness. I also used check_deps based on the comment here: https://github.com/csound/csound/pull/1756#discussion_r1359910496

bramtayl:
Will need to rebase onto #1760

util/SDIF/CMakeLists.txt
```diff
@@ -3,4 +3,5 @@
 set(sdif2ad_SRCS sdif2adsyn.c sdif.c sdif-mem.c)
 
 make_executable(sdif2ad "${sdif2ad_SRCS}" "")
+target_link_libraries(sdif2ad PUBLIC ${CSOUNDLIB})
```
bramtayl, line 6: Unrelated bugfix

Java/CMakeLists.txt
```diff
@@ -55,9 +47,15 @@ if(BUILD_JAVA_INTERFACE)
...
-        target_include_directories(${SWIG_MODULE__jcsound7_REAL_NAME}
-        PRIVATE ${JNI_INCLUDE_DIRS})
+
+        target_link_libraries(_jcsound7 PUBLIC ${CSOUNDLIB} JNI::JNI)
+        if (JNI_AWT_FOUND)
+            target_link_libraries(_jcsound7 PRIVATE JNI::AWT)
+        endif()
+        if (JNI_JVM_FOUND)
+            target_link_libraries(_jcsound7 PRIVATE JNI::JVM)
+        endif()
```
bramtayl, line 57: While I was here, I fixed up the linking a bit

tests/c/CMakeLists.txt
```diff
@@ -26,26 +26,14 @@ if(BUILD_TESTS AND BUILD_STATIC_LIBRARY AND GTest_FOUND)
...
         if (APPLE)
             set(CMAKE_CXX_FLAGS "-stdlib=libc++")
         endif()
-
-        target_include_directories(unittests PRIVATE
-            ${GTEST_INCLUDE_DIRS})
-
-        target_link_libraries(unittests PRIVATE
-            ${GTEST_BOTH_LIBRARIES} ${CSOUNDLIB_STATIC})
     endif()
```
bramtayl, line 37: While I was here, I fixed up the test code a bit

bramtayl:
On top of #1776

bramtayl:
Since this has grown to be unwieldy, I split it up in to a ton of smaller PR. Each one builds on the previous

bramtayl:
Bump

kunstmusik:
Haven't checked this, but beware that we ran into an issue from another PR about using features that were only present in later versions of CMake than what is available in some of our supported platforms (e.g., Ubuntu 22.04, Debian 11). We still need to do an audit of earliest Linux platforms we'll support but those two should be guides for the moment.

bramtayl:
Ok, well CMAKE_FIND_PACKAGE_PREFER_CONFIG was added in cmake 3.15 and debian 11 ships cmake 3.18 so I think we're good. But I added the extra check, see also https://github.com/csound/csound/pull/1760#discussion_r1380628016

bramtayl:
Part of #1811 now

PR #1741, opened 2023-09-07
debug error on warning option

bramtayl:
So you can build a debug build with compiler warnings if you want

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,6 +193,7 @@ option(LEXER_DEBUG "Build lexer with debug option" OFF)
 option(USE_GIT_COMMIT "Show the git commit in version information" ON)
 option(REQUIRE_PTHREADS "For non-Windows systems, set whether Csound will use threads or not" ON)
 option(BUILD_DOCS "Build documentation (requires doxygen)" OFF)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -193,6 +193,7 @@ option(LEXER_DEBUG "Build lexer with debug option" OFF)
 option(USE_GIT_COMMIT "Show the git commit in version information" ON)
 option(REQUIRE_PTHREADS "For non-Windows systems, set whether Csound will use threads or not" ON)
 option(BUILD_DOCS "Build documentation (requires doxygen)" OFF)
+option(DEBUG_ERROR_ON_WARNING "For a debug build, error on a compiler warning" ON)
```
stekyne, line 196: This should default to off. Until we make an effort to actually reduce the warnings. Otherwise there will be build breakages where they weren't failing before.

bramtayl: I agree, but `-Werror` is on by default and I didn't want to change things.

bramtayl: Ok, turned it off by default

stekyne (approved):

bramtayl:
bump

PR #1742, opened 2023-09-07
cmake exports

bramtayl:
cmake can automatically generate files so you can import csound from another cmake project


CMakeLists.txt
```diff
@@ -113,6 +113,7 @@ function(make_executable name srcs libs)
             OUTPUT_NAME ${ARGV3})
     endif()
     install(TARGETS ${name}
+    EXPORT CSoundExports
```
stekyne, line 116: A small thing but it's generally spelled Csound and not CSound

cmake/Modules/FindCsound.cmake
stekyne, line 1: Is this right to delete this?

bramtayl: This so, the CMake exports will automatically generate a find module already?

vcpkg
bramtayl, line 1: Oops, didn't mean to delete this

cmake/Modules/FindCsound.cmake
stekyne: Did you check if the generated one contains everything this one does?

bramtayl: Well, checked for Linux, but the generated find module will depend on the platform you're building for, and I don't have a mac, so can't check there. I can say that the existing find module had no hope of working on Windows, because `find_library` wouldn't find the dlls?

stekyne: Ok well let's try and see how it goes. Ideally we would be using exported targets

stekyne (approved):

bramtayl:
bump

kunstmusik:
I checked the output from this and want to make sure it's covering everything the previous file did.  Will need a little time to experiment and test.

bramtayl:
Ok, thanks!

bramtayl:
@kunstmusik did you get the change to try this out?

stekyne (approved):
Reading more, this seems to be the recommended way of doing this. Going to merge now

PR #1743, opened 2023-09-07
change windows plugin folder

bramtayl:
Because the windows plugins are dll executables


stekyne (commented):
I'm not sure about this change. Just because they're DLLs doesn't mean they need to be in a particular folder surely? Once it's on the path, it'll work.

bramtayl:
Ok, sure, can always move later I guess

PR #1746, opened 2023-09-07
remove alloca

bramtayl:
Because alloca isn't available on all platforms

### DIFF
--- a/Engine/envvar.c
+++ b/Engine/envvar.c
@@ -1030,21 +1030,28 @@ void *csoundFileOpenWithType(CSOUND *csound, void *fd, int type,
 #if defined(WIN32)
       /* To handle Widows errors in file name characters. */
       size_t sz = 2 * MultiByteToWideChar(CP_UTF8, 0, name, -1, NULL, 0);
-      wchar_t *wfname = alloca(sz);
+      wchar_t *wfname = malloc(sz);
       wchar_t *wmode = 0;
 
       MultiByteToWideChar(CP_UTF8, 0, name, -1, wfname, sz);
       sz = 2 * MultiByteToWideChar(CP_UTF8, 0, param, -1, NULL, 0);
-      wmode = alloca(sz);
+      wmode = malloc(sz);
       MultiByteToWideChar(CP_UTF8, 0, param, -1, wmode, sz);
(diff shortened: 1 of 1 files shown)


stekyne (approved):

bramtayl:
bump

Engine/envvar.c
```diff
@@ -1030,12 +1030,13 @@ void *csoundFileOpenWithType(CSOUND *csound, void *fd, int type,
 #if defined(WIN32)
       /* To handle Widows errors in file name characters. */
       size_t sz = 2 * MultiByteToWideChar(CP_UTF8, 0, name, -1, NULL, 0);
-      wchar_t *wfname = alloca(sz);
+      // alloca not available on all platforms
```
kunstmusik, line 1033: Don't really need this comment in code if we're removing alloca.

Engine/envvar.c
```diff
@@ -1046,6 +1047,11 @@ void *csoundFileOpenWithType(CSOUND *csound, void *fd, int type,
         }
         fullName = (char*) name;
       }
+      free(wfname);
```
kunstmusik, line 1050: Because of the goto in line 1045, it's possible for wfname and wmode to leak memory.

kunstmusik (changes requested):

Engine/envvar.c
bramtayl: Oh yeah oops. Would it be good enough to just copy the two free lines immediately before the goto?

Engine/envvar.c
bramtayl: The comment was mostly just so alloca doesn't sneak its way back in, but I can remove it too if you want

Engine/envvar.c
bramtayl: Also, is there a way to add a test for the memory leak?

Engine/envvar.c
bramtayl: Ok, well, removed the comment

Engine/envvar.c
bramtayl: Ok, copied the frees before the goto

PR #1748, opened 2023-09-07
link plugins to csound

bramtayl:
Because all the plugins require the csound library


bramtayl:
Ref #1465

stekyne (commented):
I think it makes sense but this is working currently via another mechanism.  https://github.com/csound/csound/blob/053fd4da66fb9cb8b39c4af9f324465e706cc03b/CMakeLists.txt#L1122 You should try and remove this add_library and get the build working using the target_link_libraries way.

bramtayl:
> I think it makes sense but this is working currently via another mechanism.

> You should try and remove this add_library and get the build working using the target_link_libraries way.

Hmm, don't understand?

stekyne:
The plugins are currently building regardless of this change. That's because it's linked to every target via `add_library` instead of `target_add_library` that you are adding. You should remove the more general `add_library` and make sure it still all works/compiles.

stekyne:
Ah you're right, that is what I meant. The untargeted includes will apply to all targets which is not good but the target_include will only apply to the given target. Sorry I got confused there. 

Right so this makes sense then

stekyne (approved):

stekyne:
Maybe rebase and see that the build is ok

bramtayl:
Ok rebased

kunstmusik:
@bramtayl : I think the premise of this PR is incorrect. At least last I remembered, we purposely don't link plugins with Csound and just compile with headers and that it had to do with issues on Windows.

Requesting advice on this from @rorywalsh @gogins @jpffitch @vlazzarini

vlazzarini:
@kunstmusik yes, absolutely plugins do not require linking. If they do, they're using the Host API and that's wrong.

Plugins use the public functions in the CSOUND object and only need headers to be built.

bramtayl:
Hmm I don't understand:

> Plugins use the public functions in the CSOUND object

So how could they not be linked to csound then?

vlazzarini:
Because plugin functions are passed the CSOUND object. Or, in C++, they are subclasses of CSOUND. That's how the mechanism works (once plugin libs are loaded, see in Top/csmodule.c). 
No linking, no dependencies.

bramtayl:
> plugin functions are passed the CSOUND object

Right, so they should be linked to csound? I guess I'm just very confused, because that seems like its the very definition of a dependency

bramtayl:
For reference, the Qt plugins are all linked to the main binaries:

```
~/6.5.1/gcc_64/plugins/platforms$ ldd libqxcb.so
	linux-vdso.so.1 (0x00007fff02dee000)
	libQt6XcbQpa.so.6 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libQt6XcbQpa.so.6 (0x00007f2f383d3000)
	libxkbcommon-x11.so.0 => /lib/x86_64-linux-gnu/libxkbcommon-x11.so.0 (0x00007f2f383b3000)
	libxkbcommon.so.0 => /lib/x86_64-linux-gnu/libxkbcommon.so.0 (0x00007f2f3836d000)
	libxcb-cursor.so.0 => /lib/x86_64-linux-gnu/libxcb-cursor.so.0 (0x00007f2f38000000)
	libxcb-icccm.so.4 => /lib/x86_64-linux-gnu/libxcb-icccm.so.4 (0x00007f2f38366000)
	libxcb-image.so.0 => /lib/x86_64-linux-gnu/libxcb-image.so.0 (0x00007f2f3835e000)
	libxcb-keysyms.so.1 => /lib/x86_64-linux-gnu/libxcb-keysyms.so.1 (0x00007f2f38359000)
	libxcb-randr.so.0 => /lib/x86_64-linux-gnu/libxcb-randr.so.0 (0x00007f2f38347000)
	libxcb-render-util.so.0 => /lib/x86_64-linux-gnu/libxcb-render-util.so.0 (0x00007f2f38340000)
	libxcb-shm.so.0 => /lib/x86_64-linux-gnu/libxcb-shm.so.0 (0x00007f2f3833b000)
	libxcb-sync.so.1 => /lib/x86_64-linux-gnu/libxcb-sync.so.1 (0x00007f2f38332000)
	libxcb-xfixes.so.0 => /lib/x86_64-linux-gnu/libxcb-xfixes.so.0 (0x00007f2f38327000)
	libxcb-render.so.0 => /lib/x86_64-linux-gnu/libxcb-render.so.0 (0x00007f2f38319000)
	libxcb-shape.so.0 => /lib/x86_64-linux-gnu/libxcb-shape.so.0 (0x00007f2f38314000)
	libxcb-xkb.so.1 => /lib/x86_64-linux-gnu/libxcb-xkb.so.1 (0x00007f2f382f6000)
	libxcb.so.1 => /lib/x86_64-linux-gnu/libxcb.so.1 (0x00007f2f382cc000)
	libQt6OpenGL.so.6 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libQt6OpenGL.so.6 (0x00007f2f3822a000)
	libQt6Gui.so.6 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libQt6Gui.so.6 (0x00007f2f37600000)
	libGL.so.1 => /lib/x86_64-linux-gnu/libGL.so.1 (0x00007f2f37579000)
	libQt6Core.so.6 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libQt6Core.so.6 (0x00007f2f36e00000)
	libX11-xcb.so.1 => /lib/x86_64-linux-gnu/libX11-xcb.so.1 (0x00007f2f38223000)
	libX11.so.6 => /lib/x86_64-linux-gnu/libX11.so.6 (0x00007f2f36cc2000)
	libdl.so.2 => /lib/x86_64-linux-gnu/libdl.so.2 (0x00007f2f3821e000)
	libpthread.so.0 => /lib/x86_64-linux-gnu/libpthread.so.0 (0x00007f2f38217000)
	libstdc++.so.6 => /lib/x86_64-linux-gnu/libstdc++.so.6 (0x00007f2f36a00000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007f2f37490000)
	libgcc_s.so.1 => /lib/x86_64-linux-gnu/libgcc_s.so.1 (0x00007f2f37fdc000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f2f36600000)
	libglib-2.0.so.0 => /lib/x86_64-linux-gnu/libglib-2.0.so.0 (0x00007f2f368b9000)
	libgthread-2.0.so.0 => /lib/x86_64-linux-gnu/libgthread-2.0.so.0 (0x00007f2f38210000)
	libxcb-util.so.1 => /lib/x86_64-linux-gnu/libxcb-util.so.1 (0x00007f2f38207000)
	libXau.so.6 => /lib/x86_64-linux-gnu/libXau.so.6 (0x00007f2f37fd6000)
	libXdmcp.so.6 => /lib/x86_64-linux-gnu/libXdmcp.so.6 (0x00007f2f37fce000)
	libEGL.so.1 => /lib/x86_64-linux-gnu/libEGL.so.1 (0x00007f2f36caf000)
	libfontconfig.so.1 => /lib/x86_64-linux-gnu/libfontconfig.so.1 (0x00007f2f3686b000)
	libQt6DBus.so.6 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libQt6DBus.so.6 (0x00007f2f3653a000)
	libz.so.1 => /lib/x86_64-linux-gnu/libz.so.1 (0x00007f2f36c91000)
	libfreetype.so.6 => /lib/x86_64-linux-gnu/libfreetype.so.6 (0x00007f2f36470000)
	libGLdispatch.so.0 => /lib/x86_64-linux-gnu/libGLdispatch.so.0 (0x00007f2f363b8000)
	libGLX.so.0 => /lib/x86_64-linux-gnu/libGLX.so.0 (0x00007f2f36839000)
	libicui18n.so.56 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libicui18n.so.56 (0x00007f2f35e00000)
	libicuuc.so.56 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libicuuc.so.56 (0x00007f2f35a00000)
	libicudata.so.56 => /home/brandon/6.5.1/gcc_64/plugins/platforms/./../../lib/libicudata.so.56 (0x00007f2f34000000)
	librt.so.1 => /lib/x86_64-linux-gnu/librt.so.1 (0x00007f2f3748b000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f2f3848e000)
	libpcre2-8.so.0 => /lib/x86_64-linux-gnu/libpcre2-8.so.0 (0x00007f2f3631e000)
	libbsd.so.0 => /lib/x86_64-linux-gnu/libbsd.so.0 (0x00007f2f36c7c000)
	libexpat.so.1 => /lib/x86_64-linux-gnu/libexpat.so.1 (0x00007f2f3680e000)
	libdbus-1.so.3 => /lib/x86_64-linux-gnu/libdbus-1.so.3 (0x00007f2f362d0000)
	libpng16.so.16 => /lib/x86_64-linux-gnu/libpng16.so.16 (0x00007f2f35dc8000)
	libbrotlidec.so.1 => /lib/x86_64-linux-gnu/libbrotlidec.so.1 (0x00007f2f36c6f000)
	libmd.so.0 => /lib/x86_64-linux-gnu/libmd.so.0 (0x00007f2f362c3000)
	libsystemd.so.0 => /lib/x86_64-linux-gnu/libsystemd.so.0 (0x00007f2f33f30000)
	libbrotlicommon.so.1 => /lib/x86_64-linux-gnu/libbrotlicommon.so.1 (0x00007f2f362a0000)
	libcap.so.2 => /lib/x86_64-linux-gnu/libcap.so.2 (0x00007f2f35dbc000)
	libgcrypt.so.20 => /lib/x86_64-linux-gnu/libgcrypt.so.20 (0x00007f2f33de8000)
	liblzma.so.5 => /lib/x86_64-linux-gnu/liblzma.so.5 (0x00007f2f33db6000)
	libzstd.so.1 => /lib/x86_64-linux-gnu/libzstd.so.1 (0x00007f2f33d02000)
	liblz4.so.1 => /lib/x86_64-linux-gnu/liblz4.so.1 (0x00007f2f33cdf000)
	libgpg-error.so.0 => /lib/x86_64-linux-gnu/libgpg-error.so.0 (0x00007f2f33cb9000)

```

vlazzarini:
No linking. Csound opens a dynamic lib, lookups up the payload, registers the opcodes.

The dynamic lib does not need to be linked to the Csound for this to happen. The code does not use any symbols from Csound except those defined inside the CSOUND structure as public and for that it only needs the headers.

You may need to consult the docs on plugins to understand how they work. There's a short chapter in the Csound FLOSS manual.

vlazzarini:
I don't see the relevance of QT here. I'm trying to explain how plugins work in Csound since we introduced them over 20 years ago.

bramtayl:
You're acting pretty dismissive, but I'm just trying to help!

Here's what the cmake docs say:

```
Generally, a dependency should be specified in a use of target with the PRIVATE keyword if it is used by only the implementation of a library, and not in the header files. If a dependency is additionally used in the header files of a library (e.g. for class inheritance), then it should be specified as a PUBLIC dependency. A dependency which is not used by the implementation of a library, but only by its headers should be specified as an INTERFACE dependency.
```

For example, rtpulse doesn't have any public headers, but right at the top of `rtpulse.c` is `#include <csdl.h>`. That's very clearly using csound in the implementation of the library, so it should be a private dependency.
`

gogins:
Plugins are not linked by the compiler or linker at compile time. Rather,
plugins are found, loaded, and linked at run time.

bramtayl:
Oh, I finally understand the confusion maybe. I'm talking about linking the plugins to csound, not the other way around.

So for example,

```
target_link_libraries(rtpa PRIVATE csound64)
```

not

```
target_link_libraries(csound64 PRIVATE rtpa)
```

kunstmusik:
@bramtayl You're still not getting it, but we do not link csound to plugins. It's not necessary as the headers define everything the plugin needs to compile and run, since Csound loads the plugins and passes in data (CSOUND struct) that is the same size and layout as what the plugins were compiled against. Plugins read all data and functions from the CSOUND struct that is passed into it and therefore they do not need to link to anything in another library as they only depend on the runtime data passed in. This allows plugins to be loaded with different versions of Csound. 

As Victor mentioned, this was designed decades ago and has served the community well in that time. (If I remember correctly, Richard Dobson and Michael Gogins had a lot of conversations about this on the dev list and referenced either ActiveX or COM objects as part of the design?). 

If you have further questions, please ask away, but you need to understand first that plugins works without linking to Csound and it is by design.

bramtayl:
Well, I still don't understand, but I appreciate everyone's patience trying to explain it to me. I'm 95% sure if we move to using imported targets for dependencies, like in #1739, the plugins will no longer compile, but I could be wrong. If so, I'll flag this issue again so we can figure out how to handle it.

But to check my understanding:

> the same size and layout as what the plugins were compiled against

Right, so the plugins need to be compiled again csound? So cmake needs to aware of the dependency?

> This allows plugins to be loaded with different versions of Csound

Isn't that how dynamic linking always works? We only change the library name when the API changes, so that we can dynamically link any library with the same API?

genell:
I’m REALLY no developer and have very limited programming experience so I am just as likely to make matters worse, but I was thinking that one way to look at plugins is as user opcodes. In a csd file you can define custom opcodes. In the “header” of the opcode statement you need to declare what types and how many variables that should be expected as input and as output. Then, when csound compiles the csd it identifies the “opcode” statement and reads the “header” info about input and outputs. When you use you custom opcode in the orchestra section, csound compiles that orchestra, compares to a list of known opcodes, among which is you custom one at this point, checks that input and output values are correct, and adds it to the graph. Plugins do the same, except they are not written in csound code but c or c++. So you include the csound header to tell you code how to format an “a” variable or a “f” variable of whatever, and upon compilation, it just checks those formats in the header and act accordingly. Now you have a piece of software that can be called by csound if you inside csound call you opcode by name (the internal name inside the plug-in, not necessarily the file name of the compiles plugin), and if you place it in a path where csound looks for opcodes. The, jut like for the user opcode case, the opcode is on the internet runtime list, and csound “stupidly” sends data according to the plug-in “header” and expects some other data in return. No build-time linking, only headers, and paths that csound know. Regards,Anders29 sep. 2023 kl. 13:03 skrev bramtayl ***@***.***>:﻿
Well, I still don't understand, but I appreciate everyone's patience trying to explain it to me. I'm 95% sure if we move to using imported targets for dependencies, like in #1739, the plugins will no longer compile, but I could be wrong. If so, I'll flag this issue again so we can figure out how to handle it.
But to check my understanding:

the same size and layout as what the plugins were compiled against

Right, so the plugins need to be compiled again csound? So cmake needs to aware of the dependency?

This allows plugins to be loaded with different versions of Csound

Isn't that how dynamic linking always works? We only change the library name when the API changes, so that we can dynamically link any library with the same API?

genell:
Please excuse the multitude of spelling errors - I hope it doesn’t obfuscate the message completely, even if that message is all wrong…/A29 sep. 2023 kl. 18:34 skrev Anders Genell ***@***.***>:﻿I’m REALLY no developer and have very limited programming experience so I am just as likely to make matters worse, but I was thinking that one way to look at plugins is as user opcodes. In a csd file you can define custom opcodes. In the “header” of the opcode statement you need to declare what types and how many variables that should be expected as input and as output. Then, when csound compiles the csd it identifies the “opcode” statement and reads the “header” info about input and outputs. When you use you custom opcode in the orchestra section, csound compiles that orchestra, compares to a list of known opcodes, among which is you custom one at this point, checks that input and output values are correct, and adds it to the graph. Plugins do the same, except they are not written in csound code but c or c++. So you include the csound header to tell you code how to format an “a” variable or a “f” variable of whatever, and upon compilation, it just checks those formats in the header and act accordingly. Now you have a piece of software that can be called by csound if you inside csound call you opcode by name (the internal name inside the plug-in, not necessarily the file name of the compiles plugin), and if you place it in a path where csound looks for opcodes. The, jut like for the user opcode case, the opcode is on the internet runtime list, and csound “stupidly” sends data according to the plug-in “header” and expects some other data in return. No build-time linking, only headers, and paths that csound know. Regards,Anders29 sep. 2023 kl. 13:03 skrev bramtayl ***@***.***>:﻿
Well, I still don't understand, but I appreciate everyone's patience trying to explain it to me. I'm 95% sure if we move to using imported targets for dependencies, like in #1739, the plugins will no longer compile, but I could be wrong. If so, I'll flag this issue again so we can figure out how to handle it.
But to check my understanding:

the same size and layout as what the plugins were compiled against

Right, so the plugins need to be compiled again csound? So cmake needs to aware of the dependency?

This allows plugins to be loaded with different versions of Csound

Isn't that how dynamic linking always works? We only change the library name when the API changes, so that we can dynamically link any library with the same API?

vlazzarini:
This is not really the place for such conversation. If there's a problem, a ticket is open and we look into it. If there's a question on how to write and build a plugin we have two lists, a user and a developer's list and there are lots of knowledgeable people there to help. 

I'll try to close the conversation with this. Here's a trivial plugin:

```
#include <csdl.h>

typedef struct Opcode {
 OPDS h;
 MYFLT *out, *in
} OPCD;

static int perf(CSOUND *csound, OPCD *p) {
   *p->out = *p->in;
    return OK;
}

static OENTRY localops[] = {  {"opcd", sizeof(OPCD), 0, 2, "k", "k", NULL, (SUBR) perf} };

PUBLIC int64_t csound_opcode_init(CSOUND *csound, OENTRY **ep) { 
   (void) csound; 
   *ep = localops; 
   return (int64_t) sizeof(localops);  
};

PUBLIC int csoundModuleInfo(void) {                                      \
{ 
return ((CS_APIVERSION << 16) + (CS_APISUBVER << 8) + (int) sizeof(MYFLT)); 
}
```

The plugin code does not depend on any external functions, there is no need to link to anything. To build it you only need to call the compiler to compile it (passing the header location) and the linker to create a dynamic library. If you do this, then Csound can load this plugin and you have the opcode "opcd" available to your code.

PR #1754, opened 2023-10-10
Add open file callback

nonameentername:
Add callback for open file and open file descriptor to allow extending csound api when opening files.

These changes could allow csound to open files that are not located in file system and instead add user defined code to open files from memory, zip files, etc.

### DIFF
--- a/Engine/envvar.c
+++ b/Engine/envvar.c
@@ -764,7 +764,7 @@ static SNDFILE *csoundOpenFile_Snd(CSOUND *csound, const char *path, int mode, S
     SNDFILE *sf = NULL;
 
     if(csound->OpenSoundFileCallback_ != NULL) {
-        sf = csound->OpenSoundFileCallback_(csound, path, mode, &sfinfo);
+        sf = csound->OpenSoundFileCallback_(csound, path, mode, sfinfo);
     }
 
     return sf;
--- a/Engine/envvar.c
+++ b/Engine/envvar.c
@@ -760,6 +760,61 @@ char *csoundGetDirectoryForPath(CSOUND* csound, const char * path) {
 #endif  
 }
 
+static SNDFILE *csoundOpenFile_Snd(CSOUND *csound, const char *path, int mode, SFLIB_INFO *sfinfo){
+    SNDFILE *sf = NULL;
+
+    if(csound->OpenSoundFileCallback_ != NULL) {
+        sf = csound->OpenSoundFileCallback_(csound, path, mode, &sfinfo);
+    }
+
+    return sf;
+}
+
+static FILE *csoundOpenFile_Std(CSOUND *csound, char **fullName,
+                            const char *filename, const char *mode)
+{
+    FILE *f = NULL;
+
+    /* try to open file if callback is specified */
+    if(csound->OpenFileCallback_ != NULL) {
+        f = csound->OpenFileCallback_(csound, filename, (char*) mode);
+        if (f != NULL) {
+            *fullName = (char*) filename;
+        }
+    }
+
+    /* fallback to fopen */
+    if (f == NULL) {
+#if defined(WIN32)
+      /* To handle Widows errors in file name characters. */
+      size_t sz = 2 * MultiByteToWideChar(CP_UTF8, 0, filename, -1, NULL, 0);
+      wchar_t *wfname = malloc(sz);
+      wchar_t *wmode = 0;
+
+      MultiByteToWideChar(CP_UTF8, 0, filename, -1, wfname, sz);
+      sz = 2 * MultiByteToWideChar(CP_UTF8, 0, mode, -1, NULL, 0);
+      wmode = malloc(sz);
+      MultiByteToWideChar(CP_UTF8, 0, mode, -1, wmode, sz);
+      f = _wfopen(wfname, wmode);
+      if (UNLIKELY(f == NULL)) {
+        /* csoundErrorMsg(csound, Str("csound->FileOpen2(\"%s\") failed: %s."), */
+        /*                filename, strerror(errno)); */
+        free(wfname);
+        free(wmode);
+      } else {
+        *fullName = (char*) filename;
+        free(wfname);
+        free(wmode);
+      }
+#else
+      *fullName = (char*) filename;
+      f = fopen(filename, (char*) mode);
+#endif
+    }
+
+    return f;
+}
+
 static FILE *csoundFindFile_Std(CSOUND *csound, char **fullName,
                                 const char *filename, const char *mode,
                                 const char *envList)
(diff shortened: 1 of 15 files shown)


nonameentername:
This is work in progress.  I am looking into libsndfile open virtual and how to allow using that to open sound files.

http://www.mega-nerd.com/libsndfile/api.html#open_virtual

Maybe allow a user to set their own implementation of SF_VIRTUAL_IO and use that to open sound files.

vlazzarini:
Note that we don't use libsndfile directly (so Csound can be built without it). If you need to access libsndfile, you should do it via the interface in InOut/soundfile.c

kunstmusik:
@vlazzarini , @rorywalsh , @stekyne : Could you all take a look at this?  This relates to godot-csound. @nonameentername mentioned on Discord:

```
An example on how to use the callback can be found here:
https://github.com/nonameentername/godot-csound/blob/main/src/csound_godot.cpp#L37
https://github.com/nonameentername/godot-csound/blob/main/src/csound_godot.cpp#L562
```

Also there seems to be build failures in Github Actions related to iOS and macOS+VCPKG.

vlazzarini:
Not exactly sure what godot-csound is, and I am not following discussions on Discord.

I did a first pass on this and I didn't spot any major issue but I have a few points to make:

- this works crosses with various changes that are still in PRs and have not been merged. I'd like to see these outstanding PRs dealt with and merged and then do another review on how these changes fit.

- I see the callback setting in the host API, but not  in the module API. I suppose that's correct but need to check that modules won't need or want to do this.

- I also see changes in mp3in.c, and I suppose they are ok as this is being moved to use the standard C library, but also note that this code is likely to be moved out to the plugins repo. I guess this is neither here not there, but it all needs to work as a plugin. It's already a plugin in develop.

- The final point is that this should all compile without libsndfile. I don't see why it wouldn't but it needs to be checked.

kunstmusik:
@vlazzarini I was looking at this PR and I think FileOpen2 and setFileOpenCallback can be removed as FileOpen2 doesn't appear to be used anywhere. 

Otherwise, this seems to clean up some uses of fopen() in addition to the new API functions and functionality. Code seems good but I think we should do some smoke tests with Android, Web, and Desktop. 

Any issues you foresee @vlazzarini in relation to the newapi work?

vlazzarini:
FileOpen2 used to be part of the module API, but that's just called FileOpen now. 
I think these changes do not touch the new host API, but only add a couple of functions to the module API. It would be important to make sure the merge when it is done doesn't cause any issues.

My question is: are these callback setting functions to be called by hosts or by plugins? For the former, they need to be in csound.h, for the latter in CSOUND (csoundCore.h & csound.c). I can see the function might be useful for hosts, but would it be used by plugin opcodes? We want the APIs to be slim.

Incidentally and unrelated: I see copies there are copies of csound.h and csoundCore.h in the iOS examples folders. Is that wise? I guess these are out of date now unless they have been kept in sync by someone else.

nonameentername:
> @vlazzarini I was looking at this PR and I think FileOpen2 and setFileOpenCallback can be removed as FileOpen2 doesn't appear to be used anywhere.
> 
> Otherwise, this seems to clean up some uses of fopen() in addition to the new API functions and functionality. Code seems good but I think we should do some smoke tests with Android, Web, and Desktop.
> 
> Any issues you foresee @vlazzarini in relation to the newapi work?

@kunstmusik what is the expectation for smoke tests?  Is there an example in the project that I can use as reference?

> FileOpen2 used to be part of the module API, but that's just called FileOpen now. I think these changes do not touch the new host API, but only add a couple of functions to the module API. It would be important to make sure the merge when it is done doesn't cause any issues.
> 
> My question is: are these callback setting functions to be called by hosts or by plugins? For the former, they need to be in csound.h, for the latter in CSOUND (csoundCore.h & csound.c). I can see the function might be useful for hosts, but would it be used by plugin opcodes? We want the APIs to be slim.
> 
> Incidentally and unrelated: I see copies there are copies of csound.h and csoundCore.h in the iOS examples folders. Is that wise? I guess these are out of date now unless they have been kept in sync by someone else.

The intent of the callbacks in this pr is to allow using libcsound and specify a callback that will be used before csound opens a file.  If the callback is able to open the file then libcsound would use that instead of trying to open the file.

I am not sure if this would be considered the host or plugin.  Could you clarify the difference between these?

vlazzarini:
@nonameentername a host is a software that embeds Csound and manipulates it via the functions in csound.h. In this case, the CSOUND struct is opaque and there is no access to its members. This is the host API.

A plugin is a code extension that is added to Csound via the module callbacks or as an opcode and it accesses the CSOUND structure via csdl.h. In that case, all its public members (function pointers) are accessible. This is the module API.

Some functions are present in both, some are not. For example,  csoundPerformKsmps() is in the host API but there is no counterpart in the module API (opcodes and modules are not supposed to use it). FindFT() is in the module API but it does not exist in that form in the host API (searching for function tables is something opcodes do, hosts have a higher-level form of accessing tables).

It appears to me that what you propose is something a host would use.

nonameentername:
@vlazzarini I think I understand the difference now.  Based on your previous comment, would it be correct to say that I should revert the changes for `csoundCore.h` and only update `csound.h`?

vlazzarini:
Leave it there for now, I am hoping to do a final review of the APIs before we release and if it looks like it is not needed I can remove them.

kunstmusik:
> @kunstmusik what is the expectation for smoke tests? Is there an example in the project that I can use as reference?

@nonameentername Just looking to test the new code paths using a project that loads sound and non-sound files.

nonameentername:
@kunstmusik I have verified that the changes included here are working in the following platforms: Windows, MacOS, Linux, Android, iPhone and Wasm.  I'm using the changes in this pr on my project https://github.com/nonameentername/godot-csound and everything seems to be working fine.

Here are links to an example on using the callback:
[csound_set_callbacks](https://github.com/nonameentername/godot-csound/blob/31047cb0a36b68043f141ab02a39322ed26ac91b/src/csound_godot.cpp#L38-L39)
[open_file_callback](https://github.com/nonameentername/godot-csound/blob/31047cb0a36b68043f141ab02a39322ed26ac91b/src/csound_godot.cpp#L578-L617)
[open_soundfile_callback](https://github.com/nonameentername/godot-csound/blob/31047cb0a36b68043f141ab02a39322ed26ac91b/src/csound_godot.cpp#L619-L669)

please let me know if you have any questions.

kunstmusik:
@nonameentername Thanks for your patience and diligence on this PR. I did some light testing of the branch here and it seems to work well. It's also good to see Csound and Godot working together. 🫡  

Going to go ahead and merge this now. Thanks again for your work!

kunstmusik (approved):

PR #1756, opened 2023-10-11
use intl target

bramtayl:
`cmake` comes with a prepackaged `Intl` find module which creates a handy imported target.


CMakeLists.txt
```diff
@@ -559,10 +559,7 @@ else()
   message(STATUS "Not using atomic builtins - user disabled")
 endif()
 
-find_library(VORBISFILE_LIBRARY vorbisfile)
-#check_include_file(libintl.h LIBINTL_HEADER)
-find_library(LIBINTL_LIBRARY intl HINTS /usr/local/lib)
-find_package(Gettext)
+
```
bramtayl, line 562: This line is unrelated, but `VORBISFILE_LIBRARY` isn't used anywhere, figured I'd remove it while I was here

CMakeLists.txt
```diff
@@ -559,10 +559,7 @@ else()
   message(STATUS "Not using atomic builtins - user disabled")
 endif()
 
-find_library(VORBISFILE_LIBRARY vorbisfile)
-#check_include_file(libintl.h LIBINTL_HEADER)
-find_library(LIBINTL_LIBRARY intl HINTS /usr/local/lib)
-find_package(Gettext)
```
bramtayl, line 565: The gettext package isn't actually used anywhere

bramtayl: And clearly I was wrong about that

stekyne (commented):
I'm not really sure how the internationalisation stuff is supposed to work. Maybe @kunstmusik knows

CMakeLists.txt
```diff
@@ -1216,14 +1213,12 @@ if(USE_CURL AND CURL_FOUND)
 
 endif()
 
-# Linux platforms may or may not have a separate libintl (may be a part of of libc)
-set(LIBINTL_AVAIL (LIBINTL_LIBRARY OR LINUX))
-check_deps(USE_GETTEXT LIBINTL_AVAIL GETTEXT_MSGFMT_EXECUTABLE)
```
kunstmusik, line 1221: We use check_deps to check:

* Is the flag to enable compilation set to true? (Here USE_GETTEXT)
* Are the required dependencies for the target available? 

The value gets stored back into the first variable if the dependencies are missing. 

I think to get the same behavior you'll have to move the find_package()'s out from below and restore using check_deps.

kunstmusik (changes requested):

CMakeLists.txt
bramtayl: Well it's inside the `if (USE_GETTEXT)` block so the first one is already sorted. And `find_package` should handle the second one. So I think we're good?

kunstmusik: No, please re-read what I wrote. The way it is in this PR does not follow what is done elsewhere in regards to check_deps. The code in this PR checks for packages after testing for USE_GETTEXT, which will then fail the configuration process if the libraries aren't found. The way it's done elsewhere in the code base is it checks both "Can I?" (are the deps available) and "Should I?" (the USE_GETTEXT flag).

The build uses USE_GETTEXT set to true by default and therefore out-of-the-box will try to build with it and ignore if the deps aren't found. The user then also has the option to disable its use explicitly by setting USE_GETTEXT to false. Because Csound can build without gettext, this is the behavior we want out of the build system.

bramtayl: Ok, I understand now. Sure, I can restore `check_deps`. Does seem a bit odd to allow users without `gettext` to build with `USE_GETTEXT` though.

kunstmusik:
Looks good, test on macOS worked well for USE_GETTEXT on and off, CI builds completed. Thanks!

PR #1758, opened 2023-10-23
WIP: Use target_include_directories

bramtayl:
This is more fine-grained handling of `include_directories` that specifies include directories by target.

Waiting for a bunch of other PRs


vlazzarini:
Not sure why #1748 is being referenced here. That one is closed.

@kunstmusik  when you have a chance, could you look at this Java/ subdirectory. It would be good to eliminate SWIG altogether, if you have Java bindings that you have written and could contribute.

PR #1759, opened 2023-10-23
Add samplerate module

bramtayl:


bramtayl:
Will need to rebase onto #1760

bramtayl:
Bump?

bramtayl:
Part of #1811 now

Issue #144, opened 2014-03-24
Windows environment variables not cleaned up on uninstall. [sf bugs #144]


kunstmusik:
**Originally reported by Sourceforge User: afillebrown**

When uninstalling Csound on Windows the PATH variable still contains the "bin" portion of the Csound directory, and the PYTHONPATH and RAWWAVE_PATH variables are not being removed.

PR #1760, opened 2023-10-23
Prefer config packages

bramtayl:
cmake prefers "find" modules to "config" modules by default. This is a problem because the vcpkg "config" modules are generally better, with one exception, see https://github.com/csound/csound/pull/1760#issuecomment-1785121404. So set cmake to prefer "config" modules, and force liblo to use a "find" module for now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,11 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.15)
 cmake_policy(SET CMP0077 NEW)
 
 # silence RPATH cmake warning
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
 
 
 option(USE_VCPKG "Use VCPKG to download and manage dependencies" OFF)
--- a/Opcodes/CMakeLists.txt
+++ b/Opcodes/CMakeLists.txt
@@ -117,7 +117,7 @@ if(BUILD_DSSI_OPCODES)
 endif()
 
 ## OPCODES WITH EXTERNAL DEPENDENCIES ##
-find_package(LIBLO)
+find_package(LIBLO MODULE)
 if(BUILD_OSC_OPCODES AND LIBLO_FOUND)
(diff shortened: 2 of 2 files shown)


bramtayl:
Nope, was wrong about this

bramtayl:
Ok, so the vcpkg liblo find module appears to be broken on windows. So I've removed the osc plugin from the windows installer. But I'm confused as to how it was formerly working??

stekyne:
Can you explain why it was broken?

bramtayl:
Eh, ok, the existing find module works well enough on windows it looks like. It's just the vcpkg find module that doesn't work. I'm pretty sure that this is because the vcpkg liblo version is very out of date. See #144.

bramtayl:
Ok, I've updated the title, and edited the first comment with an explanation.

stekyne:
@bramtayl you can update liblo in vcpkg. I was the one who added it in the first place. I never updated it since. I thought it automatically updated these days though if it was using github releases. I can maybe take a look

bramtayl:
Sure, in the mean time forcing liblo to use the find module works well enough though?

CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ cmake_policy(SET CMP0077 NEW)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
```
stekyne, line 8: We need to set the minimum cmake version to 3.15 here. https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_PREFER_CONFIG.html

Lets add that in this PR as well

stekyne:
I checked the version of liblo in the vcpkg and it's the latest as shown in the liblo repo itself. Maybe I can check the find config for liblo in vcpkg

bramtayl:
Was fixed by https://github.com/radarsat1/liblo/pull/123 I think, which was in Sep 7, 2022

CMakeLists.txt
bramtayl: Ok done!

stekyne:
Ah right I see. Ok then I guess nothing can be done until they tag another release.

bramtayl:
Bump

bramtayl:
Moved this back into #1739

PR #1761, opened 2023-10-23
Add portmidi module

bramtayl:


stekyne:
This should probably be tested on Mac as well to be sure

bramtayl:
Why isn't there CI on Mac?

stekyne:
It's just a bit more complicated. I was in the process of doing it but ran into some issues. I'll probably take a look again soon.

bramtayl:
Part of #1811 now

PR #1762, opened 2023-10-23
Add portaudio module

bramtayl:


platform/vcpkg/ports/portaudio/vcpkg.json
bramtayl, line 1: @stekyne was there are reason you had a custom PortAudio port? Looks similar to the vcpkg default

stekyne: Yeah, there is a missing component that has to be downloaded so I had to fork it for Csound to get the same version of Portaudio we had before VCPKG.

InOut/CMakeLists.txt
```diff
@@ -113,15 +109,14 @@ if(USE_PULSEAUDIO)
     make_plugin(rtpulse rtpulse.c ${PULSEAUDIO_LIBRARY} ${PULSESIMPLE_LIBRARY})
 endif()
 
-if(PORTAUDIO_FOUND OR USE_VCPKG)
+if(USE_PORTAUDIO AND ((TARGET portaudio_static) OR (TARGET portaudio)))
```
stekyne, line 112: Is this the recommended way to check if a target exists? Isn't it convention to set `HAVE_X` package?

bramtayl: Yeah, but the vcpkg portaudio find module doesn't set a variable like that

platform/vcpkg/ports/portaudio/vcpkg.json
bramtayl: Oh I see. Ok, ref #1765

InOut/CMakeLists.txt
stekyne: Yeah but if you have `find_package(portaudio CONFIG REQUIRED)` the whole build fails if it's not found. It seems redundant to check here in this kind of awkward way.

bramtayl: > find_package(portaudio CONFIG REQUIRED) the whole build fails if it's not found

That would make sense to me too. But @kunstmusik was saying that we want to turn variables like `USE_PORTAUDIO` off if the dependencies aren't found. I would have done this with `check_deps`, but I couldn't figure out how to get the `(TARGET portaudio)` conditions to pass through as function arguments.

bramtayl: Added `check_deps` for clarity

stekyne: Ah right. Yeah I see what you mean. There is a flag `FAIL_NOT_FOUND` or something that we use to fail the build if a dependency is missing. We could possibly use it in that case.  Maybe we will need to explicitly check for the targets so

bramtayl:
Bump?

stekyne:
What is this needed for? It's been said elsewhere, you need to explain the changes and the justification for making alterations to something that has already been working. What's the advantage of it?

bramtayl:
Using imported targets has a few benefits:

- Unify vcpkg and non-vcpkg builds

bramtayl:
Part of #1811 now

PR #1764, opened 2023-10-30
Extend support for extra pfields 

gesellkammer:
This PR adds support for extra pfields for the API and includes support for the 'passign' opcode.


gesellkammer:
I don't know why there are fails in the checks, but this fix is really necessary for some use cases where score events with large number of pfields are needed.

vlazzarini:
Thanks, checks on Windows are wasm are failing because the CI builds are broken (not an issue here).

We need a PR for develop too as csound 6 is reaching EOL.

PR #1767, opened 2023-10-31
Link sdif2ad to csound

bramtayl:
Needs to be public because csound is used in the headers


stekyne:
As stated elsewhere, can we just add this to 1760 and make it one change. Chained PRs are tough to review properly thanks

PR #1768, opened 2023-10-31
Move LIBSNDFILE_LIBRARY earlier

bramtayl:
To reference `LIBSNDFILE_LIBRARY` in vcpkg builds


stekyne:
Hey Bram, appreciate all the work you're doign but it's very hard to piece together how all these small PRs fit together (there are 23 with your name currently).  It's ok to have a resonably large PR (<400-500 lines) if it forms some holistic change to the system. Right now all of these small PRs look ok in isolation but there is no outer context to see what other side effects might occur.

stekyne:
With this PR, we should really try to just work with targets anymore and not these vars that hold either libs or headers.

bramtayl:
Ok, sorry, I thought splitting it up would make it easier to review, not harder. I did go through today and separate out commits that don't rely on any others. Any of my PRs that requires other PRs is now labeled WIP, and we can worry about those after everything else goes through?

bramtayl:
> With this PR, we should really try to just work with targets anymore

Yeah I agree, this is just a temporary fix until #1739 goes through

bramtayl:
Ok, I'll just merge this into #1739

PR #1769, opened 2023-10-31
Java fixes

bramtayl:
Not 100% sure I got this right, but the java wrapper isn't tested and I'm not convinced it actually works anyways.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,8 @@ cmake_policy(SET CMP0077 NEW)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -5,6 +5,8 @@ cmake_policy(SET CMP0077 NEW)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
+cmake_policy(SET CMP0078 NEW)
+cmake_policy(SET CMP0086 NEW)
```
bramtayl, line 9: Was getting warnings about these policies not being set so set them

kunstmusik:
Looks good, going to merge, but we're also about to move away from using SWIG Java bindings for desktop Java. I'll be pushing changes shortly that will disable the Java bindings.

PR #1770, opened 2023-10-31
Test fixes

bramtayl:
Some easy simplifications. Note the use of `check_deps`

### DIFF
--- a/tests/c/CMakeLists.txt
+++ b/tests/c/CMakeLists.txt
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.5)
 
 find_package(GTest)
-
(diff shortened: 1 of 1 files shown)


tests/c/CMakeLists.txt
```diff
@@ -1,9 +1,9 @@
 cmake_minimum_required(VERSION 3.5)
 
 find_package(GTest)
```
stekyne, line 3: How do we prefer CONFIG if it's available over MODULE here?

stekyne: Ok I remember the flag you set elsewhere. We should keep all those similar changes together. Otherwise it's hard to see what's going on

PR #1771, opened 2023-10-31
Add mpeg

bramtayl:
To have mpeg enabled for vcpkg

### DIFF
--- a/vcpkg.json
+++ b/vcpkg.json
@@ -19,7 +19,8 @@
         {
             "name": "libsndfile",
             "features": [
-                "external-libs"
+                "external-libs",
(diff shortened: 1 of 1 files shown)


stekyne:
Just wondering why this is needed or how you discovered it should be added?

bramtayl:
To be honest I don't really remember well, but it was part of #1739. It makes sense to have mp3 support if the option is on by default though I think?

stekyne:
Yeah true. No harm adding it for now. There are numerous references toward mpeg in the code

stekyne (approved):

PR #1775, opened 2023-10-31
Link plugins to csound

bramtayl:
Already tried this once but. I don't understand the problem given it passes tests. And this is necessary for #1758 and #1739.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1169,6 +1169,8 @@ function(make_plugin libname srcs)
         math(EXPR i "${i}+1")
     endwhile()
 
+    target_link_libraries(${libname} PUBLIC ${CSOUNDLIB})
+
     set_target_properties(${libname} PROPERTIES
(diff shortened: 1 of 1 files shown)


bramtayl:
In the future, it might be good to have `CSOUNDLIB_PRIVATE` or something that has all the private headers as public. Then `CSOUNDLIB`, the plugins, and the executables could link to that instead.

stekyne:
I'm not sure really what's the right solution here. As the guys stated elsewhere, there shouldn't really be any linking done. Maybe this can be modelled in Cmake with an interface library or something? https://cmake.org/cmake/help/latest/command/add_library.html#id4 I'll need to read more tbh

bramtayl:
Yeah I'm still trying to get it straight

bramtayl:
This is interesting reading though I don't understand it: https://cmake.org/cmake/help/latest/prop_tgt/INTERFACE_LINK_LIBRARIES_DIRECT.html

bramtayl:
Ok, I think I understand better what's going on...some of the plugins use internal csound stuff that's not available in the public headers.

PR #1777, opened 2023-10-31
WIP: Fix alsa

bramtayl:
Waiting for #1762


cmake/Modules/Findportaudio.cmake
```diff
@@ -0,0 +1,42 @@
...
+)
+
+find_library(portaudio_LIBRARY
+	NAMES
+		portaudio portaudio_x64
+	PATHS
+		/usr/lib
+		/usr/local/lib
+		/opt/local/lib
+		/sw/lib
```
hlolli, line 23: these hardcoded paths will fail on many linux systems, maybe it should be combined with inferred search paths from the operating system's environment before checking hard-coded paths?

bramtayl: I think cmake should do that automatically? https://cmake.org/cmake/help/latest/command/find_library.html

stekyne: Yeah these are additional hints but cmake will check the normal locations dependning on the platform. Maybe use the word HINTS instead of PATHS as I was unsure until I checked. It seems like they do the same thing?

CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ cmake_policy(SET CMP0077 NEW)
 set(CMAKE_MACOSX_RPATH 1)
 set(CMAKE_EXPORT_COMPILE_COMMANDS 1)
 set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
+set(CMAKE_FIND_PACKAGE_PREFER_CONFIG ON)
```
stekyne, line 8: Isn't this also set in another PR?

bramtayl:
Moved this back into #1739

PR #1780, opened 2023-11-02
Reenable tests

bramtayl:
This will be my last PR in a while I promise. I added as many of the tests back to CI as I could. Several of the tests are currently failing and/or timing out.

This brings test line coverage from like 10% to 40% which is...an improvement?

I think getting this done should be a priority before we make any other sizable changes, especially before any of the other tests regress

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,14 +111,9 @@ option(BUILD_DOCS "Build documentation (requires doxygen)" OFF)
 option(DEBUG_ERROR_ON_WARNING "For a debug build, error on a compiler# warning" OFF)
 option(BARE_METAL "To configure for bare metal build" OFF)
 
-
-
-if(NO_CS_SERVER)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}  -DNO_CSOUND_SERVER")
-endif()
-
 if(BARE_METAL)
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}  -DBARE_METAL")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}  -DNO_CSOUND_SERVER")
 endif()
 
 # Include this after the install path definitions so we can override them here.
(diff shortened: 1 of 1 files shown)


tests/c/CMakeLists.txt
```diff
@@ -1,56 +1,50 @@
...
-        csound_debugger_test.cpp
-        csound_message_buffer_test.cpp
-        csound_orc_compile_test.cpp
-        csound_orc_semantics_test.cpp
-        csound_type_system_test.cpp
-        engine_test.cpp
-        io_test.cpp
-        server_test.cpp
-    )
+add_executable(unittests
```
bramtayl, line 1: I just removed the condition at the start didn't change much else in this file. Git just isn't lining up the lines correctly

Python/CMakeLists.txt
```diff
@@ -4,10 +4,17 @@ option(BUILD_PYTHON_INTERFACE "Build the Python interface" ON)
...
+else()
+      set(NOT_APPLE TRUE)
+endif()
+check_deps(BUILD_PYTHON_INTERFACE NOT_APPLE PYTHON3_EXECUTABLE)
+
+if(BUILD_PYTHON_INTERFACE)
       execute_process (
-           COMMAND ${PYTHON3_EXECUTABLE} -c
-           "import site, sys; sys.stdout.write(site.getsitepackages()[0])"
+           COMMAND ${PYTHON3_EXECUTABLE} -m site --user-site
```
bramtayl, line 17: Changed python to install to the user, not the system, dir, because it's easier to get permissions

CMakeLists.txt
```diff
@@ -183,7 +183,7 @@ option(NEW_PARSER_DEBUG "Enable tracing of new parser" OFF)
 option(BUILD_MULTI_CORE "Enable building for multicore system" ON)
 option(FAIL_MISSING "Fail when a required external dependency is not present (useful for packagers)" OFF)
 option(USE_GETTEXT "Use the Gettext internationalization library" ON)
-option(BUILD_STATIC_LIBRARY "Also build a static version of the csound library" OFF)
+option(BUILD_STATIC_LIBRARY "Also build a static version of the csound library" ON)
```
bramtayl, line 186: No harm in turning this on by default, I think, and it's necessary for tests

bramtayl:
Ok, updates:

I took all the csd files in the test folder and categorized them into:

- `passing`: files that generate wav files with consistent checksums
- `indeterminate_checksum`: files that generate wav files with indeterminate checksums
- `no_output`: files that generate no output
- `hanging`: files that take too long to generate
- `linux_only`: files that only work on linux

I also consolidated the python scripts and updated it to generate checksums from within python

Each csd file now has its own ctest

bramtayl:
I bumped the tempo on the tests up to 1000! bpm so to get the tests to finish in a reasonable amount of time

bramtayl:
Also added an option to track test coverage, plus a little script to generate a coverage report

PR #1782, opened 2023-11-08
Run include what you use

bramtayl:
include-what-you-use automatically detects and adds headers you've used, so I ran it. I also added a `map.imp` file to map some system files to more general ones. Only issue I ran into was that I couldn't get `signalflowgraph` not to rely on `csoundCore`, which violates a rule in the code. So I added an exception for it for now?


bramtayl:
Ok I give up can't get this working on windows

PR #1744, opened 2023-09-07
add python option

bramtayl:
Because installing python requires admin privilege

PR #1787, opened 2023-11-15
Use the built-in cmake python module

bramtayl:
Python has a built-in python find module that does a lot of the work for us

Also added a mini-macro for evaluating boolean expressions for convenience

### DIFF
--- a/Python/CMakeLists.txt
+++ b/Python/CMakeLists.txt
@@ -1,22 +1,14 @@
 ## Python module ##
 
-option(BUILD_PYTHON_INTERFACE "Build the Python interface" ON)
+option(INSTALL_PYTHON_INTERFACE "Build the Python interface" ON)
 
-find_program(PYTHON3_EXECUTABLE python3)
+find_package(Python3)
 
-if(BUILD_PYTHON_INTERFACE AND NOT APPLE AND PYTHON3_EXECUTABLE)
-      execute_process (
-           COMMAND ${PYTHON3_EXECUTABLE} -c
-           "import site, sys; sys.stdout.write(site.getsitepackages()[0])"
-           OUTPUT_VARIABLE PYTHON3_SITE_PACKAGES
-         )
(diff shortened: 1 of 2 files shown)


Python/CMakeLists.txt
```diff
@@ -1,22 +1,14 @@
 ## Python module ##
 
-option(BUILD_PYTHON_INTERFACE "Build the Python interface" ON)
+option(INSTALL_PYTHON_INTERFACE "Build the Python interface" ON)
```
stekyne, line 3: Can't really rename this as it would break the public API that people are using

Python/CMakeLists.txt
```diff
@@ -1,22 +1,14 @@
...
-      SET(PYTHON3_MODULE_INSTALL_DIR ${PYTHON3_SITE_PACKAGES} CACHE
-       PATH "Python3 module install dir")
-       
-      if (PYTHON3_MODULE_INSTALL_DIR)
-         message(STATUS "PYTHON3 MODULE INSTALL DIR: ${PYTHON3_MODULE_INSTALL_DIR}" )
-        install(FILES ${CMAKE_CURRENT_SOURCE_DIR}/ctcsound.py
-            DESTINATION ${PYTHON3_MODULE_INSTALL_DIR})
-      endif()
+assign_bool(NOT_APPLE NOT APPLE)
+check_deps(INSTALL_PYTHON_INTERFACE NOT_APPLE Python3_FOUND)
```
stekyne, line 8: Why not on APPLE? The existing code doesn't seem to make that distinction?

Python/CMakeLists.txt
bramtayl: Well, I added the option in https://github.com/csound/csound/pull/1744 two months ago, so I doubt anyone is relying on it, but happy to use the original name

Python/CMakeLists.txt
bramtayl: Well, there is a `AND NOT APPLE` in the original code, I have no idea why it's there though. We could potentially try to get rid of it and see what happens?

PR #1789, opened 2023-12-01
Add coverage

bramtayl:
So we can see what parts of the code aren't tested and add tests for them


CMakeLists.txt
stekyne, line 1: My comments from the other PR still stand. I don't think it's a good side effect that `TRACK_COVERAGE` can disable optimisations. This is something that is not intuitive and could lead to issues. I think the coverage build should be a separate target to isolate the changes needed to the build. 

I also think the coverage should only really be applicable to the engine itself and not the opcodes. If the coverage build requires separate build options like disabling optimisations then this is the appropriate way to handle that imo.

bramtayl: > coverage should only really be applicable to the engine itself and not the opcodes

Hmm, why?

stekyne: The plugins by and large are external from the engine but we have them here out of convienence mostly. I don't think there is much unit testing that can be done for DSP algorithms outside of some of the integration tests. The engine is different.

bramtayl: > I don't think there is much unit testing that can be done

Don't sound so defeatist! We can make sure everything gets tested! I can help!

stekyne: Well I'm not sure I've ever seen DSP code being unit tested well. It's pretty hard to do seeing as it's not really logic based. It's also fairly obvious if the code is not working correctly when listening to it. The engine should take priority as it's more state driven.

bramtayl:
Added a coverage CI action and uploaded the results to codecov

bramtayl:
Example report: https://app.codecov.io/github/csound/csound/tree/bramtayl%2Fcsound%3Aadd-coverage

stekyne:
Should exclude tests folder and the opcode folder tbh

bramtayl:
Ok, excluded the tests and opcode folders for now https://github.com/csound/csound/blob/797c80793c702d33eee9894594106c7cb963763e/generate_coverage.sh#L16

CMakeLists.txt
```diff
@@ -922,17 +927,17 @@ endif()
 # Handling New Parser
 
 set(YACC_SRC ${CMAKE_CURRENT_SOURCE_DIR}/Engine/csound_orc.y)
-set(YACC_OUT ${CMAKE_CURRENT_BINARY_DIR}/csound_orcparse.c)
-set(YACC_OUTH ${CMAKE_CURRENT_BINARY_DIR}/csound_orcparse.h)
+set(YACC_OUT ${CMAKE_CURRENT_BINARY_DIR}/Engine/csound_orcparse.c)
```
bramtayl, line 931: gcov gets confused if the inputs and outputs aren't in the same folder

CMakeLists.txt
```diff
@@ -1012,6 +1017,7 @@ set(PRELEX_SCOOUT ${CMAKE_CURRENT_BINARY_DIR}/csound_prslex.c)
 add_custom_command(
     OUTPUT ${PRELEX_SCOOUT}
     DEPENDS ${PRELEX_SCOSRC}
+    COMMAND mkdir -p ${CMAKE_CURRENT_BINARY_DIR}/Engine
```
bramtayl, line 1020: Make if it doesn't exist yet

CMakeLists.txt
```diff
@@ -1034,6 +1040,8 @@ if(MSVC)
 endif()
 
 list(APPEND libcsound_private_include_dirs ${CMAKE_CURRENT_BINARY_DIR})
+list(APPEND libcsound_private_include_dirs ${CMAKE_CURRENT_BINARY_DIR}/Engine) # for csound_orcparse.h
+list(APPEND libcsound_private_include_dirs ${CMAKE_CURRENT_BINARY_DIR}/Frontends/beats) # for beats.tab.h
```
bramtayl, line 1046: Add subfolders for generated headers

CMakeLists.txt
```diff
@@ -1009,6 +1014,7 @@ set(PRELEX_SCOOUT ${CMAKE_CURRENT_BINARY_DIR}/csound_prslex.c)
 #    COMMAND ${FLEX_EXECUTABLE} ARGS -B -t ${LEX_SCOSRC} > ${LEX_SCOOUT}
 #    )
 
+file(MAKE_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/Engine)
```
stekyne: Should you really have to manaully create these folders? If it's output here, it should make it at the same time. There aren't any other manually created folders in the binary dir

.github/workflows/csound_builds.yml
```diff
@@ -251,6 +251,56 @@ jobs:
...
+          sudo apt-get install cmake libjack-dev libpulse-dev ladspa-sdk dssi-dev autoconf libtool
+
+      - name: Configure Build
+        run: cmake -B build -S . -DUSE_VCPKG=1 -DCUSTOM_CMAKE="./platform/linux/custom.cmake" -DTRACK_COVERAGE=1
+
+      - name: Build Csound
+        run: cmake --build build --config Release
+
+      - name: Run integration and unit tests
+        run: cmake --build build --target test csdtests
```
stekyne: We don't need to run tests on this. It will have already run for the normal Linux build. Remove any step that is redundant

.github/workflows/csound_builds.yml
```diff
@@ -251,6 +251,56 @@ jobs:
...
+      
+      - name: Install gcovr
+        run: sudo apt-get install gcovr
```
stekyne: Just add this into the `Install dependencies` section

.github/workflows/csound_builds.yml
```diff
@@ -251,6 +251,56 @@ jobs:
...
+
+      - name: Make a coverage folder
+        run: mkdir coverage
```
stekyne: Merge this with the below step

cmake/coverage.cmake
```diff
@@ -0,0 +1,20 @@
+check_cxx_compiler_flag("-0g" has_debug_optimization)
+if (has_debug_optimization)
+    # disable optimizations for better coverage stats
+    add_compile_options("-0g")
```
stekyne: I noticed you're building for release in the CI. Why not remove this and just run coverage on debug builds?

stekyne:
This is looking good. I'd say we can probably merge it soon. @kunstmusik what do you think?

CMakeLists.txt
bramtayl: I thought so too, was working locally for me without this but CI failed without manually making the folders. Maybe something to do with permissions?

.github/workflows/csound_builds.yml
bramtayl: We need to run tests after building with `-DTRACK_COVERAGE=1`. I could add `-DTRACK_COVERAGE=1` to the normal Linux build though?

.github/workflows/csound_builds.yml
bramtayl: Ok done!

.github/workflows/csound_builds.yml
bramtayl: Ok done!

cmake/coverage.cmake
bramtayl: Oh that's a good idea, done!

bramtayl:
Before you merge might be a good idea to create a codecov token: https://docs.codecov.com/docs/adding-the-codecov-token. Though seems to work ok without it?

.github/workflows/csound_builds.yml
stekyne: Why?

bramtayl: The point of coverage is to tell you which lines of code were run during testing?

CMakeLists.txt
stekyne: Yeah I imagine so. I have experienced this before but I can't remember how it was resolved.

.github/workflows/csound_builds.yml
stekyne: No way really? 😄 

Gcov is a static analysis though? You feed in the source files and it gives you the result. That's irrespective of running the tests or not.

stekyne: https://github.com/shenxianpeng/gcov-example Ok i see how it works now. So there are two parts to it. That makes more sense to me now. Looks good so

PR #1790, opened 2023-12-01
add include guards

bramtayl:
Add include guards to header files without include guards. This prevents accidentally including the same file twice

### DIFF
--- a/Opcodes/partikkel.h
+++ b/Opcodes/partikkel.h
@@ -1,6 +1,6 @@
 /*
 Partikkel - a granular synthesis module for Csound 5
-Copyright (C) 2006-2009 Øyvind Brandtsegg, Torgeir Strand Henriksen,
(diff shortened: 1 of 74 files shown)


Opcodes/partikkel.h
```diff
@@ -1,6 +1,6 @@
 /*
 Partikkel - a granular synthesis module for Csound 5
-Copyright (C) 2006-2009 �yvind Brandtsegg, Torgeir Strand Henriksen,
+Copyright (C) 2006-2009 �yvind Brandtsegg, Torgeir Strand Henriksen,
```
stekyne: Hmm why did this change?

bramtayl: Hmm, something must have gone wrong with git I'll fix it

bramtayl: Done

PR #1791, opened 2023-12-06
Use threads module

bramtayl:
Use the built-in cmake threads module.

- Unifies vcpkg and non-vcpkg builds (although threads isn't a vcpkg package)


stekyne:
This does look pretty good. Is there anything that is risky about this change?

@kunstmusik what do you think?

kunstmusik:
The code here looks good to me. Android and Web builds are broken at the moment but I would not expect them to be affected by this change.  

 @bramtayl: I don't understand the PR description and it looks like it's some copy/paste from another PR. Could you update the description to describe this PR's changes in case we have to look back at this ticket?
 
Once that's cleared up I think we're good to merge @stekyne.

bramtayl:
Ok, updated, but the notes about using imported targets should mostly still apply here

kunstmusik:
Merging, thanks for the contribution

PR #1793, opened 2023-12-12
Add apple modules

bramtayl:

### DIFF
--- a/InOut/CMakeLists.txt
+++ b/InOut/CMakeLists.txt
@@ -9,9 +9,8 @@ option(USE_PORTMIDI "Build the PortMIDI I/O module" ON)
 option(USE_IPMIDI "Build the IPMIDI I/O module" ON)
 option(USE_JACK "Build the jack I/O module and opcodes" ON)
 option(USE_ALSA "Build the ALSA I/O module" ON)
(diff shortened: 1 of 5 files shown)


InOut/CMakeLists.txt
```diff
@@ -9,9 +9,8 @@ option(USE_PORTMIDI "Build the PortMIDI I/O module" ON)
 option(USE_IPMIDI "Build the IPMIDI I/O module" ON)
 option(USE_JACK "Build the jack I/O module and opcodes" ON)
 option(USE_ALSA "Build the ALSA I/O module" ON)
-# option(USE_COREAUDIO "Build the CoreAudio I/O module" ON)
-# option(USE_COREMIDI "Build the CoreMIDI I/O Module" ON)
-# option(USE_AUDIOUNIT "Build the CoreAudio AudioUnit I/O module (requires CoreAudio)" ON)
+option(USE_COREMIDI "Build the CoreMIDI I/O Module" ON)
+option(USE_AUDIOUNIT "Build the CoreAudio AudioUnit I/O module (requires CoreAudio)" ON)
```
stekyne, line 13: @vlazzarini I think this was disabled a while ago. Do you remember why? Is this needed now?

bramtayl:
Part of #1811 now

PR #1794, opened 2023-12-13
Enable apple ci

bramtayl:
Just an experiment to enable CI on apple


stekyne:
That's great you got around the flex/bison issue. What else is left to get this working?

bramtayl:
I think it's probably good to merge as is. It would be good to get the disabled test working on apple at some point though

tests/c/cscore_tests.cpp
```diff
@@ -35,17 +35,18 @@ void cscore(CSOUND *cs)
     cscoreListGetUntil(cs, 100);
 }
 
-TEST (CScoreTests, testCscore)
-{
-    csoundSetGlobalEnv("OPCODE6DIR64", "../../");
-    CSOUND *csound = csoundCreate(0);
+// TODO: fix on apple
```
stekyne, line 38: What needs to be fixed?

bramtayl: It didn't pass on CI. Don't have an apple computer so can't debug

stekyne:
Nice work thanks

PR #1795, opened 2023-12-15
IWYU

bramtayl:
include-what-you-use (iwyu) is software that automatically detects the headers you use and makes sure all of them are included.

I split this up into 7 commits due to the size; I can separate them into separate PRs if desired:

1) add_iwyu: This adds iwyu to the build process. I added a mappings file, which maps private headers to public headers. I also added a shell script, fix_includes.sh, which you can run to fix includes automatically.
2) update header flags: Here, I increase the list of headers we check for, and programmatically create a flag for each of them
3) WINSOCKAPI: This puts the `_WINSOCKAPI_` flags around `#include <windows.h>`. The flag makes sure windows.h doesn't include `winsock.h`, because we want to include `winsock2.h` instead.
4) pvoc_forward: This moves the `PVOC_GLOBALS` forward declaration to its own file. This seems to have made `iwyu` less confused.
5) `REMOT_BUF`: `REMOTE_BUF` was declared in `remote.h`, but used in `csoundCore.h`, and `remote.h` includes `csoundCore.h`. So I moved `REMOTE_BUF` to `csoundCore.h`.
6) Standalone headers: I moved the main header to the top of each c file. So, for example, I moved `#include "cs_par_base.h"` to the start of `cs_par_base.c`. This helps `iwyu`, and also makes sure each header file works as a standalone. I had to include some extra headers to make sure all the headers work standalone.
7) I ran `iwyu`. This is by far the largest commit. But I think the changes should be pretty straightforward to review, because `iwyu` adds detailed comments about why to include each file.


bramtayl:
include-what-you-use (iwyu) is software that automatically detects the headers you use and makes sure all of them are included.

I split this up into 7 commits due to the size; I can separate them into separate PRs if desired:

1) add_iwyu: This adds iwyu to the build process. I added a mappings file, which maps private headers to public headers. I also added a shell script, fix_includes.sh, which you can run to fix includes automatically.
2) update header flags: Here, I increase the list of headers we check for, and programmatically create a flag for each of them
3) WINSOCKAPI: This puts the _WINSOCKAPI_ flags around `#include <windows.h>`. The flag makes sure windows.h doesn't include `winsock.h`, because we want to include `winsock2.h` instead.
4) pvoc_forward: This moves the declaration of the `PVOC_GLOBALS` forward declaration to its own file. This seems to have made `iwyu` less confused.
5) `REMOT_BUF`: `REMOTE_BUF` was declared in `remote.h`, but used in `csoundCore.h`, and `remote.h` includes `csoundCore.h`. So I moved `REMOTE_BUF` to `csoundCore.h`.
6) Standalone headers: I moved the main header to the top of each c file. So, for example, I moved `#include "cs_par_base.h"` to the start of `cs_par_base.c`. This helps `iwyu`, and also makes sure each header file works as a standalone. I had to include some extra headers to make sure all the headers work standalone.
7) I ran `iwyu`. This is by far the largest PR. But I think the changes should be pretty straightforward to review, because `iwyu` adds detailed headers about why to include each file.

stekyne (commented):
This does look interesting. Can you say anything about build times changes, binary size changes etc? Any noticeable difference at all?

bramtayl:
Same size, takes about 50% longer to build with iwyu turned on. But I'm guessing most people won't have iwyu installed so it won't make a difference in most cases?

CMakeLists.txt
```diff
@@ -513,17 +527,38 @@ else()
...
+    stdint.h
+    sys/ioctl.h
+    sys/select.h
+    sys/socket.h
+    sys/time.h
+    sys/types.h
+    termios.h
+    unistd.h
+    values.h
+    winsock.h
```
stekyne, line 569: Can you explain where all these extra headers came from?

CMakeLists.txt
```diff
@@ -111,7 +111,21 @@ option(BUILD_DOCS "Build documentation (requires doxygen)" OFF)
 option(DEBUG_ERROR_ON_WARNING "For a debug build, error on a compiler# warning" OFF)
 option(BARE_METAL "To configure for bare metal build" OFF)
 option(CUSTOM_MALLOC "Use custom malloc in bare metal builds" OFF)
-
+option(INCLUDE_WHAT_YOU_USE "Run include-what-you-use" ON)
```
stekyne: This should be off by default. Also the variable name is a bit misleading. Should be `RUN_INCLUDE_WHAT_YOU_USE` surely?

stekyne:
It doesn't seem like this should be run every time by default even if installed. We should just do it periodically. There is no point in making the build 50% slower for every developer. I was asking about the actual build time without this tool being run. With headers included directly after running this tool, was there any noticeable improvement by any metric? Otherwise, this is a large PR for seemingly little value. It's a lot of git noise which will make merging branches possibly a lot more difficult later.

CMakeLists.txt
bramtayl: Ok, turned it off by default and renamed

CMakeLists.txt
bramtayl: The additional headers are all either POSIX headers or glibc extensions. These aren't available on e.g. Windows. To see where they are used, you can search for them. For example, `arpa/inet.h` is used in `OOps/remote.c`

bramtayl:
> was there any noticeable improvement

Well, you can read about several reasons to include-what-you-use here: https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md

Some of the reasons don't apply here, but the ones about fewer recompiles, refactoring, and self-documentation do

bramtayl:
I think this needs #1791 or something like it to pass on macos

PR #1796, opened 2023-12-19
target_include_directories

bramtayl:
Uses `target_include_directories` instead of `include_directories` for more fine-grain control.

Eventually, it would be good for the plugins and utilities not to use the private include directories, but currently they do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,6 +35,9 @@ project(Csound)
 
 enable_testing()
 
+set(libcsound_private_include_dirs "")
+set(libcsound_public_include_dirs "")
+
 SET(BUILD_SHARED_LIBS ON)
 
 if (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang")
@@ -194,7 +197,7 @@ endif()
 
 if(WIN32 AND NOT MSVC)
     if(EXISTS "C:/MinGW/include")
-        include_directories(C:/MinGW/include)
+        list(APPEND libcsound_private_include_dirs "C:/MinGW/include")
     else()
         MESSAGE(STATUS "MinGW include dir not found.")
     endif()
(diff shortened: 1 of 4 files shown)


InOut/CMakeLists.txt
```diff
@@ -157,25 +155,14 @@ if(USE_IPMIDI)
...
-            ${JACK_LIBRARY} ${PTHREAD_LIBRARY})
-    elseif(WIN32)
-        check_deps(JACK_INCLUDE_PATH)
-        include_directories("${JACK_INCLUDE_PATH}")
-        list(APPEND rtjack_LIBS
-            ${JACK_LIBRARY} ${PTHREAD_LIBRARY})
-    else()
-        list(APPEND rtjack_LIBS
-            ${JACK_LIBRARY} ${PTHREAD_LIBRARY})
-    endif()
```
stekyne, line 174: Why did you remove this code?

bramtayl: Well, `list(APPEND rtjack_LIBS ${JACK_LIBRARY} ${PTHREAD_LIBRARY})` was in all 3 branches. And we'd need the jack include directory regardless because we `#include <jack/jack.h>`

CMakeLists.txt
```diff
@@ -568,14 +571,17 @@ endif()
...
 
-include_directories(./H)
-include_directories(./include)
-include_directories(./Engine)
+list(APPEND libcsound_private_include_dirs ${Csound_SOURCE_DIR}/H)
+list(APPEND libcsound_public_include_dirs 
+    $<BUILD_INTERFACE:${Csound_SOURCE_DIR}/include>
+    $<INSTALL_INTERFACE:include>
+)
+list(APPEND libcsound_private_include_dirs ${Csound_SOURCE_DIR}/Engine)
```
stekyne, line 579: These first and last lines can be combined no?

CMakeLists.txt
```diff
@@ -35,6 +35,9 @@ project(Csound)
 
 enable_testing()
 
+set(libcsound_private_include_dirs "")
+set(libcsound_public_include_dirs "")
```
stekyne, line 39: Should these be upper case just to be consistent? Could probably just be `LIBCSOUND_PRIVATE_INCLUDES` etc.

stekyne:
This does look like a good change. I think if we can fix the formatting a bit, we can merge this.

EDIT: Just going to merge this now actually seeing as it's relevant to something I'm working on also

PR #1798, opened 2023-12-27
Allow building for windows using mingw with vcpkg

nonameentername:
Changes to allow cross compiling for windows using mingw and vcpkg.


CMakeLists.txt
```diff
@@ -12,7 +12,12 @@ option(USE_VCPKG "Use VCPKG to download and manage dependencies" OFF)
 
 if (USE_VCPKG)
     # Detect correct triplet based on CMake env
-    if (WIN32)
+    if (CMAKE_SYSTEM_NAME STREQUAL "Mingw")
+        set(VCPKG_TARGET_TRIPLET "x64-mingw-static")
+        set(VCPKG_INSTALL_OPTIONS "--allow-unsupported")
+        set(PTHREAD_LIBRARY "-lpthread")
+        set(WIN32 1)
```
stekyne: This is provided by cmake no?

nonameentername: Yeah, it's provided by cmake.  I removed it from my pull request.

scripts/mingw/Dockerfile
stekyne, line 1: It might be better to move these files into the "platform/mingw" folder instead of creating this "scripts" folder.

stekyne: There is also a "Dockerfiles" folder if it makes sense to move them there

nonameentername: I moved the files to the `mingw64-linux` folder.

stekyne:
Can you add  this to the build CI as well? There is a `csound_builds.yaml` file where you can copy the template of the others with the build steps for this.

nonameentername:
I updated the `csound_builds.yaml` file to include the windows mingw build and also updated tests to run them with wine.

tests/commandline/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 cmake_minimum_required(VERSION 3.5)
 
-if(MSVC)
+if (MINGW)
+  add_custom_target(csdtests ${Python_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/test.py --use-wine=true --csound-executable=${CMAKE_BINARY_DIR}/csound --opcode6dir64=${CMAKE_BINARY_DIR}/${CMAKE_BUILD_TYPE} --source-dir=${CMAKE_CURRENT_SOURCE_DIR})
```
kunstmusik, line 4: Seems like it should work to revert the --use-wine change and then pass in using --csound-executable="wine ${CMAKE_BINARY_DIR}/csound". @nonameentername have you tried this?

CMakeLists.txt
```diff
@@ -12,7 +12,11 @@ option(USE_VCPKG "Use VCPKG to download and manage dependencies" OFF)
 
 if (USE_VCPKG)
     # Detect correct triplet based on CMake env
-    if (WIN32)
+    if (CMAKE_SYSTEM_NAME STREQUAL "Mingw")
```
stekyne: https://cmake.org/cmake/help/latest/variable/MINGW.html there is a variable for this

CMakeLists.txt
```diff
@@ -12,7 +12,11 @@ option(USE_VCPKG "Use VCPKG to download and manage dependencies" OFF)
...
+        set(VCPKG_TARGET_TRIPLET "x64-mingw-static")
+        set(VCPKG_INSTALL_OPTIONS "--allow-unsupported")
+        set(PTHREAD_LIBRARY "-lpthread")
```
stekyne: Why are these needed? I don't think they should be

platform/mingw64-linux/build_debug.sh
```diff
@@ -0,0 +1,37 @@
...
+    -DUSE_VCPKG=1 \
+    -DBUILD_JAVA_INTERFACE=OFF \
+    -DINSTALL_PYTHON_INTERFACE=OFF \
+    -DBUILD_UTILITIES=OFF \
+    -DBUILD_TESTS=OFF \
+    -DCMAKE_SYSROOT=/usr/x86_64-w64-mingw32 \
+    $src_dir
+
+make
+make install
```
stekyne, line 33: https://github.com/csound/csound/blob/master/platform/windows/Custom-vs.cmake
We normally make a custom.cmake file and feed that into the cmake command. It should be more flexible than this.

`cmake -B build -S . -DUSE_VCPKG=1 -DCUSTOM_CMAKE="./platform/mingw/custom.cmake"` The command should just end up being this then.

As for the last commands to copy files, this could possibly be added into the cmakelists files afaik

platform/mingw64-linux/dependencies.sh
```diff
@@ -0,0 +1,34 @@
...
+apt install -y g++-mingw-w64
+apt install -y git
+apt install -y mingw
+apt install -y mingw-w64-tools
+apt install -y vim
+apt install -y cmake
+apt install -y curl
+apt install -y zip
+apt install -y flex bison
+apt install -y wine
```
stekyne, line 18: Have you checked which of these are already installed on the github action runner? I suspect most are so this might be unnecessary. It might be useful for local runs though. I would avoid calling it in the CI if it's not needed.

CMakeLists.txt
nonameentername: vcpkg gives the following error message if `--allow-unsupported` is not specified when cross-compiling:
```
vcpkg-msbuild is only supported on 'native & windows', which does not match x64-linux. This usually means that there are known build failures, or runtime problems, when building other platforms. To ignore this and attempt to build vcpkg-msbuild anyway, rerun vcpkg with `--allow-unsupported`.
```

CMakeLists.txt
nonameentername: I tried using `MINGW` but it is wasn't defined.  Not sure if this is related to cross-compiling.

tests/commandline/CMakeLists.txt
nonameentername: I tried using csound-executable with wine but it did not work.  It treats the space as part of the name of the executable.

platform/mingw64-linux/dependencies.sh
nonameentername: I updated the the ci to avoid installing the dependencies.

platform/mingw64-linux/build_debug.sh
nonameentername: I updated the build to use the above command.

tests/commandline/CMakeLists.txt
kunstmusik: I think this will be fine then for the PR and we can refactor another time.

kunstmusik:
@nonameentername I ran the github action to build and it's failing on mingw/vcpkg:

```
E: Unable to locate package mingw
```

Could you look into this and correct? Other than that, I think the rest is in good shape. (@stekyne please chime in on any further review.)

nonameentername:
@kunstmusik I updated the github action.

stekyne:
Just looking at the CI logs, any cause for concern? 

```
update-alternatives: using /usr/bin/x86_64-w64-mingw32-gcc-posix to provide /usr/bin/x86_64-w64-mingw32-gcc (x86_64-w64-mingw32-gcc) in manual mode
update-alternatives: warning: skip creation of /usr/bin/x86_64-w64-mingw32-gcc-8 because associated file /usr/bin/x86_64-w64-mingw32-gcc-8.3-posix (of link group x86_64-w64-mingw32-gcc) doesn't exist
update-alternatives: using /usr/bin/x86_64-w64-mingw32-g++-posix to provide /usr/bin/x86_64-w64-mingw32-g++ (x86_64-w64-mingw32-g++) in manual mode
```

nonameentername:
Only `x86_64-w64-mingw32-gcc` and `x86_64-w64-mingw32-g++` are being used.  It should be fine if `x86_64-w64-mingw32-gcc-8` doesn't get created.

kunstmusik:
Going to merge, thanks @nonameentername for the contribution!

PR #1803, opened 2024-01-04
modify printarray to print at init time by default for string arrays

gesellkammer:


vlazzarini:
Can we have a PR for develop? csound6 is reaching EOL.

PR #1804, opened 2024-01-04
Avoid warnings

bramtayl:
These changes avoid gcc and clang compilation warnings (haven't tested msvc). I separated it into several smaller commits with similar changes.


InOut/rtalsa.c
```diff
@@ -1856,7 +1856,7 @@ int listAlsaSeq(CSOUND *csound, CS_MIDIDEVICE *list, int isOutput) {
       /* reset query info */
       snd_seq_port_info_set_client(pinfo, snd_seq_client_info_get_client(cinfo));
       snd_seq_port_info_set_port(pinfo, -1);
-      count = 0;
+      // count = 0;
```
stekyne: Is this not read at all? Then maybe delete. Shouldn't have commented code

Opcodes/OpcodeBase.hpp
```diff
@@ -374,7 +379,7 @@ class OpcodeNoteoffBase
 };
 
 #if (__cplusplus >= 201103L) && !(defined(__wasi__))
-#pragma message("Defining heap_object_manager_t.")
+// #pragma message("Defining heap_object_manager_t.")
```
stekyne: Whats the warning here?

Opcodes/bformdec2.cpp
stekyne, line 1: Are you sure the default values make sense for the params?

tests/c/csound_debugger_test.cpp
```diff
@@ -174,7 +180,10 @@ TEST_F (DebuggerTests, testBreakpointInstrument)
 int count = 0;
 static void brkpt_cb5(CSOUND *csound, debug_bkpt_info_t *bkpt_info, void *userdata)
 {
-    debug_opcode_t *debug_opcode = bkpt_info->currentOpcode;
+    (void)(csound);
+    (void)(bkpt_info);
+    (void)(userdata);
+    // debug_opcode_t *debug_opcode = bkpt_info->currentOpcode;
```
stekyne: Whats the warning here?

stekyne (approved):

InOut/rtalsa.c
bramtayl: Ok, deleted instead of commented unused assignments

tests/c/csound_debugger_test.cpp
bramtayl: debug_opcode was unused. Maybe it's needed for breakpoint or something though?

Opcodes/bformdec2.cpp
bramtayl: Nope. But they were undefined before, so accessing them would have caused bugs/undefined behavior anyways

Opcodes/OpcodeBase.hpp
bramtayl: Oops, I thought I had seen somewhere that pragma messages count as warnings, but I guess I was wrong, so I reverted.

stekyne:
@kunstmusik what do you think?

bramtayl:
Bump

PR #1805, opened 2024-01-10
use target_compile_features

bramtayl:
Use `target_compile_features` for C++ 11. I think it's a pretty safe bet that any modern compiler will support C++ 11.


stekyne (approved):
Looks good. This should be the way we do this

PR #1806, opened 2024-01-12
Separate folders

bramtayl:
This puts plugins and executables into their own separate folders. I see this as having the following advantages:

- Code organization
- We can eventually track coverage of plugins and executables separate from the main library
- We can more easily move them out of the main repo if desired
- We can more easily isolate include directories, and make sure the main library hasn't accidentally included headers from plugins/utilities and vice-versa, if desired

I've separated this into several commits, each corresponding to reorganizing one folder


jpffitch:
Why do we need to do this?  We would have to relearn where things are ;
revise our tools etc.

Surey we should be using our time to deliver cs7?

vlazzarini:
I don't agree with this change either. We don't need it and it will just get in the way of us trying to finish Csound 7.

kunstmusik:
FWIW: I don't mind this change in principle as it follows a common code organization pattern for monorepos. I might change the folder names from this PR but that's trivial. It could be useful for new devs coming to the repo to find things. 

We could perhaps delay doing this change until a later time after we get further down CS7 changes, so that comparing to CS6 is a little easier for the time being (thinking if we get any further CS6 PRs and having to get them to develop too).

rorywalsh:
I'm with Steven. I also think it would make things more accessible for new
devs.

stekyne:
I'd also support changes to make it more accessible. Issues I can see:

- 33 directories in the root. Some of these can be moved down a level to group by function
- Files that are too long. Anything beyond 500 lines is probably too much. There can be exceptions depending ofc but in general, leverage the file system and IDEs tree view of the codebase to show it's structure.
- For Cmake, the way some directories are nested does not gel well with the ideal layout for targets etc
- OOps and Opcodes folder, probably should be combined into one with an appropriate structure
- Engine and Top folder, not quite sure why these are separate
- util, util1 and util3 seems redundant to have 3
- Android, Bela, Daisy, iOS, wasm could all go into `platform` probably
- Java, Python could be put into an `interfaces` folder 

These are things that would clear up the immediate root folder I think. Based on those comments I think something like this makes some sense.

```
- csound
  - cmake
    - FindSomeLib.cmake
    - something_else.cmake
  - DockerFiles
   - <docker files per platform>
  - include (public headers)
    - libcsound
      - csound.h
  - src (engine / libcsound)
    - includes
    - engine
     - lexer?
     - (maybe split the engine code up by function here)
    - opcodes
     - internal
     - other (not sure if to merge OOps and Opcodes, maybe there is no need to divide them)
    - transport (inputs and outputs)
    - utility (stuff from Top folder)
    - translations (not sure where this should go, does it belong to libcsound or the CLI?)
  - apps (frontends)
    - beats
    - csound
    - debugger
    - utilties (code from the 3 util folders)
  - assets
   - <example CSDs>
   - <samples>
  - tests
    - unit (c)
    - integration (command line)
    - regression
    - soak
  - docs
    - <markdown/latex files>
    - html manual
    - release notes
  - scripts
    - helper.py
  - platform
    - windows
    - osx
    - linux
    - wasm
  - installer
   - windows
   - macos
   - linux
  - interfaces
   - java
   - python
```

bramtayl:
Well, a bigger reorganization would be nice too, but this is limited to just plugins and executables at the moment. Happy to change the folder names etc.

bramtayl:
Let me know if/when people want to do this and I can rebase

PR #1809, opened 2024-01-15
Changes to allow compiling for MacOS using osxcross

nonameentername:


stekyne:
Can you explain what this is for?

nonameentername:
osxcross is a MacOS cross-compiler that allows compiling Mac binaries from Linux and *BSD.
More information can be found [here](https://github.com/tpoechtrager/osxcross).

The changes in this pull request are to allow compiling csound for MacOS using the cross-compiler.  This allows creating a MacOS csound binary and static/dynamic Libraries from Linux and Windows.

stekyne (commented):
Looks like a lot of work fair play. Is this ready? I'm a bit unsure about the vcpkg overlaid ports. Can we ensure they aren't used for any non-osxcross build?

nonameentername:
Yeah, the changes are ready.  I wanted to add some tests to the github build, but building the cross compiler takes a long time.
I can update the pull request if that's acceptable.

The changes to the port files are actually simple and only related to MacOS:

```diff
diff vcpkg/ports/gettext-libintl/portfile.cmake platform/vcpkg/ports/gettext-libintl/portfile.cmake
88a89,92
> if(VCPKG_TARGET_IS_OSX AND VCPKG_CROSSCOMPILING)
>     list(APPEND OPTIONS --host ${TARGET_TRIPLET}-darwin)
> endif()
>
$ diff vcpkg/ports/mp3lame/ platform/vcpkg/ports/mp3lame/
diff vcpkg/ports/mp3lame/portfile.cmake platform/vcpkg/ports/mp3lame/portfile.cmake
101a102,105
>     if(VCPKG_TARGET_IS_OSX AND VCPKG_CROSSCOMPILING AND VCPKG_OSX_ARCHITECTURES STREQUAL "x86_64")
>         list(APPEND OPTIONS --host "${VCPKG_OSX_ARCHITECTURES}-apple-darwin")
>     endif()
>
$ diff vcpkg/ports/portmidi/ platform/vcpkg/ports/portmidi/
diff vcpkg/ports/portmidi/framework_link.patch platform/vcpkg/ports/portmidi/framework_link.patch
38c38
< +      -Wl,-framework,CoreMidi
---
> +      -Wl,-framework,CoreMIDI
45c45
< +      -Wl,-framework,CoreMidi
---
> +      -Wl,-framework,CoreMIDI
diff vcpkg/ports/portmidi/search-for-threads-in-config.patch platform/vcpkg/ports/portmidi/search-for-threads-in-config.patch
12c12
<
---
>
```
I'm planning on submitting the port changes as a pull request to vcpkg and if they are accepted then they can be removed from csound's vcpkg ports.

stekyne:
Yeah I was wondering about the amount of overlays. I guess it would be ideal to get it integrated into those ports directly. 

Who in general uses this osxcross platform? I can't really say I've heard of it before this

nonameentername:
There are some open source projects that use osxcross platform.  The godotengine uses it to create their MacOS builds.
In general its used by developers that want to target MacOS but do not own a Mac computer.

.gitignore
```diff
@@ -237,3 +237,6 @@ xcode
 *.nupkg
 *.exe
 bin
+
+#vcpkg
```
stekyne: We have VCPKG as a submodule so this needs to be here. If VCPKG is external, then it installs them in vcpkg_installed so we can have both. Ideally we should use the internal one though as I don't think the overlays will be applied right?

CMakeLists.txt
```diff
@@ -207,6 +221,12 @@ if(APPLE)
     endif()
 endif()
 
+if(OSXCROSS_TARGET)
+    set(CMAKE_C_FLAGS "-I${OSXCROSS_SDK}/usr/include ${CMAKE_C_FLAGS}")
```
stekyne, line 225: Could use `list([APPEND](https://cmake.org/cmake/help/latest/command/list.html#append) <list> [<element>...])` syntax here but not really a big deal

.gitignore
nonameentername: I think this can be removed.  The `vcpkg_installed` directory is being stored in the build directory and that is already being ignored.

CMakeLists.txt
nonameentername: I tried using the following but I get build errors.
```cmake
list(PREPEND CMAKE_C_FLAGS -I${OSXCROSS_SDK}/usr/include)
list(PREPEND CMAKE_CXX_FLAGS -I${OSXCROSS_SDK}/usr/include)
list(APPEND CMAKE_SHARED_LINKER_FLAGS -L${CMAKE_OSX_SYSROOT}/usr/lib)
```
Maybe I'm missing something?

stekyne: Hmm I think we had issues with this before. I'm not exactly sure why. Maybe just leave it. Tbh we shouldn't be setting `CMAKE_C_FLAGS ` or similar flags directly. It should be set on the target and CMake will populate all of these for us. We can worry about that later.

.github/workflows/csound_builds.yml
```diff
@@ -124,6 +124,119 @@ jobs:
...
+      - name: Build osxcross
+        if: steps.osxcross.outputs.cache-hit != 'true'
+        run: |
+          git clone https://github.com/tpoechtrager/osxcross.git
+          sudo /home/runner/work/csound/csound/osxcross/tools/get_dependencies.sh
+          OSX_SDK="MacOSX14.0.sdk"
+          OSX_SDK_URL="https://github.com/joseluisq/macosx-sdks/releases/download/14.0/${OSX_SDK}.tar.xz"
+          curl -sSL "$OSX_SDK_URL" -o "/home/runner/work/csound/csound/osxcross/tarballs/$OSX_SDK.tar.xz"
+          UNATTENDED=1 /home/runner/work/csound/csound/osxcross/build_clang.sh
+          UNATTENDED=1 /home/runner/work/csound/csound/osxcross/build.sh
```
stekyne, line 176: How big is this to cache do you know?

.github/workflows/csound_builds.yml
```diff
@@ -124,6 +124,119 @@ jobs:
...
+          done
+          mkdir artifact
+          zip -r artifact/build-$GITHUB_SHA.zip build
+
+      - name: Temporarily save osxcross build
+        uses: actions/upload-artifact@v2
+        with:
+          name: osxcross-artifact
+          path: artifact
+          retention-days: 1
```
stekyne: Not sure about this. We are hitting some usage limits and this will add to it. Why can this be the same job in the workflow to build and test?

.github/workflows/csound_builds.yml
nonameentername: The cache for this is 1.7 GB

.github/workflows/csound_builds.yml
nonameentername: The cross-compiled artifacts are MacOS binaries and it's not possible to run them on the same Linux box that built them.

nonameentername: I removed the tests workflow for osxcross to avoid creating the artifacts.

.github/workflows/csound_builds.yml
stekyne: That might be too much. There is already the cache for each VCPKG dir and we have to share that with every platform. Also I didn't realise this takes 2.5 hrs to build. Is there anyway to run this locally and cache this at a URL elsewhere?

nonameentername: Yeah, it should be possible to cache this somewhere.  Is there some location that would be preferred?

stekyne: Hmm not really sure. Somewhere publicly hosted I guess

vlazzarini:
Not quite sure where we are with this. @stekyne are we good to merge?

stekyne:
I'm not sure how to proceed with this. It requires a lot of time (2.5 hrs build) and space (1.7gb for caching) to build and we've already run into trouble with resources for GH actions. I think this should be an optional CI build only and not automatic but it should be fine to add the support for OSXcross in CMake.

vlazzarini:
ok, @nonameentername can you switch off the CI build and then we can merge it?

nonameentername:
@vlazzarini I disabled the osxcross ci build.

vlazzarini:
thanks

PR #1810, opened 2024-01-15
Add scripts to allow building Windows locally

nonameentername:


platform/windows/build_debug.ps1
```diff
@@ -0,0 +1,13 @@
...
+
+cmake -B build -S . -DUSE_VCPKG=1 -DCUSTOM_CMAKE=".\platform\windows\Custom-vs.cmake" -DCMAKE_INSTALL_PREFIX:PATH=$InstallPrefix
+
+Write-Host "Build Csound"
+
+cmake --build build --config Debug
+
+Write-Host "Install Csound"
+
+cmake --build build --target install --config Debug
```
stekyne, line 13: Not entirely sure this is the best way to solve this. It only really helps the first time you clone the repo. I imagine if devs are changing code, they don't want to regenerate the whole project but just run the build command and let cmake pick up any changes internally. Maybe this is better solved with Cmake presets? I thought about adding that recently.

platform/windows/dependencies.ps1
```diff
@@ -0,0 +1,16 @@
+Write-Host "Install csound dependencies"
+
+choco install -y winflexbison3 innosetup llvm javaruntime jdk8 swig python
```
stekyne: What is llvm needed for?

nonameentername: It's not needed.  I removed it from the install script.

platform/windows/build_debug.ps1
nonameentername: I created these mainly for reproducing the github windows build.  I had an error with another pull request, but I couldn't see the error in the github logs.  I'm open to updating these to improve them.

vlazzarini:
How are we with this PR? Can it be merged?

nonameentername:
I think the changes are ready to be merged.

@stekyne are any changes needed before this can be merged?

vlazzarini:
Great, once @stekyne approves it we can merge and close this. Thanks a lot for it.

stekyne (approved):

PR #1811, opened 2024-01-18
Imported targets for dependencies

bramtayl:
Using imported targets has some benefits:

- Packaging include dirs and libraries together for cleaner code

Ideally, imported modules should keep track of indirect dependencies. For example, SndFile has an indirect dependency of mp3lame. This has additional advantages:

I decided not to do this here for a couple of reasons:
- This isn't being done as is
- There are a ton of indirect dependencies of things like SndFile and PulseAudio
- It should really only be necessary in Windows anyway, and on Windows, we're using vcpkg. vcpkg includes find modules that usually keep better track of indirect dependencies.


CMakeLists.txt
```diff
@@ -423,37 +423,30 @@ message(STATUS "BUILD_PLUGINS_DIR set to ${BUILD_PLUGINS_DIR}.")
...
-            message(STATUS "libsndfile not found")
-            set(USE_LIBSNDFILE 0)
-        endif()
-    endif()
+if (USE_VCPKG)
+    find_package(SndFile CONFIG)
+else()
+    find_package(SndFile MODULE)
 endif()
+check_deps(USE_LIBSNDFILE SndFile_FOUND)
```
stekyne, line 437: You're not checking if the sndfile target was actually found in the VCPKG branch

CMakeLists.txt
```diff
@@ -423,37 +423,30 @@ message(STATUS "BUILD_PLUGINS_DIR set to ${BUILD_PLUGINS_DIR}.")
 # OS specific checks
 TEST_BIG_ENDIAN(BIG_ENDIAN)
 
+# public if we use the headers of the dependency in our public headers
+set(libcsound_PUBLIC_LIBS "")
```
stekyne: There is an guideline for CMake not to use variables like this and just use the targets directly. https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#avoid-custom-variables-in-the-arguments-of-project-commands

InOut/CMakeLists.txt
```diff
@@ -9,55 +9,48 @@ option(USE_PORTMIDI "Build the PortMIDI I/O module" ON)
 option(USE_IPMIDI "Build the IPMIDI I/O module" ON)
 option(USE_JACK "Build the jack I/O module and opcodes" ON)
 option(USE_ALSA "Build the ALSA I/O module" ON)
-# option(USE_COREAUDIO "Build the CoreAudio I/O module" ON)
-# option(USE_COREMIDI "Build the CoreMIDI I/O Module" ON)
-# option(USE_AUDIOUNIT "Build the CoreAudio AudioUnit I/O module (requires CoreAudio)" ON)
+option(USE_COREMIDI "Build the CoreMIDI I/O Module" ON)
+option(USE_AUDIOUNIT "Build the CoreAudio AudioUnit I/O module (requires CoreAudio)" ON)
```
stekyne, line 13: Why enable this again?

InOut/CMakeLists.txt
```diff
@@ -9,55 +9,48 @@ option(USE_PORTMIDI "Build the PortMIDI I/O module" ON)
...
 
 list(APPEND CMAKE_REQUIRED_INCLUDES "/usr/local/include")
 
 # FIND LIBRARIES AND HEADERS
 
 if(USE_PORTAUDIO)
     if (USE_VCPKG)
-        find_package(portaudio CONFIG REQUIRED)
+        # vcpkg includes a config find module
```
stekyne: Unnecessary comment I'd say

InOut/CMakeLists.txt
```diff
@@ -9,55 +9,48 @@ option(USE_PORTMIDI "Build the PortMIDI I/O module" ON)
...
+        find_package(portaudio CONFIG)
```
stekyne: Tbh for the VCPKG build I am just failing if it doesn't find the dependency. It is just a lot easier. There is no extra variables or testing if the target exists or not. The VCPKG build is kind of the CI build so I think this is ok. For the other users using their system packages for dependencies, we can silently disable features if not found.

InOut/CMakeLists.txt
```diff
@@ -70,14 +63,16 @@ endif()
...
         make_plugin(cmidi cmidi.c)
-        target_include_directories(cmidi PRIVATE ${COREMIDI_INCLUDE_PATH})
-        target_link_libraries(cmidi PRIVATE ${COREMIDI_LIBRARY} ${COREFOUNDATION_LIBRARY})
+        target_link_libraries(cmidi PRIVATE CoreMIDI::coremidi CoreFoundation::corefoundation)
+    endif()
+    if (USE_AUDIOUNIT)
         make_plugin(rtauhal rtauhal.c)
-        target_include_directories(rtauhal PRIVATE ${AUDIOUNIT_INCLUDE_PATH})
-        target_link_libraries(rtauhal PRIVATE ${AUDIOUNIT_LIBRARY} ${COREFOUNDATION_LIBRARY} ${COREAUDIO_LIBRARY})
+        target_link_libraries(rtauhal PUBLIC AudioUnit::audiounit CoreFoundation::corefoundation CoreAudio::coreaudio)
```
stekyne: Is this really necessary to set as public? Afaik we don't expose the header for audiounit in libcsound so shouldn't need to do that? I could be wrong

InOut/CMakeLists.txt
```diff
@@ -70,14 +63,16 @@ endif()
 # endif()
 
 if(APPLE)
-#    check_deps(USE_COREMIDI APPLE COREMIDI_INCLUDE_PATH COREMIDI_LIBRARY COREFOUNDATION_LIBRARY)
-#   check_deps(USE_AUDIOUNIT APPLE COREAUDIO_INCLUDE_PATH COREAUDIO_LIBRARY AUDIOUNIT_INCLUDE_PATH AUDIOUNIT_LIBRARY COREFOUNDATION_LIBRARY)
+    check_deps(USE_COREMIDI APPLE CoreMIDI_FOUND CoreMIDI_FOUND CoreFoundation_FOUND)
+    check_deps(USE_AUDIOUNIT APPLE CoreAudio_FOUND AudioUnit_FOUND CoreFoundation_FOUND)
```
stekyne: Already checked for apple on line 72/65, can omit it I guess

InOut/CMakeLists.txt
```diff
@@ -110,37 +105,21 @@ if(USE_PULSEAUDIO)
...
+if(USE_PORTAUDIO)
     make_plugin(rtpa rtpa.c)
-    if(USE_VCPKG)
+    # use the static library if available
+    if (TARGET portaudio_static)
         target_link_libraries(rtpa PRIVATE portaudio_static)
     else()
-        target_include_directories(rtpa PRIVATE ${PORTAUDIO_INCLUDE_DIRS})
-        target_link_libraries(rtpa PRIVATE ${PORTAUDIO_LIBRARIES})
+        target_link_libraries(rtpa PRIVATE portaudio)
```
stekyne: Can add the targets like this instead to avoid duplication
```
target_link_libraries(rtpa PRIVATE
            $<IF:$<TARGET_EXISTS:portaudio>,portaudio,portaudio_static>
)
```

cmake/Modules/FindAudioUnit.cmake
stekyne, line 1: Bit unsure of adding more Find.cmake files. The recommendation is to move towards config files. Can we do that for these system libraries?

cmake/cmake-utilities.cmake
```diff
@@ -40,6 +40,8 @@ endfunction()
 function(make_utility name srcs)
     make_executable(${name} "${srcs}" "${CSOUNDLIB}")
     add_dependencies(${name} ${CSOUNDLIB})
+    # we use these dependencies in our public headers so link them too
+    target_link_libraries(${name} PUBLIC ${libcsound_PUBLIC_LIBS})
```
stekyne: This doesn't seem like it should be necessary. If we're linking in the libcsound target, it should bring in any dependencies with it?

cmake/cmake-utilities.cmake
```diff
@@ -40,6 +40,8 @@ endfunction()
 function(make_utility name srcs)
     make_executable(${name} "${srcs}" "${CSOUNDLIB}")
     add_dependencies(${name} ${CSOUNDLIB})
```
stekyne: I don't think this is needed in modern cmake. It should be enough to target_link_libraries libcsound

util/CMakeLists.txt
```diff
@@ -36,16 +33,18 @@ if(BUILD_UTILITIES)
     make_utility(srconv      srconv_main.c)
 
     if (USE_VCPKG)
-        find_package(SampleRate REQUIRED)
-        set(LIBSAMPLERATE_LIBRARY SampleRate::samplerate)
-    else ()
-        find_library(LIBSAMPLERATE_LIBRARY NAMES samplerate libsamplerate-0)
+        # vcpkg includes a config find module
+        find_package(SampleRate CONFIG)
+        # set a found variable because vcpkg doesn't
```
stekyne: Again these copy+pasted comments can be omitted. They're not adding much

stekyne (commented):
Looks good. Left some comments that we can discuss

CMakeLists.txt
bramtayl: I'm not? The vcpkg find module seems to be setting SndFile_FOUND

CMakeLists.txt
bramtayl: Ok, replaced with SndFile::sndfile, thanks for the tip!

InOut/CMakeLists.txt
bramtayl: Ok, I set it back to off by default. Seems a shame though if the plugins still work

InOut/CMakeLists.txt
bramtayl: Ok, deleted it!

InOut/CMakeLists.txt
bramtayl: Nice, changed it to that, thanks!

InOut/CMakeLists.txt
bramtayl: Ok, done!

InOut/CMakeLists.txt
bramtayl: Ok, I added REQUIRED back for the vcpkg branches

InOut/CMakeLists.txt
bramtayl: Oops, changed to PRIVATE, thanks!

cmake/cmake-utilities.cmake
bramtayl: Ok, I deleted it.

cmake/cmake-utilities.cmake
bramtayl: I agree, linking to libcsound would be the sane option. But we aren't allowed to do this I guess, https://github.com/csound/csound/pull/1748#issuecomment-1740207758, so instead, we have to link all dependents to our public dependencies too

util/CMakeLists.txt
bramtayl: Ok, removed!

cmake/Modules/FindAudioUnit.cmake
bramtayl: I think only dependees are supposed to provide config files, while dependents can use find files if not?

InOut/CMakeLists.txt
bramtayl: Wait, nevermind, they were enabled before because we weren't checking for `USE_COREMIDI` and `USE_AUDIOUNIT`. But we need to have the variables so that `check_deps` will work. So reenabled.

CMakeLists.txt
stekyne: That's a good point. It's not a pattern we can rely on though afaik. Are config imports supposed to adhere to that standard as well? I'm not sure

InOut/CMakeLists.txt
stekyne: Ok so. I wasn't insinuating that they should be disabled. I don't know why they were disabled before.

cmake/Modules/FindAudioUnit.cmake
stekyne: https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#use-exported-targets-of-external-packages The advice is to use the targets directly and not variables exported. I guess that makes sense as the findX.cmake files are completely variable (depends how they were written) in how you are supposed to use them.

cmake/cmake-utilities.cmake
stekyne: The comment is correct. The opcodes themselves only need the headers to build against. They get loaded at runtime and just use the Csound API. These utilities don't have the same restriction though.

I think the solution here is that we should have a libcsound interface target that only contains the headers i.e. it's a target that produces no binary. We can use that to link against the opcodes. I'm wondering though, what libcsoudn public libs should be passed to dependent targets? That doesn't seem right to me.

CMakeLists.txt
```diff
@@ -1126,8 +1114,7 @@ if (LINUX)
     )
 endif()
 
-# TODO include target rather than libs to also bring in transitive dependencies
-set(libcsound_LIBS ${LIBSNDFILE_LIBRARY})
+set(libcsound_LIBS "")
```
stekyne, line 1117: Is this actually populated with anything?

CMakeLists.txt
bramtayl: Well, the vcpkg config module shouldn't change much. But we can test for the existence of SndFile::sndfile instead I guess.

cmake/Modules/FindAudioUnit.cmake
bramtayl: The one right below that says `Use a find module for third-party libraries that do not support clients to use CMake` though

CMakeLists.txt
stekyne: That's what I mean. We can't assume any variables will be declared by the exported target but you can test the target directly as it's a cmake construct. I'm not sure so can it export a partial/malformed target though or if it will error. I think it does throw an error.

cmake/Modules/FindAudioUnit.cmake
stekyne: Yeah, that's if they're not using Cmake though? If they're using Cmake, they should export targets which is the config method. Find modules are the desired solution if avoidable. I can't remember the talk I saw explaining this but it's true afaik.

bramtayl: Well, at least ubuntu doesn't ship cmake config files for any of the relevant packages:

https://packages.ubuntu.com/mantic/amd64/libsndfile1-dev/filelist
https://packages.ubuntu.com/mantic/amd64/portaudio19-dev/filelist
https://packages.ubuntu.com/mantic/amd64/libportmidi-dev/filelist
https://packages.ubuntu.com/mantic/amd64/libjack-jackd2-dev/filelist
https://packages.ubuntu.com/mantic/amd64/libsamplerate0-dev/filelist

CMakeLists.txt
bramtayl: Yes, several things are appended later

CMakeLists.txt
bramtayl: Ok, changed to check for the target instead!

cmake/cmake-utilities.cmake
bramtayl: Ok, well, I deleted the public libs for utilities because they're linked to libcsound

bramtayl: I think the right solution would be to build an "internal" `libcsound-internals` that has all headers available, and use that to link plugins and utilities that need internals. Then we can have a "public" `libcsound` that links to `libcsond-internals` but only has public headers available.

cmake/Modules/FindAudioUnit.cmake
stekyne: I had a thought. Why not return a target from findX files? Instead of just using variables.

bramtayl: I thought that was the point of this PR?

stekyne: Sorry, I said that wrong. I mean even with this PR, we're still testing variables. We should just test the target itself. That way we can apply the same logic to either VCPKG or legacy dependency management. I don't think any of the `x_FOUND` vars should be used after this PR.

stekyne: For instance, I'd replace this `check_deps(USE_LIBSNDFILE SndFile_FOUND)` with maybe another utility e.g. `check_deps(USE_LIBSNDFILE SndFile::SndFile <any other deps>`

bramtayl: Well, `check_deps` needs bools so that wouldn't work, but:

```
assign_bool(SndFile_FOUND TARGET SndFile::SndFile)
check_deps(USE_LIBSNDFILE SndFile_FOUND)
```

would. Can change it to that if you like, either way works

bramtayl: The cmake docs do say that `Pkg_FOUND` variables should be set in both find and config modules though: https://cmake.org/cmake/help/latest/command/find_package.html#basic-signature

stekyne: Interesting, wasn't aware of that. Is VCPKG following that though?

CMakeLists.txt
```diff
@@ -1288,8 +1275,8 @@ else()
     target_link_libraries(${CSOUNDLIB} PRIVATE ${libcsound_LIBS})
 endif()
 
-if(USE_VCPKG)
-    target_link_libraries(${CSOUNDLIB} PRIVATE SndFile::sndfile)
+if(USE_LIBSNDFILE)
+    target_link_libraries(${CSOUNDLIB} PUBLIC SndFile::sndfile)
```
stekyne, line 1279: Why is this moving from private to public? Are we exposing libsndfile headers via the API? Otherwise this should remain private. I have noticed weird issue with this library and linking libcsound though so it would be good to clear this up.

stekyne:
Going to merge this as I need some of the changes. Nice work. It's a big improvement

PR #1812, opened 2024-01-19
Less private include-dirs

bramtayl:
Most utilities and plugins don't need the private include-dirs, so we shouldn't include them by default. Just include them for the plugins/utilities that need them. This has a few advantages:

- Utilities and plugins must specifically ask for private header folders so can't include them by accident
- Utilities and plugins that don't use private header folders don't have to worry about clashes with them


cmake/cmake-utilities.cmake
```diff
@@ -9,8 +9,6 @@
 #
 function(make_executable name srcs libs)
     add_executable(${name} ${srcs})
-    target_include_directories(${name} PRIVATE ${libcsound_private_include_dirs})
-    target_include_directories(${name} PUBLIC ${libcsound_public_include_dirs})
```
bramtayl: moved the public headers line to make_utility. I don't think executables are intended to rely on csound headers. My bad from #1796

util/SDIF/CMakeLists.txt
```diff
@@ -2,5 +2,5 @@
 
 if(BUILD_UTILITIES)
 set(sdif2ad_SRCS sdif2adsyn.c sdif.c sdif-mem.c)
-make_executable(sdif2ad "${sdif2ad_SRCS}" "")
+make_utility(sdif2ad "${sdif2ad_SRCS}" "")
```
bramtayl: This should be a utility because it relies on csound I think

stekyne: Whats the difference between an executable and a utility though? To me it seems the same

util/CMakeLists.txt
```diff
@@ -14,26 +14,50 @@ endif()
...
+    target_include_directories(dnoise PRIVATE ${Csound_SOURCE_DIR}/util) # for std_util.h and utilmain.h
     make_utility(envext      env_main.c)
     make_utility(extractor   xtrc_main.c)
+    target_include_directories(extractor PRIVATE ${Csound_SOURCE_DIR}/util) # for std_util.h and utilmain.h
     make_utility(het_export  hetx_main.c)
+    target_include_directories(het_export PRIVATE ${Csound_SOURCE_DIR}/util) # for std_util.h and utilmain.h
     make_utility(het_import  heti_main.c)
+    target_include_directories(het_import PRIVATE ${Csound_SOURCE_DIR}/util) # for std_util.h and utilmain.h
     make_utility(hetro       het_main.c)
+    target_include_directories(hetro PRIVATE ${Csound_SOURCE_DIR}/util) # for std_util.h, utilmain.h, and SDF/sdif.h
```
stekyne: This seems a bit messy. I'd just add this into the `make_utility` function no? So what if it's not used by all targets, it will save all this extra code.

bramtayl: Sure, but they're not used by all utilities, like `csanalyze`. I guess we could make a new `make_std_util` function?

util/SDIF/CMakeLists.txt
bramtayl: Based on my reading, utilites rely on libcsound and executables don't. They aren't the most descriptive names though...

util/CMakeLists.txt
stekyne: Think about it. So what if there are a few extract headers included for a few utilities. We have to maintain this code so write as little of it as possible. Just include the standard csound header locations within `make_utility` and avoid adding all this extra noise in the file.

util/SDIF/CMakeLists.txt
stekyne: Ok I looked at the code. I didn't realise `make_utility` is calling `make_executable`. Really not sure if there is much value in these anymore with cmake targets. Can worry about removing them later.

util/CMakeLists.txt
bramtayl: Ok, I added separate `make_internal_utility` and `make_internal_plugin` functions that include internals too

Issue #1818, opened 2024-01-28
spout output issue with local ksmps and nchnls > 1


vlazzarini:
I have been studying the spout part of the code and I think I found an issue. Here's
the description:

1. In multichannel, output opcodes write to `csound->spraw` in a non-interleaved way.
2. With `lksmps != csound->ksmps`, make_interleave() interleaves `csound->spraw` according to `lksmps` blocks 
3. This assumption is not correct if there is a mix of instances with different local ksmps and global ksmps. The value of `lksmps` passed to make_interleaved() is of the last local ksmps instance performed.
4. Worse, since some `csound->spraw` will contain a mix of outputs with data in non-interleaved blocks of various ksmps sizes, interleaving correctly is not possible.

Leaving this here so we don't forget about it.

vlazzarini:
Fixed in PR 1819

PR #1819, opened 2024-01-28
Reorganising output and out opcodes revised 

vlazzarini:
This PR modifies the output spout so that each instrument can hold its own separate output.
It also revises all output opcodes to use a single function.

Current design in Csound 6
====

1. out opcodes etc -> mix into csound->spraw (non interleaved)
2. at every k-cycle make_interleave() interleaves spoutraw into csound->spout
3. locks in each out opcode to protect writing into csound->spraw
4. output is implemented by various opcodes.

Changes in this PR
====

1. all output opcodes share the same inner loop code.
2. locks removed
3. interleaving takes place at
a more reasonable place 

 
PROS: 
====
 - fixes issue #1818 (but needs testing)
 - this paves the ground for the implementation of oversampling in instruments.
 - out opcodes refactored should
facilitate maintenance. 

PR is now ready.


vlazzarini:
Revised this now, removed the per-instance spout because of unnecessary extra copying. This now is performing as well as the code in develop.

There's an issue with PARCS, where the mixing of various thread outputs is not right. I am trying to figure out what is going on. Probably I have the wrong assumptions.

vlazzarini:
PARCS restored and runs slightly faster than csound 6.18.

There's an oddity with PARCS in Csound 7 that it gives louder amplitudes at the end of some sections, but this varies from run to run. It's happening here and in the develop branch code. I suspect it is to do with the new parser (and how certain things were not put back into it).

As far as the output goes, it is all working correctly now.
Other things need to be tested, but I think this design is good.

jpffitch:
Sorry not to put this in a review but I cannot find how to do that.
Seems a step forward and improves efficiency.  I have not read all the code , but I trust you to get it basically right.

Extra thought; not all instruments call "out" in any of its forms, so tfor his class can skip this admin code and avoid memory allocation for instance local spout.  The semantic analysis should be able to flag to which class an instrument belongs.  Coud be done oither ways of course.

You should merge this PR

vlazzarini:
thanks, I'll do the tests and if they pass I'll report back and see if we can merge.

In the end the local spout was the wrong idea. It just introduced an extra copying layer that slowed the code down.

But I think the other changes are good.

vlazzarini:
@jpffitch @kunstmusik this is ready now for review. I have tested the code and it is working with all out opcodes.

vlazzarini:
Can we merge this?

kunstmusik:
@vlazzarini Apologies for delay, will review this evening!

vlazzarini:
thanks

vlazzarini:
I've updated it with some outstanding work I had thought of in the meantime.

Engine/musmon.c
```diff
@@ -270,7 +270,10 @@ int musmon(CSOUND *csound)
     csound->nspout = csound->ksmps * csound->nchnls;  /* alloc spin & spout */
     csound->nspin = csound->ksmps * csound->inchnls; /* JPff: in preparation */
     csound->spin  = (MYFLT *) csound->Calloc(csound, csound->nspin*sizeof(MYFLT));
-    csound->spraw = (MYFLT *) csound->Calloc(csound, csound->nspout*sizeof(MYFLT));
+    /* VL: 27.1.24 spraw redeployed as thread buffer */
+    printf("threads: %d \n", csound->oparms->numThreads);
```
kunstmusik: This looks like a debug message for testing? If so, suggest to remove.

Engine/insert.c
```diff
@@ -1881,14 +1881,14 @@ int subinstr(CSOUND *csound, SUBINST *p)
     return csoundPerfError(csound, &(p->h),
                            Str("subinstr: not initialised"));
   }
+  // This not needed since spout buffer is already managed by
```
kunstmusik: Looking at this comment, does it mean the code below should be removed?

Engine/insert.c
```diff
@@ -2668,7 +2668,8 @@ static void instance(CSOUND *csound, int insno)
     var->memBlock = (CS_VAR_MEM*)(temp - CS_VAR_TYPE_OFFSET);
     var->memBlock->value = csound->ekr;
   }
-
+  
+  //printf("alloc spout: %p \n", ip->spout);
```
kunstmusik: Remove dead code

Top/csound.c
```diff
@@ -1767,10 +1741,12 @@ int kperf_nodebug(CSOUND *csound)
     /* for one kcnt: */
     if (csound->oparms_.sfread)         /*   if audio_infile open  */
       csound->spinrecv(csound);         /*      fill the spin buf  */
-    csound->spoutactive = 0;            /*   make spout inactive   */
     /* clear spout */
     memset(csound->spout, 0, csound->nspout*sizeof(MYFLT));
-    memset(csound->spraw, 0, csound->nspout*sizeof(MYFLT));
+    memset(csound->spraw,0,
+           sizeof(MYFLT)*csound->nspout*csound->oparms->numThreads);
+    // 27.1.24 spraw is not used anymore
```
kunstmusik: Seems like this comment is a bit confusing about spraw not used anymore? Looks like it is still in use. What do you think about giving spraw a different name, or at least adding comments to the field in csoundCore.h? I suspect future devs would appreciate some documentation about the spout vs spraw.

include/csoundCore.h
```diff
@@ -582,6 +582,7 @@ typedef struct CORFIL {
                               (used by opcodes) */
     MYFLT   *spin;         /* offset into csound->spin */
     MYFLT   *spout;        /* offset into csound->spout, or local spout */
+    int      spout_flag;
```
kunstmusik: Is this necessary? I did a search through the code and found spout_flag was only ever set to 0 (unless I'm missing something).

kunstmusik:
@vlazzarini Things looked good as far as I could see though I left some comments about possible dead code and tidying matters.

vlazzarini:
Thanks I think some of the comments are not applicable anymore, as I made some later changes. spraw was removed and them later put back so that's one. I'll review today

Engine/musmon.c
vlazzarini: done.

Engine/insert.c
vlazzarini: spurious comment left over from earlier changes. removed it

Engine/insert.c
vlazzarini: done

Top/csound.c
vlazzarini: I'll think of a name change. I removed it earlier but then restored it.

include/csoundCore.h
vlazzarini: removed.

vlazzarini:
Ready now.

kunstmusik:
Looks good to me, merging now, thanks!

Issue #1820, opened 2024-02-02
Turning off an instrument that is releasing causes instrument count to become negative.


nonameentername:
I am working on creating an instrument in csound that keeps track of active instruments and turns them off to keep the maximum number of instruments under a certain number.
When turning off an instrument that has extra time the active instrument number count becomes negative.  The number keeps getting smaller every time an instrument gets turned off if it has extra time.

I'm using the following opcodes:
```csound
turnoff2 iInstr, 0, 0
kActive active iInstr, 0, 0
```
I found the following section of code decreases the instruments that are currently pending release:
https://github.com/csound/csound/blob/a1580f9cdf331c35dceb486f4231871ce0b00266/Engine/insert.c#L909-L910
The value of pending release does not get updated when using turnoff since extra time gets set to 0.
https://github.com/csound/csound/blob/a1580f9cdf331c35dceb486f4231871ce0b00266/Engine/insert.c#L1016-L1020

nonameentername:
Fixed in pr https://github.com/csound/csound/pull/1821

PR #1821, opened 2024-02-02
Update to fix negative active instrument count

nonameentername:
This pull request attempts to address the following issue: https://github.com/csound/csound/issues/1820


vlazzarini:
This looks fairly inoffensive. If it causes problems later, we can review it, but looks OK to me. @jpffitch do you agree?

jpffitch:
I  think thus PR should be accepted as is.  What I do not know is how to accept for v6.17 and cs7; how can I tell which it is  going to change?  

It ought to be simple to accept this fix for both.   **I just ask How**

vlazzarini:
it's just a case of choosing merge pr and choose merge commit

PR #1822, opened 2024-02-09
Interface libraries

bramtayl:
This uses interface libraries to encapsulate sets of include dirs for reuse by multiple targets. This has a few advantages:

- Collects include dirs into interface libraries for cleaner code
- Easier control of which targets get access to which include dirs
- Can set different compile flags for different sets of include dirs

I was able to attach a ton of flags to these new interface libraries.
- Used `target_compile_definitions` to attach compile definitions to specific targets
- Used `target_compile_features` to set the C++ standard for targets

This has an additional advantage: 

- Works for both C and C++, removing the need to duplicate code. We can use generator expressions to specifically target one or the other.

### DIFF
--- a/Top/csmodule.c
+++ b/Top/csmodule.c
@@ -1201,7 +1201,6 @@ extern int32_t date_localops_init(CSOUND *, void *);
 extern int32_t system_localops_init(CSOUND *, void *);
 extern int32_t liveconv_localops_init(CSOUND *, void *);
 extern int32_t gamma_localops_init(CSOUND *, void *);
-extern int32_t quadbezier_localops_init(CSOUND *, void *);
 extern int32_t framebuffer_localops_init(CSOUND *, void *);
 extern int32_t cell_localops_init(CSOUND *, void *);
 extern int32_t exciter_localops_init(CSOUND *, void *);
@@ -1294,10 +1293,11 @@ const INITFN staticmodules[] = { hrtfopcodes_localops_init, babo_localops_init,
 */
 typedef NGFENS* (*FGINITFN)(CSOUND *);
 
+NGFENS *quadbezier_fgens_init(CSOUND *, void *);
 NGFENS *ftest_fgens_init(CSOUND *);
 NGFENS *farey_fgens_init(CSOUND *);
 
-const FGINITFN fgentab[] = {  ftest_fgens_init, farey_fgens_init, NULL };
+const FGINITFN fgentab[] = {  ftest_fgens_init, farey_fgens_init, quadbezier_fgens_init, NULL };
 
 CS_NOINLINE int csoundInitStaticModules(CSOUND *csound)
 {


CMakeLists.txt
```diff
@@ -132,7 +134,7 @@ option(NEW_PARSER_DEBUG "Enable tracing of new parser" OFF)
 option(BUILD_MULTI_CORE "Enable building for multicore system" ON)
 option(FAIL_MISSING "Fail when a required external dependency is not present (useful for packagers)" OFF)
 option(USE_GETTEXT "Use the Gettext internationalization library" ON)
-option(BUILD_STATIC_LIBRARY "Also build a static version of the csound library" OFF)
+option(BUILD_STATIC_LIBRARY "Also build a static version of the csound library" ON)
```
stekyne: Any particular reason to default this to ON?

CMakeLists.txt
```diff
@@ -1022,8 +1025,8 @@ if(MSVC)
         COMPILE_FLAGS -DYY_NO_UNISTD_H)
 endif()
 
-list(APPEND libcsound_private_include_dirs ${CMAKE_CURRENT_BINARY_DIR})
-list(APPEND libcsound_public_include_dirs $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>)
+target_include_directories(libcsound_internals INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
```
stekyne: Looking at this, it should use the build_interface as well if it's accessing the build folder?

CMakeLists.txt
```diff
@@ -1075,6 +1078,8 @@ if(APPLE)
 endif()
 
 # Add the install target
+install(TARGETS libcsound_interface EXPORT CsoundExports)
```
stekyne: So this export is just the libcsound public headers right?

stekyne (commented):
Nice work. Let me take a closer look but it seems like the right approach to me

vlazzarini:
The interface library has been removed from Csound in 7.0. If used to be built as a separate file called libcsnd. We don't have that anymore.

This target may be removed.

CMakeLists.txt
bramtayl: Oops, that was just for testing, turned if back off

CMakeLists.txt
bramtayl: I don't think so, the `BUILD_INTERFACE` / `INSTALL_INTERFACE` bit is only necessary for public include dirs, not internals

CMakeLists.txt
bramtayl: Yes

gogins:
Was @bramtayl referring to INTERFACE scope in CMake commands, or to the
interfaces library directory?

Michael Gogins
Irreducible Productions

vlazzarini:
I read it wrong

CMakeLists.txt
stekyne: It doesn't make sense to include the bin directory for the configure though. It's including it when there is nothing there. If there is no build error then this must not be needed. Either way we should investigate it more.

bramtayl: We're putting a couple generated files into the bin folder: https://github.com/csound/csound/blob/a1580f9cdf331c35dceb486f4231871ce0b00266/include/CMakeLists.txt#L5 https://github.com/csound/csound/blob/a1580f9cdf331c35dceb486f4231871ce0b00266/CMakeLists.txt#L972. I'm not sure that's necessarily the right place to put them though

bramtayl:
> referring to INTERFACE scope in CMake commands, or to the interfaces library directory?

See https://cmake.org/cmake/help/latest/command/add_library.html#interface-libraries

CMakeLists.txt
stekyne: Ok after reading a bit more I'm not really sure now. Need to rethink which needs to be in `BUILD_INTERFACE` or not. I thought it was related to the build step but it's when exporting the build which makes sense. Lets leave it as it is and see if we encounter any problems.

bramtayl:
Ok, included the `libcsound_CFLAGS` in the interface libraries as well

CMakeLists.txt
```diff
@@ -41,8 +41,20 @@ project(Csound)
...
+add_library(libcsound_internals INTERFACE)
+
+# for users
+add_library(libcsound_interface INTERFACE)
+
+# for internal opcodes and plugins that use OpcodeBase.hpp
+add_library(opcode_internals INTERFACE)
+
+# for utils that use std_util.h and utilmain.h
+add_library(utility_internals INTERFACE)
```
stekyne: Less sure about this change. Why split them up so much?

bramtayl: Hmm, well, none of them are used exclusively together, so there's no obvious way to combine them. Which ones would you want to combine? The `libcsound_only` exists almost solely for the `__BUILDING_LIBCSOUND` flag, otherwise I'd combine it with `libcsound_internals`

stekyne: I can't see any sources being added to this `utility_internals ` though no?

CMakeLists.txt
```diff
@@ -41,8 +41,32 @@ project(Csound)
...
+    $<INSTALL_INTERFACE:include>
+)
+
+# for internal opcodes and plugins that use OpcodeBase.hpp
+add_library(opcode_internals INTERFACE)
+target_include_directories(opcode_internals INTERFACE Opcodes)
+
+# for utils that use std_util.h and utilmain.h
+add_library(utility_internals INTERFACE)
+target_include_directories(opcode_internals INTERFACE util)
```
stekyne: Is this meant to be `utility_internals ` instead of `opcode_internals`?

bramtayl: Oops yeah fixed thanks!

CMakeLists.txt
bramtayl: Ref https://github.com/csound/csound/pull/1822#discussion_r1493382104

CMakeLists.txt
```diff
@@ -1262,19 +1267,19 @@ set_target_properties(${CSOUNDLIB} PROPERTIES
 
 if(BUILD_STATIC_LIBRARY)
     add_library(${CSOUNDLIB_STATIC} STATIC ${libcsound_SRCS})
-    target_include_directories(${CSOUNDLIB_STATIC} PRIVATE ${libcsound_private_include_dirs})
-    target_include_directories(${CSOUNDLIB_STATIC} PUBLIC ${libcsound_public_include_dirs})
+    target_link_libraries(${CSOUNDLIB_STATIC} PRIVATE
+        $<BUILD_INTERFACE:libcsound_only>
+        $<BUILD_INTERFACE:libcsound_internals>
+        $<BUILD_INTERFACE:opcode_internals>
+    )
```
stekyne: Why is this `build_interface`? As far as I know, it's not common to have this with `target_link_libraries`.

CMakeLists.txt
stekyne: @bramtayl I still don't get the need to introduce 4 new targets. What benefit does that have over adding one?

CMakeLists.txt
bramtayl: Without this I get:

```
[cmake] CMake Error: install(EXPORT "CsoundExports" ...) includes target "csound64-static" which requires target "libcsound_only" that is not in any export set.
```

This makes it seems like adding `BUILD_INTERFACE` is necessary? https://discourse.cmake.org/t/cmake-error-install-export-projecttargets-includes-target-xxx-which-requires-target-yyy-that-is-not-in-any-export-set/7344/5

CMakeLists.txt
bramtayl: I see a few advantages:

- You can see at a glance which parts of the interface are required for which targets (flags, include dirs, etc.)
- Targets don't get bloated with parts of the interface they don't need

bramtayl: Eh, ok, I can see maybe this overcomplicates things. I combined stuff so now there's just two interfaces, `libcsound_internals` and `libcsound_interface`.

CMakeLists.txt
```diff
@@ -1182,77 +1179,65 @@ check_function_exists(strtod_l HAVE_STRTOD_L)
...
 
-if(HAVE__STRTOK_R)
-    add_definitions(-DHAVE__STRTOK_R)
-endif()
-if(HAVE__STRTOD_L)
-    add_definitions(-DHAVE__STRTOD_L)
-endif()
-if(HAVE__SPRINTF_L)
-    add_definitions(-DHAVE__SPRINTF_L)
-endif()
```
bramtayl: These were unused

CMakeLists.txt
```diff
@@ -1140,7 +1135,8 @@ endif()
 
 if(LINUX)
     message(STATUS "Building on Linux.")
-    add_definitions(-DLINUX -DPIPES -DNO_FLTK_THREADS -D_GNU_SOURCE -DHAVE_SOCKETS)
```
bramtayl: - `NO_FLTK_THREADS` was unused
- `HAVE_SOCKETS` is set below (ref https://github.com/csound/csound/pull/1822/files#r1501260727)

bramtayl:
Tests are failing and for the life of me can't figure out why :(

bramtayl:
Needed `_GNU_SOURCE` I guess

CMakeLists.txt
```diff
@@ -914,7 +910,6 @@ if(INIT_STATIC_MODULES)
         Opcodes/mixer.cpp
         Opcodes/signalflowgraph.cpp)
 
-    set_source_files_properties(${static_modules_SRCS} PROPERTIES COMPILE_FLAGS -DINIT_STATIC_MODULES)
```
bramtayl: Already set in libcsound_internals

CMakeLists.txt
```diff
@@ -1187,77 +1183,65 @@ check_function_exists(strtod_l HAVE_STRTOD_L)
...
-if(HAVE__STRTOD_L)
-    add_definitions(-DHAVE__STRTOD_L)
-endif()
-if(HAVE__SPRINTF_L)
-    add_definitions(-DHAVE__SPRINTF_L)
-endif()
-
 # Pass flags according to system capabilities
 
-if(HAVE_WINSOCK_H OR HAVE_SYS_SOCKETS_H)
```
bramtayl: Think `HAVE_SYS_SOCKETS_H` was a typo

CMakeLists.txt
stekyne: Alright, we need to test this to be sure. I don't fully get it but let's see

CMakeLists.txt
stekyne: Some of these are probably irrelevant after we moved out the plugins to the `csound/plugins` repo. It's good to clear that up.

CMakeLists.txt
```diff
@@ -46,8 +46,24 @@ project(Csound)
...
+    ${CMAKE_CURRENT_BINARY_DIR} # for generated private headers
+    Opcodes # for internal opcodes and plugins that use OpcodeBase.hpp
+    util # for utils that use std_util.h and utilmain.h
+)
+target_compile_definitions(libcsound_internals INTERFACE __BUILDING_LIBCSOUND)
+
+# for users
+add_library(libcsound_interface INTERFACE)
+target_include_directories(libcsound_interface INTERFACE
+    $<BUILD_INTERFACE:${Csound_SOURCE_DIR}/include>
```
stekyne: I don't see why this needs to be in the build_interface. It's the source dir so will exist at configure time no?

InOut/CMakeLists.txt
```diff
@@ -46,7 +46,7 @@ if(APPLE)
     check_deps(USE_COREMIDI CoreMIDI_FOUND CoreFoundation_FOUND)
     if(USE_COREMIDI)
         make_plugin(cmidi cmidi.c)
-        target_link_libraries(cmidi PRIVATE CoreMIDI::coremidi CoreFoundation::corefoundation)
+        target_link_libraries(cmidi PRIVATE CoreMIDI::coremidi CoreFoundation::corefoundation libcsound_internals)
```
stekyne: Why not just add that into the `make_plugin` part seeing as every plugin will need the headers? This target is just the headers right? No binary should be linked into the plugins.

stekyne (commented):
This looks a lot better. Nice work. I'm a bit worried that there is a lot of testing involved but it looks a lot cleaner for sure. Good job

InOut/CMakeLists.txt
bramtayl: Most plugins don't need the internal headers though

CMakeLists.txt
bramtayl: Hmm, I don't understand?

InOut/CMakeLists.txt
stekyne: Ok well they need the csound API. so there does need to be some header included. What is in `libcsound_internals`?

CMakeLists.txt
stekyne: Why did you add `$<BUILD_INTEFACE:` and not just leave it as `${Csound_SOURCE_DIR}/include`?

InOut/CMakeLists.txt
bramtayl: All the headers except what gets installed into the public include dir

CMakeLists.txt
bramtayl: For public include dirs, `target_include_directories` needs an absolute path for the build interface:

`$<BUILD_INTERFACE:${Csound_SOURCE_DIR}/include>`

and a relative path for the install interface:

`$<INSTALL_INTERFACE:include>`

More info here: https://cmake.org/cmake/help/latest/command/target_include_directories.html

InOut/CMakeLists.txt
stekyne: Then that's the inverse of what should happen no? The plugin should only use the header for the public facing API for Csound. It gets loaded dynamically at runtime. Why would the cmidi plugin need internal headers?

util/SDIF/CMakeLists.txt
```diff
@@ -3,4 +3,7 @@
 if(BUILD_UTILITIES)
 set(sdif2ad_SRCS sdif2adsyn.c sdif.c sdif-mem.c)
 make_utility(sdif2ad "${sdif2ad_SRCS}" "")
+if(${CMAKE_BUILD_TYPE} MATCHES "Release")
+    target_compile_definitions(sdif2ad PRIVATE NDEBUG) # needed for asserts
+endif()
```
stekyne: Why not just add this into `make_utility` and have it uniform across utilities?

InOut/CMakeLists.txt
bramtayl: cmidi uses midiops.h which isn't part of the public include dirs. I don't think that's the best design but

util/SDIF/CMakeLists.txt
bramtayl: `sdif2ad` is the only utility or plugin with asserts

tests/c/CMakeLists.txt
```diff
@@ -19,31 +19,14 @@ if(BUILD_TESTS)
...
-    if(USE_VCPKG)
-        set_property(TARGET unittests PROPERTY CXX_STANDARD 14)
-
-        target_link_libraries(unittests PRIVATE
-            GTest::gtest_main ${CSOUNDLIB_STATIC})
-    else()
-        set(CMAKE_CXX_FLAGS "-fpermissive -Wwrite-strings -std=c++0x")
-
-        if (APPLE)
-            set(CMAKE_CXX_FLAGS "-stdlib=libc++")
```
bramtayl, line 37: This was set earlier

tests/c/CMakeLists.txt
```diff
@@ -19,31 +19,14 @@ if(BUILD_TESTS)
...
     )
 
-    target_include_directories(unittests PRIVATE ${libcsound_private_include_dirs})
-    target_include_directories(unittests PUBLIC ${libcsound_public_include_dirs})
+    target_link_libraries(unittests PRIVATE libcsound_internals)
 
     # Required because we're compiling the tests written in C as C++
-    set(CMAKE_CXX_FLAGS "-fpermissive -Wwrite-strings -std=c++0x")
-
-    if(USE_VCPKG)
```
bramtayl: No differences between vcpkg and non-vcpkg branches left so can merge

CMakeLists.txt
```diff
@@ -281,38 +286,16 @@ if(WIN32)
...
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-endif()
-
-check_c_compiler_flag(-std=gnu99 HAS_GNU99)
-if (HAS_GNU99)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
-endif()
-if (HAS_CXX_VISIBILITY_HIDDEN)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-endif()
```
bramtayl, line 314: Moved to CompilerOptimizations.cmake

util1/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ make_utility(extract     sortex/xmain.c)
 if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_CLANG OR MSVC)
     make_utility(cs         csd_util/cs.c)
     if(HAVE_UNISTD_H)
-        target_compile_definitions(cs PRIVATE "-DHAVE_UNISTD_H")
```
bramtayl: The "-D" was a typo

Frontends/CMakeLists.txt
```diff
@@ -19,6 +19,20 @@ if(BUILD_CSOUND_COMMAND)
...
+
+  if(WIN32)
+    target_compile_definitions(csound-bin PRIVATE "WIN32")
+  endif()
+  if(LINUX)
+    target_compile_definitions(csound-bin PRIVATE "LINUX")
+  endif()
+  if(MACOSX)
+    target_compile_definitions(csound-bin PRIVATE "MACOSX")
+  endif()
```
stekyne: I feel like there is a more natural way of doing this surely

InOut/CMakeLists.txt
stekyne: Alright. Maybe there is a kind of internal plugin

Frontends/CMakeLists.txt
bramtayl: Yeah I think so too. Compilers should already set macros to identify the OS and we should probably use those instead. See https://github.com/cpredef/predef

bramtayl: But that can wait for another PR?

bramtayl:
Split up into four commits

cmake/CompilerOptimizations.cmake
```diff
@@ -2,52 +2,54 @@
 check_c_compiler_flag(-ftree-vectorize HAS_TREE_VECTORIZE)
 check_cxx_compiler_flag(-ftree-vectorize HAS_CXX_TREE_VECTORIZE)
 if (HAS_TREE_VECTORISE)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ftree-vectorize")
+    add_compile_options("$<$<COMPILE_LANGUAGE:C>:-ftree-vectorize>")
```
stekyne: Is it really required to include these language checks? It's the same compiler for C/C++ anyways?

bramtayl: On some platforms they might be a bit different I guess? For clang, there's `clang`  for c and `clang++` for c++.

bramtayl: Moved to #1828

Opcodes/CMakeLists.txt
```diff
@@ -22,20 +22,48 @@ option(BUILD_CPP_OPCODES "Build CPP plugins" ON)
...
-make_plugin(signalflowgraph signalflowgraph.cpp)
-make_plugin(ampmidid ampmidid.cpp)
+if(BIG_ENDIAN)
+    target_compile_definitions(bformdec2 INTERFACE "WORDS_BIGENDIAN")
+endif()
+
+make_plugin(mixer mixer.cpp libcsound_internals)
+if(INIT_STATIC_MODULES)
+    target_compile_defintions(mixer PRIVATE "INIT_STATIC_MODULES")
+endif()
```
stekyne: I don't think this works. I'd prefer that make_plugin would internally make sure of init_static_modules. Rather than manually adding this flag everywhere. But still, this is a big architecture change so I think maybe we should omit that from this PR if this is to get merged anytime soon.

bramtayl: Hmm, I'm not sure what you mean by it doesn't work...

> I'd prefer that make_plugin would internally make sure of init_static_modules

But not all plugins need `INIT_STATIC_MODULES`?

> this is a big architecture change so I think maybe we should omit that

Sure, we can just do the first commit if you want

stekyne: Ok looking at it more. I don't think it can be done another way right now. I'd maybe just group all of these target_compile_definitions in one block. Probably at the end of all these definitions. I thought there might be a more Cmake friendly idiom to create a plugin statically vs dynamically but it would create a lot of changes from what is there now.

bramtayl: Ok, done in #1838

PR #1825, opened 2024-02-19
Added homebrew MacOS build to Github Actions

stekyne:


stekyne:
@bramtayl @kunstmusik going to merge this as there is no risk. Added a homebrew based build like many are using on MacOS latest. There are older versions of MacOS as well if we felt that we need to support them explicitly.

PR #1826, opened 2024-02-20
mac-imports

bramtayl:
Reland the mac parts of #1811 reverted by https://github.com/csound/csound/commit/1e5accfa365b10e834677e3ed7bccdb0b9ded74f


bramtayl:
@vlazzarini This is working fine on macos both with and without vcpkg. I'm not sure what's different with your setup? Maybe try `set(CMAKE_FIND_FRAMEWORK "FIRST")`? https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_FRAMEWORK.html

vlazzarini:
Not  building here.  So it's not really working on MacOs without vcpkg

```
[ 72%] Linking C shared library ../libcmidi.dylib
cd /Users/victor/src/csound/debug/InOut && /Applications/CMake.app/Contents/bin/cmake -E cmake_link_script CMakeFiles/cmidi.dir/link.txt --verbose=1
/Library/Developer/CommandLineTools/usr/bin/cc -DHAVE_CLOCK_GETTIME  -ffast-math -fomit-frame-pointer -DPFFFT_SIMD_DISABLE -DBETA -Wall -Wno-missing-field-initializers -fvisibility=hidden -std=gnu99 -DHAVE_VECLIB -DHAVE_ATOMIC_BUILTIN -DHAVE_CURL -DHAVE_PTHREAD -g -arch arm64 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk -mmacosx-version-min=11 -dynamiclib -Wl,-headerpad_max_install_names -o ../libcmidi.dylib -install_name @rpath/libcmidi.dylib CMakeFiles/cmidi.dir/cmidi.c.o  /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreMIDI.framework /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreFoundation.framework 
ld: can't map file, errno=22 file '/Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreMIDI.framework' for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [libcmidi.dylib] Error 1
make[1]: *** [InOut/CMakeFiles/cmidi.dir/all] Error 2
make: *** [all] Error 2
```
as I explained in my email to develop list (I think you should join it so you can keep track of  what is going on), the command-line is wrong here

```
/Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreMIDI.framework /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreFoundation.framework 
```

It needs to be 

```
-framework CoreMIDI  -framework CoreFoundation
```

that is how frameworks are linked to in MacOs. The CMake script is not doing it correctly. This is what we should see:

```
[ 72%] Linking C shared library ../libcmidi.dylib
cd /Users/victor/src/csound/debug/InOut && /Applications/CMake.app/Contents/bin/cmake -E cmake_link_script CMakeFiles/cmidi.dir/link.txt --verbose=1
/Library/Developer/CommandLineTools/usr/bin/cc -DHAVE_CLOCK_GETTIME  -ffast-math -fomit-frame-pointer -DPFFFT_SIMD_DISABLE -DBETA -Wall -Wno-missing-field-initializers -fvisibility=hidden -std=gnu99 -DHAVE_VECLIB -DHAVE_ATOMIC_BUILTIN -DHAVE_CURL -DHAVE_PTHREAD -g -arch arm64 -arch x86_64 -isysroot /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk -mmacosx-version-min=11 -dynamiclib -Wl,-headerpad_max_install_names -o ../libcmidi.dylib -install_name @rpath/libcmidi.dylib CMakeFiles/cmidi.dir/cmidi.c.o  -framework CoreMIDI -framework CoreFoundation 
[ 72%] Built target cmidi
```

bramtayl:
Does `set(CMAKE_FIND_FRAMEWORK "FIRST")` help? https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_FRAMEWORK.html

stekyne:
"FIRST" is the default according to that documentation @bramtayl so I guess not

stekyne:
@vlazzarini what macOS version are you on? GH actions has 11-14

vlazzarini:
I am on 14. 
But this is not an OS issue, it's CMake not constructing the correct command.

vlazzarini:
The problem is here:

```
target_link_libraries(cmidi PRIVATE CoreMIDI::coremidi CoreFoundation::corefoundation)
```

This is passing the path to the framework directories 

```
/Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreMIDI.framework /Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk/System/Library/Frameworks/CoreFoundation.framework 
```

to the command, but we want the correct link flag instead.

Changing it to this fixes it

```
target_link_libraries(cmidi PRIVATE ${CoreMIDI_LIBRARY} ${CoreFoundation_LIBRARY})
```

We also need

```
target_link_libraries(rtauhal PRIVATE ${AudioUnit_LIBRARY} ${CoreFoundation_LIBRARY} ${CoreAudio_LIBRARY})
```

for the build to succeed.

bramtayl:
Sure, I'm just trying to figure out what the different is between your local build and the CI build, both of which are on MacOS and non-vcpkg. Since it's working fine on CI, there must be something different about your setup?

kunstmusik:
I tested with a fresh checkout from the PR branch, cleared my build dir, and was able to configure and build here. @vlazzarini maybe you had cached values in CMakeCache.txt? I saw MacOSX13.3sdk mentioned; my CMakeCache.txt mentions MacOSX14.2.sdk.

vlazzarini:
I removed the CMakeCache.txt and the CMakeFiles directory and did a fresh build.
Rebuilding, I get the same issue. CMake is not constructing the correct command-line. 

```
victor@firebird debug % cmake --version
cmake version 3.23.3
```

vlazzarini:
I looked at the CI output and the command is correctly constructed there.

```
/Applications/Xcode_14.2.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc -DHAVE_CLOCK_GETTIME  -ffast-math -mfpmath=sse -fomit-frame-pointer -fvisibility=hidden -std=gnu99 -DHAVE_VECLIB -DHAVE_ATOMIC_BUILTIN -DHAVE_CURL -DHAVE_PTHREAD -isysroot /Applications/Xcode_14.2.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.1.sdk -mmacosx-version-min=12.7 -dynamiclib -Wl,-headerpad_max_install_names -o ../libcmidi.dylib -install_name @rpath/libcmidi.dylib CMakeFiles/cmidi.dir/cmidi.c.o  -framework CoreMIDI -framework CoreFoundation 
```

 One difference is that I do not use Xcode, but I have the commandline tools installed instead. So my compiler SDK is installed at

```
/Library/Developer/CommandLineTools/SDKs/MacOSX13.3.sdk
```

CMake may be getting confused with it. In any case, we need a fix because it's not as if I have some weird build system installed here.

bramtayl:
Does setting `CMAKE_OSX_SYSROOT` help? https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_SYSROOT.html

vlazzarini:
I don't think so, because it is already set correctly. The -isysroot is correctly set and the location of the SDK is also found.  The frameworks are found correctly, as if we just change the target_link_libraries() it works. The issue is that the contents of the string CoreMIDI::coremidi etc are copied verbatim into the command, instead of interpreted as a
framework to be linked to.

vlazzarini:
Updated CMake to 3.29.0 and it is building. I should have thought of this earlier. 

Ok: we need a check for a minimum CMake version. I don't know what it is, but 3.23 wasn't it.

stekyne:
Nice find @vlazzarini. That is worrying as that's relatively recent. Maybe we can track down what the change in behaviour is.

bramtayl:
Maybe to do with this change? https://gitlab.kitware.com/cmake/cmake/-/issues/24946 not sure how to work around it

stekyne:
Well we either set the minimum Cmake level or else revert/alter the change

kunstmusik:
I think it's generally easy for mac devs to install a new version of cmake. If the code works for Victor now and works for everyone else, I think we can attribute it to a quirk in older cmake for mac and move on as it'll be an edge case moving forward. 

For minimum cmake we should support 3.13.4 as that is what Debian 10 has and it is still an LTS release. (See https://listserv.heanet.ie/cgi-bin/wa?A2=ind2401&L=CSOUND-DEV&P=775 for more info.)

vlazzarini:
Maybe we should set a minimum for the mac to 3.29 so others don't get tripped by this and we get blamed for a system that can't be built.

vlazzarini:
Here's my suggestion then

```
if (APPLE)
cmake_minimum_required(VERSION 3.29)
else()
cmake_minimum_required(VERSION 3.13.4)
endif()
```

bramtayl:
Ok done! I did 3.28 for mac though, I think that's when the change we need landed https://cmake.org/cmake/help/latest/release/3.28.html#platforms

vlazzarini:
merged

PR #1827, opened 2024-02-21
Use cmake functions instead of messing with flags

bramtayl:
Here I used`add_compile_options` instead of tinkering with the flags.

I see see this as having two advantages:

- Works for both C and C++, removing the need to duplicate code. We can use generator expressions to specifically target one or the other.

I've added some comments about other drive-by cleanups I made.

I didn't touch any of the `custom-platform.cmake` files unless they were tested on CI. We probably should add a CI build for these platforms and/or delete platforms we no longer support.


CMakeLists.txt
```diff
@@ -170,29 +167,23 @@ endif()
...
 elseif(${CMAKE_BUILD_TYPE} MATCHES "Debug")
     message(STATUS "Building for debug")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DBETA")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DBETA")
+    add_compile_definitions("BETA")
 endif()
 
 # set -Werror if in Debug configuration
 if(NOT MSVC AND NOT WASM)
-#ifdef BARE_METAL
```
bramtayl, line 183: Should just be `if` for cmake

CMakeLists.txt
```diff
@@ -273,37 +263,30 @@ if(WIN32)
...
     else()
-        set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -mstackrealign -static-libgcc -static")
-        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mstackrealign -static-libstdc++ -static-libgcc -static")
-        set_target_properties(${CSOUNDLIB} PROPERTIES LINK_FLAGS "-static-libstdc++ -static-libgcc -static")
-        set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -static-libstdc++ -static-libgcc -static")
+        add_compile_options("-mstackrealign")
+        add_link_options("-static"
+            "$<$<COMPILE_LANGUAGE:C>:-static-libgcc>"
+            "$<$<COMPILE_LANGUAGE:CXX>:-static-libstdc++>"
+        )
```
bramtayl, line 274: I'm pretty sure `-mstackrealign` is a compile option and the others are link options but not 100% sure

CMakeLists.txt
```diff
@@ -533,28 +515,28 @@ foreach(header ${HEADERS_TO_CHECK})
     string(TOUPPER ${header} tmp)
 
     string(REGEX REPLACE [./] "_" upper_header ${tmp})
-    check_include_file(${header} HAVE_${upper_header})
+    set(header_flag "HAVE_${upper_header}")
+    check_include_file(${header} ${header_flag})
+    if (header_flag)
+        add_compile_definitions(${header_flag})
+    endif()
```
bramtayl, line 522: We can add the compile definitions here instead of doing it individually (so we don't need to set e.g. `HAVE_DIRENT_H`)

Top/csound.c
```diff
@@ -1234,7 +1234,7 @@ static void psignal_(int sig, char *str)
 
 static void signal_handler(int sig)
 {
-#if defined(HAVE_EXECINFO) && !defined(ANDROID) && !defined(NACL)
+#if defined(HAVE_EXECINFO_H) && !defined(ANDROID) && !defined(NACL)
```
bramtayl, line 1237: Added `_H` to conform with other header flags

iOS/custom.cmake.ios
```diff
@@ -2,11 +2,3 @@
 
 # GLOBAL
 set(CMAKE_VERBOSE_MAKEFILE ON)
-#### NOTE the processor type needs setting
-include(CheckCCompilerFlag)
-set(CMAKE_C_FLAGS "")
-set(CMAKE_CXX_FLAGS "")
```
bramtayl: Not sure why this was necessary?

platform/windows/Custom-vs.cmake
```diff
@@ -36,14 +36,3 @@ set(HAVE_BIG_ENDIAN 0)
...
-# Disable the following warnings in msvc
-# - C4244 loss of data in conversion
-# - C4267 loss of data in conversion
-# - C4005 macro refinitions
-# - C4996 unsafe functions
-# - C4047 levels of indirection difference (int and void*)
-# - C4090 different const qualifiers
-# - C4477 format string type differences
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} /wd4244 /wd4267 /wd4005 /wd4996 /wd4047 /wd4090 /wd4477")
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /wd4244 /wd4267 /wd4005 /wd4996 /wd4047 /wd4090 /wd4477 /wd4251")
```
bramtayl: Fine to leave in these warnings I think

tests/c/CMakeLists.txt
```diff
@@ -21,29 +21,9 @@ if(BUILD_TESTS)
 
     target_include_directories(unittests PRIVATE ${libcsound_private_include_dirs})
     target_include_directories(unittests PUBLIC ${libcsound_public_include_dirs})
-
-    # Required because we're compiling the tests written in C as C++
-    set(CMAKE_CXX_FLAGS "-fpermissive -Wwrite-strings -std=c++0x")
```
bramtayl, line 26: - `-fpermissive` isn't necessary anymore based on CI.
- Fine to leave in `write-strings` warnings
- gtest requires C++14

tests/c/CMakeLists.txt
```diff
@@ -21,29 +21,9 @@ if(BUILD_TESTS)
...
-    if(USE_VCPKG)
-        set_property(TARGET unittests PROPERTY CXX_STANDARD 14)
-
-        target_link_libraries(unittests PRIVATE
-            GTest::gtest_main ${CSOUNDLIB_STATIC})
-    else()
-        set(CMAKE_CXX_FLAGS "-fpermissive -Wwrite-strings -std=c++0x")
-
-        if (APPLE)
-            set(CMAKE_CXX_FLAGS "-stdlib=libc++")
```
bramtayl, line 37: This was set earlier

tests/c/CMakeLists.txt
```diff
@@ -21,29 +21,9 @@ if(BUILD_TESTS)
 
     target_include_directories(unittests PRIVATE ${libcsound_private_include_dirs})
     target_include_directories(unittests PUBLIC ${libcsound_public_include_dirs})
-
-    # Required because we're compiling the tests written in C as C++
-    set(CMAKE_CXX_FLAGS "-fpermissive -Wwrite-strings -std=c++0x")
-
-    if(USE_VCPKG)
```
bramtayl, line 28: No differences between vcpkg and non-vcpkg branches left so can merge

CMakeLists.txt
```diff
@@ -273,38 +263,21 @@ if(WIN32)
...
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-endif()
-
-check_c_compiler_flag(-std=gnu99 HAS_GNU99)
-if (HAS_GNU99)
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99")
-endif()
-if (HAS_CXX_VISIBILITY_HIDDEN)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-endif()
```
bramtayl, line 306: Moved to CompilerOptimizations.cmake

CMakeLists.txt
```diff
@@ -501,10 +473,8 @@ if (NOT MATH_LIBRARY)
     set(MATH_LIBRARY "")
 endif()
 
-cmake_push_check_state()
```
bramtayl: what was this for?

CMakeLists.txt
```diff
@@ -1154,81 +1138,25 @@ endif()
...
-endif()
-if(HAVE__SPRINTF_L)
-    add_definitions(-DHAVE__SPRINTF_L)
+    add_compile_definitions("WIN32")
 endif()
 
 # Pass flags according to system capabilities
 
-if(HAVE_WINSOCK_H OR HAVE_SYS_SOCKETS_H)
+if(HAVE_WINSOCK_H OR HAVE_SYS_SOCKET_H)
```
bramtayl, line 1154: Think `HAVE_SYS_SOCKETS_H` was a typo

stekyne:
It's the right thing to avoid directly changing the global flag variables but we should be using `target_` variants and not the global functions e.g. `add_compile_definitions`. https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#declare-compile-definitions-with-target_compile_definitions

Try to taget the libcsound target for these changes if that's where they are intended. This is good work though and I had wanted to do the same myself. I think you can easily just switch the function used and use the appropriate target.

CMakeLists.txt
```diff
@@ -273,38 +263,21 @@ if(WIN32)
...
+            "$<$<COMPILE_LANGUAGE:C>:-static-libgcc>"
+            "$<$<COMPILE_LANGUAGE:CXX>:-static-libstdc++>"
+        )
     endif()
 endif()
 
 # MacOS/IOS c++11 flags
 if(APPLE)
-    set(CMAKE_CXX_FLAGS  "${CMAKE_CXX_FLAGS} -stdlib=libc++")
-    set_target_properties(${CSOUNDLIB} PROPERTIES CXX_COMPILE_FLAGS  "-std=c++11")
```
stekyne, line 286: Make sure that `target_compile_features(${CSOUNDLIB} PRIVATE cxx_std_11)` is set somewhere. I'm not entirely sure adding the `-stdlib=libc` is required. This was in the early days of C++11 when the compiler support wasn't great.

CMakeLists.txt
stekyne: Just leave it for now if we don't know what it is for. It seems to unwind any variables modified in this section. It's used when `check_function_exists` is called. Let's just leave it. Try not to add changes that could possibly blow up in our face.

CMakeLists.txt
```diff
@@ -533,28 +503,23 @@ foreach(header ${HEADERS_TO_CHECK})
     string(TOUPPER ${header} tmp)
 
     string(REGEX REPLACE [./] "_" upper_header ${tmp})
-    check_include_file(${header} HAVE_${upper_header})
+    set(header_flag "HAVE_${upper_header}")
```
stekyne: In Cmake, the variables have upper case and function have lower case. Let's stick to one convention

CMakeLists.txt
```diff
@@ -533,28 +503,23 @@ foreach(header ${HEADERS_TO_CHECK})
...
 endif()
 
-if (HAVE_EXECINFO_H)
-    add_definitions("-DHAVE_EXECINFO")
-endif()
-
-check_deps(USE_LRINT HAVE_LRINT)
-if(USE_LRINT)
-    add_definitions("-DUSE_LRINT")
-endif()
```
stekyne: Are these moved somewhere else or just deleted?

CMakeLists.txt
```diff
@@ -590,23 +555,42 @@ list(APPEND libcsound_private_include_dirs ${Csound_SOURCE_DIR}/util)
...
-     list(APPEND libcsound_CFLAGS -DHAVE_PTHREAD_SPIN_LOCK)
-  endif()
+foreach(a_function ${FUNCTIONS_TO_CHECK})
+    string(TOUPPER ${a_function} upper_function)
+    set(function_flag "HAVE_${upper_function}")
+    check_function_exists(${a_function} ${function_flag})
+    if (${function_flag})
+        add_compile_definitions(${function_flag})
+    endif()
+endforeach()
```
stekyne: Again use upper case for the variables

CMakeLists.txt
```diff
@@ -1154,81 +1138,25 @@ endif()
...
-endif()
-if(HAVE_SYS_TYPES_H)
-    list(APPEND libcsound_CFLAGS -DHAVE_SYS_TYPES_H)
-endif()
-if(HAVE_TERMIOS_H)
-    list(APPEND libcsound_CFLAGS -DHAVE_TERMIOS_H)
-endif()
-if(HAVE_VALUES_H)
-    list(APPEND libcsound_CFLAGS -DHAVE_VALUES_H)
-endif()
```
stekyne: Where have these moved to?

bramtayl:
> we should be using `target_` variants

Sure. Let's do #1822 first then. That way I can add the definitions to the new interfaces

platform/windows/Custom-vs.cmake
stekyne: How so?

bramtayl: Well, the warnings are suggestions for fixes and we probably should heed them eventually?

CMakeLists.txt
bramtayl: Did these in batch with `HEADERS_TO_CHECK`. But will probably have to do them individually if we want to specify which target to add them to.

CMakeLists.txt
bramtayl: Did these in batch with `FUNCTIONS_TO_CHECK`. But will probably have to do them individually if we want to specify which target to add them to.

CMakeLists.txt
bramtayl: I added `set(CMAKE_CXX_STANDARD 11)`, but will change it to be target specific. I think setting the stdlib is required for apple? https://stackoverflow.com/questions/14228856/how-to-compile-c-with-c11-support-in-mac-terminal

CMakeLists.txt
bramtayl: Ok done!

CMakeLists.txt
bramtayl: Ok done!

platform/windows/Custom-vs.cmake
stekyne: Just leave the warnings muted for now. The PR that fixes the warnings can delete the suppression of it. No point making a bunch of noise for now

CMakeLists.txt
stekyne: That's from 2013. That shouldn't be needed anymore

CMakeLists.txt
stekyne: These should be added to libcsound target

CMakeLists.txt
bramtayl: Moved to #1822

CMakeLists.txt
bramtayl: Moved to #1822

CMakeLists.txt
bramtayl: Ok, left it

platform/windows/Custom-vs.cmake
bramtayl: Ok, left them

bramtayl:
Ok, moved most of the changes involving compile definitions to #1822 because I could attach those to the interfaces

CMakeLists.txt
bramtayl: Ok, deleted it

Opcodes/CMakeLists.txt
```diff
@@ -73,6 +73,9 @@ if(BUILD_DSSI_OPCODES)
     check_include_file(ladspa.h LADSPA_H)
     set(dssi_SRC dssi4cs/src/load.c dssi4cs/src/dssi4cs.c)
     make_plugin(dssi4cs "${dssi_SRC}" dl)
+    if(USE_LIB64)
+        target_compile_definitions(dssi4cs PRIVATE "LIB64")
```
bramtayl, line 77: Only needed for dssi4cs

CMakeLists.txt
```diff
@@ -212,7 +206,6 @@ endif()
 option(USE_LIB64 "Set to on to set installation directory for libraries to lib64" OFF)
 if(USE_LIB64)
 	set(LIBRARY_INSTALL_DIR "${CMAKE_INSTALL_PREFIX}/lib64")
-    add_definitions("-DLIB64")
```
bramtayl, line 215: Only needed for dssi4cs

bramtayl:
Most of the things left are options that could apply to any target, so I've kept them global for now.

CMakeLists.txt
```diff
@@ -53,9 +53,7 @@ endif()
 message(STATUS "Csound directory: ${CMAKE_HOME_DIRECTORY}")
 
 # C++11 needed
-if(NOT MSVC)
-  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
-endif()
+set(CMAKE_CXX_STANDARD 11)
```
stekyne: Set this per target not globally

CMakeLists.txt
```diff
@@ -170,8 +168,7 @@ endif()
 # in Release configuration, set NDEBUG
 if(${CMAKE_BUILD_TYPE} MATCHES "Release")
     message(STATUS "Building for release")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNDEBUG")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -DNDEBUG")
+    add_compile_definitions("NDEBUG")
```
stekyne: Again set this for the libcsound target not globally

Opcodes/CMakeLists.txt
stekyne, line 77: is there really a lib called `lib64`?

iOS/custom.cmake.ios
stekyne: Removing this is just asking to introduce some regression. Are you sure there is no difference removing this?

stekyne:
Still setting some definitions globally. We need to avoid that as it makes it harder to consume this CMake project. Also, there is no point adding these flags to every target anyway. Most of them belong to libcsnd

Opcodes/CMakeLists.txt
bramtayl: Nope, this is a compile definition: https://github.com/csound/csound/blob/1a576670a6c4c8ba036f0758f3b53b5812531bdf/Opcodes/dssi4cs/src/dssi4cs.c#L1076

stekyne: Ah sorry misread that.

iOS/custom.cmake.ios
bramtayl: 90% sure, but I added it back in just in case

CMakeLists.txt
bramtayl: Ok, moved to #1822

bramtayl:
Wasn't much left so moved the rest into #1822

PR #1829, opened 2024-02-29
use os flags

bramtayl:
Use the compiler flags for OS instead of adding our own:

`_WIN32` for windows
`__APPLE__` for macosx
`__gnu_linux__` for linux

See https://github.com/cpredef/predef/blob/master/OperatingSystems.md

### DIFF
--- a/Engine/envvar.c
+++ b/Engine/envvar.c
@@ -32,7 +32,7 @@
 #include <fcntl.h>
 #endif
 
-#if defined(WIN32) && !defined(__CYGWIN__)
+#if defined(_WIN32) && !defined(__CYGWIN__)
 #  include <direct.h>
 #  define getcwd(x,y) _getcwd(x,y)
 #endif
@@ -84,7 +84,7 @@ typedef struct CSFILE_ {
 #if defined(MSVC)
 #define RD_OPTS  _O_RDONLY | _O_BINARY
 #define WR_OPTS  _O_TRUNC | _O_CREAT | _O_WRONLY | _O_BINARY,_S_IWRITE
-#elif defined(WIN32)
+#elif defined(_WIN32)
 #define RD_OPTS  O_RDONLY | O_BINARY
 #define WR_OPTS  O_TRUNC | O_CREAT | O_WRONLY | O_BINARY, 0644
 #elif defined DOSGCC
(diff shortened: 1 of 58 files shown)


include/sysdep.h
```diff
@@ -211,15 +211,15 @@ typedef uint_least16_t uint16;
 
 /* find out operating system if not specified on the command line */
 
-#if defined(_WIN32) || defined(__WIN32__)
-#  ifndef WIN32
-#    define WIN32 1
+#if defined(__WIN32__)
+#  ifndef _WIN32
+#    define _WIN32 1
```
stekyne, line 216: Hmm I'm not sure we should do this. Manually defining what is a well known symbol is asking for trouble IMO. 

Having said that, `WIN32` is defined by CMake. I think we should just be leveraging CMake completely for a lot of this system defintion stuff. https://cmake.org/cmake/help/latest/variable/WIN32.html

Engine/envvar.c
stekyne, line 1: It would be a lot less git noise if we simply defined `WIN32` when `_WIN32` is present

stekyne: Same is true for the other platforms imo

include/sysdep.h
bramtayl: Redefining `_WIN32` does seem like a bad idea, I could just delete this statement.

Using cmake or built-in definitions both work. Reducing the amount of manual definitions necessary to use our headers might be a good thing, but this would just be a drop in the bucket.

Engine/envvar.c
bramtayl: That could work too

include/sysdep.h
bramtayl: Thought about it a bit and I do think it makes more sense to use native flags instead of cmake ones. Not everyone uses cmake, and so we want to reduce the amount of compiler definitions necessary to use our headers as much as possible I think.

stekyne: How do you mean? What is the alternative to CMake here? That is the build system that Csound uses. So there is no other alternative. It's a mistake to over generalise on things that are definite. We use Cmake, we should leverage it as much as possible so long as that is the case.

bramtayl: No, what I mean is: suppose someone wants to use write a C program and link to libcsound. But suppose that someone isn't using cmake. Then they can't use our exported target. So, to use our headers, they'll have to define all the config variables by hand first.

stekyne: Ah right, sysdep is client facing. In that case we have to account for each supported compiler/platform as best as possible. I guess it is doing that already. Best we can do is to update it for latest GCC/Clang/MSVC versions.

PR #1832, opened 2024-03-04
more plugins

bramtayl:
All of these opcodes files only need public headers, so are good candidates to be plugins I think. The benefits are

- Code organization
- Easier to move out of the main repo

### DIFF
--- a/Top/csmodule.c
+++ b/Top/csmodule.c
@@ -1201,7 +1201,6 @@ extern int32_t date_localops_init(CSOUND *, void *);
 extern int32_t system_localops_init(CSOUND *, void *);
 extern int32_t liveconv_localops_init(CSOUND *, void *);
 extern int32_t gamma_localops_init(CSOUND *, void *);
-extern int32_t quadbezier_localops_init(CSOUND *, void *);
 extern int32_t framebuffer_localops_init(CSOUND *, void *);
 extern int32_t cell_localops_init(CSOUND *, void *);
 extern int32_t exciter_localops_init(CSOUND *, void *);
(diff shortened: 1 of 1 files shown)


vlazzarini:
This PR changes the organisation of opcodes by making internal opcodes as plugins.
We can't accept this PR because there are platforms that cannot use plugins and this will break Csound in those platforms.

bramtayl:
Oh, that's too bad. Which platforms can't use plugins?

vlazzarini:
The only ones that can are the desktop OSs - linux, MacOS, Windows.

stekyne:
@bramtayl Afaik mobile (iOS, Android), web assembly

https://github.com/csound/csound/blob/7d3808e0f007332610e072bd1c6a2b3cf78a8253/CMakeLists.txt#L914 There is a flag to build some opcodes statically for these kinds of platforms. Maybe this PR could come up with a more general solution for that? I haven't really looked into it much.

bramtayl:
Ok, I changed it so that plugins are included statically if `INIT_STATIC_MODULES` is true. I also added a CI build to test that `INIT_STATIC_MODULES` is working.

> Maybe this PR could come up with a more general solution?

I added an `ADD_INIT_FUNCTION` macro to help. Still a bit clunky though

vlazzarini:
This will very likely break things in places. I spent a lot of time getting it to build and run on embedded baremetal and this will cause problems there

I do not see a good enough reason for this PR.

bramtayl:
Is there any way to add tests for embedded baremetal?

bramtayl:
Continued in #1837

PR #1837, opened 2024-03-06
More plugins

bramtayl:
Accidentally closed #1832, continued here:

All of these opcodes files only need public headers, so are good candidates to be plugins I think. The benefits are

- Reduce size of csound library on platforms with plugins
- Code organization
- Easier to move out of the main repo


rorywalsh:
I have concerns about incorporating additional plugin opcodes because they increase the complexity of packaging Csound-based host software, especially on Windows. However, it seems like they are here to stay, for better or worse. Regarding the current PR, it's crucial to ensure that we can statically link to all these plugins and that this linking is enabled by default. Victor has highlighted potential issues this might pose with bare-metal builds; what about Android? As far as I know, Android doesn't use CMake, so could this have implications there? We need to thoroughly investigate these aspects before merging. I think iOS should be Ok.

bramtayl:
Ok, let's add CI for bare-metal and android first then?

bramtayl:
What are the issues with packaging plugins on windows?

rorywalsh:
In the past, I've encountered problems where hosts would load Csound based plugins, but refuse to load any plugins, or simply not find them for some strange reason. Mot of it could be resolved, but as far as I'm concerned the less dlls the better 

Additionally, if a plugin statically links to Csound, it may still attempt to load the system opcodes found in CSOUND6DIR64. This poses a risk that future updates to the API and newer versions of the plugin opcodes could cause issues with plugins still using an older build of Csound. On Mac I can relink to the plugin opcodes so this is not an issue.

bramtayl:
> problems where hosts would load Csound based plugins, but refuse to load any plugins, or simply not find them

We could try to fix that if you have any debugging info?

> if a plugin statically links to Csound, it may still attempt to load the system opcodes

We could just disable plugin loading if `INIT_STATIC_MODULES` is true. I think it would be unexpected for a static library to load plugins anyways?

rorywalsh:
That was an edge case that I could only recreate with Windows 10 Home and Live. I found a workaround which did the trick in the end. But it's still a pain to have to distribute endless dlls with a plugin. 

> I think it would be unexpected for a static library to load plugins anyways?

I believe that is the case here, although I could be wrong 🤔

bramtayl:
Well, since we have exported targets now, it's not too hard to do something like:

```
install(IMPORTED_RUNTIME_ARTIFACTS CSound::rtpa
    RUNTIME_DEPENDENCY_SET MY_DEPENDENCIES
)
install(RUNTIME_DEPENDENCY_SET MY_DEPENDENCIES
```

See https://github.com/bramtayl/Justly/blob/master/CMakeLists.txt for example

> that is the case here

Not quite, if you set `INIT_STATIC_MODULES`, then code from the plugins will be included directly in the csound library, and there won't be external plugins.

rorywalsh:
`Not quite, if you set INIT_STATIC_MODULES, then code from the plugins will be included directly in the csound library, and there won't be external plugins.`

Sorry, I meant that's the case if you don't statically link to these opcodes. Yes, that's the most important thing here, that we can set INIT_STATIC_MODULES 👍

rorywalsh:
Btw, I don't think the wasm build of Csound uses Cmake either (@hlolli, @kunstmusik?). And I'm not sure it can support plugins, so we have to be careful these changes don't break things there.

rorywalsh:
@bramtayl, have you any experience with CMake for Android projects? I just opened [this](https://github.com/csound/csound/issues/1843) issue.

bramtayl:
Nope, but I can take a look!

rorywalsh:
Great :)

PR #1838, opened 2024-03-06
Interface libs

bramtayl:
Accidentally closed #1822 so here it is again (sorry)

This uses interface libraries to encapsulate sets of include dirs for reuse by multiple targets. This has a few advantages:

- Collects include dirs into interface libraries for cleaner code
- Easier control of which targets get access to which include dirs
-  Can set different compile flags for different sets of include dirs


bramtayl:
Commits:

- interface libs: add the basic interface libs
- target_compile_features: use only target_compile_features to add compile features
- target_compile_definitions: use only target_compile_definitions to add compile definitions
- fix src_conv bug: fix a bug in src_conv involving internationalization
- less __BUILDING_LIBCSOUND: reduce usage of __BUILDING_LIBCSOUND, relying on clearly defined interface instead
- cwindow_internal: move the internal functions from cwindow.h to their own private header file

bramtayl:
Got a little ahead of myself, just the first commit. Will save other commits for follow-up PRs.

stekyne (commented):
This looks great. If the testing is all good I'm happy with it

PR #1839, opened 2024-03-09
Add utility tests

bramtayl:
I added tests for as many of the utilities as I could. There are still some missing, if anyone has an idea for how to test those I'd appreciate it


bramtayl:
These tests take 2 seconds to run

bramtayl:
Part of #1846

PR #1840, opened 2024-03-10
graphic display

fggp:
The third parameter of  the callback function set by csoundSetMakeGraphCallback is a label set by the init function of the respective opcode in OOps/disprep.c:
"fft" for the opcode dispfft (line 301),
"display" for the opcode pvsdisp (line 62),
"display" for the opcode display (line 120).
We could modify this so that the label is the opcode name: "dispfft" instead of "fft" on line 301 and "pvsdisp" instead of "display" on line 120.


vlazzarini:
I think this can't be merged easily as it is coming from the master branch (6.18) into develop (7) and they have diverged a lot. Can you do another PR starting from develop?

PR #1842, opened 2024-03-10
Graphic display

fggp:
The third parameter of the callback function set by csoundSetMakeGraphCallback is a label set by the init function of the respective opcodes in OOps/disprep.c:
"fft" for the opcode dispfft (line 301),
"display" for the opcode pvsdisp (line 62),
"display" for the opcode display (line 120).
We could modify this so that the label is the opcode name: "dispfft" instead of "fft" on line 301 and "pvsdisp" instead of "display" on line 62.

### DIFF
--- a/Android/CsoundAndroid/jni/version.h
+++ b/Android/CsoundAndroid/jni/version.h
@@ -40,7 +40,7 @@
 #define CS_PACKAGE_VERSION  VERSION
 #define CS_VERSION          (6)
 #define CS_SUBVER           (18)
-#define CS_PATCHLEVEL       (0)
+#define CS_PATCHLEVEL       (1)
 
 
 #define CS_APIVERSION       4   /* should be increased anytime a new version
--- a/Android/CsoundAndroid/jni/version.h
+++ b/Android/CsoundAndroid/jni/version.h
@@ -40,7 +40,7 @@
 #define CS_PACKAGE_VERSION  VERSION
 #define CS_VERSION          (6)
 #define CS_SUBVER           (18)
-#define CS_PATCHLEVEL       (1)
+#define CS_PATCHLEVEL       (0)
 
 
 #define CS_APIVERSION       4   /* should be increased anytime a new version
--- a/Engine/fgens.c
+++ b/Engine/fgens.c
@@ -35,23 +35,29 @@
 #include <stdlib.h>
 /* #undef ISSTRCOD */
 
+static inline int32_t byte_order(void)
+{
+    const int32_t one = 1;
+    return (!*((char*) &one));
+}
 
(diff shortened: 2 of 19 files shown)


vlazzarini:
this again is from master into develop, it's not possible to merge as the two codebases have diverged. You need to make your PR from develop to develop.

fggp:
I got it. I was trying to do the PR from my access to the Csound repo and
not from a fork! I've now created a fork so that if I do other PRs it will
work from my fork develop branch to the Csound develop branch. For this PR
I did manually the change directly in the Csound develop branch and I
closed the PR.

I'm actually doing intensive tests on the Csound API from Python with
ctcsound and from Go with go-csnd (https://github.com/fggp/go-csnd). Here
are some thoughts about the API:

1) keep it simple
2) show C functions so that it's easy for FFI (Foreign Functions Interface)
of different programming languages to translate these functions
3) keep the object structure of the API with an opaque pointer to the
Csound structure as first argument of the functions so that it's easy to
implement the binding as a class in the foreign language if this language
supports OOP.
4) implement getter/setter functions for the main properties of Csound to
avoid exposing directly the internal structures of Csound. The advantage of
this is that we can modify the structures in the future without modifying
the API
5) simplify the channel part of the API
6) avoid exposing variadic functions
7) refactor the thread part of the API (CsoundPerformanceThread)? This has
to be discussed with the devs. In any case, the thread part of the API
should be apart from the common part of the API. For example the Go
programming language has its own concurrent mechanism (coroutines) that can
be very efficient.
8) divide the API in two or three parts: common functions (used by
everyone), specialized functions (used in particular cases), exotic
functions.
9) list the features we don't want to support anymore in the API because
they're not part of the core Csound lib. For example, do we have to support
graphics displays ? Wouldn't it be better to replace this feature by
dedicated streams of data?

What  do you think of this? Do you have any other ideas?

Regards

François


Le dim. 10 mars 2024 à 19:04, vlazzarini ***@***.***> a
écrit :

vlazzarini:
Regarding making PRs, the way I do it is to 
1) branch from develop
2) make the changes to the branch and push it to github
3) do a PR from that branch
That way I don't need to create a fork of the project. It's much simpler.

======
I think the list of things you are proposing is good. Also we have to bear in mind that there are 2 APIs,
the host one (csound.h) and the plugin/module (csdl.h, which is the exposed parts of CSOUND structure). 

One thing in my mind is to make sure that for the latter we are only the exposing what is
safe to expose. Currently, csound.performKsmps() is offered there, and that is an example of a dangerous thing. Someone who does not know what they're doing can lock up csound by using it in a plugin. But that is a separate consideration. What you propose for the host API sounds very good.

PR #1844, opened 2024-03-16
Local SR

vlazzarini:
This PR introduces a local sr to INSDS with a view to supporting oversampling in UDOs.
The macro CS_ESR has been redefined. A new macro defining CS_ONEDSR has been introduced.

- csound->onedsr still needs to be replaced in all opcodes using it. 
- A check for the use of csound->esr and csound->GetSr() in opcodes needs to be made.
- We will probably have to remove GetSr(), GetKr(), GetKsmps() to avoid further use in opcodes.

Not ready to be merged yet.

### DIFF
--- a/Engine/insert.c
+++ b/Engine/insert.c
@@ -37,6 +37,10 @@
 #include "csound_orc_semantics.h"
 #include <inttypes.h>
 
+static SR_CONVERTER *src_init(CSOUND *, int, float, int);
+static void src_deinit(CSOUND *, SR_CONVERTER *);
+static int src_convert(CSOUND *, SR_CONVERTER *, MYFLT *, MYFLT *);
+
 static  void    showallocs(CSOUND *);
 static  void    deact(CSOUND *, INSDS *);
 static  void    schedofftim(CSOUND *, INSDS *);
@@ -3067,8 +3071,26 @@ int csoundKillInstanceInternal(CSOUND *csound, MYFLT instr, char *instrName,
 }
 
 // sample rate conversion
+// src_init() - allocate and initialise converters
+// src_convert() - convert
+// src_deinit() - deallocate converters
+// Upwards: integer convertion ratios are used  
+// Input is read on first of N calls to src_convert(), N = ratio.
+// Output is split into N blocks, taking N calls to drain.
+// Downwards: an input of size N is taken and N/ratio
+// samples are output.
+struct _SR_CONVERTER
+{
+  float *bufferin, *bufferout;
+  float   ratio;
+  int     size;
+  int     cnt;
+  int     mode;
+  void   *data;
+};
+
 // Basic linear converter 
-SR_CONVERTER *src_linear_init(CSOUND *csound, int mode, float ratio, int size) {
+static SR_CONVERTER *src_linear_init(CSOUND *csound, int mode, float ratio, int size) {
   IGN(mode);
   SR_CONVERTER *pp = (SR_CONVERTER *) csound->Calloc(csound, sizeof(SR_CONVERTER));
   pp->data = csound->Calloc(csound, sizeof(MYFLT));
@@ -3079,7 +3101,7 @@ SR_CONVERTER *src_linear_init(CSOUND *csound, int mode, float ratio, int size) {
   return pp;
 }
 
-void src_linear_deinit(CSOUND *csound, SR_CONVERTER *pp) {
+static void src_linear_deinit(CSOUND *csound, SR_CONVERTER *pp) {
   csound->Free(csound, pp->bufferin);
   csound->Free(csound, pp->data);
   csound->Free(csound, pp);
@@ -3108,6 +3130,7 @@ void src_linear_process(SR_CONVERTER *pp, MYFLT *in, MYFLT *out, int outsamps){
   *((MYFLT *) pp->data) = in[incnt-1];
 }
 
+static
 int src_linear_convert(CSOUND *csound, SR_CONVERTER *pp, MYFLT *in, MYFLT *out){
   IGN(csound);
   int size = pp->size, cnt = pp->cnt;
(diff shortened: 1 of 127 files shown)


vlazzarini:
- I've replaced csound->GetSr(), GetKr(), GetKsmps() and GetKcounter() with versions that return the local values. This prevents any opcode to try to use the global values. Internal opcodes may still access esr etc directly so that is still a vulnerability.

- I've removed all direct access from opcodes to esr, ekr, onedsr, onedkr, and almost all direct access to ksmps (there's a few cases in bus.c where it makes sense to use it). I noted that there were quite a few cases where ekr, onedkr or ksmps were being used and that's a bug since ksmps can be local. Not sure if we should bother fixing these cases in Csound 6.19.

I think I have the infrastructure ready to be implement oversampling in UDOs. I thought of undersampling too, but I think I will leave that for the moment.

vlazzarini:
The code is complete now, with oversampling implemented using SRC. It is also possible to implement it using other sampling rate conversion code if we have it. 

The outline is as follows:
1. Oversampling is controlled by an opcode "oversample" which sets a local (higher) local sr.
2. Oversampling is done by an integer factor.
3. This can only be used in UDOs without local ksmps or with no audio/control array arguments. These limitations may be removed in the future with more work.
4. If the local sr diverges from the global sr, then code in useropcd2() will convert all audio/control input arguments to a higher sampling rate. Control rate changes accordingly, ksmps is kept the same, but the opcodes will be run more than once on the upsampled audio/control signal (2 or more blocks). The output is then downsampled and copied to the udo output buffers.

It has not been tested at all yet.

vlazzarini:
I forgot there was a csound->sicvt that was dependent on the sampling rate. Added a local version. I also did  tests that confirmed it is working as expected. Added a "sr" local variable and confirmed that the oversample opcode throws an error if

- used in an instrument;
- local ksmps is used;
- a control/audio array argument is used;

This is now ready for review @kunstmusik @jpffitch 
Also started a version 7.0 of the manual in a new branch and pushed a manual page plus some changes to the setksmps and opcode pages to remove the hidden ksmps parameter which has been taken off the code.

bramtayl:
Don't really understand these changes, but when I was working on #1837, I noticed using the internal esr field was one of the most common reasons that a certain opcode file couldn't be turned into a plugin.

vlazzarini:
> Don't really understand these changes, but when I was working on #1837, I noticed using the internal esr field was one of the most common reasons that a certain opcode file couldn't be turned into a plugin.

yes, that is because csound->esr is not available to plugins. In any case, ALL opcodes now are forbidden to access this. CS_ESR or csound->GetSr(INSDS *p) are used instead. This has been implemented as part of this PR.

vlazzarini:
I implemented a linear src converter as a fallback in cases where Secret Rabbit Code is not available. It also replaces the SRC linear converter if that mode is selected.
Also noticed that argument checks in xin were being made against inputs, which led to crashes if a constant was used for k-rate. I changed this to check against the outputs, which seems to be a better idea (since they have to be variables).

vlazzarini:
@kunstmusik @jpffitch 
This is ready for your review. It would be good to merge this soon as the work on #1848 is likely to cause lots of merge conflicts as it progresses.

vlazzarini:
Just a note: in view of the work on redesigning the module API (PR 1852), I have removed the GetLocalXXX() functions from CSOUND. Whether we will need functions for global Csound attributes like sr etc, maybe is a question for that PR.

kunstmusik:
@vlazzarini I took a brief look through and wasn't sure this code works for structs (User-Defined Types). Can you confirm whether it does or not? I could imagine use cases where a user may want to implement a filter or oscillator using UDTs and UDOs and also employ oversampling.

vlazzarini:
I have not contemplated UDTs, so I actually don't know, probably not. 

I think supporting oversampling/undersampling for these will need to be considered in a later PR, beyond 7.0. At the moment, I don't know enough about them to be able to do it.

With local SR in place, we can expand the oversampling/undersampling support as required. Arrays can also be considered then.

Engine/insert.c
```diff
@@ -1634,28 +1630,63 @@ int useropcd(CSOUND *csound, UOPCODE *p)
...
+  bufs = udo->ar + inm->outchns;
   tmp = buf->iobufp_ptrs; // this is used to record the UDO's internal vars
   // for copying at perf-time
   current = inm->in_arg_pool->head;
   for (i = 0; i < inm->inchns; i++) {
     void* in = (void*)bufs[i];
     void* out = (void*)p->args[i];
     tmp[i + inm->outchns] = out;
-    current->varType->copyValue(csound, current->varType, out, in);
+    // DO NOT COPY K or A vars
```
kunstmusik, line 1657: I think doing type-specific code here is going to cause limitations for supporting all data types, and that support should be done before this is merged into develop. 

What about adding an oversampling or converter argument to copyValue? Seems like there needs to be the input argument var at source SR and local version of var at target SR.  xinset/xoutset/useropcd should target copying the value from source SR to target SR variables. Adding that to the CS_TYPE gets us the ability to handle per data type, and things like UDTs can walk their members to over/undersample.

vlazzarini: I think that is a good idea but can be done at a second stage, it is beyond the scope of this work. Note that all the type-specific code was there already, I just moved things around (I did not add anything particular). So it seems to me that you are right but it's a different PR.

I would like to merge this as soon as possible as it is holding up other work that depends on the local SR definitions. I am also not sure about the type system and would need to learn more in order to make such changes. It would actually be better if you did these changes as I am unsure of the design you have in your mind.

kunstmusik:
Filed https://github.com/csound/csound/issues/1869 to track further work for Local SR and UDTs, arrays, and type system.

include/csoundCore.h
```diff
@@ -559,13 +559,15 @@ typedef struct CORFIL {
     double   offbet;
     /* Time to turn off event, in seconds (negative on indef/tie) */
     double   offtim;
-    /* Python namespace for just this instance. */
-    void    *pylocal;
```
kunstmusik, line 563: I assume removing this will break the Python opcodes. Noting this here that those may need to be updated for CS7.

kunstmusik (approved):
Large PR; the code changes look good as far as scope for this PR. I did some searches in branch code to see if everything that should be replaced with local versions of sr, kr, ksmps, onedsr, etc. were done. I'm not 100% confident in double-checking all of these so we'll have to be on the lookout for missed changes. 

I added a comment about removal of pylocal which will need to be addressed in changes for Python opcodes. (I think it's good to remove this and find more generic solutions for those opcodes to use.)

Further work scheduled in PR #1869 to address UDTs, arrays, and type system.

vlazzarini:
yes, I think this should be solved with the global variable mechanism.

PR #1846, opened 2024-03-18
Revamp tests

bramtayl:
Combines #1839 and #1788 plus a few more changes:

- Add the files from the examples folder of the manual to soak
- Sort tests targeting specific opcodes into folders based on where the opcode was from


bramtayl:
@tjingboem seems to have updated/added a lot of the examples

bramtayl:
I know this is a big PR, but I think the number of test files that no longer seem to work, only work on some platforms, or yield indeterminate output should be enough to convince anyone of the dire need for more thorough testing. Since I'm not actually touching any of the code, just reenabling tests, I don't think it needs quite as much scrutiny as a regular PR.

Tests take 2 min to run.

I've tried to split up the commits to make it easier to review. The commits are:

- add examples: add the contents of the examples folder of the manual to soak
- moving and copying: organizing files into separate folders for each opcode file and utility
- remove old opcode tests: remove tests that use opcodes that no longer exist
- remove duplicate tests: remove tests that were duplicated elsewhere
- remove unused inputs: remove inputs that weren't used by any test
- test fixes: fix a few miscellaneous test issues
- new infrastructure: add the infrastructure for integration tests
- add opcode tests: add tests for opcodes
- add utility tests: add tests for utilities
- checksums: add checksums for tests

stekyne:
Looks like some good changes. Bit concerned by the number of new files and the checksums. I'll let the others share their opinion.

PR #1847, opened 2024-03-20
Build on android

bramtayl:
Add a CI android build. Notes:

- I used vcpkg for dependencies. Unfortunately, several dependencies (PortMidi, PortAudio, SndFile, LibLO) aren't getting built correctly by vcpkg, so I've disabled them for now.
- This just builds the libraries, not a fully fledged android app.
- I used an ANDROID_ABI of `x86_64`, is this what we want?
- There's a pre-existing top-level Android folder with a bunch of stuff in it, but I'm not really sure what it's all for tbh
- No tests for now. Can't do much without SndFile


bramtayl:
@rorywalsh

rorywalsh:
Thanks for this @bramtayl Can libSndfile just be built using Cmake (https://github.com/libsndfile/libsndfile/blob/master/Building-for-Android.md). That's the most important one. I'm not sure if the port* libs are really essential. The CsoundAndroid folder seems to be the one that looks after building the CsoundAndroid shared library. The CsoundForAndroid folder contains some AndroidStudio projects. I don't think these can be built using CMake as AndroidStudio uses Cmake itself internally. There is a CSD player project that Victor wrote which builds without AndroidStudio, but I'm not sure that method of building is still supported. I'm afraid I don't have that much experience with Android. I think support for x86_64 is most important, but it would be nice to support 32-bit systems too.

vlazzarini:
portmidi and portaudio are only relevant for deskop OS. Not even sure they are supported anywhere else.

giovannibedetti:
I think all the ANDROID_ABIs should be in, see https://developer.android.com/ndk/guides/abis
[armeabi-v7a](https://developer.android.com/ndk/guides/abis#v7a)
[arm64-v8a](https://developer.android.com/ndk/guides/abis#arm64-v8a)
[x86](https://developer.android.com/ndk/guides/abis#x86)
[x86_64](https://developer.android.com/ndk/guides/abis#86-64)

bramtayl:
Probably easiest just to fix the vcpkg android build of sndfile: https://github.com/microsoft/vcpkg/issues/37572

Actually, sndfile seems to work fine so long as we disable the mp3 part

bramtayl:
Not sure if we can test because we can't use the android libraries from ubuntu I don't think

bramtayl:
I think for testing we'll need java bindings? Android studio seems to require java bindings anyways.

bramtayl:
Ok, added four different Android ABIs.

bramtayl:
Dependency issues:

https://github.com/microsoft/vcpkg/issues/37572
https://github.com/microsoft/vcpkg/issues/37602
https://github.com/microsoft/vcpkg/issues/37603

PR #1850, opened 2024-03-23
Less __BUILDING_LIBCSOUND

bramtayl:
Will update with more in depth description


vlazzarini:
you can't remove that token. It is there for a reason. These types of changes need to be done with a deep understanding of the codebase. Before you do start them, you will need to bring the topic on the developer's list.
Adding "internal" versions of the headers is only complicating things for developers.

I am stopping this right now and I am not inviting a discussion here.

bramtayl:
Tests passed fine on Linux, just need to fix the other platforms? Can you reopen so I can finish? I marked it as a draft on purpose

bramtayl:
My reasoning here was as follows:

We already have private headers (in H) and public headers (in include). So for the most part, we can just move parts of the code in `__BUILDING_LIBCSOUND` to the private headers. The only exception is the csound struct, which there are two versions of, one for internals, and one for public use. I think it makes sense for developers to choose which one they need directly. I didn't duplicate any code, I had a macro that inserted the common fields into both of them.

bramtayl:
I was also really careful not to change `#include "csoundCore.h"` to `#include "csoundCore_internal.h`" unless the internals were strictly necessary.

bramtayl:
I also think it's a real footgun to use the inclusion of `CSOUND_CSDL_H` to demarcate the public from private sections. What order you include headers shouldn't change the behavior of code.

bramtayl:
I've honestly never seen a project where people are so hostile to new ideas. It's pretty clear that this project is dying, based on the proportion of untested code, the huge number of disabled tests in soak that fail, and the lack of people working on it. Which is a real shame because of all the people who have put so much work into this project over the years. It's honestly disrespectful to them to treat people who are only trying to help (like me) this way.

DaveSeidel:
Your ideas may be good, but you need to DISCUSS THEM IN ADVANCE before jumping in and doing the work. This is absolutely essential, especially in open-source projects like this where the participants are all remote. It's not enough to simply mark a PR as draft, especially when the changes you propose affect so many files, and particularly if you happen to be the newest person on the team with the least experience with the codebase. I don't know if you're a professional programmer; it's fine if you are not, but it's possible that you have never had to deal with this reality. There is an etiquette to working successfully on a distributed software project, and it doesn't exist just for the sake of politeness. Good communication is an absolute must.

Another point: just because something is in general a good idea or even a best practice in the industry does not mean that it's necessarily the right thing for the project. This is especially true for long-running projects. The older and more complex the system, the more likely it is that some compromises will need to be made. This may be hard to accept, but there it is.

bramtayl:
@vlazzarini literally said he doesn't want to disccuss:

> I am not inviting a discussion

But I'm done trying to help anyways

joachimheintz:
i agree with dave, and i understood victor mentioning ths developer list 
as the right place to discuss changes like the one you wanted to propose.

your activities appear in the moment when the developers concentrate on 
the release of csound 7 which is a huge and very important work, with a 
deadline in september, for the csound conference in vienna.

so apart from anything else, the conflict curently is simply: to spend 
time with your suggestions on re-organizing the build system, or to go 
ahead with the cs7 development.  from your point of view this is 
probably not a contradiction, but for the limited time of the devs to 
spend in the project, and what your pull requests require as occupation 
and discussion, it is.

DaveSeidel:
@bramtayl You are missing my main point. The right time to start a discussion is BEFORE you start doing the work. This allows the team to review your idea and the approach you propose to take. There's always the possibility that the answer will be "no". Jumping in without any discussion at all, and communicating very little about what you're doing is just passive aggressive, antithetical to working as part of a team, and setting yourself up for failure.

If you decide to participate in any other established projects, I urge you to spend some time getting a feeling for the dynamic of the participants and finding a way to contribute that fits in.

bramtayl:
> Good communication

In what world is good communication closing a PR, without even considering it, giving someone the chance to finish it, and refusing to discuss it? And why has no one even bothered to discuss the merits of the PR itself?

bramtayl:
Works fine on my fork now https://github.com/bramtayl/csound/pull/20

Issue #29, opened 2014-03-24
Many params in ftgen causes problem [sf bugs #29]


kunstmusik:
**Originally reported by Sourceforge User: mbechard**

System: Win XP Pro SP 2, Csound 5.04, 2ghz CPU, 2ghz RAM

Synopsis: usage of ftgen using GEN02 with many (256) parameters causes strange problem (ftgen used outside instrument declaration).

Compile the attached CSD; should give following error:

INIT ERROR in instr 1: Invalid ftable no. 0.000000
kamp    envlpx  1       .01     p3      #i0     gitbatk_1       .85     .001
0
B  0.000 - note deleted.  i1 had 1 init errors

Comment out "gitbstif_1    ftgen..." and it should compile fine.

PR #1852, opened 2024-03-23
Module API redesign

vlazzarini:
Started working on consolidating and rationalising the module API (aka public functions in CSOUND).
First step was to consolidate the real FFT interface.


vlazzarini:
Some more work on this:
- Removed unused functions 
- Consolidated some 
- Moved all OPDS-related functions out of the structure as inline: the rationale is twofold, first these do not depend on CSOUND but strictly operate on OPDS, secondly, they are minimal. We can add these without affecting the CSOUND structure.

bramtayl:
I like the renaming to make things easier to read/understand, maybe we should add deprecation messages though?

How much benefit is there from these public functions being part of the csound struct, instead of global functions? Global functions might be a bit easier to organize, and reduce the need for so much `csound->`. Is the idea that opcodes can swap out the functions for alternatives?

vlazzarini:
Anything that depends on CSOUND needs to be inside CSOUND otherwise external opcodes will not be able to use them. This is because external opcodes and other modules are loaded into a running instance and use its functionality without having to be linked to libcsound. Internal opcodes can access CSOUND entirely and all the public libcsound functions, but external modules/plugins cannot. The module API is for the benefit of external opcodes.  

All opcodes in ../Opcodes should be external and all opcodes in ../OOps should be internal. Not all external are built as plugins because of the constraints of platforms we support. Code in 6.x drifted to accessing certain internal parts of the library because of this, but I am working on fixing this. We don't open the CSOUND structure because access then cannot be controlled and we are open to vulnerabilities. This has to be tightly controlled to avoid externals bringing the system down.

No need for deprecation messages at this point because version 7.0 does not have API compatibility with  6.x.

bramtayl:
> use its functionality without having to be linked to libcsound

Why are we avoiding linking plugins to libcsound though? Seems like it would be easier than having to carry around all these function definitions in the struct?

vlazzarini:
We are not avoiding anything. 
This is how the system is designed, plugins are dynamically loaded modules, Csound loads them and resolves all symbols from inside. It is how a plugin system should work.

bramtayl:
Hm, well, on my branch, I've started to move these out to global functions, and tests, including the plugin tests I've added, pass fine. https://github.com/bramtayl/csound/pull/29. I do link the plugins to libcsound though.

kunstmusik:
Despite trying to explain that's not how plugin architectures are done
multiple times the past months, it seems you're not interested in
learning. Linking to libcsound will "work" for tests and running on
your own system, it won't work for an ecosystem of hosts, csound
libraries, and plugins (particularly sharing compiled plugins with
other users). It's also particularly problematic for linking issues on
Windows when distributing csound or csound-based apps and plugins.

bramtayl:
> sharing compiled plugins with other users

Hm, that doesn't seem like the best idea anyways?

> problematic for linking issues on Windows

Can we add a test for that?

kunstmusik:
We've already tested it through 20 years of this plugin system working and
working with users and developers to solve these problems years ago. We
won't be adding tests for this because it's not practical and would cost
too much money to run. But if you want to set up an automated test system
that installs different versions of Windows, with different user
configurations, compiles on one system and loads csound-based plugins in
various commercial DAWs on other systems, you or others are more than
welcome to do it.

As for sharing compiled plugins, it's not only a good idea, it's essential.
It'd be foolish to create a system and ecosystem that only works for
programmers/developers. Most of our users don't compile csound or plugins.
And in general, most musicians do not compile their own applications or
plugins.

bramtayl:
I'm just looking for a way to trigger the issue, that's all. For example, github actions has several different OS versions to choose from: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories, we could just download the same plugin and try it on different versions.

> Most of our users don't compile csound or plugins

Oh sure, of course users would want to download csound and the plugins all at once. I'm just skeptical of the wisdom of a user sharing a plugin with another user, especially because the two users could have totally different operating systems, c standard libraries, etc.

vlazzarini:
@bramtayl  why don't you do a fork of Csound using your own ideas? Provided you keep to the licensing, it's Free software and you can do whatever you want with it. 

We're happy for you to pursue your own design ideas, but for Csound 7.0. we are keeping to the library and API design we have put together since 2003.

kunstmusik:
As I said, you're welcome to implement something elsewhere, but we won't be
integrating that and running it here due to costs. I've given you example
scenarios where these kinds of problems manifest and they're real scenarios
because we've worked through them before here.

As for sharing plugins, of course you're not going to share Windows plugins
with Mac users or even ones compiled on Ubuntu with a SuSE user. That's
obvious. Compiling a plugin on Windows or Mac and distributing it to other
users on those systems, however, is a use case that is common and one we
support.

"Of course users would want to download ... all at once"

That's a very poor assumption and it would defeat one of the main reasons
to have a plugin system, which is to allow users to extend their programs
via 3rd parties.

BTW: You should know people have tried linking plugins to libcsound before.
You're not the first person to do this. It's always been a failure where it
"works for me" for one person and stops working when people start
distributing binaries. If that's the scope of what you want Csound to be,
then you're welcome to go that path, though I wouldn't advise it.

bramtayl:
> why don't you do a fork of Csound using your own ideas?

Ok, I'll work on my on fork for now. As of [#29](https://github.com/bramtayl/csound/pull/29) I've moved all public member functions from CSound to global functions, which means for the most part I can just use a forward declaration for the struct. The only remaining need for a second definition of the struct is for the C++ wrapper in plugin.h, but I'm working on dropping that too.

> Compiling a plugin on Windows or Mac and distributing it to other users on those systems

I still don't understand why linking libcsound in plugins would cause a problems in these use cases. Any breadcrumbs you have to specific previous issues would be helpful.

vlazzarini:
Just to let you know that we will not be accepting any such changes in the Csound code structure, but of course you are welcome to fork and develop your own version.

kunstmusik:
@vlazzarini Could you update this PR with the latest from develop? Will review once this is building through GH actions.

vlazzarini:
will do

vlazzarini:
It's all merged now.

include/csoundCore.h
```diff
@@ -660,6 +662,199 @@ typedef int32_t (*SUBR)(CSOUND *, void *);
...
+   * Returns true if argument is a string code
+   */
+  static inline int IsStringCode(MYFLT f){
+    return isstrcod(f);
+  }
+  
+  /**
+   * Returns the number of input arguments for opcode 'p'.
+   */
+  static inline int GetInputArgCnt(void *p){
```
kunstmusik: API-wise, seems clearer to me to make the argument OPDS* and make calling code case to OPDS* if necessary. (This applies to this function and others below.)

include/csoundCore.h
```diff
@@ -1072,24 +1265,48 @@ typedef struct _message_queue_t_ {
...
+    /** Get number of output channels */
     uint32_t (*GetNchnls)(CSOUND *);
     /** Get number of input channels */
     uint32_t (*GetNchnls_i)(CSOUND *);
     MYFLT (*Get0dBFS) (CSOUND *);
+
+    /** Get number of control blocks elapsed */
+    uint64_t (*GetKcounter)(CSOUND *);
+    MYFLT (*GetA4)(CSOUND *);
+    int (*GetTieFlag)(CSOUND *);
```
kunstmusik, line 1300: Shouldn't this get the tie-status from the INSDS*? Seems like getting it from CSOUND* would interfere with parallel processing. (Same comment for GetReinitFlag)

kunstmusik: Noticing these functions were like this and re-organized in this PR, so the above shouldn't hold this PR back, but noting the oddity here.

vlazzarini: I think the GetKcounter should be removed as it is INSDS-based, not sure about 
the tie flag function though. The GetA4() is ok.

include/csoundCore.h
vlazzarini: ok sounds good to me

Engine/fgens.c
```diff
@@ -1755,7 +1730,7 @@ static int gen30(FGDATA *ff, FUNC *ftp)
     for (i = 0; i < l2; i++)
       x[i] = xsr * f2[i];
     /* filter */
-    csound->RealFFT(csound, x, l2);
+    csoundRealFFT(csound, x, l2);
```
kunstmusik: Not a blocker, but I thought we had moved to using csound->APIFunc() in opcodes and fgens to have parity with external opcodes, and used csoundAPIFunc() for core code and host code. I may be recalling this incorrectly though. Is there a particular reason to change to use csoundRealFFT here instead of csound->RealFFT? (@vlazzarini)

Top/csound.c
```diff
@@ -334,466 +303,476 @@ static inline const CS_TYPE *StringType(CSOUND *csound) {
 }
 
 static inline const CS_TYPE *AsigType(CSOUND *csound) {
-    return csound->asigType;
+  return csound->asigType;
```
kunstmusik, line 301: Curious, why were these types added here as part of the API? Were these not accessible from csound_standard_types.h?

H/prototyp.h
```diff
@@ -186,59 +186,7 @@ int csoundSetReleaseLength(void *p, int n);
  */
 MYFLT csoundSetReleaseLengthSeconds(void *p, MYFLT n);
 
-/**
```
kunstmusik, line 189: These functions here look like they were previously a part of the API. Are there alternative ways now to for opcodes to access this information?

Top/csound.c
vlazzarini: When I added the option to build opcodes as plugins, I got undefined symbols for these. The best solution I could find was to make them available through the Csound struct. This was originally done in another PR.

H/prototyp.h
vlazzarini: They don't seem to be used anywhere in the code and they are not public either. I don't know if we should add them to the module API. I tend to try and keep it to what is being used, at this stage I think we should know more or less what opcodes require. Happy to add it if you think it needs to be there.

Engine/fgens.c
```diff
@@ -1769,7 +1744,7 @@ static int gen30(FGDATA *ff, FUNC *ftp)
       x[i] = FL(0.0);
     x[1] = x[l1];
     x[l1] = x[l1 + 1] = FL(0.0);
-    csound->InverseRealFFT(csound, x, l1);
+    csound->RealFFT(csound, csound->RealFFTSetup(csound,l1,FFT_FWD), x);
```
kunstmusik: Checking here: the older code called csound->InverseRealFFT(), but this is now calling csound->RealFFT(). Is this correct?

vlazzarini: yes, the interface is common to both directions, it gets set up differently. Just noted that it's FFT_INV, will fix.

util/cvanal.c
```diff
@@ -73,9 +73,9 @@ static int32_t cvanal(CSOUND *csound, int32_t argc, char **argv)
         case 's':
           FIND(Str("no sampling rate"))
 #ifdef USE_DOUBLE
-          csound->sscanf(s, "%lf", &sr);
+          sscanf(s, "%lf", &sr);
```
kunstmusik: I don't know the history here and would have to check, but curious we had a version of sscanf as part of the CSOUND struct API. Was there a reason for this and does moving to sscanf directly cause any potential issues on some platform?

vlazzarini: There were to do with localisation in CSOUND, LocalizeString, sprintf, sscanf, strtod, strtok_r. Only sscanf and strod were used in plugins. I suppose we need to offer them, but since they did not depend on CSOUND, I thought we could find a better location. I'll put them back for now.

Opcodes/dsputil.c
```diff
@@ -40,10 +40,11 @@
     /* where callers pass s/2+1, this recreates the parent fft frame size */
 
 /* assumes that FFTsize is an integer multiple of 4 */
-void Polar2Real_PVOC(CSOUND *csound, MYFLT *buf, int32_t FFTsize)
+void Polar2Real_PVOC(CSOUND *csound, MYFLT *buf, void *setup)
```
kunstmusik: Could we change this from `void* setup` to `CSOUND_FFT_SETUP *setup`?

kunstmusik (changes requested):

Opcodes/dsputil.c
vlazzarini: done

vlazzarini:
I also completed the work with overhauling the FT access in the module API, removing obsolete and redundant functions.

Opcodes/hrtferx.h
```diff
@@ -41,4 +41,5 @@ typedef struct {
   MYFLT         outl[BUF_LEN], outr[BUF_LEN];
   MYFLT         x[BUF_LEN], yl[BUF_LEN], yr[BUF_LEN];
   MYFLT         bl[FILT_LENm1], br[FILT_LENm1];
+  void *setup, *isetup;
```
kunstmusik, line 44: Note: we should get the formatting branch merged in sometime soon, there's a lot of inconsistencies in recent PRs

Opcodes/oscbnk.c
```diff
@@ -2017,7 +2017,7 @@ static void vco2_calculate_table(CSOUND *csound,
   /* inverse FFT */
   fftbuf[1] = fftbuf[table->size];
   fftbuf[table->size] = fftbuf[(int32_t) table->size + 1] = FL(0.0);
-  csound->InverseRealFFT(csound, fftbuf, (int32_t) table->size);
+  csound->RealFFT(csound,csound->RealFFTSetup(csound,table->size,FFT_INV),fftbuf);
```
kunstmusik: This call to RealFFTSetup looks awkward inline. Would prefer this be made into a separate variable that's initialized on a separate line.

Opcodes/oscbnk.c
```diff
@@ -2271,7 +2274,7 @@ static int32_t vco2init(CSOUND *csound, VCO2INIT *p)
       tp.w_fftbuf = (MYFLT*) csound->Malloc(csound, sizeof(MYFLT) * (i + 2));
       for (j = 0; j < ftp->flen; j++)
         tp.w_fftbuf[j] = ftp->ftable[j] / (MYFLT) (ftp->flen >> 1);
-      csound->RealFFT(csound, tp.w_fftbuf, (int32_t) ftp->flen);
+      csound->RealFFT(csound, csound->RealFFTSetup(csound, ftp->flen, FFT_FWD), tp.w_fftbuf);
```
kunstmusik: Same here, prefer RealFFTSetup on separate line with separate variable.

Opcodes/padsynth_gen.cpp
```diff
@@ -550,7 +550,7 @@ static int padsynth_gen(FGDATA *ff, FUNC *ftp) {
     spectrum[complexI].imag(real * std::sin(random_phase));
   };
   spectrum[0].imag(0);
-  csound->InverseRealFFT(csound, ftp->ftable, N);
+  csound->RealFFT(csound, csound->RealFFTSetup(csound,N,FFT_INV),ftp->ftable);
```
kunstmusik: Same here, prefer RealFFTSetup on separate line

Opcodes/phisem.c
```diff
@@ -193,1106 +193,1106 @@ static MYFLT noise_tick(CSOUND *csound)
...
+  p->coeffs00        = -WUTR_RESON * FL(2.0) *
+    COS(WUTR_CENTER_FREQ0 * CS_TPIDSR);
+  p->coeffs11        = WUTR_RESON * WUTR_RESON;
+  p->coeffs10        = -WUTR_RESON * FL(2.0) *
+    COS(WUTR_CENTER_FREQ1 * CS_TPIDSR);
+  p->coeffs21        = WUTR_RESON * WUTR_RESON;
+  p->coeffs20        = -WUTR_RESON * FL(2.0) *
+    COS(WUTR_CENTER_FREQ2 * CS_TPIDSR);
+  /* Note On */
+  p->shakeEnergy     = *p->amp * CS_DBFS_FLOAT * MAX_SHAKE * FL(0.1);
```
kunstmusik: Our other macros like CS_ONEDKSMPS and CS_ONEDKR use ONED for 1/x values. Seems like this macro should be renamed CS_ONEDDBFS_FLOAT. (Or CS_ONEDDBFS? Not sure what having _FLOAT differentiates)

Opcodes/physmod.c
```diff
@@ -35,6 +35,20 @@
 #include "brass.h"
 #include <math.h>
 #include "interlocks.h"
+
+static inline double intpow_(MYFLT x, uint32_t n)
```
kunstmusik: Did a search and we have intpow, intpow_, intpowint, and intpow1 in our codebase. Think we could consolidate all of these further into one function.

vlazzarini: reduced to two inline functions.

include/csoundCore.h
```diff
@@ -1098,74 +1341,52 @@ typedef struct _message_queue_t_ {
...
-    char *(*GetInputArgName)(void *p, int n);
-    int (*GetOutputArgCnt)(void *p);
-    char *(*GetOutputArgName)(void *p, int n);
     char *(*GetString)(CSOUND *, MYFLT);
-    int32 (*strarg2insno)(CSOUND *, void *p, int is_string);
-    char *(*strarg2name)(CSOUND *, char *, void *, const char *, int);
-
+    int32 (*StringArg2Insno)(CSOUND *, void *p, int is_string);
+    char *(*StringArg2Name)(CSOUND *, char *, void *, const char *, int);
+    const CS_TYPE *(*StringType)(CSOUND *csound);
```
kunstmusik, line 1354: Not for this PR, but I think if we're adding these here we could create and add these to a "StandardTypes" struct. Also in general, as CS7 progresses, we might benefit to partition CSOUND into more sub-types, brainstorming things like:

* csound.mem (memory)
* csound.sndfile (sound file I/O)
* csound.ftables
* etc.

Opcodes/sndloop.c
```diff
@@ -1242,15 +1241,9 @@ static int32_t pvsvoc_process(CSOUND *csound, pvsvoc *p)
           ceps[i] = fenv[i/2];
           ceps[i+1] = 0.0;
         }
-        if (!(N & (N - 1)))
-          csound->InverseComplexFFT(csound, ceps, N/2);
-        else
-          csound->InverseComplexFFTnp2(csound, ceps, tmp);
+        csound->InverseComplexFFT(csound, ceps, N/2);
         for (i=coefs; i < N-coefs; i++) ceps[i] = 0.0;
```
kunstmusik, line 1245: Noticed this for-loop, we should probably replace lines like this with memset

H/prototyp.h
kunstmusik: I guess if they weren't public then no worries; if someone comes back that they were using it we'll revisit.

kunstmusik (approved):
I think this PR is in condition to merge. There's some outstanding issues to investigate but I have filed some tickets in the tracker and we can proceed to handle in future PRs.

Issue #1689, opened 2023-01-29
event opcode with "e" statement ignores time as p1


joachimheintz:
this works in the score to finish the performance after ten seconds:
e 10

so i expected this to do the same:
instr 1
  event_i("e", 10)
endin
schedule(1,0,2)

but it **immediately** terminates.

this works instead:
instr 1
  event_i("e", 0, 10)
endin
schedule(1,0,2)

but it seems not consistent to me. (why having 0 as p1 if this is not the case for a score statement?)

can we leave the current behaviour for backwards compatibility, but add the `event_i("e",10)` possibility?

it would mean: if there is only one parameter following "e", read it as time until to terminate csound.

vlazzarini:
That's how it works and is documented. We can't change it now

https://csound.com/manual/e.html

joachimheintz:
i think it is how it is documented, but not how it works.

the example on this manual page has "e10", not "e 0 10"

when i have an empty score, and put "e 10" as statement, it will 
terminate after ten seconds.

when i put in what is written as "syntax" on the manual page:
     e 0 10
then it terminates immediately.

so i think the manual page is wrong.

vlazzarini:
it terminates immediately as it is. So that's wrong too.

vlazzarini:
This is now fixed in PR#1858.

PR #1858, opened 2024-04-01
dealt with issue #1689

vlazzarini:
The realtime event "e" now behaves as in the score,

```
event "e", 10
```
closes Csound at 10 seconds.

The previous case

```
event "e",  0, 10
```
continues to work


vlazzarini:
Fixed for API string events now 92a51fc4062fd7cf92bf12a08ea189ab3b5554b3

PR #1859, opened 2024-04-02
Deinit functions

vlazzarini:
New deinit system for opcodes
================

- The third function slot in OENTRY has been redeployed to support a deinit function. 
- This function has the same signature as init() and perf() functions.
- The old deinit callback mechanism has been removed
- Opcodes have been converted to using the new system
-  include/plugin.h and Opcodes/OpcodeBase.hpp have been modified to operate with the new system


gogins:
Good,
Mike

Michael Gogins
Irreducible Productions

vlazzarini:
As discussed in the developer's list, I have now removed the thread field from OENTRY.

The information it carried can be reproduced by

```
oentry.iopadr && oentry.kopadr  ? 3 : 
(oentry.iopadr ? 1 : 2)
```

vlazzarini:
Completed the PR with renaming: iopadr to init, opadr/kopadr to perf and dopadr to deinit.

kunstmusik:
@vlazzarini This looks to have conflicts now after merging in the oscils branch. Could you take a look at merging develop into this? I'll proceed to review after GH actions can build this branch again.

vlazzarini:
I will. I was anticipating this.

vlazzarini:
This is done now.

Engine/csound_orc_compile.c
```diff
@@ -1935,7 +1935,8 @@ if (engineState != &csound->engineState) {
       continue;
     if (PARSER_DEBUG)
       csound->DebugMsg(csound, "Instr 0 check on opcode=%s\n", bp->t.opcod);
-    if (UNLIKELY((thread = oentry->thread) & 06 ||
+    thread =  oentry->init && oentry->perf ? 3 : (oentry->init ? 1 : 2);
+    if (UNLIKELY((thread & 06) ||
```
kunstmusik: This logic seems odd given that thread can only be 1,2, or 3. Not sure how either & 06 or !thread would ever be true.

vlazzarini: I think this was not changed when we removed aopadr and thread type 4. The check can be replaced by a simple check for perf. I'll do that now

vlazzarini: I had a look and I can't see that in the current code there's a situation where

```
!thread && bp->t.pftype != 'b'
```

either. The only "thread 0" opcodes are instr, endin etc.

Engine/csound_orc_compile.c
```diff
@@ -1935,10 +1935,14 @@ if (engineState != &csound->engineState) {
...
     if (PARSER_DEBUG)
       csound->DebugMsg(csound, "Instr 0 check on opcode=%s\n", bp->t.opcod);
-    if (UNLIKELY((thread = oentry->thread) & 06 ||
-                 (!thread && bp->t.pftype != 'b'))) {
-      csound->DebugMsg(csound, "***opcode=%s thread=%d pftype=%c\n",
-                       bp->t.opcod, thread, bp->t.pftype);
+    /* VL: now the check is simply for oentry->perf, which is the
+       only condition possible for perf-time code 
+    */
+    if (UNLIKELY((oentry->perf != NULL) ||
```
kunstmusik: This check still seems off to me, if perf != NULL, it'll never make it to the second conditional check.

Engine/insert.c
```diff
@@ -918,13 +918,39 @@ static void schedofftim(CSOUND *csound, INSDS *ip)
 extern  int     csoundDeinitialiseOpcodes(CSOUND *csound, INSDS *ip);
 int     useropcd(CSOUND *, UOPCODE*);
 
+
+void deinit_pass(CSOUND *csound, INSDS *ip) {
+  int error = 0;
+  /*
```
kunstmusik: Shall this commented code be removed? Unsure if this means more work is necessary for protecting in multithreaded context or not.

OOps/diskin2.c
```diff
@@ -294,7 +294,7 @@ int32_t soundin(CSOUND *csound, DISKIN2 *p){
     return ret;
 }
 
-int32_t diskin2_async_deinit(CSOUND *csound, void *p);
+//int32_t diskin2_async_deinit(CSOUND *csound, void *p);
```
kunstmusik: remove commented code

OOps/diskin2.c
```diff
@@ -470,7 +471,7 @@ static int32_t diskin2_init_(CSOUND *csound, DISKIN2 *p, int32_t stringname)
         *start = 1;
       }
 #endif
-      csound->RegisterDeinitCallback(csound, p, diskin2_async_deinit);
+      //csound->RegisterDeinitCallback(csound, p, diskin2_async_deinit);
```
kunstmusik: remove commented code

OOps/diskin2.c
```diff
@@ -1062,7 +1066,7 @@ static int32_t sndo1set_(CSOUND *csound, void *pp, int32_t stringname)
     if (q->fd != NULL)                  /* if file already open, */
       return OK;                        /* return now            */
 
-    csound->RegisterDeinitCallback(csound, pp, soundout_deinit);
+    //csound->RegisterDeinitCallback(csound, pp, soundout_deinit);
```
kunstmusik: remove commented code

OOps/diskin2.c
```diff
@@ -1760,8 +1765,8 @@ static int32_t diskin2_init_array(CSOUND *csound, DISKIN2_ARRAY *p,
         csound->CreateThread(diskin_io_thread_array, *top);
       }
 #endif
-      csound->RegisterDeinitCallback(csound, (DISKIN2 *) p,
-                                     diskin2_async_deinit_array);
+      //csound->RegisterDeinitCallback(csound, (DISKIN2 *) p,
```
kunstmusik: remove commented code

OOps/goto_ops.c
```diff
@@ -41,8 +41,10 @@ int32_t kgoto(CSOUND *csound, GOTO *p)
   CS_PDS = p->lblblk->prvp;
   /* VL 16.2.23 fix for UDOs where the
      label gets confused */
-  if(CS_PDS->insdshead == NULL)
```
kunstmusik, line 44: Note to self: review history of this to see what this originally fixed and whether it is still a problem.

Opcodes/OSC.c
```diff
@@ -188,8 +188,8 @@ static int32_t osc_send_set(CSOUND *csound, OSCSEND *p)
     if (UNLIKELY(p->multicast)) lo_address_set_ttl(p->addr, 1);
     p->cnt = 0;
     p->last = 0;
-    csound->RegisterDeinitCallback(csound, p,
-                                   (int32_t (*)(CSOUND *, void *)) oscsend_deinit);
+    //    csound->RegisterDeinitCallback(csound, p,
```
kunstmusik: remove commented code

Opcodes/OSC.c
```diff
@@ -628,8 +628,8 @@ static int32_t osc_listener_init(CSOUND *csound, OSCINIT *p)
     pp->nPorts = n + 1;
     csound->Warning(csound, Str("OSC listener #%d started on port %s\n"), n, buff);
     *(p->ihandle) = (MYFLT) n;
-    csound->RegisterDeinitCallback(csound, p,
-                                   (int32_t (*)(CSOUND *, void *)) OSC_deinit);
+    //csound->RegisterDeinitCallback(csound, p,
```
kunstmusik, line 629: remove commented code

Opcodes/OSC.c
```diff
@@ -666,8 +666,8 @@ static int32_t osc_listener_initMulti(CSOUND *csound, OSCINITM *p)
                     Str("OSC multicast listener #%d started on port %s\n"),
                     n, buff);
     *(p->ihandle) = (MYFLT) n;
-    csound->RegisterDeinitCallback(csound, p,
-                                   (int32_t (*)(CSOUND *, void *)) OSC_deinit);
+    //    csound->RegisterDeinitCallback(csound, p,
```
kunstmusik: remove commented code

Opcodes/OSC.c
```diff
@@ -789,8 +789,8 @@ static int32_t OSC_list_init(CSOUND *csound, OSCLISTEN *p)
     p->c.method = lo_server_thread_add_method(p->port->thread,
                                               p->c.saved_path, p->c.saved_types,
                                               OSC_handler, p->port);
-    csound->RegisterDeinitCallback(csound, p,
-                                   (int32_t (*)(CSOUND *, void *)) OSC_listdeinit);
+    // csound->RegisterDeinitCallback(csound, p,
```
kunstmusik: remove commented code

Opcodes/OSC.c
```diff
@@ -1069,8 +1069,8 @@ static int32_t OSC_alist_init(CSOUND *csound, OSCLISTENA *p)
     p->c.method = lo_server_thread_add_method(p->port->thread,
                                               p->c.saved_path, p->c.saved_types,
                                               OSC_ahandler, p->port);
-    csound->RegisterDeinitCallback(csound, p,
-                                   (int32_t (*)(CSOUND *, void *)) OSC_listadeinit);
+    // csound->RegisterDeinitCallback(csound, p,
```
kunstmusik: remove commented code

Opcodes/afilters.c
```diff
@@ -32,732 +32,732 @@
...
     return OK;
+  }
+
+  if (UNLIKELY(offset)) memset(out, '\0', offset*sizeof(MYFLT));
+  if (UNLIKELY(early)) {
+    nsmps -= early;
+    memset(&out[nsmps], '\0', early*sizeof(MYFLT));
+  }
+
+  /* if (p->afc[0] != p->lkf)      { */
```
kunstmusik: There's some blocks of commented out code in a few places in this file that looks like they were from before CS6 and can be removed.

Opcodes/cpumeter.c
```diff
@@ -108,7 +108,7 @@ int32_t cpupercent_init(CSOUND *csound, CPUMETER* p)
     p->cnt = (p->trig = (int32_t)(*p->itrig * CS_ESR));
     // Would it be better to add a deinit process so the closing happens in perf?
     //csound->CreateFileHandle(csound, &p->fp, CSFILE_STD, "/proc/stat");
-    csound->RegisterDeinitCallback(csound, (void *) p, deinit_cpupercent);
+    // csound->RegisterDeinitCallback(csound, (void *) p, deinit_cpupercent);
```
kunstmusik: remove commented code

Opcodes/date.c
```diff
@@ -183,7 +183,7 @@ static int32_t readf_init_(CSOUND *csound, READF *p, int32_t isstring)
     }
     if (UNLIKELY(p->fd==NULL))
       return csound->InitError(csound, "%s", Str("readf: failed to open file"));
-    return csound->RegisterDeinitCallback(csound, p, readf_delete);
+    return OK; // csound->RegisterDeinitCallback(csound, p, readf_delete);
```
kunstmusik: remove commented code

Opcodes/dssi4cs/src/dssi4cs.c
```diff
@@ -923,10 +923,10 @@ int32_t dssictls_init(CSOUND * csound, DSSICTLS * p)
 #endif
 
 //    if ((int) csound->GetInputArgAMask(p) & 4) {
```
kunstmusik: remove commented code

Opcodes/mp3in.c
```diff
@@ -860,1622 +854,22 @@ static int32_t sprocess3(CSOUND *csound, DATASPACE *p)
 
 }
 
-#ifdef ANDROID
```
kunstmusik, line 863: I see a lot of code here removed for Android. Could you comment on this @vlazzarini so I can understand what's going on here?

Opcodes/pvsops.cpp
```diff
@@ -448,7 +448,6 @@ struct TPrint : csnd::Plugin<0, 1> {
   PrintThread t;
 
   int init() {
-    csound->plugin_deinit(this);
```
kunstmusik, line 451: Maybe this commented out opcode should just be deleted?

Opcodes/vbap_zak.c
```diff
@@ -27,580 +27,4 @@
...
-    p->spread_base.y = p->cart_dir.z;
-    p->spread_base.z = -p->cart_dir.x;
-    vbap_zak_moving_control(csound,p);
-    for (i=0;i<n;i++) {
-      p->beg_gains[i] = p->updated_gains[i];
-      p->end_gains[i] = p->updated_gains[i];
-    }
-    return OK;
-}
+// code moved to vbap.c
```
kunstmusik, line 30: vbap.c got quite big from these merges; suppose could have kept these separate and moved into a vbap folder, but not a big issue either way.

Top/argdecode.c
```diff
@@ -1557,8 +1557,11 @@ PUBLIC int argdecode(CSOUND *csound, int argc, const char **argv_)
               break;
             }
 #endif
-            if (!decode_long(csound, s, argc, argv))
-              csound->LongJmp(csound, 1);
+           if (!decode_long(csound, s, argc, argv))
+              // shouldn't need to long jump here -
+              // - it's crashing if csoundSetOption fails
+              //csound->LongJmp(csound, 1);
+             csound->Message(csound,"\n ...ignoring \n");
```
kunstmusik, line 1564: Think we need to check the behavior here. If a user supplies an invalid option, this function should return an error, yes? That way the commandline app can exit gracefully and API apps can handle that situation. (If I'm reading this code correctly...)

include/csoundCore.h
```diff
@@ -317,12 +317,11 @@ typedef struct CORFIL {
         char    *opname;
         uint16  dsblksiz;
         uint16  flags;
-        uint8_t thread;
         char    *outypes;
         char    *intypes;
-        int32_t     (*iopadr)(CSOUND *, void *p);
-        int32_t     (*kopadr)(CSOUND *, void *p);
-        int32_t     (*aopadr)(CSOUND *, void *p);
+        int     (*init)(CSOUND *, void *p);
```
kunstmusik: Any reason for the move from int32_t to int?

include/plugin.h
```diff
@@ -1128,13 +1130,13 @@ template <std::size_t N, std::size_t M> struct Plugin : OPDS {
   /** check if this opcode runs at init time
   */
   bool is_init() {
-    return this->iopadr ? true : false;
+    return this->init ? true : false;
```
kunstmusik: could be `return this->init != NULL`

include/plugin.h
```diff
@@ -1128,13 +1130,13 @@ template <std::size_t N, std::size_t M> struct Plugin : OPDS {
...
   }
 
   /** check if this opcode runs at perf time
   */
   bool is_perf() {
-      return this->opadr ? true : false;
+      return this->perf ? true : false;
```
kunstmusik: could be `return this->perf != NULL`

util/mkdb.c
```diff
@@ -123,7 +123,7 @@ typedef struct oentry {
         char    *intypes;
         int     (*iopadr)(void *, void *p);
         int     (*kopadr)(void *, void *p);
-        int     (*aopadr)(void *, void *p);
+        int     (*dopadr)(void *, void *p);
```
kunstmusik: I'm not familiar with this utility, but this struct still has thread which is removed now. Would rather see init, perf, deinit here too.

Engine/insert.c
vlazzarini: done

Engine/csound_orc_compile.c
vlazzarini: yep, it's much simpler.

OOps/diskin2.c
vlazzarini: done

OOps/diskin2.c
vlazzarini: done

OOps/goto_ops.c
vlazzarini: it seemed to be wrong fix as it failed in deinit.

Opcodes/OSC.c
vlazzarini: done

Opcodes/afilters.c
vlazzarini: removed

Opcodes/dssi4cs/src/dssi4cs.c
vlazzarini: done

Opcodes/mp3in.c
vlazzarini: they need to go. They've already gone in other branches.

Opcodes/vbap_zak.c
vlazzarini: I preferred to keep it all together. It was hard to keep track of these opcodes.

Top/argdecode.c
vlazzarini: this was a bug fix. It was crashing csound when the API function was called on non-existent option strings. We can look at this later. Sometimes exiting is not the right thing. It can still carry on, with a warning. Long jumps will cause trouble.

include/plugin.h
vlazzarini: it can

util/mkdb.c
vlazzarini: yep

Engine/csound_orc_compile.c
```diff
@@ -1935,10 +1935,13 @@ if (engineState != &csound->engineState) {
...
     if (PARSER_DEBUG)
       csound->DebugMsg(csound, "Instr 0 check on opcode=%s\n", bp->t.opcod);
-    if (UNLIKELY((thread = oentry->thread) & 06 ||
-                 (!thread && bp->t.pftype != 'b'))) {
-      csound->DebugMsg(csound, "***opcode=%s thread=%d pftype=%c\n",
-                       bp->t.opcod, thread, bp->t.pftype);
+    /* VL: now the check is simply for oentry->perf, which is the
+       only condition possible for perf-time code 
+    */
+    if (UNLIKELY(oentry->perf != NULL && bp->t.pftype != 'b')) {
```
kunstmusik: Rereading the original code, it looks like it checked if it was perf, or thread 0 and bp->t.pftype != 'b'. Guessing this should be:

```
if (UNLIKELY(oentry->perf  != NULL || (oentry->init == NULL && bp->t.pftype != 'b')) {
```

OOps/goto_ops.c
kunstmusik: Marking as not resolved as I haven't looked into it yet.

include/plugin.h
kunstmusik: (I accidentally resolved this one when it wasn't)

Top/argdecode.c
kunstmusik: I have no problem with getting rid of longjmps as I don't care for it, but this fix for setOption breaks the commandline IMO. I just tried running csound with an invalid long option and it not only kept running, it looks like it tried to parse the flag twice:

```
stevenyi@stevens-MacBook-Pro build % ./csound ~/work/csound/live-code/livecode.csd --asdsdfsdf 
unknown long option: '--asdsdfsdf'
 ...ignoring 
UnifiedCSD:  /Users/stevenyi/work/csound/live-code/livecode.csd
unknown long option: '--asdsdfsdf'
 ...ignoring 
scoreless operation
```

I can't think of a CLI app that doesn't fail on invalid args, so this is strange, plus the double "unknown long option" is concerning. 

Took a look at the code a bit further, there's on a few uses of argdecode in the codebase. Even though argdecode returns an int, it only ever returns 1. csoundSetOption then flips the result to a 0 with a strange ternary. Additionally, there are other calls to dieu(), which in turn calls longjmp. 

It looks like at one point argdecode *did* actually return a value as main.c checks the result of argdecode and will call longjmp there (which is fine as it's the CLI app), but that actually never happens since argdecode was calling longjmp first. 

I think we should use argdecode's return value in these areas of code and have the code here return 0 for a failure. We should also replace calls to dieu() with return 0.

OOps/goto_ops.c
kunstmusik: Looks like this code was added in this commit: https://github.com/csound/csound/commit/91b1ed0d4006ec15b79798f25c37ee027e1574e4 

Which mentions it was done to fix loops in UDOs. I found the mailing list thread that this was about:

https://listserv.heanet.ie/cgi-bin/wa?A2=ind2302&L=CSOUND&P=70026

I think this needs a little more investigation to address both the loops bug as well as deinit problem.

vlazzarini: can we add an issue to do this and move on?

Top/argdecode.c
vlazzarini: this issue is really unrelated to deinit(). I saw the bug and prevented the crash. Can we file a ticket and continue? I think it needs to be considered separately.

include/csoundCore.h
vlazzarini: no. fixed

Engine/csound_orc_compile.c
vlazzarini: going with your solution

Opcodes/pvsops.cpp
vlazzarini: done

OOps/goto_ops.c
vlazzarini: Ok I got to the bottom of this after barking at wrong trees:

- INDS and OPDS are often cast one to the other (BAD IDEA!)
- When I modified OPDS to include the deinit chain, a misalignment was caused
- in this case when we reassigned CS_PDS->insdshead, we were writing to an INSDS object and overwriting it caused the nxtoff to be rewritten 
- that made timexpire() go into an infinite loop.

Still figuring out the best solution. I fixed it by adding a padding member, but that's a horrible hack.

vlazzarini: There are two solutions:

1) Don't allow writing to CS_PDS->insdshead: that was always overwriting something, which previously was symptomless, but still wrong. One wonders how it could have ever worked, but the insdshead pointer seems to have been carried through correctly.

2) Add an insdshead to INSDS at the same byte offset to OPDS: it's a bit of a hack but should ensure that it can be used to carry insdshead.

Now the REAL solution is to disentangle INSDS and OPDS. This is a major piece of work that requires a dedicated PR. 

Incidentally, I think the original cause of the bug is due to INSDS being read as OPDS in the case of UDOs. I have not found the exact point, but I'd say we should find it in one of the useropcd() functions.

Engine/insert.c
```diff
@@ -982,6 +998,7 @@ static void deact(CSOUND *csound, INSDS *ip)
   if (ip->fdchp != NULL)
     fdchclose(csound, ip);
   csound->dag_changed++;
+  csound->Message(csound, "deact end\n");
```
kunstmusik: Looks like this should be removed

Top/argdecode.c
kunstmusik: I think that's fine. I added https://github.com/csound/csound/issues/1878 to track this with reference to this thread here. Shouldn't take too much to sort out I think but happy enough to get this PR done first and address this in that issue.

OOps/goto_ops.c
kunstmusik: I looked and didn't see casts between the two, mostly casting of void* to OPDS*. Casting between the two definitely seems wrong. Could you point me to somewhere that CS_PDS->insdshead is being written to? I'm not keen on #2 as I think it'd make this issue even more difficult to disentangle in the future. 

At this point, if the code is restored, does that still cause deact to go into a loop? Seems we also need to add a test CSD for both the loops in UDOs as well as deact. 

And how would you like to proceed? I won't mind if we file an issue and move forward, as I want to get to the PRs that restore the opcodes that were removed (e.g., "active", which my live code system uses and I need to better test things).

vlazzarini: There's the casting of INSDS to OPDS in various places, for example in insert.c:2728,  `instance()`, we have

```
prvids = prvpds = prvpdd = (OPDS*) ip;
```
In the case of UDOs, `p->lblblk->prvp`  is NULL, but it is an `INDS *` not a `OPDS *`. `CS_PDS` is `p->h.insdshead->pds`, which is a `OPDS *`. That assignment is OK if we just use it for jumping, but in the case of the fix I applied, it assumes we have an `OPDS *` when we have an `INSDS *` instead. 

It was a fluke that it worked, because it carried the correct data for jumping and it overwrite something that did not cause a problem. When I added the deinit members in the OPDS, that caused the overwriting to be in a place where it mattered (for deact)

So, we can remove the wrong fix and try a better one for the case of loops in UDOs, or we can embrace the OPDS/INSDS collusion by making sure it works in this case.

This is a case of the old code coming back to bite us, I think. Not sure yet how hard it is to disentangle it.

My opinion is that we should remove the fix (or keep it commented out as it is) and open an issue on the while loop bug for UDOs.

Engine/insert.c
vlazzarini: yes!

OOps/goto_ops.c
kunstmusik: I think at this point we need to get through a lot of PRs and better to take on the regression. I've filed an issue here:

https://github.com/csound/csound/issues/1879

and assigned it to the two of us since we've been looking at it. 

Going to approve and merge this PR and move on to the next one.

kunstmusik (approved):

kunstmusik:
Note: a couple of regressions resulted from this PR:

* https://github.com/csound/csound/issues/1879
* https://github.com/csound/csound/issues/1878

We will need to address these but it's worth it to merge and move forward.

Issue #1865, opened 2024-04-05
Build opcodes as plugins


vlazzarini:
At the last Csound conference a discussion was had as to how to make Csound builds of varying sizes by excluding opcodes. For Csound 7.0 we should attempt to do this by separating opcodes into two types

1. Internal: these opcodes may use functions from the Csound library directly as they are built into the system. The code for these resides in ./OOps 

2. External: these opcodes use the module API only. They can be provided as built-in for systems that do not have dynamic loading available or as plugin libraries if that is more convenient. This allows us to provide builds of various sizes.

vlazzarini:
Most of the work to enable this is already in place in PR#1844, PR#1848, PR#1852, and PR#1859. 

We need now to adjust the build system to allow for plugin libraries to be built as an option, plus providing a mechanism to implement this alternative.

vlazzarini:
Work completed in PR #1867

PR #1866, opened 2024-04-05
Bult-in opcodes build

vlazzarini:
As a first step towards #1865, this PR builds all opcodes that are not deprecated or do not have external dependencies into the Csound Library. This is the full system, when it is either not convenient or not possible to load plugins.

The second step (in a different PR) is to allow an option to build all external opcodes (in ./Opcodes) as plugins.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,8 @@ else()
     cmake_minimum_required(VERSION 3.13.4)
 endif()
 
+set (CMAKE_CXX_STANDARD 11)
+
 cmake_policy(SET CMP0077 NEW)
 
 include(cmake/cmake-utilities.cmake)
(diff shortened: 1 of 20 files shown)


CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ else()
     cmake_minimum_required(VERSION 3.13.4)
 endif()
 
+set (CMAKE_CXX_STANDARD 11)
```
stekyne, line 7: Generally it's advised to set this on the cmake target rather than globally. It should be set for libcsound at least. Maybe the opcodes just need it set in make_plugin or something

stekyne (approved):
Looks good

CMakeLists.txt
vlazzarini: yes, I suspected that, but it did not seem to be defined because the CI failed for iOS. I think it's OK to use C++11 as our baseline though.

vlazzarini:
thanks for the review

kunstmusik:
Isn't this going to cause problems with Debian with the scansyn opcodes linked in?

vlazzarini:
I don't think so. They will have the alternative to build it as plugins. PR1867 restores that possibility.

Issue #1864, opened 2024-04-05
Handling of extended guard point and sizes of power-of-two+1 tables


vlazzarini:
Two related issues have been identified

1) There is no mechanism to request an extended guard point to a GEN 
2) Tables using power-of-two+1 size have power-of-two length, which is confusing for users expecting the length to be equal to the size requested.

Proposed solutions

1) Use a non-zero fractional value for table size to request an extended guard point

```
f1 0 2000.1 7 0 2000 1   ; extended guard point
```

2) Use a negative size to request an actual size, disregarding the power-of-two+1 size mechanism.

```
f1 0 -1025 7 0 2000 1   ; flen = 1025 with standard guard point
```

vlazzarini:
I did a code review and the description of the problem was slightly wrong. Currently (e.g. 6.x)

1) Negative size tables *always* set an extended guard point.
2) Negative size tables keep their requested size (e.g. no change from pow2 + 1 to pow2).

So the problem is just how to give the user an option to suppress (1), and set a normal guard point with a copy of the value at position 0. This is done by using a table size with decimal part > 0 as proposed.
Note that fractional sizes are rounded to the nearest integer and this behaviour has not been touched.

 Implemented in PR #1872

joachimheintz:
thanks for this, victor.  (and of course all the big work on the CS7 
transition.)
you wrote two times "negative" in your two items, but i guess you mean 
one of it as "positive"?

vlazzarini:
No, I mean negative in both.

```
ifn ftgen  1,0,-1025,10,1
print ftlen(ifn)
```
gives currently

```
ftable 1:
instr 0:  #i2 = 1025.000
```

joachimheintz:
this is really a complicated area ...
let me ask if i am right now in my understanding:

a) (posiive table size)
print(ftlen(ftgen(0,0,8,2,0)))
print(ftlen(ftgen(0,0,9,2,0)))
both return 8 as table size, which is against what a user would expect 
in setting the table sizes.
the reason is that the 9nth value in size=9 is interpreted as extended 
guard point, but the table size itself is 8.

instead,
b) (negative table size)
print(ftlen(ftgen(0,0,-8,2,0)))
print(ftlen(ftgen(0,0,-9,2,0)))
return 8 and 9 as table size, which is what a user would expect.

but in the latter case, the guard point (which is always there but 
somehow invisible for the user) is always an extended one.  with your 
implementation, you give the option to have the guard point as copy of 
[...]

PR #1872, opened 2024-04-16
Feature/guardpoint

vlazzarini:
This PR completes the work for #1864


vlazzarini:
@kunstmusik this PR is still outstanding (from April). I was reviewing the f docs in the manual and realised it had not been merged yet.

kunstmusik:
Seems good to me. I do think something like this would be good to have a unit test for (either C or csd). Will go ahead and merge and we can note for the future.

kunstmusik (approved):

Issue #1880, opened 2024-05-13
<CsScore> #include "file.sco" returns stack pointer = 0


Nandos0804:
# Step to recreate the problem
-  Csd that invokes two file, an orchestra and a score
```
<CsoundSynthesizer>
    
    <CsOptions>
    -o test.wav -W -m 96 --m-amps=1
    </CsOptions>

    <CsInstruments>
        sr = 44100
        kr = 44100
        nchnls = 1
        0dbfs = 1

        giSine ftgen 0, 0, 2^10, 10, 1

        #include "prova.orc"

    </CsInstruments>
    
    <CsScore>
        #include "prova.sco"
        e
    </CsScore>

</CsoundSynthesizer>
```
- Orc file with code from poscil3 example
```
instr 1

krnd  randomh 40, 440, 1	; produce random values
ain   poscil3 .6, krnd, giSine
kline line    1, p3, 0    	; straight line
aL,aR pan2    ain, kline	; sent across image
      outs    aL, aR

endin
```
- Score from the same example
```
i1 0 10
e
```

- Alle the file are stored in the same dircetory (non admin path). From windows 10 terminal invoking 
```
csound prova.csd
```
returns 
```
UnifiedCSD:  .\prova.csd
rtaudio: PortAudio module enabled ...
using callback interface
rtmidi: PortMIDI module enabled
setting path from . to C:\Users\giuse\Desktop
sorting score ...
stack pointer = 0
```
# Tried without success
- giving full path for the #include statement 
- using new #includestr
- tried "\filename.sco" and ".\filename.sco"

The include in the orchestra is working flawlessly

vlazzarini:
I can't reproduce it using csound 7.0 beta or 6.18.1

```
victor@firebird debug % csound prova.csd
UnifiedCSD:  prova.csd
rtaudio: PortAudio module enabled ...
using callback interface
rtmidi: PortMIDI module enabled
sorting score ...
stack pointer = 0
	... done
--Csound version 7.0 beta (double samples) May 14 2024
[commit: 37e9b38ce41f2d5fd8032c34b9c20fa3a078e3de]
libsndfile-1.2.2
graphics suppressed, ascii substituted
sr = 44100.0, kr = 44100.000, ksmps = 1
0dBFS level = 1.0, A4 tuning = 440.0
ftable 101:
ftable 101:	1024 points, scalemax 1.000
               .-''''''-._
            _-'           '.
          _-                '.
         -                    '_
       .'                       -
[...]

Issue #1879, opened 2024-05-12
Regression: fix issue with loops in UDOs


kunstmusik:
A regression was created in the deinit PR due to a complication regarding updating OPDS struct definition and has an impact related to loops in UDOs that was previously fixed. The issue is a little complex and requires some larger refactoring, but since we have a queue of PRs to get through we're taking on the regression and noting it here to fix.

The background to the issue is here: https://github.com/csound/csound/pull/1859#discussion_r1596201351

The comments there have links to the relevant mailing list thread that discusses the UDO looping issue.

vlazzarini:
This has now been fixed in #1883

PR #1883, opened 2024-05-15
 FIX loops in UDO (#1879) and crash with no newline in score include (#1880) 

vlazzarini:
This PR addresses two issues

loops in UDO (https://github.com/csound/csound/issues/1879)
score include crash (https://github.com/csound/csound/issues/1880)
The first problem is a longstanding issue with loops in UDOs, which has never been addressed completely. The solution found was to make the perform code inside UDOs as similar as possible
to the one in instrument perform (performKsmps). This fixes the issue and makes the opcode
behave as expected.

The second problem is that the score lexer was crashing on the absence of a newline in an
included file. Adding a newline fixes the problem.


kunstmusik (approved):
Tested the Test0 example from the mailing list that was the original issue reported and it worked here with this branch. I noticed we still have csound->ids as a global used for init-time gotos, which is something we should probably look to remove and make local to the instrument. 

Going to merge this in.

Issue #1878, opened 2024-05-12
Regression: Fix issue with argdecode


kunstmusik:
argdecode was modified to fix an issue in https://github.com/csound/csound/pull/1859 with setOption failures calling longjmp. This however caused issues with commandline csound not exiting early as it should on a failed argument. Diagnosing this, argdecode should be rewritten to use return values and calling code updated to use the return values. (See comment in https://github.com/csound/csound/pull/1859/files#r1596198649).

vlazzarini:
Fixed in https://github.com/csound/csound/pull/1885.

PR #1885, opened 2024-05-18
fixing #1878

vlazzarini:
A simple fix for the regression issue. The function returns 0 (indicating failure) on a non-existing long option.


kunstmusik (approved):

kunstmusik:
Tested and fix appropriately failed for commandline with unknown arg.

PR #1892, opened 2024-05-31
Fixes for UDO regression

vlazzarini:
This PR fixes the UDO regression issue where

- `setksmps` could not be placed after `xin` 
-  `K` type was leading to type check failure because it was checked against `k`

### DIFF
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -1457,6 +1457,20 @@ void insert_opcodes(CSOUND *csound, OPCODINFO *opcodeInfo,
   }
 }
 
+static int inargs_check(OPCODINFO *opinfo, char *inargs) {
+  char *c = opinfo->intypes;
+  int i;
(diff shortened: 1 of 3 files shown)


kunstmusik:
Is there a missing commit? I see the code for K fix but nothing for setksmps. (Tested with transit.csd and got a crash and reported issue with setksmps after xin)

vlazzarini:
I had tested it in another branch and thought I had made the change here. Now transit.csd runs to the end, but there is a warning that csound->Free is called with an invalid pointer. That needs to be investigated but I think it is a separate issue to this.

vlazzarini:
That issue is fixed now, it was a problem in the ftgentmp deinit where the opcode object was being freed when it shouldn't have.

kunstmusik (approved):

kunstmusik:
Thanks, project is now working for me too. Merging and closing.

PR #1893, opened 2024-06-07
Jpff2

jpffitch:


vlazzarini:
The commit history of this PR is confusing, did you make your branch off develop? It has several commits by @hlolli which should not have been there and some mentions of csound6 branch merges. I don't think we can do much with it.
@kunstmusik what do you think went wrong here and what advice can we give?

stekyne:
The branch is based off of master instead of develop. Just do `git rebase develop` in this branch `jpff2`. Ideally should name the branch after the ticket number or something to indicate what it's solving.

kunstmusik:
Sounds like @stekyne figured out the situation. However, it makes me wonder if there are some wasm-build commits that never made it to CS7. (Been working on getting the wasm build going again for CS7, I'll check the above commits.) I'd suggest following @stekyne 's suggestions; one of us can also clone and do this and submit a new PR.

Issue #1598, opened 2022-03-04
Distribute Csound for MacOS as a .bundle


giovannibedetti:
Hi,

To be able to import Csound in Unity correctly for the MacOS platform, I need it in the form of a ".bundle", instead of a ".framework".

Could it be added to the build pipeline maybe?

Here is a bash script with the steps to convert a ".framework" into a ".bundle":

[convert_Csound_Framework_To_Bundle.sh.zip](https://github.com/csound/csound/files/8188472/convert_Csound_Framework_To_Bundle.sh.zip)

Thanks!

vlazzarini:
Can't you just rename the framework as a .bundle, create a MacOS dir inside it and put a symlink to CsoundLib64 there?

Then put in the info.plist and nothing else.

@rorywalsh do you know anything about this? I thought you used Csound in unity before.

giovannibedetti:
Yes sure that is what the bash script is doing ;)
The issue with Unity is that when building from Unity the .framework is not copied in the exported build, instead the .bundle is, so with the .bundle we have working MacOS builds.

vlazzarini:
The script does a lot of copying and changing links. You may not need all that, just a symlink.

giovannibedetti:
Thanks, I will try to use a symlink!
Btw the script is from the Unity support team, I asked them for help for the build issues we had on MacOs with the .framework.

rorywalsh:
@giovannibedetti and I have been doing it in a more brute force fashion. If having it as a .bundle means better integration into Unity, that's the way to go. @giovannibedetti , let me know how you get on. I will try to test the iOS stuff this week 👍

giovannibedetti:
Have a look here, this is the whole conversation about the issue:

https://fogbugz.unity3d.com/default.asp?1311509_ukb6c3vemh2mq1d4

kunstmusik:
Scheduling this for 7.0.0. It seems like it should not take too much work to add this to our build scripts as well as to github actions so that we're deriving a second artifact from our standard build.

vlazzarini:
@giovannibedetti 
I am incorporating your script into our release building scripts. Just a question about this line:

```
cp CsoundLib64.framework.meta CsoundLib64.bundle.meta
```

what is `CsoundLib64.framework.meta`?

vlazzarini:
A PR has been created to address this request.

giovannibedetti:
Thanks for this Victor!

The meta files should be generated by Unity (a .meta file exists for every
[...]

PR #1896, opened 2024-06-11
Building bundle from framework for release on MacOS

vlazzarini:
This PR addresses #1598, where it is requested that we build a bundle as part of the MacOSX release. The release script is also updated for 7.0.

### DIFF
--- a/installer/macosx/release-build.sh
+++ b/installer/macosx/release-build.sh
@@ -210,7 +210,7 @@ sudo chmod -R 755    CsoundApps64/Package_Contents/usr
 echo "creating bundle ..."
 
 # this step converts the framework into a bundle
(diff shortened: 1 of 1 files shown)


stekyne:
Would be great if this could be added as CMake code. https://stackoverflow.com/questions/53560288/how-to-create-a-macos-app-bundle-with-cmake I had a quick look and it seems like it might be possible. Ideally we get rid of platform specific scripts and just use CMake. I can try and take a look eventually but maybe this is good to merge for now

stekyne (approved):

vlazzarini:
We need both framework and bundle. At the moment, cmake gives us a framework, and we have this script to make a bundle. I guess next step is integrate it inside the cmake script.

PR #1897, opened 2024-06-12
Fix csound wasm

kunstmusik:


.github/workflows/csound_wasm.yml
vlazzarini, line 1: all looks ok. Since the API will change, there will be further changes needed, but it is good to bring wasm in line with current lib code now.

vlazzarini (approved):

PR #1900, opened 2024-06-23
Handle quoted strings in option reader

scottsitar:
When reading an option like `--env:FOO="bar baz"`, the first quotation mark would be replaced by ETX while the second quotation mark would get replaced by a NULL.  This has the effect of appending an ETX to the start of the variable's value.

As a first pass to addressing the parsing of quoted strings, this adds a post-processing step once we've detected the end of the current argument which compacts down the string, removing all the special characters we inserted while finding the end of the string.

This addresses an issue with quoted paths in the INCDIR variable reported on the mailing list.


scottsitar:
After thinking about it, I wasn't happy with the initial patch in my message on the mailing list, since it exposed the control characters we inserted during option parsing outside of the function that needed to know about them.  So, I've taken a different approach here by adding a post-processing step to each argument to get rid of these characters.

I tried to make this change as minimally invasive as I could, as I noticed a few other issues while working on this that I didn't want to address all at once:

1) If there's a quoted string argument, it terminates the current line at the end and passes the results on to argdecode - this misses any other arguments that may have come after it on the line
2) There are a few out-of-bounds accesses on the line buffer as we're processing a line

I'd be happy to take a closer look at the argument parsing to clean it up and put some tests around it, if people think that would be worthwhile.  I've only added a very basic test here, and I'm not sure if this is the right way to test this function.  Please let me know if there's a better way to do it!

vlazzarini:
This looks good to me. Sorry for the long delay. I wasn't sure it was ready. Could bring it in sync with develop?

scottsitar:
No worries!  Reading over my comment, it was definitely ambiguous, so sorry about that.  I've merged in the latest develop and fixed up the test according to the new api.

vlazzarini (approved):
All looks good, thx.

PR #1901, opened 2024-06-23
Update readme template to avoid cmake run changes

scottsitar:
I found this will working on my other change - scratching my head wondering if I'd messed up syncing my fork with the main repository.  Doing a cmake generation will create the main readme from this template, so I've updated it with the new blurb about the develop branch that was added to the main one.


vlazzarini (approved):
all ok thanks

vlazzarini:
Thanks. I forgot all about the readme generation and edited the generated file.

PR #1907, opened 2024-07-01
dynamically construct path to csound.exe

### LINKED ISSUE #1906: invalid escape sequence in commandline test
The commandline test contains the following lines:
https://github.com/csound/csound/blob/b4e8bc58d864d9ec3ed391fcb38bd9d909444ab1/tests/commandline/test.py#L227-L228 where the string `"..\csound.exe"` contains the invalid escape sequence `"\c"`.

Obviously this is just a typo, and what is meant here is the literal backslash, rather than an escape sequence.

As of Python3.12, this is an error.

You could either use a raw-string (`r"..\csound.exe"`), escape the backslash (`"..\\csound.exe"`) or construct the path on the fly (`os.path.join("..", "csound.exe")`)

umlaeute:
this way, we do not have to use the dreaded backslash in strings and accidentally forget to escape it (something that Python3.12 does not like at all).

I went for the `os.path.join()` route, as the relevant code is protected by https://github.com/csound/csound/blob/b4e8bc58d864d9ec3ed391fcb38bd9d909444ab1/tests/commandline/test.py#L227

at least in theory this could mean that the path seperator is **not** a backslash, so we shouldn't use it...

Closes: https://github.com/csound/csound/issues/1906

### DIFF
--- a/tests/commandline/test.py
+++ b/tests/commandline/test.py
@@ -225,7 +225,7 @@ def runTest():
             print(command)
             retVal = os.system(command)
         elif(os.sep == '\\' or os.name == 'nt'):
-            executable = (csoundExecutable == "") and "..\csound.exe" or csoundExecutable
+            executable = (csoundExecutable == "") and os.path.join("..", "csound.exe") or csoundExecutable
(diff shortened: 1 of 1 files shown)


vlazzarini (approved):
Looks good, thank you! I'll wait for CI results before merging.

PR #1908, opened 2024-07-01
Remove executable bit from C source files

ndim:
Remove the executable bit from the following *.c source files:

    util/lpc_main.c
    util/mixer.c
    util/pvc_main.c
    util1/sortex/smain.c
    util1/sortex/xmain.c

These are C source files, not scripts of any kind, so it makes no sense for them to have the x bit set.

PR #1909, opened 2024-07-01
Remove executable bit from non-executable files

ndim:
Remove the executable bit from the following *.c source files:

    util/lpc_main.c
    util/mixer.c
    util/pvc_main.c
    util1/sortex/smain.c
    util1/sortex/xmain.c

These are C source files, not scripts of any kind, so it makes no sense for them to have the x bit set.

This PR was originally erroneously misfiled against the `master` branch at https://github.com/csound/csound/pull/1908. This change can be cherry-picked into the `develop` branch as well.


ndim:
There are few more files which should not have the executable bit set, but those did not get caught by the automatic rpm checks for executable source code files. I am examining those, and will be adding them as well.

ndim:
Commit d08aada00f6bb6e652f8105c474398201a81acf1 appears to be complete for the `csound6` branch.

PR #1914, opened 2024-07-22
fix: removing armv7;armv7s from libsndfile build for iOS

rorywalsh:


vlazzarini (approved):

vlazzarini:
looking good. Rerunning the vcpkg job to see if it clears now.

PR #1915, opened 2024-08-07
Changes to allow compiling for iOS using ioscross(cctools-port)

nonameentername:
The following changes allow cross-compiling Csound for iOS using a Linux docker image.


.github/workflows/csound_builds.yml
```diff
@@ -101,6 +101,53 @@ jobs:
...
+        run: ./vcpkg/bootstrap-vcpkg.sh
+
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install build-essential liblzma-dev libxml2-dev libz-dev libbz2-dev cpio ninja-build
+
+      - name: Download ioscross
+        run: |
+          curl -L "https://drive.usercontent.google.com/download?id=1COyrQ7ypTTVBiI9dBIOvswIm1CWbNFa0&confirm=xxx" | tar xvz
```
stekyne: Not a great solution if this is outside of the control of the Csound org or admins. If the link goes dead or whatever other reason, then the build breaks. It's also a bit opaque, not sure what is being executed here unless we look inside. If there are proprietary files then that's probably not great either.

.github/workflows/csound_builds.yml
```diff
@@ -101,6 +101,53 @@ jobs:
...
+
+      - name: Configure Build
+        run: |
+          $(eval /home/runner/work/csound/csound/ioscross/ioscross_conf.sh)
+          cmake -B build/arm64 -S . -DUSE_VCPKG=1 -DCUSTOM_CMAKE=./platform/ioscross/custom-ios.cmake \
+              -DIOS=1 \
```
stekyne: Maybe put all these extra flags inside the `custom-ios.cmake` ? Can clean this up a bit

CMakeLists.txt
```diff
@@ -20,14 +20,24 @@ option(USE_VCPKG "Use VCPKG to download and manage dependencies" OFF)
...
-        elseif(CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")
-            set(VCPKG_TARGET_TRIPLET "arm64-osxcross")
+        if (IOS)
+            if(CMAKE_OSX_ARCHITECTURES STREQUAL "arm64")
+                set(VCPKG_TARGET_TRIPLET "arm64-ioscross")
+            endif()
+            set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE ${CMAKE_CURRENT_SOURCE_DIR}/platform/ioscross/ios.cmake)
+            set(VCPKG_CMAKE_SYSTEM_NAME iOS)
+            set(VCPKG_TARGET_ARCHITECTURE "${CMAKE_OSX_ARCHITECTURES}")
+            set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS}  -DIOS=1")
```
stekyne: I think at this point you should create a vcpkg triplet file to encapsulate all these flags and whatever logic is needed

platform/ioscross/README.md
```diff
@@ -0,0 +1,52 @@
+Cross compiling iOS from Linux
+==============================
+
+This directory contains scripts for compiling a iOS version 
+of Csound from Linux.  These scripts were developed using Ubuntu.
+
+**[Please ensure you have read and understood the Xcode license
+terms first before using it.](https://www.apple.com/legal/sla/docs/xcode.pdf)**
```
stekyne, line 8: I'm pretty sure this says you must develop for Apple machines using Apple products and this cross compilation (or use of their SDKS on non-Apple machines) is breaking the EULA. That's one thing for a dev to do it as a hobby project but for an organisation like Csound it might mean something else. Can you explain if this is legally gray or otherwise?

platform/ioscross/README.md
```diff
@@ -0,0 +1,52 @@
...
+sudo apt install docker.io
+```
+
+2. Add user to docker group:
+
+```bash
+sudo usermod -a -G docker $USER
+```
+
+3. Download XCode with the iOS SDK (you must be logged into an Apple ID to download Xcode).
```
stekyne, line 24: Are we using the SDKs or Xcode for this cross compilation at some point?

platform/ioscross/custom-ios.cmake
```diff
@@ -0,0 +1,13 @@
...
+set(BUILD_STATIC_LIBRARY ON)
+set(BUILD_JAVA_INTERFACE OFF)
+set(INSTALL_PYTHON_INTERFACE OFF)
+set(BUILD_UTILITIES OFF)
+set(BUILD_TESTS OFF)
+set(HAVE_TERMCAP OFF)
+set(USE_DOUBLE 0)
+
+set(CMAKE_CXX_COMPILER $ENV{OSXCROSS_TARGET_DIR}/usr/bin/arm-apple-darwin11-clang++)
+set(CMAKE_C_COMPILER $ENV{OSXCROSS_TARGET_DIR}/usr/bin/arm-apple-darwin11-clang)
```
stekyne: Can these not be determined by Cmake? It seems a bit too machine specific. This build will break if those files aren't found.

platform/ioscross/ios.cmake
stekyne, line 1: Not really sure what is happening here. What is this needed for? It looks like messing with VCPKG internal cmake functions

platform/vcpkg/ports/mp3lame/disable-ios-capterm.patch
stekyne, line 1: Not hugely in favour of adding custom patches to packages within this repo. I would occasionally update our VCPKG packages by simply fetching the latest remote for the git submodule and I'd rather not have to encounter a build failure because this patch no longer works with the updated version of mp3lame. 

Can you get this merged in VCPKG? We used to have a lot of platform specific patches that were a nightmare to maintain but they're all removed now and we just rely on VCPKG mostly. I had to add some dependencies myself and fix issues within ports at times as well.

platform/vcpkg/triplets/arm64-ioscross.cmake
```diff
@@ -0,0 +1,15 @@
...
+set(VCPKG_OSX_ARCHITECTURES arm64)
+
+set(CMAKE_OSX_SYSROOT $ENV{OSXCROSS_SDK})
+
+set(VCPKG_C_FLAGS "-I${CMAKE_OSX_SYSROOT}/usr/include")
+set(VCPKG_CXX_FLAGS "-I${CMAKE_OSX_SYSROOT}/usr/include -stdlib=libc++")
+set(VCPKG_LINKER_FLAGS "-L${CMAKE_OSX_SYSROOT}/usr/lib")
+
+get_filename_component(TOOLCHAIN_FILE "../platform/ioscross/ios.cmake" REALPATH BASE_DIR "${CMAKE_SOURCE_DIR}")
+set(VCPKG_CHAINLOAD_TOOLCHAIN_FILE ${TOOLCHAIN_FILE})
```
stekyne, line 11: Afaik this should be in the `CMakeLists.txt` and not this triplet file. Look at the triplet files in the VCPKG folder and you'll see they are mostly settings and not referencing external files.

platform/vcpkg/ports/mp3lame/disable-ios-capterm.patch
stekyne: Originally, I added only `Portaudio` as an overlay to fix a problem but this folder was never intended to include more ideally. Anytime we update VCPKG packages by fetching the latest remote git version, these packages remain the same and have to be manually updated with the new files. It means any custom patches stored here have to be re-written for the newer version if they break. I would rather we had zero overlays because it is must less maintenance. I'm just realising there are now four which I don't think is a great idea.

stekyne:
Thanks for your continued work. This looks like a good quality PR in terms of the code written and adherring to the patterns used already. I'm a bit worried about the legality of cross-compiling Apple software even though I know we added the OSXcross platform already. 

I've left comments for specific issues I found. Overall, I'm a bit skeptical of these cross-compiling platforms. It requires some tweaks that are not quite clean and require maintenance when updating VCPKG packages. The Cmake files are marginally more complicated but not terribly so. I've outlined the points in the comments anyways. 

Are these builds being used downstream in a particular project or can you say why they are needed? Seeing as they require some maintenance in this repo, mostly the port overlays, but also if the builds start breaking due to various external reasons, it's hard to accept these in if there are very few people using them. 

A few years ago the build system and the number of external dependencies in this repo meant that a lot of dev time was fixing the platform and not doing features. I think it's been pretty good the last few years especially seeing as VCPKG has abstracted a lot of that. I think adding too many platforms and platform specific CMake code introduces some of that complexity back when we're trying to simplify it further.

.github/workflows/csound_builds.yml
nonameentername: I do agree that this is not a great solution but was the suggested approach in the past for a previous pr.
The steps to build the tar are found in the [README](https://github.com/nonameentername/csound/blob/dcc5108cf9986141816a1621fd0b989f1973d493/platform/ioscross/README.md#build-github-targz).
To avoid linking to a file outside of the Csound org and admins, I updated my pr to instead use the location specified in the variable `vars.IOSCROSS_DOWNLOAD_URL`.  The Csound admins can chose to upload this to a desired location and the set the variable to enable the job.

.github/workflows/csound_builds.yml
nonameentername: I moved as many of these flags as I could into the custom cmake files.  I wasn't able to move all of them since the vcpkg config happens before the custom cmake file is read.

CMakeLists.txt
nonameentername: I moved the logic to the vcpkg triplet.

platform/ioscross/README.md
nonameentername: There are large open source organizations that allow cross-compiling for Apple machines.
The Godot project has [documentation](https://docs.godotengine.org/en/stable/contributing/development/compiling/cross-compiling_for_ios_on_linux.html) on how to cross compile their project for ios using cctools-port.  Some of the changes here are based on their documentation.
They do have a [disclaimer](https://docs.godotengine.org/en/stable/contributing/development/compiling/cross-compiling_for_ios_on_linux.html#disclaimer) in their documentation that specifies the following:

> While it is possible to compile for iOS on a Linux environment, Apple is very restrictive about the tools to be used (especially hardware-wise), allowing pretty much only their products to be used for development. So this is not official. However, in 2010 Apple said they relaxed some of the [App Store review guidelines](https://developer.apple.com/app-store/review/guidelines/) to allow any tool to be used, as long as the resulting binary does not download any code, which means it should be OK to use the procedure described here and cross-compiling the binary.

platform/ioscross/README.md
nonameentername: Yes, Xcode and the SDKs are used but the end user is responsible getting an Apple ID and downloading Xcode.  The scripts themselves do not download Xcode to avoid re-distribution of the Apple SDK.

platform/ioscross/custom-ios.cmake
nonameentername: I updated these to make them less platform specific.

platform/ioscross/ios.cmake
nonameentername: I removed the VCPKG internals and only specify the settings needed for cross compilation.

platform/vcpkg/ports/mp3lame/disable-ios-capterm.patch
nonameentername: I agree that having these patches is not ideal.  I'm looking into how to avoid having to patch these files for this pr, but I think since these ports already existing it doesn't hurt to modify them.

platform/vcpkg/triplets/arm64-ioscross.cmake
nonameentername: I tried removing this file but when using a toolchain vcpkg ignores most of the vcpkg flags and relies on the toolchain to set the values.
The vckpg [documentation](https://learn.microsoft.com/en-us/vcpkg/users/triplets#vcpkg_chainload_toolchain_file) it specifies `VCPKG_CHAINLOAD_TOOLCHAIN_FILE` being an option for triplets.  Not sure how to go about cross compiling without specifying this value.

nonameentername:
@stekyne Thanks for taking the time to review my pr.

I’ve added cross-compilation changes to facilitate smoother integration with the Csound project. This allows contributors to make and test changes on one machine while ensuring compatibility across different platforms. It also benefits contributors without Apple computers by letting them address build issues locally rather than relying solely on GitHub Actions, which can be tricky to debug.

Additionally, I’ve included instructions for building Docker containers to help with local Csound builds. The goal of these changes is to streamline the process for contributors, making it easier to verify their updates and reproduce the build environment for troubleshooting.

These changes are being utilized in my own project [godot-csound](https://github.com/nonameentername/godot-csound). Previously, I’ve invested considerable time setting up development environments on both Windows and Apple computers to address compilation errors. With cross-compilation, many of these issues could have been resolved more efficiently.

.github/workflows/csound_builds.yml
stekyne: You can add platform specific code to where the VCPKG code is in the main cmakelists.txt as well

.github/workflows/csound_builds.yml
stekyne: I suggested caching the build files to reduce the 2.5 hour build time but it wasn't ideal like here either. Maybe it would be better to create a docker image that contains the build tool binaries already. At least we can have a record of how it's built and version the images.

platform/ioscross/README.md
stekyne: Ok that sounds a bit better so

platform/vcpkg/ports/mp3lame/disable-ios-capterm.patch
stekyne: Did you read my point about updating those packages afterwards though? Patches create a lot of maintenance to update these overlay packages in the future.

.github/workflows/csound_builds.yml
nonameentername: I updated the pr to use a docker image instead of an external file.

rorywalsh (commented):
I'm going to approve this. I understand the concerns raised by @stekyne, but I think it's worth letting this through for now.

rorywalsh (approved):

rorywalsh:
Merging

vlazzarini:
this is not merged yet, any issues?

rorywalsh:
Sorry, should be merged now

vlazzarini:
I see the iOS ioscross build is skipped. Is that how it should be?

nonameentername:
@vlazzarini, some minimal configuration is required in GitHub for the iOS ioscross build to work. I can assist with the configuration if you'd like to enable it.

vlazzarini:
can we have a PR to test it? Also if we could have the Android build, that'll complete the set. We can then decide if we want to enable these only on develop.

nonameentername:
@vlazzarini, the configuration for these needs to be done in the GitHub UI. There are no further changes needed in the code. I have already configured these to enable them in my mirror/fork of Csound, and they are all passing. Here is the [link](https://github.com/indiegamedevs/csound/actions/runs/12119947489) to this PR.

vlazzarini:
what exactly needs to be done?

nonameentername:
@vlazzarini The Android and iOS osxcross build steps are similar, but building for iOS requires additional steps since the Apple SDK needs to be downloaded and cannot be redistributed.

The current iOS osxcross build process is done in two steps:
1. **Build the Docker image using the Apple SDK**. This is done in the csound build images [workflow](https://github.com/csound/csound/blob/develop/.github/workflows/csound_build_images.yml#L45), and includes the following steps:
   - Download the Apple SDK from the configured location (`IOSCROSS_DOWNLOAD_URL`).
   - Decrypt the file using the configured secret (`IOSCROSS_PASSWORD`).
   - Build the Docker image and publish it as an artifact.
   - Set the workflow run ID in the variable `IOSCROSS_WORKFLOW_RUN_ID` to allow retrieving the Docker image. A token with permissions to store variables (`SERVICE_ACCOUNT_PAT`) is required to update the workflow ID.

2. **Use the Docker image to build csound for iOS**. Once the image is built, it is used in the csound build [workflow](https://github.com/csound/csound/blob/develop/.github/workflows/csound_builds.yml#L119), which does the following:
   - Download the Docker image using the workflow run ID from `IOSCROSS_WORKFLOW_RUN_ID`.
   - Use the Docker image to build csound for iOS.

### Required changes to enable the iOS build:
1. Download Xcode version 15.1 with the iOS SDK from the [Apple Developer portal](https://developer.apple.com/download/).
2. Encrypt Xcode using the following command in the terminal:
   ```bash
   read -s PASSWORD
   cat Xcode_15.1.xip | openssl aes-256-cbc -pbkdf2 -a -salt -pass pass:$PASSWORD > Xcode_15.1.xip.enc
   ```
3. Upload the encrypted Xcode binary to an accessible location.
4. Create a GitHub fine-grained private access token with the following permissions for the csound repository:
   - Read access to metadata
   - Read and write access to actions variables
5. In the csound project, create the following secrets:
   - `IOSCROSS_PASSWORD`: The password used to encrypt and decrypt the Xcode binary.
   - `SERVICE_ACCOUNT_PAT`: The private access token from the previous step.
6. In the csound project, create the following variable:
   - `IOSCROSS_DOWNLOAD_URL`: The location of the encrypted Apple SDK.
   
Once the above steps are complete, the Docker build images can be run manually. The variables `ANDROID_WORKFLOW_RUN_ID` and `IOSCROSS_WORKFLOW_RUN_ID` will be set and will enable the csound build workflows.

vlazzarini:
Thanks, it seems quite involved. What if I add you to the development team so you can make these changes directly? You know what to do, and I don't really.

nonameentername:
@vlazzarini yeah, that works for me.  I can make the changes.

vlazzarini:
ok, I'll do it later this evening thanks.

vlazzarini:
ok, done @nonameentername, I've added you the developers team.

nonameentername:
@vlazzarini I tried configuring these workflows, but it seems like I do not have access to the Csound settings in Github after joining the developers team.  I can pair with someone that does have access to help get these configured.

vlazzarini:
Can you try again? I elevated the role for developers.

nonameentername:
@vlazzarini thanks.  I have access to the **Settings** now but the **Secrets and variables** section is not visible.

vlazzarini:
Ok you should be able to do this now.

PR #1917, opened 2024-08-09
Fix event() in ctcsound

mesca:
The event() method in ctcsound.py did not parse the arguments correctly. This PR fixes this.
Sorry, my editor fixed the blank lines, but the only change is line 1304.


mesca:
@fggp Can you review this?

PR #1920, opened 2024-08-15
fixed setting of ksmps, kr and sr in UDOs

vlazzarini:
This PR fixes an issue where the Csound variables ksmps, kr and sr never got updated when the UDO is initialised, so they carried their old initial values instead.

I have restored the code from 6.x that had been removed and added the setting of sr.

NB: these variables are also updated by setksmps and oversample/undersample.


kunstmusik (approved):

kunstmusik:
Looks good to me, I'll go ahead and merge.

Issue #1921, opened 2024-08-15
Csound sfont opcodes fail when built with BUILD_PLUGINS=OFF


nonameentername:
In the development branch of Csound the sfont opcodes are currently not working when built with BUILD_PLUGINS=OFF.
When running csound the following error is displayed:
```bash
$ csound test.csd
UnifiedCSD:  test.csd
Elapsed time at end of orchestra compile: real: 0.001s, CPU: 0.001s
sorting score ...
        ... done
Elapsed time at end of score sort: real: 0.001s, CPU: 0.001s
--Csound version 7.0 beta (double samples) Aug 15 2024
[commit: 626e21354245c17774402b24a548812f5043e5ed]
libsndfile-1.2.2
graphics suppressed, ascii substituted
sr = 44100.0, kr = 1378.125, ksmps = 32
0dBFS level = 1.0, A4 tuning = 440.0
INIT ERROR in instr 0 (opcode sfload) line 15: sfload: could not open globals

 from file test.csd (1)
isf     sfload  "assets/FluidR3_GM.sf2"

csound command: Segmentation fault
inactive allocs returned to freespace
end of score.              overall amps:  0.00000  0.00000
           overall samples out of range:        0        0
0 errors in performance
Elapsed time at end of performance: real: 0.251s, CPU: 0.001s
no sound written to disk
```

PR #1922, opened 2024-08-15
Fixing #1921

nonameentername:
This PR fixes the issue with sfont opcodes. #1921


vlazzarini (approved):
all good, thanks for spotting and fixing this bug.

Issue #1843, opened 2024-03-13
CMake and Csound on Android


rorywalsh:
It would be nice if we could have CMake manage the Android library. I know that projects built with Android Studio generate their own CMakeFiles.txt file which can cause issues with existing CMake builds, but in this instance the build is done with `ndk-build`, so it might be manageable?

kunstmusik:
Was this done?

DaveSeidel:
Well, I can say that I have not had any trouble building Csound 7 on
Android for a while now. So I guess it was fixed (probably by accepting a
lower CMake version number).

FYI, I also wrote up a bug for the CMake team on this issue, which they
accepted, but I don't believe that it's been addressed yet.

kunstmusik:
I don't see CMake being used in the Android build.

DaveSeidel:
I don't know what happens on Github, but I have (since cs6) used CMake to
build, and I'm still doing so.

kunstmusik:
Interesting, what command are you using for this to pickup the Android NDK
toolchain and libs? If CMake works we should remove the older custom NDK
build files and migrate to that for the release build.

DaveSeidel:
Oh gosh, very sorry -- I was confusing this with the Raspberry Pi CMake
build issue. Please ignore me, and sorry for the distraction!

vlazzarini:
There's a CI build, merged. So yes, it was done. I see cmake commands in the PR.

vlazzarini:
> I don't see CMake being used in the Android build.  n

What's this line doing then?
https://github.com/csound/csound/blob/e36ebaa59ffcd7ece2d14b2e1de2843a73e1bc51/.github/workflows/csound_builds.yml#L532

kunstmusik:
That's new to me. If we're good to go with this, we should remove the older
build files and update build instructions.

vlazzarini:
We're definitely good, the build works.
@nonameentername could you do what @kunstmusik is asking?

nonameentername:
@vlazzarini @kunstmusik I can update the documentation and remove the old build files.

In the folder structure, there is an `Android` folder that I believe contains the old build files and instructions. Should the `CsoundAndroid` and `CsoundForAndroid` folders be updated to use the CMake Android artifacts, or should they be removed?

PR #1924, opened 2024-08-24
Changes to allow compiling for Android using vcpkg

nonameentername:
Implementation for #1843


stekyne:
@kunstmusik do you know who uses/maintains the android build?

kunstmusik:
I think @rorywalsh is maintaining Android now.

rorywalsh:
It's true, I put my hand up a some point. This PR looks good to me, but I've no experience with docker images.

rorywalsh (commented):
I've gone over this and it looks good. I can't accept PRs, Stephen/Steven do you guys have any concerns about this?

stekyne:
It looks ok to me but it's disabled in the CI. We need to run it at least once to make sure it's all good

nonameentername:
In this PR, there are two new jobs for the CI/CD pipeline:

Docker Image Creation Job:
- This job builds a Docker image.
- It saves the Docker image as an artifact.
- It stores the workflow run ID for future reference.

Android CI Job:
- This job uses the stored workflow run ID to download the Docker image artifact.
- It then executes the build process.

To enable Android CI:
- Manually run the Docker build job.
- Create a variable in the CI settings named ANDROID_WORKFLOW_RUN_ID, and set its value to the workflow run ID of the Docker image build job.

Additional Configuration:

- I have added functionality to automatically set the workflow run ID when building the Docker images.
- To use this feature, a fine-grained access token is required with the following permissions:
 1. Read access to metadata
 2. Read and Write access to actions variables
- Once the token is created, it can be added as a secret named SERVICE_ACCOUNT_PAT.

stekyne:
Ok lets see how it works out maybe. It sounds pretty well reasoned.

stekyne (approved):

Issue #1890, opened 2024-05-31
Refactor API for standard types


kunstmusik:
https://github.com/csound/csound/pull/1852 introduced exposing of csound's standard variable types via CSOUND struct. I think there's an opportunity to make this a little tidier than adding all of the types to the CSOUND struct, perhaps using a sub struct. Adding this issue to further investigate.

vlazzarini:
@kunstmusik I pushed a new branch containing a refactor. The main idea is

- A struct containing standard type pointers (`CS_STANDARD_TYPES`)
- A pointer to this struct added to `CSOUND` replacing the individual type pointers
- A single module API function `GetType()` to return a type given a type name char (a, k, etc)

This tidies up the way we expose types and also should allow us to add new types to the interface
without breaking backward ABI compatibility.

It may be possible to extend the refactor further, but I think it would now be mostly to do with internal changes, so it may be pushed to a later date.

kunstmusik:
Seems like GetType and csoundGetTypeWithVarTypeName() should be merged and only one used (preferring the latter). csoundGetTypeWithVarTypeName() is an ugly long name, but it works better IMO since it takes in char* and works for all type names.

vlazzarini:
I'll look at it right away.

vlazzarini:
So that looks much neater, it was a great suggestion.
I  made GetType an interface to GetTypeWith... to avoid having to expose the internal typePool.

vlazzarini:
I think this is completed now. In the end, following your suggestions, we did not need to expose the types via the API anymore since we have them in the the typePool and can access them easily through GetType(). This also led me to move the standard types header to the internal headers directory since none of it needs to be exposed via the API: the add standard types is only used internally and the const types themselves can be obtained via csoundGetTypeWithVarTypeName().

vlazzarini:
I've added access to the typePool so that the type system functions can be useful in the API. If this is a bit dangerous we should make it const and remove the add functions from the API.

Now we have to decide about varPool functions:

```
  PUBLIC CS_VAR_POOL* csoundCreateVarPool(CSOUND* csound);
  PUBLIC void csoundFreeVarPool(CSOUND* csound, CS_VAR_POOL* pool);
  PUBLIC char* getVarSimpleName(CSOUND* csound, const char* name);
  PUBLIC CS_VARIABLE* csoundFindVariableWithName(CSOUND* csound,
                                                 CS_VAR_POOL* pool,
                                                 const char* name);
[...]

PR #1925, opened 2024-08-25
Feature/apitypes

vlazzarini:
PR for #1890


vlazzarini:
WASM build failure probably due to moving standard types out of public headers. I don't know enough about WASM to be able to fix this.

kunstmusik (dismissed):

kunstmusik:
Taking a look at WASM build now

kunstmusik:
There was an issue with moving csound_standard_types.h out to H folder, since it was #include'd in csoundCore.h. I removed that from csoundCore.h and included where necessary which got things to build. The problem showed up in WASM when building a plugin library. This should be sorted out now. Going to go ahead an merge this in. We can look at exposing VAR POOLs in a separate PR.

kunstmusik (approved):

PR #1926, opened 2024-08-26
Fixing Type Copy Value functions to take account of context

vlazzarini:
Currently the create variable and copy value functions do not take account of the context they are called into. This is a problem with the asig type, where the size of the vector depends on the local value of ksmps. In 6.x, this leads to memory overuse, particularly with arrays, in cases where the local ksmps is significantly less  than the global ksmps.

This PR fixes the problem by adding an optional context argument to the type createVariable / copyValue functions. For asig types, these receive a pointer to the OPDS in which the copy is taking place if local ksmps is to be used, or NULL if the creation/copy is to use the global ksmps value.


kunstmusik:
@vlazzarini Took a quick look, is there a reason to use void* rather than OPDS*?

vlazzarini:
It should be possible but because of the way OPDS is declared in csoundCore.h I would need need to move definitions around. I could not simply include csoundCore.h as I would get circular includes. Maybe I could just do a typedef in csound_type_system.h and then just define the struct in csoundCore.h.

I also thought that in the future context may be something else so in that case the signature does not need to be changed.

EDIT: a simple forward declaration does it. Should I change it?

vlazzarini:
Looking at the csound_standard_types.c, I see this

```
void updateAsigMemBlock(CSOUND* csound, CS_VARIABLE* var) {
    int ksmps = csound->ksmps;
    var->memBlockSize = CS_FLOAT_ALIGN(ksmps * sizeof (MYFLT));
}
```

Is this OK or should we take account of local ksmps here too?

kunstmusik:
> Looking at the csound_standard_types.c, I see this
> 
> ```
> void updateAsigMemBlock(CSOUND* csound, CS_VARIABLE* var) {
>     int ksmps = csound->ksmps;
>     var->memBlockSize = CS_FLOAT_ALIGN(ksmps * sizeof (MYFLT));
> }
> ```
> 
> Is this OK or should we take account of local ksmps here too?

I think we technically can, but because local ksmps is always smaller or equals to global ksmps, it's only a minor efficiency issue.

kunstmusik:
Looks good to me, going to merge this.

kunstmusik (approved):

Issue #1919, opened 2024-08-12
Build fails with LTO due to ODR violation / type mismatch


eli-schwartz:
I tried to build with the following *FLAGS to optimize the build: `-flto=4 -Werror=odr -Werror=lto-type-mismatch -Werror=strict-aliasing`

Link-Time Optimization is a massively global compiler optimization pass which is pretty handy for producing faster executables. It also has the interesting property that because the compiler does whole-program analysis using bytecode, it can save type information and perform error checks that it normally doesn’t have enough insight for. In particular, checking for ODR issues and checking function type signature mismatches.

Note that *all* the -Werror=* flags are used to help detect cases where the compiler tries to optimize by assuming UB cannot exist in the source code -- if it does exist, ordinarily the code would be miscompiled, and this says to make the miscompilation a fatal error.

I got this error:
```
FAILED: libcsound.so.7.0 
[...]

PR #1930, opened 2024-08-31
New Host API

vlazzarini:
This PR introduces the New Host API, with changes mostly affecting public headers but also including some internal changes.

Some changes in this PR also tidy up the build of opcodes as internal or as plugins, which were including the wrong headers at times (csdl.h instead of csoundCore.h). This problem was preventing the application of link time optimisation (#1919).

NB: we have liaised with the developers of CsoundQT and Cabbage and the API has already been integrated in a new version of CsoundQT. The API has also been integrated in the new version of csound for Pd (csound7~).

A set of examples in C and in C++ has been added to the csoundAPI_examples repository.


vlazzarini:
This should be ready for review now: @kunstmusik

fggp:
I've still some work to do on ctcsound.py. Maybe one or two more days...

vlazzarini:
ok!

CMakeLists.txt
```diff
@@ -1381,10 +1384,13 @@ if(USE_LIBSNDFILE)
 endif()
 
 if(SampleRate_FOUND)
+    if(NOT IOS)
   target_link_libraries(${CSOUNDLIB} PUBLIC SampleRate::samplerate)
+      endif()
 endif()
 
 if(NOT IOS)
+   target_link_options(${CSOUNDLIB} PUBLIC ${CSOUNDLIB_SHARED_LINK_OPTIONS})
```
kunstmusik, line 1402: It looks like CSOUNDLIB_SHARED_LINK_OPTIONS is set to "" on line 8. Was this temporary work that should be removed now?

Engine/csound_orc_semantics.c
```diff
@@ -1467,8 +1468,8 @@ void add_arg(CSOUND* csound, char* varName, char* annotation, TYPE_TABLE* typeTa
   var = csoundFindVariableWithName(csound, pool, varName);
   if (var == NULL) {
     if (annotation != NULL) {
-      type = csoundGetTypeWithVarTypeName(csound->typePool, annotation);
-      typeArg = type;
+      type = (CS_TYPE *) csoundGetTypeWithVarTypeName(csound->typePool, annotation);
```
kunstmusik: Rather than cast to `(CS_TYPE *)` everywhere, shouldn't type be changed to `const CS_TYPE*`?

kunstmusik: (and same comment for all places csoundGetTypeWithVarTypeName is being used)

CMakeLists.txt
vlazzarini: no, I am using this in Custom.cmake. It is   there so it can be filled or not later
I don't know of a better way to set this up,  it's not an option.

Engine/csound_orc_semantics.c
vlazzarini: I tried that but there was a point where it was awkward. I can try again.

CMakeLists.txt
vlazzarini: but I think I can make it into an option. It's for lto.

Engine/csound_orc_semantics.c
vlazzarini: I see why it was awkward, I had to enforce everywhere. That's the right thing though:
9aea5349a0c7ed51910a88d3ffd82853adce4671 
bd70b3254d11515e9c40cc514ad9a99d6364239f

CMakeLists.txt
vlazzarini: yes, that's in there now
068238cee4203071262e9d019bb0d40a93a10a6e

OOps/bus.c
```diff
@@ -2164,3 +2274,419 @@ int32_t outvalsetSgo(CSOUND *csound, OUTVAL *p)
     // if (ans==OK) ans = koutval(csound,p);
     return ans;
 }
+
+/* ARRAY channels  implementation - VL 7.08.24 */
+static void copy_array(ARRAYDAT *out, const ARRAYDAT *in, spin_lock_t *lock) {
```
kunstmusik: Pretty sure this won't work for all array types; see array_copy_value in csound_standard_types.c. We should reuse the type system's copy mechanism here, otherwise I think we'll get unexpected data sharing.

include/circular_buffer.h
```diff
@@ -0,0 +1,96 @@
+/*
+  udp_server.h: Csound UDP server API
```
kunstmusik: file name and description incorrect here

include/cs_threads.h
```diff
@@ -0,0 +1,239 @@
+/*
```
kunstmusik: This and other headers seems to have PUBLIC functions that aren't getting included into Doxygen.  We need to audit and make sure the generated doxygen docs include everything we're making PUBLIC.

include/csoundCore.h
```diff
@@ -575,7 +781,7 @@ extern int ISSTRCOD(MYFLT);
     uint64_t kcounter;
     MYFLT    esr, sicvt, pidsr;                  /* local sr */
     MYFLT    onedsr;
-    int     overmode;
+    int     in_cvt, out_cvt;
```
kunstmusik: BTW: I'm trying to decipher what in_cvt and out_cvt means. It's not clear to me what it stands for.

include/csoundCore.h
```diff
@@ -1,7 +1,7 @@
 /*
-  csoundCore.h:
+  csoundCore.h: somet modifications to match the new API
```
kunstmusik: The file description here seems off with "somet" and just in general. Looks like and odd typo?

include/csoundCore_new.h
```diff
@@ -0,0 +1,2283 @@
+/*
```
kunstmusik: Can this file be removed?

tests/c/CMakeLists.txt
```diff
@@ -10,15 +10,13 @@ endif()
 if(BUILD_TESTS)
     add_executable(unittests
         channel_tests.cpp
-        cscore_tests.cpp
-        csound_circular_buffer_test.cpp
```
kunstmusik, line 14: I see cscore was dropped, but it seems like csound_circular_buffer_test.cpp should be restored here.

tests/c/cscore_tests.cpp
```diff
@@ -39,7 +39,7 @@ void cscore(CSOUND *cs)
 // TEST (CScoreTests, testCscore)
```
kunstmusik, line 39: seems like this file can be deleted if cscore is removed

OOps/bus.c
vlazzarini: absolutely
ac236c38b

include/csoundCore.h
vlazzarini: these are converter modes for in and out instead of a single overmode. Labelled for reference.

include/csoundCore_new.h
vlazzarini: removed

tests/c/CMakeLists.txt
vlazzarini: added back.

tests/c/cscore_tests.cpp
vlazzarini: done

include/cs_threads.h
vlazzarini: updated doxyfile

kunstmusik:
I think we're in good enough shape. I'm going to go through to check wasm build and then, assuming all's well, will merge. Will work on updating CsoundFFM (java binding) for newapi after that.

vlazzarini:
Sounds good to me. Thanks for reviewing.

kunstmusik (approved):

kunstmusik:
Did some basic live coding tests with wasm csound and things looked good for API. Merging now, well done everyone!

PR #1934, opened 2024-09-07
fix for phase modulo issue

vlazzarini:
I noticed that the phase modulo 1 macro was used incorrectly in some places as it assumed a single value input. I improved the code making it an inline function which is now correct for all cases.


kunstmusik (approved):

kunstmusik:
Straightforward change, merging now.

vlazzarini:
thanks

PR #1936, opened 2024-09-09
fix: set install RPATH for built executables on macos to use Framework install directory

kunstmusik:
This change fixes an issue for locally-built csound where LC_RPATH was cleared after running `make install`. This allows running csound and other executables from /usr/local/bin


kunstmusik:
@vlazzarini Requesting a review of this. This fixed up the issue here I was seeing where I could not run Csound from anywhere except the build directory.

vlazzarini (approved):

Issue #717, opened 2016-11-28
Update Build.md


berntisak:
Build.md needs to be updated at least for OS X, which only has Homebrew instructions (that probably should be removed as long as it's not updated?)

nwhetsell:
Seconded, and even better (perhaps) would be incorporating the build instructions into README.md.

kunstmusik:
Assigning to 7.0.0 milestone so that we have visibility on this and ensure we take care of it before we release CS7.

vlazzarini:
this is done now in #1940, a fresh set of instructions for the basic platforms. Further contributions are welcome, but I will close this ticket now.

PR #1940, opened 2024-09-11
New Build Instructions

vlazzarini:
Starting new build instructions #717. Please contribute to it.


vlazzarini:
@rorywalsh I think this is in good shape to merge. Could you review/approve it? I'll wait for the actions to complete then merge once it's approved.

rorywalsh:
This looks good Victor, I wonder if git should be mentioned somewhere at the top of the page? I mention its use in the Windows instructions, so I guess it could be mentioned elsewhere too? Also, can I remove all but the custom cmake file in platform/windows? I think leaving those scripts there is a little confusing as they are no longer needed. I'm happy to make the changes, just want to get a heads up first.

vlazzarini:
I think so, you can do both things.

rorywalsh (approved):
I think this is cleaner now with those files removed and build instructions for the 3 major OSs consolidated into a single file.

Issue #728, opened 2016-12-12
Array variables should not be allowed to have types redefined


kunstmusik:
From the Csound users mailing list discussion, there is a problem in that the parser allows variables to be defined as arrays a second time, ignoring the array definition and using the already registered type for the variable name.  The following CSD highlights the problem:

```csound-csd
<CsoundSynthesizer>
<CsOptions>
</CsOptions>
; ==============================================
<CsInstruments>
sr = 48000
ksmps = 1
;nchnls = 2
0dbfs = 1
instr 1
karr[] fillarray 1,2,3,4
karr[] = karr + 4
kindx = 0
while kindx < 4 do
  printk2 karr[kindx]
  kindx += 1
od

endin

instr 2
karr init 1
karr[] = karr + 4
printk2 karr

endin
</CsInstruments>
; ==============================================
<CsScore>
i1 0 0.001
i2 0.1 0.001
</CsScore>
</CsoundSynthesizer>
```

The above operates, but it probably shouldn't. In instr 2, the karr[]
in the second line functions as just karr, which is defined first in
the line above it as a k-rate scalar value.  The second line is
misleading in instr2, and I would imagine it *should* be considered a
bug to redeclare a variable's type a second time for arrays (and
explicitly-typed vars in CS7).  This would be in-line with errors that
would be reported in C-like languages, such as:

```
int x = 4;
int x = 5;  // ERROR
```

nwhetsell:
@kunstmusik This is off-topic, but if you write

````markdown
```csound-document
<CsoundSynthesizer>…
```
````

or

````markdown
```csound-csd
<CsoundSynthesizer>…
```
````

instead of

````markdown
```
<CsoundSynthesizer>…
```
````

then [GitHub Linguist](https://github.com/github/linguist) will highlight it for you:

```csound-csd
<CsoundSynthesizer>
<CsOptions>
</CsOptions>
; ==============================================
<CsInstruments>
sr = 48000
ksmps = 1
;nchnls = 2
0dbfs = 1
instr 1
karr[] fillarray 1,2,3,4
karr[] = karr + 4
kindx = 0
while kindx < 4 do
  printk2 karr[kindx]
  kindx += 1
od

endin

instr 2
karr init 1
karr[] = karr + 4
printk2 karr

endin
</CsInstruments>
; ==============================================
<CsScore>
i1 0 0.001
i2 0.1 0.001
</CsScore>
</CsoundSynthesizer>
```

You can also use

````markdown
```csound
[...]

PR #1797, opened 2023-12-21
Changes to allow building wasm csound using emscripten, cmake and vckpg

nonameentername:
Changes to allow building wasm csound using emscripten, cmake and vckpg

PR #1942, opened 2024-09-12
Fixing copy of arrays issue in #728

vlazzarini:
This PR address one half of isse #728, where the generic copy of variables was mixed up.
The following code is a test case, it should increment from an initial set of values

```
<CsoundSynthesizer>
<CsOptions>
</CsOptions>
; ==============================================
<CsInstruments>
sr = 48000
ksmps = 10
;nchnls = 2
0dbfs = 1
instr 1
karr[] fillarray 1,2,3,4
karr[] = karr + 4
kindx = 0
while kindx < 4 do
  printk  0,karr[kindx]
  printk  0, kindx
  kindx += 1
od
endin

</CsInstruments>
; ==============================================
<CsScore>
i1 0 0.001
</CsScore>
```

The PR includes two fixes: (a) to  =.generic opcode (b) to k[] op k opcodes where op is -, * and / . The + case was already correct.


vlazzarini:
I have now fixed the other issue in #728  of a variable being redefined as an array. This required a check to be added to the parser so that a new array with the same name as an existing variable cannot be created.

The test:

```
<CsoundSynthesizer>
<CsOptions>
</CsOptions>
; ==============================================
<CsInstruments>
instr 2
karr init 1
karr[] = karr + 4
printk2  karr
endin
</CsInstruments>
; ==============================================
<CsScore>
i2 0.1 0.001
</CsScore>
```
now gives the error

```
error:  Array variable name 'karr' used before as a different type
Line 8
 from file arrdef.csd (1)
Could not parse type for argument karr
cannot compile orchestra
```

kunstmusik:
@vlazzarini Could you add the test csd files to the tests/commandline folder and add to test.py there so we have this as part of our test suite?

vlazzarini:
How do we test for an expected error? Not sure if I know how to add these two examples.

vlazzarini:
The failures have to do with opcodes not being found. I managed to reproduce this locally and the solution seemed to be to fix the tests to find the opcodes correctly and to build plugins. But it is still failing.

vlazzarini:
I see tests are failing, I need to go back to this now.

vlazzarini:
Added the tests above.

Engine/csound_orc_semantics.c
```diff
@@ -1212,10 +1223,10 @@ char* get_arg_string_from_tree(CSOUND* csound, TREE* tree,
 
   while (current != NULL) {
     char* argType = get_arg_type2(csound, current, typeTable);
-    //FIXME - fix if argType is NULL and remove the below hack
+
     if (argType == NULL) {
-      argsLen += 1;
-      argTypes[index++] = cs_strdup(csound, "@");
+      // if we failed to find argType, exit from parser
+      csound->Die(csound, "Could not parse type for argument %s", current->value->lexeme);
```
kunstmusik: I think this will not work as a solution as it would make live coding very difficult if one makes a type mistake.

vlazzarini: it just a does a long jump to orcparse as far as I can see and returns from there. What would be the problem?

vlazzarini: I looked and on error it should jump back to csoundCompileOrcInternal() and return an error from there. Csound should carry on performing. Line 2017 of csound_orc_compile.c is where it jumps to.

"Die()" is the wrong name for that function, it just prints an error and jumps to the last point set.

kunstmusik:
I think the WASM build does not support setjmp/longjmp but I need to check.
Perhaps I was overreacting to the "Die" :)

vlazzarini:
I wonder what happens then. There are several points where control is passed from an inner point outwards when there is an error. CsoundCompile, argdecode, etc, all use that. The thought came into my head earlier. Maybe it just continues onwards, instead of jumping back.

vlazzarini:
Emscripten seems to support it. Do we use emscripten or something else?

https://emscripten.org/docs/porting/setjmp-longjmp.html#:~:text=WebAssembly%20Exception%20Handling%2Dbased%20setjmp%2Dlongjmp%20Support,-Alternatively%2C%20you%20can&text=To%20enable%20it%2C%20pass%20%2DsSUPPORT_LONGJMP,and%20catching%20exceptions%20to%20WebAssembly.

kunstmusik:
Uses clang + wasi, but an older version. Might be updating would get setjmp/longjmp but I'm not certain. (Paging @hlolli as he may have more information on this.)

vlazzarini:
I don't see any compile time guards for setjmp/longjmp. How can the code compile if these are not available on wasi?

vlazzarini:
Looking at code in Engine, there are lot of instances of csound->LongJmp() in the parser (mainly in the preprocessor lexer, but also in new_orc_parser.c). If this does not work in wasi, then we're in trouble. 

This is Csound 6 code though, so I can't see there is a problem. Should I rename csoundDie to something less dramatic like csoundRaiseException?

kunstmusik:
csoundRaiseException might be a better name, but I'm hesitant on that as setjmp/longjmp has differences to exceptions in something like C++. I have not worked with setjmp/longjmp deeply so am not the best to comment. If we change the name though let's do it in another PR. 

You can see the patching of setjmp/longjmp in the wasm build here in csound.nix:

https://github.com/csound/csound/blob/bugfix/arraycopy/wasm/src/csound.nix#L147-L153

I looked around and couldn't find definitive information on support for setjmp/longjmp in wasi, but I suspect it's possible. Let me investigate further and get back to you on it. I think what I'll do is try this branch out live coding on desktop and if all works we can merge and move to investigating setjmp/longjmp for wasi build separately. 

Also to note, there's the longstanding PR #1797 that needs to be considered which gets Csound building again with Emscripten. @hlolli and I were going to evaluate and see how optimized the Emscripten build is is compared to wasi and make some decisions around that.

kunstmusik:
I tested here and get a segfault immediately when trying to live eval code:

```
schedule(1, 0, 2, cpsmidinn(0 + iroot), 0.25)
```

iroot isn't even an array variable but it gets:

```
error:  Array variable 'iroot' used before defined
Line 16

error:  Unable to verify arg types for expression '##add'
Line 17


csound command: Segmentation fault
                   overall amps:  0.00000  0.00000
           overall samples out of range:        0        0
```

So we need a different fix.

kunstmusik:
Should not when I tested the above and got the segfault, I had not yet defined iroot.

vlazzarini:
I get a different result with not defining the variable.

code:

```
schedule(1, 0, 2, cpsmidinn(0 + iroot), 0.25)

instr 1
endin
```

Result

```
error:  Variable 'iroot' used before defined
Line 4

 from file testas.csd (1)
error:  Unable to verify arg types for expression '##add'
Line 5

 from file testas.csd (1)
error:  error: opcode 'cpsmidinn' for expression with arg types @ not found, line 5

 from file testas.csd (1)
error:  Unable to find opcode entry for 'schedule' with matching argument types:

Found:
  (null) schedule ccc@c

Candidates:
   schedule iiim
   schedule.N iiiN
   schedule.S Siim
   schedule.SN SiiN
   schedule.array i[]

Line: 5
 from file testas.csd (1)
Stopping on parser failure
cannot compile orchestra
		   overall amps:      0.0
1 errors in performance
````

**Comment by vlazzarini** [comment]
I will revert the longjmp as it is causing the segfault. Without it, the code exits without crashing.

**Comment by vlazzarini** [comment]
I run it in the debugger and the problem was the message printing looked for a lexeme in a non-existing value of a tree, and so it crashed. Removed that and the crash does not happen. Can you test again? It was nothing to do with longjmp.

**Comment by kunstmusik** [comment]
It doesn't crash now, but it still doesn't make sense to me why it's going through that branch of code for T_ARRAY_IDENT:

```
error:  Array variable 'iroot' used before defined
```

I think there's still another matter to be solved here.

**Comment by vlazzarini** [comment]
Ok, at least now we know the exit from that function with longjmp is OK. I'll check why the code is going through that branch.

**Comment by vlazzarini** [comment]
It's picked up here

```
    if (UNLIKELY(var == NULL)) {
      synterr(csound, Str("Array variable '%s' used before defined\n"
                          "Line %d\n"),
              tree->value->lexeme, tree->line - 1);
      do_baktrace(csound, tree->locn);
      return NULL;
    }`
```

I think the message is just wrong.

**Comment by vlazzarini** [comment]
While we're at that, when this error occurs, we get the "could not find type for variable" too. I suppose this is fine, but if we exit from that earlier point we just get one message.

```
victor@firebird debug % csound testas.csd -n             
UnifiedCSD:  testas.csd
error:  Variable 'k1' used before defined
Line 4
 from file testas.csd (1)
syntax error
cannot compile orchestra
		   overall amps:      0.0
2 errors in performance
no sound written to disk
```

It looks cleaner to me.

kunstmusik:
FYI: @vlazzarini @hlolli @nonameentername 

I investigated the Emscripten build to see how it fares these days. I was able to build on Linux (macOS had some problems) and it seems that the wasm build was 40mb uncompressed. I tried some flags like -Oz and -no-rtti to try to get it smaller and got it down to 30mb. In contrast, the nix build with clang + wasi is 5mb uncompressed, and compresses down to 1.4mb. 

I think the size difference is significant. The current situation though is that the nix build is very complicated to maintain. It does "just work" once it's setup, but it's been a little tricky to update clang + wasi in the nix files. I think it's still worth pursuing to see if the updated wasi can work with setjmp, so will continue to see about updating those.

vlazzarini:
What happens in wasi if a longjmp is reached? There's lots of those in the codebase.

kunstmusik:
I put a link to what the nix build does earlier: 

https://github.com/csound/csound/blob/bugfix/arraycopy/wasm/src/csound.nix#L147-L153

It preprocesses the code and replaces it with a no-op or return statement (depending on context of longjmp)

kunstmusik:
I think at this point this PR is in good enough shape to merge and we can address the issue with setjmp/longjmp and WASM separately as it's a longstanding issue.

kunstmusik (approved):

PR #1943, opened 2024-09-12
fixing fill statement

vlazzarini:
Fixing fill statement so it does not raise a warning.


rorywalsh (approved):
Thanks Victor, this was tripping up a frontend project on Windows where warnings were treated as errors.

Issue #1801, opened 2024-01-01
multiple assignments not working in CS7


joachimheintz:
this works in CS6 but not in CS7 (develop):
`iMin,iMax = 5,10`

not sure it is already expected to work but want to mention it for compatibility.

kunstmusik:
I was not aware of this syntax in CS6. Is this commonly used? I'm wondering about the long-term language design implications of this and whether it'll cause issues in the future. 

Would love to hear @jpffitch and @vlazzarini 's thoughts on this.

vlazzarini:
I wasn't aware either. Looks like tuples to me but I never knew we supported them in the old parser.

joachimheintz:
just to note that it is documented in the manual:
https://csound.com/docs/manual/assign.html

vlazzarini:
well then, looks like we need to keep this in the new parser. This goes back to 2010.

joachimheintz:
i just checked whether multiple assignments work for init rather than 
for equal.
good news that it already works also in CS7, for instance:

instr 1
   ia,ib init 1,2
   print(ia,ib)
endin

not sure if this helps to implement it for `ia,ib = 1,2`, too.

jpffitch:
I looked more carefilly and multiole assinment iin cs6 is just the usual call of an opcode.  There is no fuss about diskin2 have multiple outputs and more than 1 input. so there is issue in =or init,. The syntax is just like elsewhere
-  { "=.i",    S(ASSIGNM),0, 1,      "IIIIIIIIIIIIIIIIIIIIIIII", "m",
    minit, NULL, NULL, NULL  },
  { "=.k",    S(ASSIGNM),0, 2,      "zzzzzzzzzzzzzzzzzzzzzzzz", "z",
    NULL, minit, NULL, NULL },
In the parser3 however = is treated specially and so needs a rewrite to match the simplicity of cs6

vlazzarini:
This is now fixed in PR #1944

PR #1944, opened 2024-09-12
fixing #1801

vlazzarini:
This PR fixes the issue regarding multiple assignments, where backward compatibility with csound 6 was lost. Test3.csd in the test suite is now passing.

```
<CsoundSynthesizer>

<CsInstruments>
;
sr=44100
ksmps=1
nchnls=2


	instr 1	;untitled

iamp,ifreq = 10000,440

aout	vco2 iamp, ifreq

	outs aout, aout
	endin


</CsInstruments>

<CsScore>

i1	0.0	2	
e

</CsScore>

</CsoundSynthesizer>

```

### DIFF
--- a/Engine/csound_orc.y
+++ b/Engine/csound_orc.y
@@ -386,7 +386,7 @@ statement_list : statement_list statement
                   }
                 ;
 
-statement : out_arg_list assignment expr NEWLINE
+statement : out_arg_list assignment expr_list NEWLINE
                 {
                   $$ = (TREE *)$2;
                   $$->left = (TREE *)$1;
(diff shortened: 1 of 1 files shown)


kunstmusik:
Makes sense but I do wonder if there are other issues that might arise from this. Think better to merge and monitor for other issues that could arise. Merging now, thanks!

kunstmusik (approved):

kunstmusik:
@vlazzarini I think I jumped the gun on this one. I just built locally and test3.csd still fails, so I think there's more work to be done.

kunstmusik:
Nevermind, there was a linking problem on my computer and my build linked with an older version of CsoundLib.framework. I reinstalled and saw that the code does indeed work now. Sorry for the noise!

Issue #1886, opened 2024-05-20
Regression: newline in function syntax


vlazzarini:
Reporting here a regression in the parser. The following code parses and runs on 6.18 but fails on
7.0 beta

```
instr 1
  kfreq = p5 
  kdet = p6 
  a1 = vco2(p4, kfreq * (1+kdet)) +
  vco2(p4, kfreq * (1+kdet*2))
  out a1
endin 
```

This is the error message

```
error: syntax error, unexpected NEWLINE  (token "
") from file test-newline.csd (1)
 line 12:
>>>  a1 = vco2(p4, kfreq * (1+kdet)) + <<<
Parsing failed due to 1 syntax error!
```

So the new parser cannot deal with newlines in an expression.

vlazzarini:
Fix has been submitted as a PR: #1946

PR #1946, opened 2024-09-12
Fixing Regression: newline in function syntax #1886

vlazzarini:
As the title says, this PR fixes the issue where newlines after binary ops (+ etc) were not being parsed correctly as they did in Csound 6.


kunstmusik:
@vlazzarini I made a change to use a new optnewline rule to simplify the change you did earlier a bit. (Less repetition.) Looks to work here and tests are passing. If you're okay with this change here, I think we can merge.

kunstmusik (approved):

vlazzarini:
Thanks!

Issue #1894, opened 2024-06-11
cs7: cmake generations fails without "-DUSE_GETTEXT=0"


DaveSeidel:
This is on Raspberry Pi (raspian buster). If I don't add this option to the cmake command line, I get

```
CMake Error at CMakeLists.txt:1416 (add_library):
  Target "csound64-static" links to target "Intl::Intl" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?


CMake Error at CMakeLists.txt:1198 (add_library):
  Target "csound64" links to target "Intl::Intl" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

vlazzarini:
Thanks. A PR has been submitted to fix this.

PR #1948, opened 2024-09-12
turning get text OFF by default to fix build issues #1894

vlazzarini:
Making gettext OFF by default to avoid build issues in platforms where it cannot be found correctly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -178,7 +178,7 @@ option(BUILD_UTILITIES "Build stand-alone executables for utilities that can als
 option(NEW_PARSER_DEBUG "Enable tracing of new parser" OFF)
 option(BUILD_MULTI_CORE "Enable building for multicore system" ON)
 option(FAIL_MISSING "Fail when a required external dependency is not present (useful for packagers)" OFF)
-option(USE_GETTEXT "Use the Gettext internationalization library" ON)
+option(USE_GETTEXT "Use the Gettext internationalization library" OFF)
 option(BUILD_STATIC_LIBRARY "Also build a static version of the csound library" OFF)
 option(USE_LRINT "Use lrint/lrintf for converting floating point values to integers." ON)
 option(USE_CURL "Use CURL library" ON)


kunstmusik:
I don't think turning this off by default is a good way to solve this issue. What platform had a problem with gettext? Seems better to me to fix the CMake code we have for finding gettext.

vlazzarini:
It was on raspberry PI (see issue https://github.com/csound/csound/issues/1894). I can't test it so I thought it was best to shut it down and if anyone wants to use gettext on RPi, they can submit a PR.

rorywalsh:
With regards to #1948 I wonder if @DaveSeidel tried installing gettext ala `sudo apt-get install gettext` in order to resolve the issue? If so, and there is still an issue, perhaps we can disable for RPi whilst we look into it? 

```cmake
if(LINUX)
    if(CMAKE_SYSTEM_PROCESSOR MATCHES "armv7l|armv6l|aarch64")
        #assume OS is RPi
        option(USE_GETTEXT "Use the Gettext internationalization library" OFF)
    endif()
endif()
```

rorywalsh:
Or as Steven put it in another discussion, Csound "should gracefully not use gettext if it's not found."

vlazzarini:
> Or as Steven put it in another discussion, Csound "should gracefully not use gettext if it's not found."

yep, that's the idea, but how can we test on RPi?

rorywalsh:
Am I right in assuming one would see the same issues reported in #1894 on any system where gettext is not present? Therefore we don't need a RPi to test?

vlazzarini:
I turned USE_GETTEXT=1 here and I had no issues with it. It just did not find it and it wasn't switched on.

vlazzarini:
I installed gettext and turned it on, it's configured OK. So I am not so sure what the problem is.

DaveSeidel:
Just test it on the RPi using the latest develop. Apparently I already have
gettext installed

gettext is already the newest version (0.19.8.1-9)

And I still get the same failure unless I use -DUSE_GETTEXT.

Now I have another issue:

CMake Error at tests/c/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.20 or higher is required.  You are running version 3.16.3

This device is still on Raspian buster, and apparently I have the latest
Cmake that's available for this release. I suppose I will have to upgrade
my Raspian version to get the new version, which is a little scary but
shouldn't be too bad.

rorywalsh:
Hi Dave, U nderstand your reluctance to upgrade your syste, especially if you have everything working the way you like it. Can you not clone the cmake repo and build yourself locallay to get the latest version?

vlazzarini:
> Just test it on the RPi using the latest develop. Apparently I already have gettext installed gettext is already the newest version (0.19.8.1-9) And I still get the same failure unless I use -DUSE_GETTEXT. Now I have another issue: CMake Error at tests/c/CMakeLists.txt:1 (cmake_minimum_required): CMake 3.20 or higher is required. You are running version 3.16.3 This device is still on Raspian buster, and apparently I have the latest Cmake that's available for this release. I suppose I will have to upgrade my Raspian version to get the new version, which is a little scary but shouldn't be too bad.

I think the problem may be related to CMake versions. The failing line is perhaps depending on things that are only present in > 3.20. When you set USE_GETTEXT=0, that's bypassed. I think the minimum version was changed when something like that failed for someone and it was realised that CMake min version needed to be bumped.

DaveSeidel:
I'm in the process of upgrading to bullseye now (fingers crossed). Assuming
I still have an operational system when it's done (should be soon), I will
test again.

DaveSeidel:
I've updated to bullseye. Cmake was still at the earlier version, so I am
building Cmake 3.30.3 from source. Assuming that works, I will try the
Csound build again.

DaveSeidel:
I haven't gotten CMake to build yet. Some linker error that I don't have
time to figure out right now.

DaveSeidel:
No luck with CMake yet. The latest Raspbian version is bullseye, and the
CMake available for bullseye is 3.18.1. Do we really need Cmake 3.2+?

vlazzarini:
You can edit the requirement in CMakeLists.txt, build and report to us. If it passes then we can see to bring it down to that version.

DaveSeidel:
I changed it from 3.20 to 3.18, and now I get the error:

CMake Error at tests/c/CMakeLists.txt:11 (add_executable):
  Target "unittests" links to target "GTest::gtest_main" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?

It seems odd that the main CMakeLists.txt has a minimum version of 3.13.4
(except Apple), and it's just the tests that require 3.20

DaveSeidel:
I can work around it by commenting out "add_subdirectory(tests/c)" in the
main CMakeLists.txt. No gettext errors. So I guess this validates
the gettext fix.

vlazzarini:
It does look weird; there were a lot of changes made to these files which I thought were unwarranted. 

The Intl::Intl issue, is that ok with 3.18? I mean if you enable gettext, does it build?

DaveSeidel:
Yes, built without errors once I excluded the tests.

vlazzarini:
ok, I'll look at the tests then, see what can be done.

vlazzarini:
Other people had the BUILD_TESTS issue on Debian. I put a PR with a fix, so that

- if we are not building tests, BUILD_TEST=0 will now avoiding looking for gtest package (and fail on Debian)
- if we are not building tests, BUILD_TEST=0 should allow you to build Csound without needing CMake version to be 3.20 or higher.

The PR is https://github.com/csound/csound/pull/1957

PR #1949, opened 2024-09-12
Hotfix/tests

vlazzarini:
Trying to enable csdtests failure.


CMakeLists.txt
```diff
@@ -178,7 +178,7 @@ option(BUILD_UTILITIES "Build stand-alone executables for utilities that can als
 option(NEW_PARSER_DEBUG "Enable tracing of new parser" OFF)
 option(BUILD_MULTI_CORE "Enable building for multicore system" ON)
 option(FAIL_MISSING "Fail when a required external dependency is not present (useful for packagers)" OFF)
-option(USE_GETTEXT "Use the Gettext internationalization library" ON)
+option(USE_GETTEXT "Use the Gettext internationalization library" OFF)
```
kunstmusik: Not sure this should be set to OFF by default. It should gracefully not use gettext if it's not found.

kunstmusik:
Note: This is now failing correctly and exposes that our macOS CI builds have problems running tests.

vlazzarini:
yes, both brew and VCPKG CI have problems.

vlazzarini:
The failures have to do with opcodes not being found. I managed to reproduce this locally and the solution seemed to be to fix the tests to find the opcodes correctly and to build plugins. But it is still failing.

vlazzarini:
OK this seems to have been fixed now.

CMakeLists.txt
kunstmusik: @vlazzarini can you revert this change for this PR as it doesn't have to do with fixing the tests? We can track it in the other PR.

vlazzarini: done

kunstmusik (approved):

PR #1950, opened 2024-09-13
Fix build RPATH for macOS

kunstmusik:
Removes use of CMAKE_BUILD_WITH_INSTALL_RPATH which made the built executables link set RPATH to the install RPATH. Removing this makes the RPATH set to the build folder version of CsoundLib.framework and only sets it to the CMAKE_INSTALL_RPATH when `make install` is called.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,7 +114,6 @@ else()
 endif()
 
 if(APPLE)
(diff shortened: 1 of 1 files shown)


kunstmusik:
@vlazzarini This was the cause of my bad testing for #1944 as testing the built csound in my local folder actually linked at runtime against a prior build of CsoundLib.framework which I had installed to the system.

vlazzarini (approved):

PR #1953, opened 2024-09-16
attempt to prioritize function calls over expressions with parentheses

### LINKED ISSUE #1697: k() in cs7 not working as in cs6
for sending a constant value, say 1, to a k-channel in csound 6, i had to use k(1).
this is not any more working in cs7 develop (--Csound version 7.0 (double samples) Mar  9 2023, [commit: 2503fc0b439f70973a6ee9fdd810fda6d111882c]).

the csd below returns this error:

> error:  error: opcode 'k' for expression with arg types cS not found, line 14
> from file test.csd (1)
> error:  Unable to find opcode entry for 'chnset' with matching argument types:
> Found:
>  (null) chnset @
> Candidates:
>   chnset.i iS
>   chnset.k kS
>   chnset.a aS
>   chnset.S SS

in my opinion, it would be good to have both possibilities as legal, either
  chnset 1, "bla"
or
  chnset k(1), "bla"

```csound
<CsoundSynthesizer>
<CsOptions>
-o dac
</CsOptions>
<CsInstruments>
sr = 44100
ksmps = 64
nchnls = 2
0dbfs = 1

chn_k "bla",2

instr 1
 chnset k(1), "bla"
endin

</CsInstruments>
<CsScore>
i 1 0 1
</CsScore>
</CsoundSynthesizer>
```

scottsitar:
This is an attempt to fix #1697 .  Not for immediate merging - just to suggest a path forward.

It reintroduces the `T_IDENTB` style tokens from cs6 to coerce the parser to reduce an expression of the form

`ident '(' expr ')'`

into a function call rather than an opcode followed by an expression in parentheses.  It *almost* works.  With this change, there are two csd test failures:

```
[FAILED TESTS]

[FAIL] - Test 48: test labels in instr0 space (test_instr0_labels.csd)
	Return Code: 1	Expected: 0
[FAIL] - Test 89: test declare keyword (CS7) (test_declare.csd)
	Return Code: 1	Expected: 0
```

The first looks like trouble with the `if` token, and I'll have to defer to someone more familiar with UDOs than me to say whether this would be tweakable to work.


vlazzarini:
Thanks for this. Test 48 seems to fail because if ( ) is treated as a function instead of an IF statement.

```
if(i1<10) goto lp
```

resulting in 

```
>>>i1<10) goto <<<
```

Removing the parenthesis (which are not necessary but should not break the code either) fixes the problem.

I see it is being picked up here

```
             | identifierb expr_list ')'
              { $$ = $1;
                $1->type = T_FUNCTION;
                $1->right = $2; }
```

with the 'if' as identifierb.

scottsitar:
Ah yes, I see, thanks!  It looks like the udo case is similar, where the the start of it is being picked up by identifierb rather than `identifier udo_arg_list`.  I suspect both could be modified to make it work, but can you think of any other cases that this might have broken that wouldn't be covered by the tests?

vlazzarini:
Hard to say. I think first stop is to get all tests passing, then we can reassess. Steven would have a better idea.
This is a tricky one. The two other known parser issues (multiple assignments, newline in expressions) were easy fixes.

scottsitar:
Sorry for the delay!  I've just pushed up a further commit which takes care of the failing "if()" test case in a somewhat clean way.  It takes advantage of flex's trailing context rules to prioritize things like "if(" as IF_TOKEN followed by a parenthesis rather than an IDENTB.

I'll get to work on the UDO test case failure now.

scottsitar:
Okay, I fixed up the test_declare test failure by adding another flex start condition for declare statements.  With that, all tests now pass, and I've added a regression test into the commandline tests with the snippet from the initial bug report.

kunstmusik:
Thanks @scottsitar for this. I  do want to say I would rather find a different solution than using IDENTB as this starts to bring back a lot of noise to the parser definition from CS6. However, if we have to go this route then we will. I need a little time to review and experiment to see if there's other paths to take.

scottsitar:
Okay, thanks, sounds good!  If I end up trying out any other ideas for a fix I'll put it in a separate PR, just so this can stay here as a reference/potential fix.  When I previously tried doing it without the IDENTB tokens it always led to other unexpected parses, and the main problem I always seemed to hit was not being able to remember where the parentheses had been once we got to convert_statement_to_opcall().  Another interesting difference would be that with the IDENTB token, there is not allowed to be a space between the function name and the opening parenthesis (like in cs6), but with separate tokens, having a space in between could also allowed by the parser.

vlazzarini:
I think I might have uncovered another parsing error that could be related to this bug: see #2013

scottsitar:
I tried out your printk2 example:

```
instr 1
k1 = 1
k2 = 1
k3 = 2
k4 = 3
printk2 (k1+k2)*k3 + k4
endin
```

on this branch, and it appears to give the correct answer (7) rather than the misparse (10).  I haven't had time to dig into exactly how or why the expression seems to be getting split at the multiplication (or in fact that this branch is getting the right answer for the right reason!), but I think you're right.  The two issues do appear to be related.

I'll hopefully have time this weekend for a more detailed analysis, which I can leave in the other issue so that it's closer to the problem report.

vlazzarini:
If you run csound -v ... you will be able to see the synthetic expression. You can compare it to the wrong one I give in the issue.

scottsitar:
Very briefly, it looks like the parsing of `printk2 (k1+k2)*k3 + k4` is going something like this:

- first printk2(k1+k2) gets reduced into a function_call (function = printk2, args = k1+k2)
- then the `*` is shifted onto the parser stack
- then each of k3, `+`, and k4 are shifted onto the parser stack (so, it seems like it's not getting the relative precedence of `+` and `*` correct here)
- then it reduces `k3+k4` to an expr
- this leaves it with `function_call * expr`

Then, something happens which I don't understand yet (`make_opcall_from_func_start()`, I suppose?) which converts what it's got above:

`printk2(k1+k2) * (k3 + k4)`

into this:

`printk2( (k1+k2) * (k3+k4) )`

I haven't definitively found where in the code this happens yet, although I can certainly imagine what that transformation is trying to accomplish.  But yes, at the root, there seem to be two issues:

- mistakenly reducing `printk2(k1+k2)` as a complete function call (cs6 gets this right because it insists on no spaces between the function name and the opening parenthesis)
- not recognizing the relative precedence of `+` and `*` during the middle of the parse (I suspect because of the `opcall : function_call '*' expr_list` production, so that it doesn't have enough context to properly apply the precedence rules)

vlazzarini:
Thanks for the investigation @scottsitar .

Am I right in saying that of the two issues you list at the end, if the first one is resolved, the second one doesn't arise (in the same way that in the previous forms of this bug, the mis-parse of the arguments, #1697, gets resolved if the function parsing is right)?

If so, then the focus continues to be solving that. Do you think there may be another way to resolve it without reintroducing the rule from the old parser, since @kunstmusik thinks that's problematic?

scottsitar:
I think that's a fair assessment, @vlazzarini .  Recognizing function starts vs an "opcode ( expr )" combination seems to be the common thread here.  I think the precedence level is an independent issue that will still exist after fixing the function start issue, though, since the grammar is organized in a way where a shift/reduce conflict will not actually arise.  The function call doesn't seem to get reduced to an expr before the rest of the expression gets shifted onto the parser stack, meaning that bison won't have a chance to apply the precedence values to resolve the conflict.  I haven't tried, but I suspect I could invent other expressions that will also demonstrate this behaviour outside the context of the original bug (maybe something like `cpspch(p4)*2 + 1`?  I can try after work today).

To be honest, I'm not sure how to properly differentiate between function starts and opcode/expr combinations without using the intervening whitespace.  Since the lexer won't return that whitespace to the parser, I'm not sure if it can be fixed directly with grammar and post-processing transformations.

scottsitar:
After playing around a bit more, I can confirm that the precedence issue only occurs when the statement being parsed starts with the ambiguous function_call vs opcode/expr combination.

kunstmusik:
Sorry for the delay, I'll be catching up and testing some things here now. Hope to provide some feedback soon.

kunstmusik:
@scottsitar @vlazzarini Sorry I've been very slow on this! 

I think at this point, we'd probably be better off merging this in and having the expected behavior for Csound parsing, and we can always revisit to see if there's a different path we can take later. If that works for everyone, I think we'd need this PR to have the latest from develop merged in before we can test and merge back into develop.

scottsitar:
@kunstmusik No worries!  What's the preferred way to get this up to date with develop?  Would you prefer that I make an additional commit to merge develop into this branch and push that up, or do a rebase locally against an up-to-date develop branch and then do a force push to update this PR branch on github?  I'm happy either way.

vlazzarini:
I normally do a merge from develop into it. Since this involves only a few files it should be generally pain free. A rebase would work as well I guess. Looks like the only conflict is in tests.py

scottsitar:
Sounds good.  I've brought in the latest develop branch and verified that everything is still working as expected.

kunstmusik:
Thanks @scottsitar, going to merge this in now.

kunstmusik (approved):

PR #1954, opened 2024-09-18
bug fixed in csoundInitArrayChannel

fggp:
This PR request fixes the array_channel size calculation when the channel is created from the host API by csoundInitArrayChannel


vlazzarini:
I think this has already been merged in develop. Can you check?

fggp:
It's the same bug but it was in chn_opcode_init_ARRAY. This one is in csoundInitArrayChannel. Depends if the array channel is created by the opcode chn_array or from the host API.

vlazzarini (approved):

PR #1955, opened 2024-09-19
Bugfix/startup

vlazzarini:
This PR fixes a startup issue where Csound failed to run if supplied with `--port=N` and no code to compile. The fix checks for a code file, and if it does not find it, Csound starts as normal.
If the compilation fails, then Csound exits, unless  the option `--daemon` or `--port=0` (daemon mode) is used.

A test may be added, but we would need to adapt the test script so it can close Csound from an external signal.

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -14,7 +14,7 @@ Optional component requirements:
 - Git: A distributed version control system to let you clone and acquire
 the Csound sources. Although you can download the source as a zip from GitHub,
 having git installed makes it easier to interact with the Csound source. It
-also makes it easier to sync with the very latest source.  
+also makes it easier to sync with the very latest source.
 
 - C++ compiler toolchain: a few components are written in C++. For
 these to be built, a complete installation of C++ development tools is
--- a/BUILD.md
+++ b/BUILD.md
@@ -100,7 +100,7 @@ and then build Csound.
 ```
 brew install --only-dependencies csound
 brew install bison flex asio jack googletest
-cmake -B build -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake
+cmake -B build -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake"
 cmake --build build --config Release
 ```
 
--- a/Frontends/csound/csound_main.c
+++ b/Frontends/csound/csound_main.c
@@ -39,6 +39,8 @@
 extern int set_rt_priority(int argc, const char **argv);
 #endif
 
+extern int csoundPerform(CSOUND *);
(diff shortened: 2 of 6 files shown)


vlazzarini:
Also fixed the Windows issue where exit was not clean, because of missing exception handling.

stekyne (dismissed):

kunstmusik:
@vlazzarini I did a test of latest on Windows and am still getting a segfault after ctrl-c. Will investigate further.

vlazzarini:
ok!

kunstmusik:
Still diagnosing the windows segfault on ctrl-c. I found that commit 58b687949 is a "good" commit and that there was no segfault there when using ctrl-c. I tried doing a bisect session between that and this branch's latest commit but the session wasn't fruitful as things did not build for a lot of the commits I went through. My concern is that in updating to the new API and removing csoundStop() and csoundCleanup() that something was lost. From debugging, It looks to me that ctrl-c triggers then the app segfaults as kperf keeps running even though csoundCleanup is run. Will need to do further triage.

vlazzarini:
I will look at it. I am pretty sure both functions are redundant. It may be that I need to add some treatment of signals to the frontend for the Windows case.

UPDATE: I noted that the difference was that the signal handler sets in an explicit flag (through csoundStop()) to stop the csoundPerform(). So I reverted my code and now set a flag to get the loop exited before invoking csoundDestroy(). csoundStop() is only there for csoundPerform(), it's non-op for other cases. Pushed 2efab4690..

vlazzarini:
@stekyne could you try this branch on Windows and test if Csound exits cleanly from the command line on ctrl-C?

stekyne:
@vlazzarini I'm afraid I only have my macbook until next Monday when I'm back home.

vlazzarini:
ah, no problem, I thought you were on Windows. Thanks.

DaveSeidel:
I can try it in a few minutes, will let you know.

DaveSeidel:
Well, it doesn't look like there's an installable build, and I'm not set up
to build on Windows. Please let me know if there's an installer and I
missed it.

rorywalsh:
I have a windows build I can test with. I'll let you know in about any hour.

rorywalsh (approved):
I've tested this here and Ctrl+C now exists gracefully on Windows.

vlazzarini:
good to hear, thanks.

kunstmusik:
Just to follow up, can confirm the segfault is gone after the latest.

vlazzarini:
thanks.

PR #1957, opened 2024-09-19
working with tests cmake

vlazzarini:
The cmake script for tests is failing on Debian (and other linux systems) as findpackage is not working correctly. Also the minimum required cmake version is too high for some systems.
This PR is trying to provide fixes for this,


vlazzarini:
Also added gtest package to Ubuntu CI and confirmed tests are now properly built and run.

.github/workflows/csound_builds.yml
```diff
@@ -26,13 +26,13 @@ jobs:
...
           sudo apt-get update
-          sudo apt-get install cmake libsndfile1-dev libasound2-dev libjack-dev portaudio19-dev libportmidi-dev libpulse-dev swig liblua5.1-0-dev default-jdk libfltk1.1-dev libfluidsynth-dev liblo-dev fluid ladspa-sdk libpng-dev dssi-dev libstk0-dev libgmm++-dev bison flex libportsmf-dev libeigen3-dev libcunit1-dev gettext libsamplerate0-dev
+          sudo apt-get install cmake libgtest-dev libsndfile1-dev libasound2-dev libjack-dev portaudio19-dev libportmidi-dev libpulse-dev swig liblua5.1-0-dev default-jdk libfltk1.1-dev libfluidsynth-dev liblo-dev fluid ladspa-sdk libpng-dev dssi-dev libstk0-dev libgmm++-dev bison flex libportsmf-dev libeigen3-dev libcunit1-dev gettext libsamplerate0-dev
 
       - name: Configure build
         run: |
           mkdir build
           cd build
-          cmake .. -DUSE_MP3=0
+          cmake .. -DUSE_MP3=0 -DBUILD_TESTS=1 -DBUILD_STATIC_LIBRARY=1
```
stekyne, line 35: `BUILD_TESTS` are enabled by default afaik

tests/c/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
-cmake_minimum_required(VERSION 3.20)
 
-if(USE_VCPKG AND NOT ANDROID)
+if(BUILD_TESTS)
+cmake_minimum_required(VERSION 3.20)
```
stekyne, line 3: This was probably a good idea fair play. The high version for Cmake was to resolve some issue on Mac for GTest as far as I can remember. I agree it's a bit high for some platforms. Maybe we can set it lower except on Mac to overcome that. I think Steven said that 3.16 was the common highest. I'll see if I can recall the issue and work on dropping it down.

tests/c/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
...
+ if(USE_VCPKG AND NOT ANDROID)
     find_package(GTest CONFIG REQUIRED)
-else()
+ else()
     find_package(GTest)
     check_deps(BUILD_TESTS BUILD_STATIC_LIBRARY)
+ endif()
```
stekyne, line 9: Are these auto formatting change?

stekyne (approved):
Looks good to me. I think I enabled tests only for one of the linux builds as they should be mostly testing Csound logic and not the platform. We can enable for both apt-get and vcpkg builds though no problem. I think it runs fast anyways.

CMakeLists.txt
```diff
@@ -183,7 +183,7 @@ option(USE_LRINT "Use lrint/lrintf for converting floating point values to integ
 option(USE_CURL "Use CURL library" ON)
 option(BUILD_RELEASE "Build for release" ON)
 option(BUILD_INSTALLER "Build installer" OFF)
-option(BUILD_TESTS "Build tests" ON)
+option(BUILD_TESTS "Build tests" OFF)
```
stekyne, line 186: Ah I missed this above. Any reason to do so? Just to avoid build breakages for normal users?

.github/workflows/csound_builds.yml
```diff
@@ -166,7 +166,7 @@ jobs:
         run: brew install bison flex asio jack
 
       - name: Configure build
-        run: cmake -B build -S . -DUSE_VCPKG=1 -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake"
+        run: cmake -B build -S . -DBUILD_TESTS=1 -DUSE_VCPKG=1 -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake"
```
stekyne, line 169: I think that most custom cmake files like`custom-osx.cmake` have the tests enabled anyway

.github/workflows/csound_builds.yml
vlazzarini: I changed that because some linux (rpi and debian) had problems configuring the gtests

CMakeLists.txt
vlazzarini: yes, and because tests are more important for us than the normal users

tests/c/CMakeLists.txt
vlazzarini: probably

PR #1961, opened 2024-09-23
Use vcpkg jack2 port for CI macOS vcpkg builds

docEdub:
I'm not sure why the jack opcodes are not enabled for the CI macOS builds, but this change turns them on and updates the CI macOS vcpkg builds to use the vcpkg jack2 port instead of brew's jack package.


vlazzarini (approved):
Thanks for this, looks good to me and it's passing the CI for the two builds in question.

PR #1962, opened 2024-09-23
Remove brew asio package from CI macOS builds and BUILD.md docs

docEdub:
I'm not sure why the asynchronous I/O (asio) package is being referenced, but it is not needed as far as I can tell. Maybe it was needed before but is not needed now? Or maybe it was mistaken for the Windows asio sdk?

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -129,7 +129,7 @@ jobs:
       - name: Install dependencies
         run: |
           brew install --only-dependencies csound
-          brew install bison flex asio jack googletest
+          brew install bison flex jack googletest
 
       - name: Configure build
         run: cmake -B build -DBUILD_TESTS=1 -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake"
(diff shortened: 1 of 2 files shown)


vlazzarini (approved):
I always thought that was odd alright

kunstmusik:
I thought it was necessary for building in ASIO support for portaudio? Would be good for windows user to check.

vlazzarini:
Not on Mac, ASIO has no bearing on building portaudio there.

kunstmusik:
Ahhh, I'm going too quickly and missed that it was on macOS builds. 😅  Ignore my comment, that was very odd then!

stekyne:
I'm not sure how that got in there. Either myself or Bram probably copied pasted it in. It seems it was commented out for a long time. Anyway, it's gone now.

docEdub:
Just a note here that `asio` is not the correct vcpkg port to be using to bring in ASIO support for portaudio. The correct vcpkg is `asiosdk`. The `asio` package removed by this PR is for asynchronous I/O, and Csound is not using it on any platform afaik.

kunstmusik:
Thanks for that clarification!

Issue #1734, opened 2023-07-27
-Mm flag & ctrl7 midi channels - out of range errors


boonier:
Observed on Csound version 6.18 (double samples) Oct 20 2022

So when I was testing out the `-Mm` flag to enable access to multiple ports it looks like the ctrlN family of opcodes won't accept a midi channel greater than 1-16. This would also affect other midi input opcodes I think.

Can confirm that the error is being sent in here: https://github.com/csound/csound/blob/a1580f9cdf331c35dceb486f4231871ce0b00266/Opcodes/midiops2.c#L326-L329

More background info here: 
https://forum.csound.com/t/how-to-make-use-of-mm-flag-for-multiple-input-ports/1281?u=boonier
https://discord.com/channels/784849854270537790/786239959342121021/1133325632982499378

Cheers

vlazzarini:
Fixed in PR #1966

PR #1966, opened 2024-09-26
fixing midi channels for multiple ports #1734

vlazzarini:
This is a fix for #1734 where checks for midi channels did not allow multiple ports to be mapped to channels above 16.

### DIFF
--- a/OOps/midiops2.c
+++ b/OOps/midiops2.c
@@ -323,9 +323,9 @@
     if (UNLIKELY((ctlno = (int32) *p->ictlno) < 0 || ctlno > 127)) {
       return csound->InitError(csound, Str("illegal controller number"));
     }
-    else if (UNLIKELY((chan=(int32_t) *p->ichan-1) < 0 || chan > 15)) {
+    else if (UNLIKELY((chan=(int32_t) *p->ichan-1) < 0)) {
       return csound->InitError(csound,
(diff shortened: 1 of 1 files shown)


OOps/midiops2.c
```diff
@@ -323,9 +323,9 @@
     if (UNLIKELY((ctlno = (int32) *p->ictlno) < 0 || ctlno > 127)) {
       return csound->InitError(csound, Str("illegal controller number"));
     }
-    else if (UNLIKELY((chan=(int32_t) *p->ichan-1) < 0 || chan > 15)) {
+    else if (UNLIKELY((chan=(int32_t) *p->ichan-1) < 0)) {
```
kunstmusik: If I understand correctly, the max channel is MIDIMAXPORTS * 16 = 1024, so we may want to add back the upper bounds check for < 1024 throughout this PR. What do you think @vlazzarini?

vlazzarini: yes, good idea

kunstmusik (approved):
Did not test running this but code looks good.

vlazzarini:
Maybe we can add a test?

tjingboem:
Does [csound/csound] fixing midi channels for multiple ports #1734 (PR #1966)
go for both input and output of PortMidi?

Issue #1750, opened 2023-10-02
feature request: several midi ports


tjingboem:
I have 4 physical midi ports ( 1x M-Audio and 3 Edirol UM). A DAW like Reaper has access to all of these ports, so i can send midi data to all of my synths easily.

Csound is lacking such a feature; only one midi out (or in) can be specified. This means re-cabling of my synths as well as only one port available for 16 midi channels. Not practical.

Since i am on Linux, I have tried working with the Jacko opcodes, but these are not transparent and i did not succeed in getting them to work. Also, Jack is (mostly) Linux only.

Some sort of midi detection system for Windows, Mac and Linux and all their different drivers, and the option to set more than one midi in/out from the command line would be something to have?
And if one instrument wants to send midi data to one port, and another instrument to a second midi port, an opcode making this possible could be part of this system. This opcode could be put in the instrument. Some sort of MidiPortAssign.

tjingboem:
Addendum from Victor: 
_with portmidi you can. Each port is mapped to successivel groups of channels > 16.

From the manual

"With PortMIDI, DEVICE is either a number for single port input, 'a' for listening to all input ports, or 'm' for mapping ports to MIDI channels above 16. In this case, device 0 uses 1-16, device 1 uses 17-32, ... device N uses (N+1)*channel. The options 'a' and 'm' are also convenient when you don't have devices as they will not generate an error."_

vlazzarini:
Multi-port midi in exists with portmidi. 

Output needs to be looked. FWIW, midi output has never been implemented very completely and it has always been low priority.

tjingboem:
yes i understand this issue is low priority- but still it is a priority although low. I think not many Csound users have been missing such a feature, but it happens to be on my path and stumble on it. (And i expected Csound to  have this implemented.)

Narrowing my request:
- PortMidi, being multiplatform for Mac, Windows and Linux, to be multi-port (ins and outs)
- an opcode that assigns the particular instrument to a certain midi-port, so that in one csd different instruments can send and receive midi data to different midi-ports

tjingboem:
I have success with playing 2 keyboards that are attached to different midi input ports. They both play the csd
I am using -+rtmidi=PortMIDI -M a to address all input ports.

```
<CsoundSynthesizer>
<CsOptions>
-+rtmidi=PortMIDI -M a --midi-velocity=p5 --midi-key=p4
</CsOptions>
<CsInstruments>
sr=48000
ksmps=128
nchnls=2
[...]

PR #1967, opened 2024-09-27
adding MIDI channel test

vlazzarini:
Added a test for #1966


tjingboem:
Does [csound/csound] fixing midi channels for multiple ports #1734 (PR
#1966)
go for both input and output of PortMidi?

vlazzarini:
No, as I explained in your feature request (#1750), multiple ports and mapping into channels > 16 are only implemented for inputs. MIDI output only supports 16 channels in a single port.

kunstmusik (approved):
Test looks good, thanks!

PR #1973, opened 2024-09-29
Feature/messages

vlazzarini:
These are changes made in August that somehow never got done as a PR. 

- request to improve UDO error messaging
- fix to ftgen so that GEN 44 can also have string data.
- fix scanu/2 messages


kunstmusik (approved):
Looks good to me

PR #1975, opened 2024-09-29
fixing baremetal build + daisy instructions

vlazzarini:
This branch fixes some int - int32 discrepancies to allow baremetal build to succeed and adds instructions to build for the Daisy platform.


rorywalsh (approved):
Thanks for this Victor. It would be great if @amanjagwani can add some more details to the readme if he gets a chance.

vlazzarini:
The instructions there are enough to build a Csound library for baremetal. We should add an example code to the directory.

PR #1976, opened 2024-09-30
Making int32_t use uniform across the codebase

vlazzarini:
The use of `int32_t` has been adopted to clarify the size of integers across all platforms. However, this has not been applied consistently across the codebase with the result that in some platforms this inconsistency leads to compilation errors. This PR rectifies the issue, applying the type uniformly to the code.


Daisy/Custom.cmake
```diff
@@ -20,5 +20,5 @@ set(BUILD_OSC_OPCODES OFF)
 set(BUILD_DEPRECATED_OPCODES OFF)
 set(BUILD_UTILITIES OFF)
 
-set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wno-error -Wno-deprecated-register -Wno-deprecated-declarations -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-function -Wno-unused-variable -Wno-format -Wno-char-subscripts")
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-error -Wno-missing-field-initializers -Wno-unused-parameter -Wno-unused-function -Wno-unused-variable -Wno-format-Wno-char-subscripts")
```
stekyne, line 24: Seems like a big enough change. Any particular reasons for removing them?

stekyne (approved):

Daisy/Custom.cmake
vlazzarini: because there's a lot of spurious warnings I can't shut down. We're picking the important ones in the other builds.

PR #1979, opened 2024-10-01
conversion 64 to 32 warnings fixed

vlazzarini:
This PR reviewed all 64 to 32 implicit conversion warnings, fixing where they required attention and explicitly casting otherwise, when the conversion is benign.

Also added the relevant warning flags to cmake.


Opcodes/syncgrain.c
```diff
@@ -555,7 +555,8 @@ static int32_t filegrain_process(CSOUND *csound, filegrain *p)
               pos += hdataframes;
               csound->SndfileSeek(csound,p->sf,pos,SEEK_SET);
 
-              items = csound->SndfileRead(csound, p->sf, datap, hdatasize/chans);
+              items = (int32_t)
+                csound->SndfileRead(csound, p->sf, datap, hdatasize/chans);
```
stekyne, line 559: Looks a bit odd on the new line. Is your editor configured to a small line length? I think there is an editorconfig file in this repo. If you can use that, it would keep it consistent with anyone using Vscode or any editor that supports that standard.

vlazzarini: It's just how it ended up when I was editing it, probably an artefact of me having to zoom in to read the text.

We have a branch with a make format, which I'll hope we'll use to make things uniform. For now, I am not concerned.

kunstmusik (approved):
Did light code review and did build and ran unit tests for WASM build locally and it all looked good. Going to merge this one now.

PR #1980, opened 2024-10-02
long max table length (2^31) as default

vlazzarini:
This pushes the size to maximum possible without having to change the codebase to use 64bit ints for phase variables.

### DIFF
--- a/Engine/fgens.c
+++ b/Engine/fgens.c
@@ -2563,7 +2563,7 @@ static int32_t gen01raw(FGDATA *ff, FUNC *ftp)
     else ftp->nchanls  = 1;
     ftp->flenfrms = ff->flen / ftp->nchanls;  /* VL fixed 8/10/19: using table nchnls */
     ftp->gen01args.sample_rate = (MYFLT) p->sr;
-    ftp->cvtbas =  p->sr * csound->onedsr; /* VL - LOFACT factor (1024) removed */
+    ftp->cvtbas =  p->sr * csound->onedsr; 
     {
       SFLIB_INSTRUMENT lpd;
       int32_t ans = sflib_command(fd, SFC_GET_INSTRUMENT, &lpd, sizeof(SFLIB_INSTRUMENT));
--- a/Engine/fgens.c
+++ b/Engine/fgens.c
@@ -2563,7 +2563,7 @@ static int32_t gen01raw(FGDATA *ff, FUNC *ftp)
     else ftp->nchanls  = 1;
     ftp->flenfrms = ff->flen / ftp->nchanls;  /* VL fixed 8/10/19: using table nchnls */
     ftp->gen01args.sample_rate = (MYFLT) p->sr;
-    ftp->cvtbas = LOFACT * p->sr * csound->onedsr;
+    ftp->cvtbas =  p->sr * csound->onedsr; /* VL - LOFACT factor (1024) removed */
     {
       SFLIB_INSTRUMENT lpd;
       int32_t ans = sflib_command(fd, SFC_GET_INSTRUMENT, &lpd, sizeof(SFLIB_INSTRUMENT));
--- a/include/csoundCore.h
+++ b/include/csoundCore.h
@@ -225,8 +225,10 @@ static const uint32_t PHMASK = (1U << 31) - 1U;
 #define OCTRES     8192
 #define CPSOCTL(n) ((MYFLT)(1<<((int32_t)(n)>>13))*csound->cpsocfrc[(int32_t)(n)&8191])
 
+/* VL this seems to be a leftover without any purpose now 
+  (previously used in ugens3.c, loscilx.c, mp3in.c, fgens.c)
 #define LOFACT     1024
-
+*/
 #ifdef USE_DOUBLE
   extern int64_t MYNAN;
 #define SSTRCOD    (double)NAN
--- a/include/csoundCore.h
+++ b/include/csoundCore.h
@@ -225,10 +225,6 @@ static const uint32_t PHMASK = (1U << 31) - 1U;
 #define OCTRES     8192
 #define CPSOCTL(n) ((MYFLT)(1<<((int32_t)(n)>>13))*csound->cpsocfrc[(int32_t)(n)&8191])
 
-/* VL this seems to be a leftover without any purpose now 
-  (previously used in ugens3.c, loscilx.c, mp3in.c, fgens.c)
-#define LOFACT     1024
-*/
 #ifdef USE_DOUBLE
   extern int64_t MYNAN;
 #define SSTRCOD    (double)NAN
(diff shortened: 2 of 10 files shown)


include/csoundCore.h
```diff
@@ -203,11 +203,11 @@ extern "C" {
...
-#define MAXLEN     0x10000000
-#define FMAXLEN    ((MYFLT)(MAXLEN))
-#define PHMASK     0x0fffffff
-#else
+#ifndef  SHORT_TABLE_LENGTH  // long max table length is the default
+static const uint32_t MAXLEN = 1U << 31;  
+static const double FMAXLEN = (double) (1U << 31);
+static const uint32_t PHMASK = (1U << 31) - 1U;
+#else   // this is the original max table length
 #define MAXLEN     0x1000000L
```
kunstmusik: shouldn't this branch also use static const definitions?

include/csoundCore.h
```diff
@@ -203,11 +203,11 @@ extern "C" {
 #define CURTIME (((double)csound->icurTime)/((double)csound->esr))
 #define CURTIME_inc (((double)csound->ksmps)/((double)csound->esr))
 
-#ifdef  B64BIT
-#define MAXLEN     0x10000000
-#define FMAXLEN    ((MYFLT)(MAXLEN))
-#define PHMASK     0x0fffffff
-#else
+#ifndef  SHORT_TABLE_LENGTH  // long max table length is the default
+static const uint32_t MAXLEN = 1U << 31;
```
kunstmusik: General question why we're using uint32_t over uint64_t. Kind of odd we're mixing 32-bit ints with MYFLT and double operations.

include/csoundCore.h
```diff
@@ -225,12 +225,10 @@ extern "C" {
 #define OCTRES     8192
 #define CPSOCTL(n) ((MYFLT)(1<<((int32_t)(n)>>13))*csound->cpsocfrc[(int32_t)(n)&8191])
 
-#define LOBITS     10
+/* VL this seems to be a leftover without any purpose now
```
kunstmusik: Maybe this should just be removed altogether and also the comments in other files about LOxxx removed can also be removed? Seems we can just look it up in git history.

include/csoundCore.h
vlazzarini: I wanted to preserve the original code as it was (mostly for archaeology reasons). Not worried if you feel we ought to change it.

include/csoundCore.h
vlazzarini: Ok so the table length is used in the fixed point computing and that's strictly a 32-bit size (that's why the 2^31 limit). If we change this here, then we will need to make changes throughout the codebase (to avoid warnings at least I think but more so to be able to use 64 bis so push MAXLEN higher). For now, I think this is a good compromise.

FMAXLEN needs to accommodate an exponent size that is potentially bigger than in float, so I thought it should be double in order not to loose precision. This is also the case in the original code, but I am not proposing changing that.

include/csoundCore.h
vlazzarini: yes I missed that one. It's not used anywhere else. This is a leftover from a long tim
e ago.

include/csoundCore.h
kunstmusik: With the #ifdef it implies we are going to support SHORT_TABLE_LENGTH. If the following is there just for historic purposes, then I'd suggest removing the #ifdef and just leaving the original code commented out.

Engine/fgens.c
```diff
@@ -2563,7 +2563,7 @@ static int32_t gen01raw(FGDATA *ff, FUNC *ftp)
     else ftp->nchanls  = 1;
     ftp->flenfrms = ff->flen / ftp->nchanls;  /* VL fixed 8/10/19: using table nchnls */
     ftp->gen01args.sample_rate = (MYFLT) p->sr;
-    ftp->cvtbas = LOFACT * p->sr * csound->onedsr;
+    ftp->cvtbas =  p->sr * csound->onedsr; /* VL - LOFACT factor (1024) removed */
```
kunstmusik: These are the types of comments I think can be removed. It's noisy and it's only going to confuse future developers IMO.

include/csoundCore.h
kunstmusik: I think this comment as a whole can be removed too. It's just going to add questions to future developers and if we've moved on from it.

Engine/fgens.c
vlazzarini: will remove

include/csoundCore.h
vlazzarini: I think there's a dual purpose, documenting and supporting (maybe some systems may need this? Although I think the only purpose is to make it compatible with single-float precision). I can change it to consts if you wish.

TriangleBubble:
unsubscribe

vlazzarini:
So while the max table len tests passes in my computer, it fails on CI, here's the linux result

```
memory allocate failure for -17179869176
```

I guess we can't guarantee it will work. I'll remove the test.

kunstmusik:
@vlazzarini I tested this on Ubuntu and the test max len CSD fails for me here. Using debugger, I see on line 198 of fgens.c (https://github.com/csound/csound/blob/feature/longtables/Engine/fgens.c#L198), flen has the value 2147483648 and ff.flen has the value -2147483648. I'm not sure exactly what's going on yet but I think we should get this sorted out before releasing this branch.

vlazzarini:
It d

> @vlazzarini I tested this on Ubuntu and the test max len CSD fails for me here. Using debugger, I see on line 198 of fgens.c (https://github.com/csound/csound/blob/feature/longtables/Engine/fgens.c#L198), flen has the value 2147483648 and ff.flen has the value -2147483648. I'm not sure exactly what's going on yet but I think we should get this sorted out before releasing this branch.

Yes, I see it now. The problem is that we were assuming flen to be uint32_t but it's int32_t, so the conversion goes wrong there (it does not have enough precision). I brought it down to 2^30, which is the maximum we could get without having to go through every single GEN to make it work with int64_t. Re-enabled the test.

kunstmusik:
I still don't understand what's different about Linux here. The value was correct prior to MYFLT2LRND and casting to int32. It also worked on other Mac which make me think we need to investigate further.

vlazzarini:
> I still don't understand what's different about Linux here. The value was correct prior to MYFLT2LRND and casting to int32. It also worked on other Mac which make me think we need to investigate further.

I have done all that investigation. MYFLT2LRND behaves different on the Mac, and I think it's to do with arm64. What it does is that it rounds down to 2147483647. On Linux, it rounds up to 2147483648, which overflows to -2147483648. 

I was able to reproduce this here with simple cast, and got the problem. I then fixed it by making ff.flen int64_t but then it crashed in the GEN2 code, because there was code there using int32 and the overflow crashed it. I fixed that and managed to make the example work again.

However, it is likely that all GENS need reviewing for in64_t flen and I think this is work for later. 

EDIT: I checked the reference documentation and if x in lrint(x) produces an out of range result, its value is unspecified and a domain error is raised. That's what happens when x is 2^32. In linux, it outputs 0x80000000 but on the mac it gives 0x7FFFFFFF. Either way it's an error.

kunstmusik (approved):

kunstmusik:
Thanks for the explanation, let's move forward now.

PR #1981, opened 2024-10-03
Soundfile API

vlazzarini:
This PR does two things:

- Changes the code to open soundfiles using *only* the module API, as opposed to accessing `sflib_` functions directly.
- Adds a callback setting function so these soundfile operations can be implemented by hosts instead of the internal `sflib_` functions (which default to libsndfile if present or to non-op stubs otherwise).

This will enable among other things the access to soundfiles in platforms where libsndfile is not present.

### DIFF
--- a/include/csound_files.h
+++ b/include/csound_files.h
@@ -37,8 +37,7 @@ extern "C" {
     int32_t   channels ;
     int32_t   format ;
   } SFLIB_INFO;
-  
-
+ 
   /** 
    * Soundfile interface callbacks
    */
--- a/include/csound_files.h
+++ b/include/csound_files.h
@@ -28,6 +28,38 @@
 extern "C" {
 #endif
 
+  /**
+   * Soundfile interface data
+   */
+  typedef struct sflib_info {
+    long  frames ;     
+    int32_t   samplerate ;
+    int32_t   channels ;
+    int32_t   format ;
+  } SFLIB_INFO;
+  
+
+  /** 
+   * Soundfile interface callbacks
+   */
+  typedef struct sndfileCallbacks {
+    void *(*sndfileOpen)(CSOUND *csound, const char *path, int32_t mode,
+                        SFLIB_INFO *sfinfo);
+    void *(*sndfileOpenFd)(CSOUND *csound,
+                          int32_t fd, int32_t mode, SFLIB_INFO *sfinfo,
+                          int32_t close_desc);
+    int32_t (*sndfileClose)(CSOUND *csound, void *);
+    int32_t (*sndfileWrite)(CSOUND *, void *, MYFLT *, int32_t);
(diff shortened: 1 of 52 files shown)


vlazzarini:
This needs to be adapted for WASM.

kunstmusik:
Did a quick glance and it seems straightforward but I want to do a deeper check into the code base. For now, I'm wondering how this works with the work @nonameentername did for allowing to set file open callbacks. @nonameentername could you look through this and see if it works for csound-godot?

vlazzarini:
This works at layer below his callbacks, at the sflib_* calls level, which are ultimately how the various parts of the system operate on soundfiles specifically. I think these changes are independent from that API call.

nonameentername:
With the changes from this branch I am getting an error building plugins.  It seems like the file `soundfile.h` is no longer included in the installation but is still being referenced by `csoundCore.h`.

Error message:

```bash
In file included from include/csound/csdl.h:111,
                 from include/csound/plugin.h:32,
                 from src/strToFile.cpp:1:
include/csound/csoundCore.h:57:10: fatal error: soundfile.h: No such file or directory
   57 | #include "soundfile.h"

```

vlazzarini:
I'll check that

vlazzarini:
Fixed now.

vlazzarini:
Wasm still needs adjusting

kunstmusik:
@vlazzarini I saw some additional commits, is this complete now and ready for review or are there any other planned changes?

vlazzarini:
ready for your review. There was the question of normalisation that I had not considered correctly before but it's good now.

include/csound_files.h
```diff
@@ -176,6 +207,19 @@ extern "C" {
...
+   * Sets external callbacks for the soundfile interface
+   * replacing any default callbacks. These functions are used by
+   * Csound to perform soundfile IO throughout the system.
+   * These are set to sndfile.h functions if USE_LIBSNDFILE is defined
+   * otherwise they are set to non-op stubs.
+   * Callbacks are passed via the SNDFILE_CALLBACKS struct.
+   * Any NULL callback pointer will disable setting of the specific callback
+   * If p is NULL then all callbacks are reverted to default values.
+   **/
+  PUBLIC void csoundSetSndfileCallbacks(CSOUND *csound, SNDFILE_CALLBACKS *p);
```
kunstmusik, line 221: This makes me think we might follow a similar pattern for graph callbacks rather than individual API calls.

kunstmusik (approved):

kunstmusik:
There's a lot of changes here but from what I could see it all looked good. Did some smoke tests using manual examples to check reading/writing soundfiles. Think it's in good shape and if there's any issues that pop up we can address in future PRs.

PR #1984, opened 2024-10-05
Slidertable bug fixed and APIVERSION corrected

### LINKED ISSUE #1983: midi controller slider16table cause segmentation violation (at least on windows).
slider8table works fine, however slider16table crashes immediately.

The issue looks to be in Engine/entry1.c where slider16table, slider32table and slider64table are defined to use SLIDER8t structs instead of SLIDER16t, SLIDER32t, SLIDER64t.

Also in OOps/midiops3.c there's a hard-coded (j < 8) instead of (j < n) in SLIDER_I_TABLE_INIT macro.

SLIDERIF macro seems to have the same hard-coded (j < 8) issue.

vlazzarini:
This PR fixes #1983 and also corrects the APIVERSION - APISUBVER constants.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,8 +97,6 @@ message(STATUS "Csound directory: ${CMAKE_HOME_DIRECTORY}")
 
 
 # Project definitions
-set(CS_APIVERSION "7")
-set(CS_APISUBVER "0")
 set(CS_VERSION "7")
 set(CS_SUBVER "0")
 set(CS_PATCHLEVEL "0")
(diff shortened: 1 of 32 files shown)


vlazzarini:
@hlolli  can you fix this? It looks like the WASM build does not do the replacement automatically as in the other builds (maybe because it does not use CMake?)

```
../Top/csmodule.c:223:35: error: expected ')'
    if (majorVersion != (int32_t) CS_APIVERSION ||
                                  ^
../include/version.h:46:31: note: expanded from macro 'CS_APIVERSION'
#define CS_APIVERSION       (${CS_APIVERSION})   
                              ^
../Top/csmodule.c:223:35: note: to match this '('
../include/version.h:46:29: note: expanded from macro 'CS_APIVERSION'
#define CS_APIVERSION       (${CS_APIVERSION})   
                            ^
../Top/csmodule.c:223:35: error: use of undeclared identifier '$'
    if (majorVersion != (int32_t) CS_APIVERSION ||
                                  ^
../include/version.h:46:30: note: expanded from macro 'CS_APIVERSION'
#define CS_APIVERSION       (${CS_APIVERSION})   
```

CMakeLists.txt
```diff
@@ -97,7 +97,8 @@ message(STATUS "Csound directory: ${CMAKE_HOME_DIRECTORY}")
 
 
 # Project definitions
-set(APIVERSION "7.0")
+set(CS_APIVERSION "7")
```
kunstmusik: I'm a bit concerned we have API version vs. the regular version, subver, patchlevel. With semantic versioning, we should really just use the latter and remove APIVERSION IMO. App developers should be able to depend on the major, minor, patch to handle backwards compatibility.

vlazzarini: I'm happy with this. I can remove the API version reporting from the host API and we're done with it.

kunstmusik (approved):
Looks good to me, glad to see the simplification to just version, subversion, patch!

PR #1985, opened 2024-10-05
Overhaul of oparms access and documentation

vlazzarini:
`csound->oparms` is used to keep configuration data set in argdecode so it can be accessed by various parts of the system. In 6.x, we used an intermediary public data structure to hold this information instead of making the OPARMS structure public. 

In this re-organisation, we now have only one data structure, `OPARMS`, which is public, and are granting read-only access to the `csound->oparms` object in both host and module API. The advantages are:

- extra copying of data is avoided.
- increased safety as configuration can only be changed via `SetOptions` with error and consistency checking.

The OPARMS structure has also been documented and tidied up.


OOps/aops.c
```diff
@@ -2219,9 +2219,7 @@ int32_t hw_channels(CSOUND *csound, ASSIGN *p){
 int32_t inRange_i(CSOUND *csound, INRANGE *p)
 {
   p->narg = p->INOCOUNT-1;
-  OPARMS oparms;
-  csound->GetOParms(csound, &oparms);
-  if (UNLIKELY(!oparms.sfread))
+  if (UNLIKELY(!csound->oparms->sfread))
```
kunstmusik: This is reading oparms directly while elsewhere the code was updated to use csound->GetOParms(csound). Any reason for this here?

Opcodes/fout.c
```diff
@@ -399,22 +396,20 @@ static int32_t outfile_set_S(CSOUND *csound, OUTFILE *p)
...
-  csound->GetOParms(csound, &oparms);
   
   memset(&sfinfo, 0, sizeof(SFLIB_INFO));
   format_ = (int32_t) MYFLT2LRND(*p->iflag);
   if (format_ >= 51)
     sfinfo.format = AE_SHORT | TYP2SF(TYP_RAW);
   else if (format_ < 0) {
-    sfinfo.format = FORMAT2SF(oparms.outformat);
-    sfinfo.format |= TYPE2SF(oparms.filetyp);
+    sfinfo.format = FORMAT2SF((csound->GetOParms(csound))->outformat);
```
kunstmusik: A lot of repeated csound->GetOparms(csound) here in this file. Seems better to have:

```OPARMS oparms;
  oparms = csound->GetOParms(csound);
```
  
once and use oparms throughout this function. (Or 'O' if we want to match up with usage elsewhere.)

util/hetro.c
```diff
@@ -21,7 +21,7 @@
     02110-1301 USA
 */
 
-#include "std_util.h"                                   /*  HETRO.C   */
+#include "std_util.h"                                   /*  HETRO->C   */
```
kunstmusik: looks like this was accidentally changed from a find/replace

util/sndinfo.c
```diff
@@ -22,7 +22,7 @@
     02110-1301 USA
 */
 
-#include "std_util.h"                               /*  SNDINFO.C  */
+#include "std_util.h"                               /*  SNDINFO->C  */
```
kunstmusik: Another funny change of . to ->

util/srconv.c
```diff
@@ -219,20 +219,20 @@ static int32_t srconv(CSOUND *csound, int32_t argc, char **argv)
     char        outformch = 's';
     unsigned    outbufsiz = 0U;
     SNDFILE     *outfd = NULL;
-    OPARMS      O;
+   const OPARMS *     O;
```
kunstmusik: This looks odd as it's declaring a pointer then assigning ->outformat a few lines down. I suspect this and the other utils (which are memcpy'ing OPARMS) may have problems as I don't see how they would work when OPARMS was made read-only now. I see in some files there's a ->Calloc of sizeof OPARMS.

vlazzarini: The oparms here and in the other utilities is used only locally to pass options from one function to another and not to the rest of Csound. So it takes in the options from Csound (which are readonly and copies them locally, making some local changes.

OOps/aops.c
vlazzarini: good catch, fixed for consistency.

Opcodes/fout.c
vlazzarini: Fixed.

util/srconv.c
vlazzarini: Yes, I should have allocated and copied this but I fixed it now.
The oparms here and in the other utilities is used only locally to pass options from one function to another and not to the rest of Csound. So it takes in the options from Csound (which are readonly and copies them locally, making some local changes.)
NB: the function used to be `csound->GetOParms(csound, OPARMS *O)` and
we used it as `OPARMS O; csound->GetOParms(csound, &O);`

vlazzarini: I also noted some unused OPARMS code in another file and removed it. mkdb.c which we don't compile had also been the subject of a bad replacement.

kunstmusik (approved):
Think we can move forward with this now. Thanks!

PR #1986, opened 2024-10-06
Feature/vanillabuild

vlazzarini:
trying to introduce a vanilla build with artefacts for download.


vlazzarini:
This is done now. With this PR we have a vanilla build artefact for MacOS. This contains a build with libsndfile but no portaudio, portmidi or liblo. Coreaudio and coremidi provide the IO.

Libsndfile is statically linked, making the build self-contained.

vlazzarini:
This now builds and has correctly-linked csound binary. However it can't be run because of codesign issues. This is what I get

```
victor@firebird bin % ./csound         
dyld[7740]: Library not loaded: @rpath/CsoundLib64.framework/Versions/7.0/CsoundLib64
  Referenced from: <C6E5E5E0-6561-3060-B856-4A7D5C7E9509> /Users/victor/Downloads/csound_install/bin/csound
  Reason: tried: '../CsoundLib64.framework/Versions/7.0/CsoundLib64' (code signature in <AE6A27B8-0BE3-3943-B226-B0A84AF5FC18> '/Users/victor/Downloads/csound_install/CsoundLib64.framework/Versions/7.0/CsoundLib64' not valid for use in process: library load disallowed by system policy), '../CsoundLib64.framework/Versions/7.0/CsoundLib64' (code signature in <AE6A27B8-0BE3-3943-B226-B0A84AF5FC18> '/Users/victor/Downloads/csound_install/CsoundLib64.framework/Versions/7.0/CsoundLib64' not valid for use in process: library load disallowed by system policy)
zsh: abort      ./csound
```

EDIT: Rory suggested trying to make an installer. I wanted to avoid that, but I guess with the current difficulties, it may be the only way forward.

docEdub:
Does `target_link_options(${CSOUNDLIB} PRIVATE LINKER:-adhoc_codesign)` fix it?

vlazzarini:
> Does `target_link_options(${CSOUNDLIB} PRIVATE LINKER:-adhoc_codesign)` fix it?

It's worth a try.

vlazzarini:
Building an installer does the trick. Once it's installed, it can be run.
For the moment, this overwrites the 6.x installation, so that will need to be fixed.

vlazzarini:
This is now fully working. The Vanilla build is installed under /Applications:

- The framework
```
/Applications/Csound/CsoundLib64.framework
```

- The csound command

```
/Applications/Csound/csound
```

For the official full release, we will need to write a script to keep the installed 6.0 tree inside the framework and overlay the 7.0 tree (as "current") alongside it.

vlazzarini:
The library path is absolute so to link to the framework all we need to do is pass the path to it.
For example 7.x c API example from csoundAPIExamples can be built with the command

```
cc -o example1 example1.c -F /Applications/Csound -framework CsoundLib64
```

rorywalsh (approved):
Just tested here. All's working fine. I'm able to run Csound7 from the command line, and I'm able to link to CsoundLib64 without issues. I have to right-click to run the installer, but that's the same behaviour we had in Csound6.

vlazzarini:
thanks, merging now, this should help us getting people using Csound 7

kunstmusik:
For the full release, will it still install to /Library/Frameworks?  I think I'll have to update the CsoundFFM binding to looking /Applications/Csound/CsoundLib64.framework.

vlazzarini:
Full release goes into the usual place. It needs a more complicated set of scripts to overlay 7 on top of 6 so the two versions can coexist. 

This is a compromise solution for development. We may review it in the future.

PR #1987, opened 2024-10-08
pass-by-ref

kunstmusik:
Implements pass-by-ref for new-style UDOs when UDO instance's sr and ksmps matches parent INSDS.


vlazzarini (approved):
All good, pulled it, built, run tests locally and played some code. Looks ok to me.

PR #1998, opened 2024-10-14
Instrument Types and Variables

vlazzarini:
This PR introduces three elements:

- the `Instr` type to hold a reference to compiled instrument in the code
- the read-only local variable `this` to hold a reference to the instrument enclosing it.
- read-only global variables for named instruments, with variable names set to the instrument names defined in the code.

Plus

- overloads for typical instrument-related opcodes such as schedule, event, nstance, turnoff2, turnoff3 etc
- overload for `nstrnum` returning the number from an `Instr` variable
- overload for assignment to string, returning the name of an instrument from an `Instr` variable
- overloads for `init` and `=` to initialise instrument refs (from a number or from other instrument refs).

With this we can have the following code:

```
instr test
 test2:Instr = test
 i2 nstrnum test2
 S2 = test
 prints "instr %d %s\n", i2, S1
 event_i "i", test2, 1, 1
endin
event_i "i", test, 0, 1
```

and 

```
instr New
print p1
schedule this, 0.1 , 0.1
endin
schedule New, 0, 0.1
```

etc

A test was added to the suite.


Engine/linevent.c
```diff
@@ -575,6 +589,17 @@ int32_t eventOpcodeI_(CSOUND *csound, LINEVENT *p, int32_t insname, char p1)
         for (i = 2; i <= evt.pcnt; i++)
            evt.p[i] = *p->args[i];
       }
+      if (insname == 2) {
```
kunstmusik:
```suggestion
      else if (insname == 2) {
```

Seems like this should be `else if` here, in case insname == 1 so that the following else doesn't trigger

Engine/linevent.c
```diff
@@ -488,6 +489,15 @@ int32_t eventOpcode_(CSOUND *csound, LINEVENT *p, int32_t insname, char p1)
         evt.p[1] = (MYFLT) res;
         evt.strarg = NULL; evt.scnt = 0;
       }
+      if (insname == 2) {
```
kunstmusik:
```suggestion
      else if (insname == 2) {
```

kunstmusik:
Is "this" copied to UDOs in the same way that pfields and p1 are?  The "this" concept will get a little tricky if we ever do passing opcodes as values, as a UDO might capture the this variable of the calling context.

vlazzarini:
> Is "this" copied to UDOs in the same way that pfields and p1 are? The "this" concept will get a little tricky if we ever do passing opcodes as values, as a UDO might capture the this variable of the calling context.

I have to check but I don't think it is copied. That's in the instrument var pool and that does not get shared afaik. It is just a local var as any other the user may declare.

The instr name however is global.

NB: I think this work shows that with a little work we can have an opcode ref type too.

Engine/linevent.c
vlazzarini: I'll check but looks like you're right.

vlazzarini:
Confirmed: `this` is local to instrument and does not get passed to UDOs. Since UDOs are also instruments, they will have their own `this` with a high number but it can't be used to schedule it (we get an init/perf error which is the correct behaviour).

Engine/linevent.c
vlazzarini: done

Engine/linevent.c
vlazzarini: done

vlazzarini:
On reflection the name `this` is probably not appropriate, as it makes me expect an instance ref rather than instrument ref. However I still think this is a useful variable, particularly for recursion. So maybe I should rename it to `this_instr` or something similar?

vlazzarini:
New Opcodes
=====

I have completed this with three opcodes, two of which depend on the new type

```
var:Instr  createinstr  Scode     // create an instrument from Csound code
lambda  Instr [,ip3, ip4, ip5, ip6, ...]   // run the instrument code immediately with or without parameters
lambda  Scode  [ip3, ip4, ip5, ip6,...]   // create an anonymous instrument and run it immediately
```

I was struggling to think of opcode names, but I think these make sense. Here are some
examples

```
myinstr:Instr createinstr "out oscili(p4,p5)"
schedule(myinstr,0,1,0dbfs/2,440)
```

~~once the other PR goes in with the fixes in the semantic/expressions parsing,~~ we will also be able to do

```
schedule(createinstr("out oscili(p4,p5)"), 0 , 1 , 0dbfs/2,440)
```

this is already possible with 

```
lambda("out oscili(p4,p5)", 1, 0dbfs/2,440)
```

There's also an overload that takes instruments

```
myprint:Instr createinstr "print 1"
lambda(myprint)
```

With lambda, there is no use for p1 or p2, and p3 is optional. The first is the instrument or anonymous instr, the second is always zero. Because of the fact that the anonymous instrument is volatile and can be overwritten, execution is always immediate. For delayed execution, we need to schedule an instrument.

Note that this can actually change the way we write Csound code. We can go ahead and create our instruments, get a reference and then schedule them. For example the code may look like this:

```
synth:Instr createinstr {{
  a1 oscili p4, p5
  a2 linen  a1, 0.1,p3,0.01
  out a2
  }}

n:i = 0
while n < 12 do
 schedule(synth, n/2, 0.5, 0.5, cpsmidinn(n+60))  
 n += 1
od
```

NB: the instruments are created dynamically and will always take the first free slot in the instrument table.

EDIT: I brought in the fix from the other branch to allow inlining 
EDIT2: following a suggestion from Rory, I changed the opcode name to follow current conventions.

vlazzarini:
We may need to discuss the opcode names before we go ahead with this, so that any future opcodes along these lines follow similar ideas. Some thoughts:

- We should decide whether `create***` is the right word. An alternative is `new***`
- Should we use the current conventions or introduce camel case or underscore?
- Should we prepare overload `create` (or `new`), as in `create:Instr()` so that later
we can apply to other types? (NB: this code `create:Instr(S1)` does not currently 
find the correct opcode so this needs to be fixed in the semantic analysis. I suspect
the type hint is only currently working for single-letter types).

kunstmusik:
This PR got a lot bigger in scope. I prefer snake case for opcodes like create_instr if we go that route. I think we should also have a convention of Capital for data types and lower_case for opcodes. 

There's also the issue of differentiating an instrument definition vs and instrument instance. Could have InstrDef vs Instr here. (I suppose SuperCollider folks might be familiar from SynthDef and Synth.) 

Another alternative is "def" or "define" (if overloading) or definstr.

rorywalsh:
As much as I hate the current form of lower camelcase, it's used pretty consistently across all opcodes name. I think using a new convention could lead to confusion and frustration. So I'd go with `createinstr`, or `newinstr` or whatever, to keep it in line with opcodes like `compilestr`, `compilerorc`, etc.

joachimheintz:
+1

vlazzarini:
> This PR got a lot bigger in scope. I prefer snake case for opcodes like create_instr if we go that route. I think we should also have a convention of Capital for data types and lower_case for opcodes. 
> 
> 
> 
> There's also the issue of differentiating an instrument definition vs and instrument instance. Could have InstrDef vs Instr here. (I suppose SuperCollider folks might be familiar from SynthDef and Synth.) 
> 
> 
> 
> Another alternative is "def" or "define" (if overloading) or definstr. 

So InstrRef (or InstrDef) seems ok to me instead of Instr, but in my mind an instance is not an instrument (as I drum it into my students), so I would not like to use Instr for an instance.

I prefer Instr for an instrument reference or definition because it is what we've always used in Csound coding and so it sounds more natural to me.

I'm not keen on following supercollider conventions to be honest as they seem alien to Csound ;)

I do think we need an instance reference, currently we pass a raw handle whenever we need to do it, but we have better means now that gives us more options for expansion.

How do we call the type? I don't know, but I'd just go for Instance. It's a separate PR anyway.

vlazzarini:
I've also added an overload for "remove" that uses the new type. It occurred to me that a perf-time version might be useful, so I added "deleteinstr" to cover that.

Currently there is no way to refer to Instr variables we created outside the local scope, but I've opened an issue on this #2003 .

vlazzarini:
This is complete from my side.

kunstmusik (commented):
I think the code looks good but I think the name should be changed to InstrDef for the type. The difference between InstrDef and Instr can be akin to SC3's SynthDef and Synth. 

With def/instances it would afford something similar to what is possible with SC3, which would be to dynamically order processing of instruments (and eventually opcodes using OpcodeDef and Opcode). Users could keep these in arrays and reorder for example. 

Adding a Node class to add instrument instances would complete the work for dynamic processing order. (Though, users should be able to create their own node systems with user code given InstrDef/Instr).

vlazzarini:
Sounds good to, InstrDef. Happy to go with it. I'll make the change and commit.

Engine/entry1.c
```diff
@@ -135,6 +135,7 @@ OENTRY opcodlst_1[] = {
   { "ihold",  S(LINK),0,          "",     "",     ihold, NULL, NULL, NULL  },
   { "turnoff",S(LINK),0,           "",     "",     NULL,   turnoff, NULL, NULL },
   /* VL: 10.2.22 this was thread  but with parser3 we need to make string assignment on threads 1 & 2 */
+  { "=.instrS", S(IREF_NUM) ,0,  "S", ":InstrDef;", (SUBR) get_instr_name},
```
kunstmusik: Curious about this one. Seems odd to have an assignment overload here for InstrDef to String. One thing we might need moving forward from here is a convention for opcodes such as to_string(dataType) or str(datatype). So we could have:

S1 = str(myInstr)
S2 = to_string(myInstr)

(something like that).

Engine/entry1.c
```diff
@@ -143,6 +144,7 @@ OENTRY opcodlst_1[] = {
   {  "=.T",   S(STRCPY_OP),0,     "S",    "i",
      (SUBR) strcpy_opcode_p, (SUBR) NULL, (SUBR) NULL, NULL                 },
   { "=.r",    S(ASSIGN),0,        "r",    "i",    rassign, NULL, NULL, NULL },
+  { "=.instri", S(IREF_INIT) ,0,  ":InstrDef;", "i", (SUBR) init_instr_ref},
```
kunstmusik: Also odd here for me as I'm not sure what this means to overload = for i to InstrDef. Prefer to have some utility opcode like find_instr_def(inum).

Engine/entry1.c
```diff
@@ -1322,6 +1345,12 @@ OENTRY opcodlst_1[] = {
   { "compileorc",  S(COMPILE), 0,  "i", "S",  (SUBR) compile_orc_i, NULL, NULL },
   { "compilecsd",  S(COMPILE), 0,  "i", "S",  (SUBR) compile_csd_i, NULL, NULL },
   { "compilestr",  S(COMPILE), 0,  "i", "S",  (SUBR) compile_str_i, NULL, NULL },
+  { "createinstr",  S(CINSTR), 0,  ":InstrDef;", "S",  (SUBR) compile_instr, NULL, NULL },
+  { "deleteinstr",  S(DELETEIN), 0, "", ":InstrDef;",  NULL, (SUBR) delete_instr, NULL },
+  { "remove",  S(DELETEIN), 0, "", ":InstrDef;",  (SUBR) delete_instr, NULL },
+  { "lambda",  S(CARINSTR), 0,  "", "Sm",  (SUBR) compile_and_run_instr, NULL, NULL },
```
kunstmusik: I think we should remove these lambda opcodes here or at least rename it as it's not really a lambda. I don't think it adds much compared to using schedule with createinstr. We should probably reserve lambda for anonymous opcodes and it should be a datatype rather than an opcode.

Engine/entry1.c
vlazzarini: what about S()? Following the convention we already use.

Engine/entry1.c
vlazzarini: There's an nstrnum overload, so I'll just remove this overload.

Engine/entry1.c
kunstmusik: I think str() would be good, considering familiarity with python by a large part of our community. 

This would also be easy to overload for anyone introducing types, whether native or user-defined. S() seems odd to me and I didn't even know it existed. (I've always just used sprintf or some other formatted string opcode.) 

Rather than have a __str__ method as in Python, users could be directed to implement a UDO or opcode like:

```
struct MyNote start:i, dur:i, noteNum:i, veloc:i

opcode str(note:MyNote):S 
  return sprintf("[MyNote] %g %g %g %g", note.start, note.dur, note.noteNum, note.veloc)
endop
```

and the above could be a common pattern for user-defined types.

kunstmusik:
Thanks for accommodating on str(). I think I'm good to proceed for merging. 

I think we may want to look at the names for createinstr, deleteinstr, and remove. I think I'm not so fond of all lowercase like createinstr these days and would prefer create_instr and delete_instr. I think remove makes sense as it overloads a historical opcode. That said, we can experiment with it and revise in the future. 

One thing I was thinking we might want to do is mark this as a preview feature in some way, which would give us some flexibility to make changes if necessary but also get it into the community's hands to experiment with. I'm thinking of this from the way Java has been introducing new features as previews, which sometimes have to be explicitly enabled via a flag. They've been doing this with major features that have continued to undergo revisions in later releases until finally the feature is out of preview. We might not need to go so far as have flags, but we could create a practice where new features that need community experience before we can finalize can be marked as preview features in the manual and we can do votes for raising from preview to final versions. (Just a thought!)

I'll go ahead and merge this now, thanks!

kunstmusik (approved):

PR #1999, opened 2024-10-16
Complex type

vlazzarini:
This PR introduces a complex type to Csound, named `Complex` and a set of complex arithmetic operators, plus some fundamental operations such as conjugation, absolute value, argument, and real/imaginary access.
I have confirmed that arrays of complex numbers are working. We should add all the complex operations for array types as well and other things like FFTs etc, but that is work for another PR.

For the moment, we can have code such as

```
instr 1
Ca:Complex[] init 10    // array of Complex numbers
ca:Complex init 2,1     // a single complex number 2 + j (assignment works similarly) 
Ca[0] = ca * ca         // binary ops on arrays are all implemented (except exponentiation)
cc:Complex = Ca[0]      //  arrays work as usual
print abs(2*cc)         // complex and real numbers can be intermixed in expressions
endin
```

and so on. 

Complex types can operate at both i-time and perf-time. Expressions involving complex numbers will typically
be computed at both times, but expressions involving only complex and i-time reals will only run at i-time and
expressions involving only 1 complex number and k-var reals are going to be computed only at k-time.

If it appears necessary, a `Complexi` type may be introduced to limit computation at i-time. That should be able to use the infrastructure already in place.

Similarly, I am wondering whether a complex audio signal may be useful (for analytic signals etc). I have not yet made my mind up but that's another PR anyway.


vlazzarini:
So I moved the code to separate files, added support for polar representation and a test.

tests/commandline/test_complex_numbers.csd
```diff
@@ -0,0 +1,22 @@
...
+instr 1
+Ca:Complex[] init 10
+ca:Complex = 2,1
+print real(ca)
+cb:Complex = polar(ca)
+print arg(cb)
+Ca[0] = ca * ca
+cc:Complex = Ca[0]
+print abs(cc)
+endin
```
kunstmusik, line 25: General note: our tests generally check that they compile correctly but not so much that opcodes run as expected. I added a couple of assert UDOs here: https://github.com/csound/csound/blob/develop/tests/commandline/udo/pass_by_ref.csd#L12-L24

that allow for testing values and failing if not true. I think we could benefit from extracting that out to a reusable library and start to add use of it in our test suite.

vlazzarini: right, I hadn't come across exitnow(), very useful for this.

kunstmusik (approved):
I think the original intention for convert_internal_to_external() and convert_external_to_internal() was that it'd be a single place to deal with those conversions. Seems like the logic has spread out to create_out_arg a bit, but I think for we now we should merge this in and move forward. We can always revisit and refactor later. 

Otherwise great to see the type system being used here. After Csound 7 goes out, we'll have to be careful about introducing new types that might conflict with UDTs that people create. It'll probably spur us on to get to namespaces.

PR #2001, opened 2024-10-16
fix ios/macos build in actions to use macos-latest

vlazzarini:
macos-12 runner will be removed, this PR moves all macos builds to macos-latest.

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -102,7 +102,7 @@ jobs:
       - name: Build Csound iOS xcframework
         working-directory: iOs
         run: |
-           brew install bison flex
+          brew install bison flex
           ./build.sh
 
       - name: Create package zip
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -84,7 +84,6 @@ jobs:
   iOS_build:
     name: iOS build
     runs-on: macos-latest
-    if: vars.IOS_WORKFLOW_RUN_ID
     steps:
       - name: Install dependencies
         run: brew install bison
@@ -103,6 +102,7 @@ jobs:
       - name: Build Csound iOS xcframework
         working-directory: iOs
         run: |
+           brew install bison flex
           ./build.sh
 
       - name: Create package zip
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -102,7 +102,6 @@ jobs:
       - name: Build Csound iOS xcframework
         working-directory: iOs
         run: |
-          brew install bison flex
           ./build.sh
 
       - name: Create package zip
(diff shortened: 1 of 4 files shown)


vlazzarini:
I had to disable the iOS build because

1 - it failed on macos-latest
2 - it was actually linking to the wrong libsndfile under macos-12. For some reason there was an installation of libsndfile in the runner that was not the one built for iOS and it was finding it and linking to it.

I checked and here the same thing happens, never mind that the xcframework is built, it can't be found at all (or at least ahead of an installed libsndfile. I am supposing it links OK because it's the same arch (arm64) but it won't work on a device because of the missing libsndfile.dylib link.

I had a look around and it seems CMake can't find or link to xcframeworks and so this is not currently workable.
So I need to disable it now. Bring it back when it is fixed.

vlazzarini:
OK this is now fixed. The solution was to abandon the libsndfile xcframework build, and just make a standard static library (x86 and arm64. With that Csound built and got linked correctly. Since it's a static lib it's all in CsoundLib and we don't need to worry about it. The Csound xcframework should work as expected now.

rorywalsh:
I can vaguely recall doing this before, but ending up switching to a xframework for libsndfile so that I could run apps in the iOS emulator. I'll test it and let you know.

vlazzarini:
if you can force CMake to find the xcframework, then we can revert to it.

On macos-latest the build failed because there was no libsndfile in /usr/local and the issue of linking to xcframework became apparent.

rorywalsh (approved):
I've tested this locally, and with a CI triggered build. The resulting .xcframework bundles work as expected.

stekyne:
@vlazzarini I think I used 12 at the time because it was intel and the 13 was arm. I can't remember exactly but it's not a bad idea to have some coverage of older macos versions I guess.  It might be a good idea to have at least arm/intel builds. 

https://cmake.org/cmake/help/latest/release/3.28.html It does seem CMake supports xcframeworks at least partially. We can maybe incorporate that instead of the external shell file at some stage.

```
On Apple platforms, .xcframework folders are now supported:

The [find_library()](https://cmake.org/cmake/help/latest/command/find_library.html#command:find_library) command now finds .xcframework folders.

The [target_link_libraries()](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#command:target_link_libraries) command now supports linking against a .xcframework folder.

The [IMPORTED_LOCATION](https://cmake.org/cmake/help/latest/prop_tgt/IMPORTED_LOCATION.html#prop_tgt:IMPORTED_LOCATION) target property of an imported library target may now be the path to a .xcframework folder.
```

rorywalsh:
I tried on several occasion to get CMake to find xcframeworks over the past year and and failed miserably, even though it should be possible. But yes, it would be nice to have this at some point.

PR #2014, opened 2024-10-29
fixed csoundCreate with opcodedir override

### LINKED ISSUE #2012: Problem with csoundCreate
csoundCreate(NULL, "some opcodir path") causes Csound to freeze.

If I comment out line 126 (CSOUND_MEM_SPINLOCK) and line 132 (CSOUND_MEM_SPINUNLOCK) in csound/Engine/memalloc.c, then Csound does not freeze...

vlazzarini:
This PR fixes #2012, where`cs_strdup` was used before the memory alloc system had been started.

### DIFF
--- a/Top/csmodule.c
+++ b/Top/csmodule.c
@@ -607,7 +607,12 @@ int32_t csoundLoadModules(CSOUND *csound)
 
   /* opcodedir GLOBAL override **experimental** */
   if (csound->opcodedir != NULL) {
-    dname = csound->opcodedir;
+    char *opcdir = cs_strdup(csound, csound->opcodedir);
+    // csound->opcodedir was strdup'd so we free it here now
+    // after we copied the data.
+    free(csound->opcodedir);
+    csound->opcodedir = opcdir;
+    dname = opcdir;
     csound->Message(csound, "OPCODEDIR overridden to %s \n", dname);
   }
   size_t pos = strlen(dname);
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -1419,14 +1419,12 @@ PUBLIC CSOUND *csoundCreate(void *hostdata, const char *opcodedir)
   if (init_done != 1) {
     if (csoundInitialize(0) < 0) return NULL;
   }
-
   csound = (CSOUND*) malloc(sizeof(CSOUND));
   if (UNLIKELY(csound == NULL)) return NULL;
   memcpy(csound, &cenviron_, sizeof(CSOUND));
   init_getstring(csound);
   csound->oparms = &(csound->oparms_);
   csound->hostdata = hostdata;
-  csound->opcodedir = cs_strdup(csound, (char *) opcodedir);
   p = (csInstance_t*) malloc(sizeof(csInstance_t));
   if (UNLIKELY(p == NULL)) {
     free(csound);
@@ -1437,6 +1435,10 @@ PUBLIC CSOUND *csoundCreate(void *hostdata, const char *opcodedir)
   p->nxt = (csInstance_t*) instance_list;
   instance_list = p;
   csoundUnLock();
+  // no cs_strdup yet so we use strdup and
+  // free it later.
+  if(opcodedir != NULL) 
+    csound->opcodedir = strdup(opcodedir);
   csoundReset(csound);
   csound->API_lock = csoundCreateMutex(1);
   allocate_message_queue(csound);


Top/csmodule.c
```diff
@@ -607,7 +607,12 @@ int32_t csoundLoadModules(CSOUND *csound)
 
   /* opcodedir GLOBAL override **experimental** */
   if (csound->opcodedir != NULL) {
-    dname = csound->opcodedir;
+    char *opcdir = cs_strdup(csound, csound->opcodedir);
```
kunstmusik, line 610: I'm wondering if this would have an unintended side effect where if you call csoundReset and try to reuse an existing csound instance it'd lose it's opcodedir value?

fggp (approved):

Top/csmodule.c
vlazzarini: the opcodedir value is not lost? 

The problem may be calling free() on it instead of csound->Free(), since it has been allocated with  cs_strdup(). 

First I want to see if this stops the reported crash.

kunstmusik: First time you create it's strdup in csoundCreate, then it's replaced with one with cs_strdup. If 
Then if you call csoundReset iI imagine it would csound->Free() that value. Now if you try to start CSOUND again after reset, won't it try to read an invalid pointer from csound->opcodedir?

kunstmusik:
Wish this wasn't merged until after the code question was resolved...

fggp:
I'll have a look tomorrow and propose some tests...

Top/csmodule.c
vlazzarini: No, once it's created with cs_strdup(), it then becomes part of the Csound memory mechanism and if it's freed, then csound->opcodedir will be NULL and the code will not run because of the check. So there's no problem with this code at all (what I said about calling free on it was wrong). The Csound engine would lose the override, but that is what Csound 6 did anyway, since reset() would clear every member back to defaults (NULL, in this case).

However, it may make sense to keep the override under reset() since there is no separate function to set it now (we only do it in csoundCreate()). In this case, we just free() the string in csoundDestroy(). I've pushed a PR with this change (#2015).

vlazzarini:
> Wish this wasn't merged until after the code question was resolved...

It is resolved, there is no problem with this code. I have sent in a PR for retaining the opcodedir override. However, that is a separate issue. This PR was to fix the crash and it did that without changing the behaviour.

vlazzarini:
> I'll have a look tomorrow and propose some tests...

I think we should not waste any more time with this issue.

vlazzarini:
Just to confirm what I've said, here's a program to test it. Creates Csound with an opcodedir override, then resets it and runs. No issues.

```
int main(int argc, const char *argv[]) {
  /* Create the Csound engine instance */
  CSOUND *csound = csoundCreate(NULL, ".");
  if(csound != NULL) {
    int res = CSOUND_SUCCESS, i;
    csoundReset(csound);
    if(res == CSOUND_SUCCESS) {
      /* Compile code from string, synchronously */
      res = csoundCompile(csound, argc, argv);
      if(res == CSOUND_SUCCESS) {
        csoundStart(csound);
       /* compute audio blocks */
        csoundPerformKsmps(csound);
        while(res == CSOUND_SUCCESS) {       
          res = csoundPerformKsmps(csound);
        }
      }
    }
    /* Destroy the engine instance */
    csoundDestroy(csound);
    return 0;
  }
  return -1;
}
```

and the console showing the run

```
OPCODEDIR overridden to . 
resetting Csound instance
		   overall amps:      0.0
0 errors in performance
Elapsed time at end of performance: real: 0.000s, CPU: 0.000s
UnifiedCSD:  hommage_new.csd
Elapsed time at end of orchestra compile: real: 0.002s, CPU: 0.002s
sorting score ...
	... done
Elapsed time at end of score sort: real: 0.002s, CPU: 0.002s
--Csound version 7.0 beta (double samples) Oct 29 2024
[commit: 9165382022a3dd59f565c5e033ecb6a96a1a189d]
libsndfile-1.2.2
displays suppressed
0dBFS level = 1.0, A4 tuning = 440.0
orch now loaded
audio buffered in 1024 sample-frame blocks
AuHAL Module: found 6 output device(s):
1: BlackHole 2ch (2 channels)
2: MacBook Pro Speakers (2 channels)
3: Microsoft Teams Audio (2 channels)
4: ZoomAudioDevice (2 channels)
5: Multi-Output Device (2 channels)
6: Aggregate Device (4 channels)
selected output device: MacBook Pro Speakers
***** AuHAL module: output device open with 1024 buffer frames
writing 2048 sample blks of 64-bit floats to dac
SECTION 1:
ftable 1:
ftable 4:
ftable 21:
ftable 22:
ftable 23:
ftable 3:
B  0.000 ..  0.500 T  0.500 TT  0.500 M:  0.00000  0.00000
...
```

The override is lost, but that was also the case with Csound 6, so no change there.

Issue #2003, opened 2024-10-19
Global variables using explicit types 


vlazzarini:
As far as I can see, global variables cannot be created using explicit types. This is a problem because it limits global variables to implicit type variable declaration. New standard types and UDTs do not use implicit types.

Ideal scenario
-------
In my view, the ideal scenario is this

```
var:type init ...   // variable in instr 0 pool

instr 1
import var         // variable from instr 0 pool added to local pool
export var2:type, ...    // variable exported to instr 0 pool  (optional arguments can be used to initalise it)
...
endin

instr 2
import var2    // importing var2 created in instr 1
...
endin
```

The export case will need to make sure the variable does not already exist in the  pools etc. and treat these cases accordingly. 

I think that importing and exporting go well with explicit types

For a scheme like this to work we will need changes to semantic analysis and to the parser. While we can implement `import` and `export` now without problems, any code using these will fail semantic analysis.
The left-hand side will be checked and the global variables used in instr 1 and instr 2 will not be found at  all! Currently the semantic analysis looks *only* for reserved names or names starting with `g` in the global pool and so the variables given to import and export are not going to be found. 

Alternatives
----------

If we want to avoid the complications of changing the parser, there is an alternative.
For export we could  simply implement it as

```
var:type export [...]   // create variable in the global pool and optionally initialise it
```

which does not seem problematic.

For import, while we could do 

```
var:type import
```

we have an issue: the compiler will have created the variable already in the local pool. The import opcode could look for the variable in the global pool but what should we do with the local variable? If it is used elsewhere in the instrument, it is all tied up in the code (instead of the global var)

We could shadow it with the global variable, which as far as I know can be done by

- pointing the local var CS_VAR_MEM to the global CS_VAR_MEM so they share memory. We need to make sure when the instance allocation is cleared (eg by orcompact()) that the global var memory is not washed away (I don't think it is, but needs checking). 

Unless there is a better way to achieve this, that's it. The local variable memory is left unused (and this is wasteful but possibly can be tolerated). 

From my perspective, while `export` looks good and perhaps that's the best way of implementing it anyway, `import` looks wrong.

Thoughts?

rorywalsh:
I think it would cleaner if we could avoid using import. `export` seems ok, but I wonder if it might lead to confusion for people coming from other languages. Anything to be said for:

`var:type globalinit [...]   // create variable in the global pool and optionally initialise it`

vlazzarini:
[...]

PR #2017, opened 2024-11-01
Global variables for explicit types

vlazzarini:
Currently there is no syntax for declaring global variables for explicit types. This PR introduces this, following the discussion in #2003

Here is the main outline of the solution:

-  We would like to provide a global declaration format that parallels the one used for implicit types.
-  Global variables can be added anywhere in the code 
-  The annotation `@global` to a variable has the same effect as `g` to an implicit type. 
-  This is added to the variable name when the type is declared.

Here's a test code as an example:

```
myvar@global:i init 0
myarr@global:i[] fillarray 1,2,3

instr 1
print myvar
print myarr[0]
myarr[1] = 0
endin
```


vlazzarini:
this is complete from my side.

vlazzarini:
Fixed the issue in the develop code with const instr refs

Engine/csound_orc.lex
```diff
@@ -77,7 +77,7 @@ int get_next_char(char *, int32_t, struct yyguts_t*);
 %option nounput
 
 IDENT           [a-zA-Z_][a-zA-Z0-9_]*
-TYPED_IDENTIFIER  [a-zA-Z_][a-zA-Z0-9_]*:[a-zA-Z_][a-zA-Z0-9_]*
+TYPED_IDENTIFIER  [a-zA-Z_][a-zA-Z0-9_@]*:[a-zA-Z_][a-zA-Z0-9_]*
```
kunstmusik: I think this rule is off as it permits multiple @ symbols. I need to check but it should probably use a ? for optional @global, something like :

[a-zA-Z_][a-zA-Z0-9_]*(@global)?:[a-zA-Z_][a-zA-Z0-9_]*

kunstmusik:
Wanted to inquire, as I thought I had floated this idea, but why do we need @global? Could we instead just have scoping rules such that if you declare a variable in instr 0 space it's automatically in the global var pool and accessible by any instrument? Compiling of instruments should be able to lookup the var from the local var space first, then fallback to global, and you shouldn't need to have @global added. 

This would follow scoping rules of something like Javascript. Local names could shadow global ones and that would maintain backwards compatibility.

Engine/csound_orc_semantics.c
```diff
@@ -254,16 +292,24 @@ CS_VARIABLE* find_var_from_pools(CSOUND* csound, char* varName, char* varBaseNam
...
                                        varBaseName);
   } else {
+    // now we check for local variables
     var = csoundFindVariableWithName(csound, typeTable->localPool,
                                      varBaseName);
+    // then check for global variables that may have been explicitly defined
+    if(var == NULL) 
+    var = csoundFindVariableWithName(csound, csound->engineState.varPool,
+                                     varBaseName);
+    if(var == NULL)
```
kunstmusik, line 311: Seems like we already checked the global pool earlier? (line 299)

kunstmusik:
Went further through the code changes and saw that there was indeed a scoping system added to check local then global pools. I still wonder if we couldn't do without @global annotation and just use this to get explicitly typed variables in instr0 to work as global vars.

vlazzarini:
This would break compatibility since we can have local variables in instr0 (all the ones without g), so we need to make a distinction between them. 

I commonly use local variables (counters etc) in instr0, which have the same names as ones instruments. So we could not export them by default. The use of g ensures that. I have to check, but I think instr 0 local vars are not persistent from compilation to compilation.

Having that rule from implicit types makes it hard for us to have a different rule for explicit without it being confusing. We need an annotation as we have in 'g' for implicit types.

I don't think we can avoid an annotation: if I have

icount init 0 ; var in instr0

instr 1
icount init 0 ; different var in instr0
endin

it needs to work the same as

count:i  init 0

instr 1
count:i init 0
endin

and without an annotation or an export symbol/opcode/function we can't make a distinction.

vlazzarini:
Also: we can declare global variables inside instruments. We would have a different rule for explicit variables where we cannot,  that to me is also confusing.

Engine/csound_orc_semantics.c
vlazzarini: only for certain implicit types and reserved words, so we need to check again.

vlazzarini: only for certain implicit type vars and reserved words, so we need to check again.

Engine/csound_orc.lex
vlazzarini: you are right, improved.

vlazzarini:
Furthermore, consider you have this code

```
ivar = 440
instr 1
ivar = p4
a1 oscili 0.1,ivar
out a1
endin
schedule(1, 0, 1, ivar)
schedule(1, 0, 1, ivar*1.5)
```

and you learned about explicit-type vars and want to modernise your
code, 

```
var:i = 440
instr 1
var:i = p4
a1 oscili 0.1, var
out a1
endin
schedule(1, 0, 1, var)
schedule(1, 0, 1, var*1.5)
```

then you get all confused as it doesn't work as before.

kunstmusik:
I see, I suppose in JS they got around this once they introduce let and const to declare variables locally. If we did something similar (not saying we should), it'd look like:

```
var:i = 440
instr 1
  let var:i = p4
  a1 oscili 0.1, var
  out a1
endin
schedule(1, 0, 1, var)
schedule(1, 0, 1, var*1.5)
```

to differentiate the global vs locally scoped vars. 

Python goes a different route and needs to have global added to reference the global var:

```
var = 3
def x():
  global var
  print var
```

and we'd have something like:

```
var:i = 440
instr 1
  global var
  var2:i = p4
  a1 = oscili(0.1, var)
  a1 += oscili(0.1, var2)
  out a1
endin
schedule(1, 0, 1, var)
schedule(1, 0, 1, var*1.5)
```

Interesting to think about the different approaches. I suppose it's strange that we can define global variables within instruments but we're stuck with it due to legacy implementation. (Probably not a good practice!)

One concern that popped up in my head was what would all of this look like if/when we add namespaces and want to reference a namespaced-scoped variable.

vlazzarini:
I proposed that kind of approach in the open issue discussing this, but people there seem not to like it because it was different from how we use "g" and they thought it was too complicated. They supported the use of annotations, so I went with it.

rorywalsh:
I think the fact that we don't need to introduce any new scoping opcodes to our instruments in one big advantage of using annotations. @joachimheintz do you have any thought on this?

joachimheintz:
i agree!

rorywalsh:
@vlazzarini @kunstmusik will the use of annotations mess up future namespace features? I guess if it does this needs to be revisited, and if not, we can maybe push through and accept this.

vlazzarini:
Just thinking out loud, since we have g-vars would we not end up with issues when introducing namespaces (if any arise) which would be the same in case we introduce annotations for explicit vars?

All we doing is rolling out an already existing mechanism to work with explicit vars.

kunstmusik:
I think the way I'm seeing it is if we have "@global" then it would suggest we do something similarly to access namespaced data since we're doing a similar operation of accessing data from a different scope. We wouldn't have to and could always have "@global" and another system for accessing a namespaced value, but want to make sure we don't design ourselves into a hole.

vlazzarini:
I would think we could accommodate a suitable namespace annotation with this scheme. Notice that @global only matters at variable declaration, not access.

giovannibedetti:
why not:

var:gi = 440

it seems very clear but I'm not sure how feasable it will be with the parser

kunstmusik:
@giovannibedetti globalness is part of the variable, not the datatype. (That's how it's in my head at least.)

vlazzarini:
that's how I understand it too. Also it wouldn't look great with the new types like `var:gInstrDef` or `var:gComplex`

vlazzarini:
An alternative is to add the annotation at the front,```global:var:i```, but it may complicate the parsing.

Engine/csound_orc_compile.c
```diff
@@ -839,7 +779,8 @@ INSTRTXT *create_instrument0(CSOUND *csound, TREE *root,
 
   if (csound->ksmps != DFLT_KSMPS) {
     reallocateVarPoolMemory(csound, engineState->varPool);
-    // csound->Message(csound, "recalculate varpool\n");
+    if(csound->GetDebug(csound))
```
kunstmusik, line 782: Sidenote: this seems like a common pattern, we might benefit from a macro or function such as `debugMessage(csound, "some string")`

Engine/csound_orc_compile.c
```diff
@@ -1455,15 +1381,16 @@ void insert_opcodes(CSOUND *csound, OPCODINFO *opcodeInfo,
...
+        engineState->instrtxtp = (INSTRTXT **)
+          csound->ReAlloc(csound, engineState->instrtxtp,
+                          (1 + engineState->maxopcno) * sizeof(INSTRTXT *));
         /* Array expected to be nulled so.... */
         while (++i <= engineState->maxopcno)
           engineState->instrtxtp[i] = NULL;
       }
       inm->instno = num;
-      // csound->Message(csound, Str("UDO INSTR NUM: %d\n"), num);
+      if(csound->GetDebug(csound))
```
kunstmusik, line 1392: Related to earlier comment, we might find it better to introduce a csound->Log(csound, log_level, formatted string, args...) method that introduces different log levels like info, warn, error, etc. This is pretty common and would hide things like checking GetDebug() and simplify the codebase IMO.

kunstmusik:
@vlazzarini Suggest we merge but consider this a preview feature and continue discussion and experimentation, perhaps following up with a later PR if we decide on a different path. I think there's some fixes for other things in this branch that would be to get in now. Sound good to you?

vlazzarini:
sounds good, that'll allow me to do tests with global vars for implicit types and if we decide to re-implement this another, we can just replace the mechanism.

kunstmusik (approved):

Issue #2018, opened 2024-11-02
[possible bug] Is it possible to schedule named instances?


chrbala:
I have the following test csound file:
```csound
<CsoundSynthesizer>
<CsInstruments>
sr=44100
ksmps=64
nchnls=2
0dbfs=1

instr instrument, 1
  iName = p1
  print(iName)
endin

instr schedule_tester
  schedule(1.21, 0, 1)
  schedule("instrument.22", 0, 1)
endin

</CsInstruments>

<CsScore> 
i 1.11 0 1
i "instrument.12" 1 1
i "schedule_tester" 0 1

</CsScore>

</CsoundSynthesizer>
```

When I run `csound ./sound.csd -o=dac`, I get the following output:
```
UnifiedCSD:  ./sound.csd
rtaudio: PortAudio module enabled ...
using callback interface
rtmidi: PortMIDI module enabled
instr instrument uses instrument number 2
instr schedule_tester uses instrument number 3
Elapsed time at end of orchestra compile: real: 0.001s, CPU: 0.001s
sorting score ...
        ... done
Elapsed time at end of score sort: real: 0.001s, CPU: 0.001s
--Csound version 6.18 (double samples) Nov 24 2022
[commit: a1580f9cdf331c35dceb486f4231871ce0b00266]
libsndfile-1.1.0
graphics suppressed, ascii substituted
sr = 44100.0, kr = 689.062, ksmps = 64
0dBFS level = 1.0, A4 tuning = 440.0
orch now loaded
audio buffered in 1024 sample-frame blocks
writing 4096-byte blks of shorts to =dac (AIFF)
SECTION 1:
new alloc for instr instrument:
instr 1:  iName = 1.110
new alloc for instr schedule_tester:
new alloc for instr instrument:
instr 1:  iName = 1.210
new alloc for instr instrument:
instr 2:  iName = 2.000
B  0.000 ..  1.000 T  1.000 TT  1.000 M:  0.00000  0.00000
instr 2:  iName = 2.120
B  1.000 ..  2.000 T  2.000 TT  2.000 M:  0.00000  0.00000
Score finished in csoundPerform().
inactive allocs returned to freespace
end of score.              overall amps:  0.00000  0.00000
           overall samples out of range:        0        0
0 errors in performance
Elapsed time at end of performance: real: 0.004s, CPU: 0.004s
2048 4096 sample blks of shorts written to =dac (AIFF)
```

Notably, I see:
```
instr 1:  iName = 1.110
[...]

PR #2020, opened 2024-11-02
Fixed schedule/event/etc for named instruments with instance annotation

vlazzarini:
This PR fixes issue https://github.com/csound/csound/issues/2018 where schedule (schedulek etc) and event_i (event) ignored the instance annotation on named instrument strings.


kunstmusik:
Code looks good to me but could you add a test.csd?

vlazzarini:
good idea

kunstmusik (approved):

kunstmusik:
Thanks for the test, used it to check the functionality and all seems good. Merging now.

PR #2023, opened 2024-11-03
Reducing warnings on Windows  

vlazzarini:
There seem to be over 50 warnings on Windows. This PR aims to either deal with all of these or at least reduce this number.

A lot of these had to do with unsigned/signed comparisons and the majority pointed at comparisons with MAXLEN. Since this sets the max table size, and table sizes are signed int32s, then it did not make sense to have MAXLEN as unsigned. It's actually the max power-of-two that can be represented as 32bit signed int positive value. I also spotted am typo in the definitions for short table sizes (PHMASK was not defined).

Some changes were also made to fix warnings in the mingw build.


vlazzarini:
I think this is complete now, down to a couple of benign warnings.

stekyne (approved):

PR #2025, opened 2024-11-04
Portaudio callback module memory exception 

vlazzarini:
Using address sanitizer, I discovered that the portaudio callback IO module did not exit cleanly on SIGINT, causing a memory exception (heap-buffer-overflow). This PR fixes the issue (and also makes the code more efficient using a local variable).


kunstmusik (approved):
This is great, I think this fixes a crashing issue I had in Blue that I had been meaning to dig into further. Thanks!

vlazzarini:
I've been able to pick up a few niggly things. The good thing is that the code seems solid otherwise, so that's nice to confirm.

PR #2026, opened 2024-11-04
Remove sprintf from the codebase

vlazzarini:
Use of sprintf() is dubious and it has been deprecated in recent clang versions. I have replace all the extant instances of sprintf() with snprintf(). Without these changes -Werror fails because of -Wdeprecated. Even if this is not the case in other compilers (e.g. gcc) I feel we ought to remove this vulnerability.

Curiously there was a check for HAVE_SNPRINTF in OOps/str_ops. This was only checking for C99, and this was not tested for anywhere else in the codebase (where snprintf is used widely). I removed this as well as it gives the wrong idea. 

If there is a problem with snprintf() not existing somewhere, then we need to deal with it code-wide. I can't see that arising.


stekyne (dismissed):

InOut/rtjack.c
```diff
@@ -1292,7 +1292,7 @@ static int32_t midi_in_open(CSOUND *csound,
       (RtJackMIDIGlobals*) csound->QueryGlobalVariableNoCheck(csound,
                                                               "_rtjackMIDIGlobals");
 
-    sprintf(clientName, "%s_in", pm->clientName);
+    snprintf(clientName, MAX_NAME_LEN+3, "%s_in", pm->clientName);
```
kunstmusik: Checking, should this be +4 to account for _in and NULL terminator?

InOut/rtjack.c
```diff
@@ -1405,7 +1405,7 @@ static int32_t midi_out_open(CSOUND *csound, void **userData,
     pm =
       (RtJackMIDIGlobals*) csound->QueryGlobalVariableNoCheck(csound,
                                                               "_rtjackMIDIGlobals");
-    sprintf(clientName, "%s_out", pm->clientName);
+    snprintf(clientName, MAX_NAME_LEN+4, "%s_out", pm->clientName);
```
kunstmusik: Checking, should this be +5 to account for _out and NULL terminator?

InOut/rtjack.c
vlazzarini: The array was allocated to MAXLEN+3, so that's the length for snprintf.

I am assuming this was right and did not change it.  In the other place there is one extra character and it is +4.

InOut/rtjack.c
vlazzarini: As before, there is only 4 added bytes in the array.

InOut/rtjack.c
scottsitar: It does look like there's an inconsistency between two different versions of these fields in two different structures:

- https://github.com/csound/csound/blob/develop/H/cs_jack.h#L43
- https://github.com/csound/csound/blob/develop/InOut/rtjack.c#L1127

Based on the header file version, it feels like the intention of the `MAX_NAME_LEN` constant was the number of characters in the fields not including the terminating null.  In that case, this could potentially truncate the final "n" if the original name had the maximum length.  However, based on the jack documentation, it seems like their max length constant (returned by `jack_client_name_size()` was inclusive of the terminating null.  That being said, when we query that function and clamp to our constant, we add an extra 1:

- https://github.com/csound/csound/blob/develop/InOut/rtjack.c#L1221

So, while there's a chance this might not cause any immediate issues, it's not obvious that it's correct, either.

vlazzarini: Based on this, there's two ways to go:

 1) remove the +1 from the configuration variable length so that it can only go to MAX_NAME_LEN, also take - 4 from the port output name size (it was 3 but it needs accommodate 4 chars)

 2) add + 1 to the jack MIDI struct and to the array/snprintf in the open device functions.

The fact that the MAX_NAME_LEN is different for the audio and midi modules is neither here nor there, so either of the two options are equally correct. I'll go for the first, which uses minimal changes.

The main thing here is that the config variable max length is the same as what we expect when we are doing the snprintf(). However, if the user types a long string, it will be truncated anyway, when the configuration variable is entered. Because of the + 1, it will then be further truncated by snprintf(). Still, the objective of this change was to avoid buffer overflow, which in this case could have actually happened with sprintf() (by 1 or 2 bytes).

vlazzarini:
I have also taken the opportunity to fix the test warnings and generally tighten up the warnings/error checks.

kunstmusik (approved):
Think this is in good shape, merging now, thanks!

PR #2027, opened 2024-11-09
Bug fix in the vco init phase

fggp:
The pointer to the sine table (ftp) was kept local to the init procedure of the vco opcode, which caused a segfault when used from the VCO struct ftp field.


vlazzarini (commented):
Looks good to me.

vlazzarini (approved):

PR #2029, opened 2024-11-10
ftgenonce segfault bugfix

vlazzarini:
This PR fixes the segfault caused by a local assignment of sstrcod.


kunstmusik (approved):
Tested locally on macOS and segfault looks resolved. Merging now, thanks!

Issue #2031, opened 2024-11-11
Changes in the API: FindOpcode


gesellkammer:
In Csound6, a call to `csound->FindOpcode` (in csound6 this is called "find_opcode_new") with an opcode with signature
`opcode "test", a, ak` would be resolved correctly even if called with an init arg instead of a k arg. In csound7 this is no longer the case, nothing is found. (The opcode is only resolved if the variable is explicitly a k variable)

Example: given an opcode with an OENTRY defined as 

```c
    { "testopc", S(TESTOPC), 0, "a", "ak", (SUBR)testopc_init, (SUBR)testopc_perf, NULL, NULL}
```
I would assume that calling `csound->FindOpcode(csound, "testopc", "a", "ai")` would find the opcode since a call within csound in the form of "aout testopc ain, 0.5" would be valid, but this is not the case, null is returned instead. The opcode is correctly found when called as `csound->FindOpcode(csound, "testopc", "a", "ak")`. The former call using "ai" as input signature is successful in csound6 with  `csound->find_opcode_new(csound, "testopc", "a", "ai")`.

Any hints as to what was changed here?

vlazzarini:
That is because the function used is `find_opcode_exact()`. If it is more useful to have `find_opcode_new()`, it's an easy replacement.

gesellkammer:
Oh, that's good to know, thanks for looking into it. Would it be possible to have both, maybe in the form of a flag argument to FindOpcode?

vlazzarini:
Flag is a good idea. I want to avoid the multiplication of similar functions. I'll make a PR.

vlazzarini:
PR submitted.

PR #2032, opened 2024-11-11
added new csoundFindOpcode function

vlazzarini:
Following the discussion in #2031, a new function is introduced to select either find_opcode_exact or find_opcode_new

```
const OENTRY* (*FindOpcode)(CSOUND*, int32_t exact, char*, char* , char*);
```

### DIFF
--- a/include/csoundCore.h
+++ b/include/csoundCore.h
@@ -1648,7 +1648,7 @@ static inline double PHMOD1(double p) {
                         int32_t (*perf)(CSOUND *, void *),
                         int32_t (*deinit)(CSOUND *, void *));
     int32_t (*AppendOpcodes)(CSOUND *, const OENTRY *opcodeList, int32_t n);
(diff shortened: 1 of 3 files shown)


include/csoundCore.h
```diff
@@ -1652,7 +1652,7 @@ static inline double PHMOD1(double p) {
                         int32_t (*perf)(CSOUND *, void *),
                         int32_t (*deinit)(CSOUND *, void *));
     int32_t (*AppendOpcodes)(CSOUND *, const OENTRY *opcodeList, int32_t n);
-    OENTRY* (*FindOpcode)(CSOUND*, char*, char* , char*);
+    const OENTRY* (*FindOpcode)(CSOUND*, int32_t exact, char*, char* , char*);
```
kunstmusik, line 1655: As an aside, seems like we could be using bool more in the codebase.

kunstmusik (approved):
looks good to me

PR #2035, opened 2024-11-18
New tests

vlazzarini:
This PR introduces some new tests and as a result of runnings also a couple of bugfixes.


vlazzarini:
I think this is the complete set of tests for now. We can add more in another PR.

vlazzarini:
Following from the unit test, I made it possible for modules to add data new custom data types. Added an example plugin data type to the example set.

README.md
```diff
@@ -4,24 +4,16 @@ Version 7.0.0 (beta)
 This is the the develop branch of the Csound main code repository. At
 the moment, we are developing the next main version (moving from 6.x
 to 7.0), and this branch reflects current work-in-progress. This is
-still undergoing changes and fixes until the first release. The
```
kunstmusik: Odd, looks like the README is somehow changed back to what it was prior to the other PR here?

vlazzarini:
I think this happens because it is always rewritten by CMake. I needed to have merged before doing a build.

README.md
vlazzarini: I think this happens because it is always rewritten by CMake. I needed to have merged before doing a build.

kunstmusik:
I'm getting a test error when I run here:

```
59/60 Test #59: SndfileTests.testReadSndfile .........................***Failed    0.03 sec
      Start 60: NewTypeTests.testAddOpcodeC
60/60 Test #60: NewTypeTests.testAddOpcodeC ..........................   Passed    1.92 sec

98% tests passed, 1 tests failed out of 60

Total Test time (real) =  52.76 sec

The following tests FAILED:
         59 - SndfileTests.testReadSndfile (Failed)
Errors while running CTest
```

When I look at the test log I see:


```
[misfinit: cannot open test.raw[m
[m/Users/stevenyi/work/csound/cs7-prs/tests/c/csound_test_sndfile.cpp:171: Failure
Value of: result == 0
  Actual: false
Expected: true
```

vlazzarini:
Is test.raw on disk? It should be produced by the previous test. The error reported is file not found, see code in libsnd.c:

```
fullName = csoundFindInputFile(csound, sfname, "SFDIR;SSDIR");
if (UNLIKELY(fullName == NULL)) /* if not found */
        csoundDie(csound, Str("isfinit: cannot open %s"), sfname);
```


You can run tests/c/unittests directly to get a better output (instead of make test)

Tests are passing here and in CI. So I am not sure what the issue is.
Here is the output from that command for the two sndfile tests

```
[----------] 2 tests from SndfileTests
[ RUN      ] SndfileTests.testWriteSndfile
--Csound version 7.0 beta (double samples) Nov 20 2024
[commit: 0e8d6acc20b7d8f2ba4c1e9c1ed256756aa3946c]
libsndfile-1.2.2
0dBFS level = 32768.0, A4 tuning = 440.0
orch now loaded
audio buffered in 1024 sample-frame blocks
WARNING: Output file already exists: Will be overwritten.

writing 8192-byte blks of double floats to test.raw (raw)
SECTION 1:
new alloc for instr 1:
B  0.000 ..  1.000 T  1.000 TT  1.000 M:      0.5
End of Performance inactive allocs returned to freespace
		   overall amps:      0.5
0 errors in performance
Elapsed time at end of performance: real: 0.004s, CPU: 0.004s
1024 8192 sample blks of double floats written to test.raw (raw)
[       OK ] SndfileTests.testWriteSndfile (68 ms)
[ RUN      ] SndfileTests.testReadSndfile
--Csound version 7.0 beta (double samples) Nov 20 2024
[commit: 0e8d6acc20b7d8f2ba4c1e9c1ed256756aa3946c]
libsndfile-1.2.2
0dBFS level = 1.0, A4 tuning = 440.0
orch now loaded
audio buffered in 1024 sample-frame blocks
WARNING: audio_in test.raw has sr = 0, orch sr = 44100
WARNING: audio_in test.raw has 0 chnls, orch 1 chnls_i
reading 1024-byte blks of unknown from test.raw (unknown)
AuHAL Module: found 6 output device(s):
1: BlackHole 2ch (2 channels)
2: MacBook Pro Speakers (2 channels)
3: Microsoft Teams Audio (1 channels)
4: ZoomAudioDevice (2 channels)
5: Multi-Output Device (2 channels)
6: Aggregate Device (4 channels)
selected output device: MacBook Pro Speakers
***** AuHAL module: output device open with 1024 buffer frames
writing 1024 sample blks of 64-bit floats to dac
SECTION 1:
new alloc for instr 1:
B  0.000 ..  1.000 T  1.000 TT  1.000 M:  0.50000
End of Performance inactive allocs returned to freespace
		   overall amps:  0.50000
0 errors in performance
Elapsed time at end of performance: real: 0.939s, CPU: 0.932s
AuHAL module: device closed
44 1024 sample blks of 64-bit floats written to dac
[       OK ] SndfileTests.testReadSndfile (1069 ms)
[----------] 2 tests from SndfileTests (1138 ms total)
```

kunstmusik:
I figured out the issue. I have the following .csound7rc:

```
-o dac -+rtaudio=pa_bl
```

I have no idea why, but it works if I move that out of the way.

vlazzarini:
> I figured out the issue. I have the following .csound7rc:
> 
> ```
> -o dac -+rtaudio=pa_bl
> ```
> 
> I have no idea why, but it works if I move that out of the way.

ah, this overrides the output from the previous test, so test.raw is not written to disk, and then it's not found by the test.

kunstmusik:
Should SetOption override what is in .csound7rc? I see it does not but it seems like it should:

```
[----------] 2 tests from SndfileTests
[ RUN      ] SndfileTests.testWriteSndfile
rtaudio: PortAudio module enabled ...
using blocking interface
--Csound version 7.0 (double samples) Nov 21 2024
[commit: e0fc3210fd4f0400ab18bee47a957d0708b711b8]
libsndfile-1.2.2
0dBFS level = 32768.0, A4 tuning = 440.0
orch now loaded
audio buffered in 1024 sample-frame blocks
PortAudio V19.7.0-devel, revision 147dd722548358763a8b649b3e4b41dfffbcfbb6
   0: dac0 (BlackHole 2ch [Core Audio, 2 in, 2 out])
   1: dac1 (MacBook Pro Speakers [Core Audio, 0 in, 2 out])
   2: dac2 (GroundControl 16ch [Core Audio, 16 in, 16 out])
   3: dac3 (GroundControl 2ch [Core Audio, 2 in, 2 out])
   4: dac4 (GroundControl 64ch [Core Audio, 64 in, 64 out])
   5: dac5 (Microsoft Teams Audio [Core Audio, 1 in, 1 out])
   6: dac6 (ZoomAudioDevice [Core Audio, 2 in, 2 out])
PortAudio: selected output device 'MacBook Pro Speakers'
writing 1024 sample blks of 64-bit floats to dac
SECTION 1:
new alloc for instr 1:
B  0.000 ..  1.000 T  1.000 TT  1.000 M:      0.5
End of Performance inactive allocs returned to freespace
                   overall amps:      0.5
0 errors in performance
Elapsed time at end of performance: real: 3.013s, CPU: 0.048s
closing device
44 1024 sample blks of 64-bit floats written to dac
[       OK ] SndfileTests.testWriteSndfile (3745 ms)
```

vlazzarini:
I don't think it ever did.

kunstmusik:
Surely it must have. I always thought the precedence was:

.csound7rc (or .csound6rc)
<CsOptions>
Commandline (or setOptions if using API)

from lowest to highest. If not, a user could never override anything in .csound7rc which certainly can't be the case. I've used this for .csound6rc for years where it makes csound default to realtime and I would add `-o test.wav` to commandline whenever I needed to write to disk.

vlazzarini:
Nothing has changed in this; csoundSetOptions() is called before the environment var is looked at in csoundStart so that will override it.

It's a different order if you are passing command line arguments and CSD though because a different API function is used. In that case, argdecode is called to override.
Also in that case, there is a check  in csoundStart() that stops the environment variables being looked at again. 

I guess this behaviour can be modified so that a call to csoundSetOptions does the same. It's going to be different to 6.x

kunstmusik:
I need to re-test this with CS6 because I'm having a hard to believing this isn't a regression of some sort. I don't see how Blue or other API apps would have worked if .csound6rc took precedence over SetOption.

vlazzarini:
you will need to check the exact sequence of calls in the test. Other calls and sequences may have different results.
Also note that csound 6 setOption only supported single options afaik

Btw it's easy to check the code in main.c, envvar options are set by checkOptions() and you can see where this happens. You'll see it's called in csoundCompileArgs() and in csoundStart() (if a CSD was not used). That is the same in both 6.x and 7.0.

kunstmusik:
I think I found out what is happening and why I never encountered this with Blue. In main.c:csoundStart(), it has:

```
PUBLIC int32_t csoundStart(CSOUND *csound) // DEBUG
{
    OPARMS  *O = csound->oparms;
    int32_t     n;

    /* if a CSD was not used, check options */
    if (csound->csdname == NULL)
          checkOptions(csound);
```

That checkOptions there only gets called if there was no CSD given. Blue calls csoundCompile with a CSD so it doesn't run into this issue. (I assume CsoundQt and Cabbage also do the same?). I think you're right that this has been the way it's been but I think it's problematic to have .csound7rc potentially override values from SetOption(). 

I think the best thing here is to merge this in and I'll file an issue as this is a longstanding issue that was surfaced from these tests.

kunstmusik:
Added #2042 to track the issue. Will proceed to merge now.

kunstmusik (approved):

Issue #2037, opened 2024-11-19
pconvolve not looking at --env:SSDIR in CsOptions


joachimheintz:
in csound6 i could use `--env:SSDIR+=..` also for the impulse response file used in the pconvolve opcode.  
this is not any more working in csound7 as the attached example shows:
`INIT ERROR in instr 1 (opcode SimpConv) line 27: pconvolve: error while impulse file`

diskin is working as expected as the example shows once you cd in the csd directory.
[pconvolve.zip](https://github.com/user-attachments/files/17812029/pconvolve.zip)

vlazzarini:
Fixed in #2038

PR #2038, opened 2024-11-19
Replaced open function to allow for ssdir etc awareness

vlazzarini:
This PR fixes issue #2037


fggp (approved):
Tested here: it works now.

PR #2041, opened 2024-11-19
added csoundNewOpcodeList etc to API

vlazzarini:
Added opcode listing functions to the API (in csound_misc.h)

csoundNewOpcodeList and csoundDisposeOpcodeList


kunstmusik:
Seems straightforward. Only question I have is if python bindings should be updated here.

rorywalsh (approved):
This looks good. I'll approve. Maybe @fggp can look into updating the Python bindings when he gets a chance.

PR #2047, opened 2024-11-27
Instance ref type and operators

vlazzarini:
This PR introduces an instance ref type, some overloads, and new opcodes.

Instance ref type
----
Following on from the introduction of the instrument definition ref type, we know have an instance ref type that can be used to manipulate instances running in Csound. The type name is `Instr`.

- it takes over from the use of plain k- and i-rate variables to hold instances,
- it allows new operations on instances.

The use of MYFLT to hold instance pointers was always problematic because these types do not have enough size to accommodate a pointer in some situations (e.g. floats version running in 64bit architectures). Compounding the problem, the code in 6.x only used the mantissa bytes to hold the pointer and so that was problematic also in doubles version running on 64bit systems. Since the use of this was limited to two opcodes (`nstance` and `turnoff`), we can now deprecate the i-/k- versions.

Instance chains
-----------

Csound maintains lists (chains) of instances so the performance order can be established and memory can be managed. There are two important chains:

- active list: instances on this list are performed in the order they have been linked into. The standard order is by ascending p1.

- free instance list: instances on this list are free to be inserted into the active list. Csound uses this list to manage memory allocation. 

With the instance reference type we can manage instances in those lists and also run instances that are not in either of them.

Overloads of existing opcodes
-------
We overload the schedule opcode so it can return an instance

```
var:Instr  schedule ...
```

We overload turnoff so it can act on an instance

```
turnoff  Instr   // i-time
turnoff  Instr, koff   //k-rate with off trigger
```

New Opcodes
--------

A set of standard and more experimental opcodes are put forward

```
var:Instr  play  InstrDef[,p4, ...] 
```

creates, inits, and performs an instance indefinitely (i-time) and immediately. It adds the instance to the end of the active list, regardless of instr number, so it preserves the instantiation order.

```
pause Instr, koff
```

pauses an instance if `koff = 1` (perf-time)

```
setp Instr, n, val
```

sets an instance parameter n to val (perf-time)

```
var:Instr create InstrDef
```
creates an instance (i-time), but does not add it to the free instance or active chains (so `orcompact()` etc are not aware of it). We can later add opcodes that will insert this instance at specific places in the active list.

```
err:i init Instr[,p4, ...]
```
runs an init pass on an instance (i-time), it sets the instrument active flag but does not add it to the active chain.

```
err:k perf Instr
```

runs a perf pass on an instance (perf-time).
 
```
delete Instr
```

deletes instance (deinit-time).  If instance is active, it turns it off. If it has been linked into the free instance chain, do not delete it so it can be reused. The instance is also unlinked from the active list.

These last four opcodes allow us to run a create-init-perf-delete cycle inside an instrument. They also pave the way for us to start offering means of setting arbitrary performance orders. For example, 

```
instr 1

 // run at i-time
  myInstr:InstrDef create {{
        out linenr(oscili(p4,k(p5)),0.1,0.1,0.01)
        }}
  myInstance:Instr create myInstr
  err1:i init myInstance,0.5,440

 // run at perf-time
  err2:k perf myInstance
  glid:k expon 440, p3, 880
         setp myInstance, 5, glid

  // run at deinit time
   delete myInstance 
   delete myInstr

endin
```

NB: having a `create` for `Instr` also seems to make it logical to overload it for `InstrDef` and remove `createinstr` which was never ideal.


vlazzarini:
Active list splicing
-------------

It is now possible to splice an instance into the active list in front or after another instance.

```
err:i  splice  Instr1, Instr2, mode
```

when mode = 0, Instr1 is spliced into the active list *before* Instr2 and if mode != 0, it is spliced *after* Instr2. This works for any instances whether they are already linked into the active list or not. The following example shows an unliked instance being spliced into the list

```
instr 3
 printk 0.1, p1
 myInst:Instr create create({{ printk 0.1, p1 }})
 err:i init myInst
 err splice myInst, this, 0
endin
```

prints

```
i   5 time     0.00000:     5.00000
i   3 time     0.00000:     3.00000
````

A linked instance can also be re-spliced:

```
instr 4
 printk 0.5, p1
 myInst:Instr play create({{ printk 0.5, p1 }})
 err:i splice this, myInst, 1
endin
```

prints

```
 i   5 time     0.00000:     5.00000
 i   4 time     0.00000:     4.00000
```

`this` is a readonly instance reference to self.

File: Engine/csound_standard_types.c [hunk 1]
```diff
@@ -385,12 +403,16 @@ const CS_TYPE CS_VAR_TYPE_ARRAY = {
   array_free_var_mem, NULL, 0
 };
 
-
 const CS_TYPE CS_VAR_TYPE_INSTR = {
   "InstrDef", "instrument definition reference", CS_ARG_TYPE_BOTH,
   createInstrRef, instrRef_copy_value, NULL, NULL, 0
 };
 
+const CS_TYPE CS_VAR_TYPE_INSTANCE = {
```
kunstmusik: The name CS_VAR_TYPE_INSTANCE seems very generic, suggest this changes to CS_VAR_TYPE_INSTR_INSTANCE.

tests/commandline/test_create_init_perf_delete.csd
```diff
@@ -0,0 +1,34 @@
+<CsoundSynthesizer>
+<CsOptions>
+-n
+</CsOptions>
+<CsInstruments>
+0dbfs = 1
+
+opcode Osci(a:k,f:k):a
+xout linenr(oscili(p4,k(p5)),0.1,0.1,0.01)
```
kunstmusik: Odd example that this opcode is taking in two params but not using them. Is the naming of a and f for arg names meant to test something in particular?

vlazzarini: done

Engine/csound_standard_types.c
vlazzarini: done

tests/commandline/test_create_init_perf_delete.csd
```diff
@@ -0,0 +1,34 @@
+<CsoundSynthesizer>
```
kunstmusik, line 1: These tests have some awkard indentations and also use older-style opcode lines. Suggest to use:

```
<CsoundSynthesizer>
<CsOptions>
-n
</CsOptions>
<CsInstruments>
0dbfs = 1

opcode Osci(a:k,f:k):a
    xout linenr(oscili(a,f),0.1,0.1,0.01)
endop

instr 1
    // run at i-time
    myInstr:InstrDef = create({{ out Osci(p4,k(p5)) }})
    myInstance:Instr =create(myInstr)
    err1:i = init(myInstance,0.5,440)

    // run at perf-time
    err2:k = perf(myInstance)
    slid:k = expon(440, p3, 880)
    setp(myInstance, 5, slid)

    // run at deinit time
    delete(myInstance)
    delete(myInstr)
endin

</CsInstruments>
<CsScore>
i1 0 2
</CsScore>
</CsoundSynthesizer>
```

kunstmusik:
I think some things that are confusing to me:

1. If you schedule an instrument to play at x time, it's returning an Instr reference. From there can you call play/pause/delete/splice/etc., even though it potentially isn't running yet? 

2. Should we also have an is_playing opcode to get the status of an Instr? 

3. Is it possible to work completely outside of the active lists altogether? I.e., create instrument instances and put them into an array, then let the user deal with calling perform on an instrument instances, reorder the instances, etc. Feels like permitting complete control over instrument ordering would be ideal. 

4. The init opcode here to run init-pass is a little awkward compared to all of the other definitions of init opcodes (i-time assignment, instantiating user-defined types). That said, I'm not sure of a better name to use here either. 

5. Using "create" seems a bit strange to me looking at the example code for defining an InstrDef. What about "compile" or "compile_instr"? Also "alloc" vs "create" for Instr instances?  Or "new"?  Using new could be interesting as we could also change how UDTs work and remove overloading init there and instead have code like:

```
myVar:TestType = new(TestType, 1,2,3,"test")
```

or

```
myVar:TestType = new TestType(1,2,3,"test")
```

if we make it a keyword.

vlazzarini:
> I think some things that are confusing to me:
> 
> 1. If you schedule an instrument to play at x time, it's returning an Instr reference. From there can you call play/pause/delete/splice/etc., even though it potentially isn't running yet?

If it is not running yet, it can't be spliced because the instance is NULL. Nothing happens. Originally I thought it wasn't an error but now I've changed it so that splice returns -1 in this case. Other opcodes also check for this.

> 2. Should we also have an is_playing opcode to get the status of an Instr?

`is_playing` might need some additions to INSDS or INSTANCEREF, and some thought in terms of instances not on active list, but we can check if it has been instantiated very easily. Maybe I can implement it and leave the `is_playing` for a later PR?

> 3. Is it possible to work completely outside of the active lists altogether? I.e., create instrument instances and put them into an array, then let the user deal with calling perform on an instrument instances, reorder the instances, etc. Feels like permitting complete control over instrument ordering would be ideal.

Absolutely. That's the point of some of the opcodes I've introduced such as `create`, `init` and `perf`. When we use these, instances are outside of all chains. 

> 4. The init opcode here to run init-pass is a little awkward compared to all of the other definitions of init opcodes (i-time assignment, instantiating user-defined types). That said, I'm not sure of a better name to use here either.

I think we can keep `init` or we can also add an `initinstr` (and `perfinstr`) overload.

> 5. Using "create" seems a bit strange to me looking at the example code for defining an InstrDef. What about "compile" or "compile_instr"? Also "alloc" vs "create" for Instr instances?  Or "new"?  Using new could be interesting as we could also change how UDTs work and remove overloading init there and instead have code like:
> 
> ```
> myVar:TestType = new(TestType, 1,2,3,"test")
> ```
> 
> or
> 
> ```
> myVar:TestType = new TestType(1,2,3,"test")
> ```
> 
> if we make it a keyword.

I like `new` I think we can replace `create`  everywhere for `new` and also introduce this opcode for arrays as an overload to `init`.

```
myInstrDef:InstrDef  = new({{ out oscili(p4, p5) }})
myInstr:Instr  = new(myInstrDef)
err1:i  = init(myInstr, 0.5, 440)
err2:k  = perf(myInstr)
``` 

This looks very neat to me.

Thanks!

kunstmusik:
I think we may need a bigger discussion of syntax for "new" and/or introducing constructor opcodes, i.e., if you define a struct or instrDef, it might be easier to use if we create instances by using their names, such as:

```
struct ABC val:i, val2:i

myVal:ABC = ABC(1,2)
// or myVal:ABC = new ABC(1,2)
// or myVal:ABC = new(ABC, 1, 2)
```

One concern I have here with Instr instances is one that arises with structs as well as Opcode instances, which is variable data capturing. It's not a simple problem IMO, but for these things all to work together, it should be possible to construct an struct, opcode, or instrument instance in one context and use it in another context. For example, an instrument instance might be created and added to a global array of instrument instances, while the array is processed by another performance instrument. The data pointers used for calls to init() (as it's currently defined for Instr instances) might go out of scope between when it's set and when it's used. 

Perhaps not as much an issue with Instr instances in their current form where init arg types are using the same as what we support from the score, but we might want to extend that further in the future to accepts things like opcode instances, arrays, etc. 

That in turn brings up issues of initialization in one audio context (i.e., local ksmps or local sr) and execution in another context. (Again, more of an issue for structs and opcode instances.) For Instr instances, we'll also need to have clarity on how they interact with reinit as well as release time.

Thinking of release time, would it be possible to call init() multiple times on an instr instance, as if it was being processed as tied notes? i.e., could call init() with negative p3 to turn on the instrument indefinitely, then later call init() again with positive p3. (Is that supported already?) I think besides `is_playing`, we might also need `is_releasing` as an opcode. (Trying to think through what features would be necessary to do full management of voices from user-code using this system.)

vlazzarini:
> I think we may need a bigger discussion of syntax for "new" and/or introducing constructor opcodes, i.e., if you define a struct or instrDef, it might be easier to use if we create instances by using their names, such as:
> 
> ```
> struct ABC val:i, val2:i
> 
> myVal:ABC = ABC(1,2)
> // or myVal:ABC = new ABC(1,2)
> // or myVal:ABC = new(ABC, 1, 2)
> ```
> 

While I understand this, I think we need to make a decision, otherwise we will never move on to finish the work. If new is not good, let's take create as proposed.


> One concern I have here with Instr instances is one that arises with structs as well as Opcode instances, which is variable data capturing. It's not a simple problem IMO, but for these things all to work together, it should be possible to construct an struct, opcode, or instrument instance in one context and use it in another context. For example, an instrument instance might be created and added to a global array of instrument instances, while the array is processed by another performance instrument. The data pointers used for calls to init() (as it's currently defined for Instr instances) might go out of scope between when it's set and when it's used.

The way it works at the moment is that if data pointers refer to instantiated instruments they can be used. If however they have been deleted, trying to use them throws an error. So an instrument can create an instance, it can be copied into a global var and passed to another instr. If that first instr deletes the instance, any attempts to act on it in the other instr will return an error. 

We can create an instr, etc in one context and use it in another context, provided that

- an instr has not been initialised
- if initialised, then the context matches the one in which if was initialised

That seems the logical thing to do. To support divergent contexts would require a very complicated mechanism. It does not seem reasonable to me.

> 
> Perhaps not as much an issue with Instr instances in their current form where init arg types are using the same as what we support from the score, but we might want to extend that further in the future to accepts things like opcode instances, arrays, etc.
> 

Opcode types are treated in a similar form (but that's for the other PR). 

> That in turn brings up issues of initialization in one audio context (i.e., local ksmps or local sr) and execution in another context. (Again, more of an issue for structs and opcode instances.) For Instr instances, we'll also need to have clarity on how they interact with reinit as well as release time.

The context of the initialisation is key to define all init aspects - release time, ksmps, sr etc. A context check (which actually I should add as I did to opcode types) should be made and yield an error in the case of diverging ~~ksmps~~/sr. ~~The release time should be treated in the same way actually~~. I will add this check to perf(), which is the only one where this is relevant because in this case init is done by separate opcode.

EDIT: release time should not be treated as a mismatch. There are two situations:
- release time less than calling instrument: this is a normal occurrence in an instrument as we may have opcodes with different release times and the large overall sets the ip->xtratim.
- release time longer than calling instrument: also not an issue because the result is a click, an ordinary user error that can arise in many other situations and we should not make a big fuss about it.

Diverging ksmps is not a problem here because it is handled correctly when opcodes are run. The only issue is diverging sr, which can occur only in an UDO at the moment and will not be handled correctly as over/undersampling will be applied by calling perf() more or fewer times than it should. I have added code to handle this now.

EDIT END.

Reinit works as before, when that is called at perf time an init pass is run. Because we require that the perf happens in a matching context to the init call, then this is not going to raise any issues. However, I think we need to pass the reinit flag value from caller at perf time so that any reinit in the instrument can also be triggered if the calling instr is in reinit. I will investigate that.

> 
> Thinking of release time, would it be possible to call init() multiple times on an instr instance, as if it was being processed as tied notes? i.e., could call init() with negative p3 to turn on the instrument indefinitely, then later call init() again with positive p3. (Is that supported already?) I think besides `is_playing`, we might also need `is_releasing` as an opcode. (Trying to think through what features would be necessary to do full management of voices from user-code using this system.)

That is possible. The init() opcode can be called any number of times for an instance. As I said before - is_playing() is beyond the scope of this PR (there's some thought needed), but we already have is_releasing(), it's called release().

kunstmusik:
Yes I think we can address create/new in a separate issue and we should as it would touch upon things outside this PR. 

Also I think there's a misunderstanding, as I don't see a release() opcode that works with an instrument instance like:

```
if(is_releasing(instrInstance) == 0) then
  ;; now remove from my self-managed instr instance list
endif
```

Regardless, I think we can merge what's here and address further concerns in a separate PR as the context issues look settled here. I'll file a couple issues in the issue tracker for new/create and further opcodes for instr instance management then approve/merge this one.

kunstmusik:
Added #2092 and #2093 to follow up from this PR.

kunstmusik (approved):

vlazzarini:
Thanks. There is delete() that does that, called at deinit time, that is after the opcode is released. Is that what you mean?

PR #2048, opened 2024-11-28
Bugfix: boolean with explicit types

vlazzarini:
There was a nasty bug created by the fact that an old function `argtyp2()` was still being used in the codebase despite the fact that it only works for implicit types.

This was used to select the correct boolean type for a synthetic value and since it could not track the type from the first letter, it was always selecting `b` (i-time boolean). As a result, this code failed miserably (even though there is nothing wrong with it)

```
instr 1
  err:k init 0
  if err == 0 then
    printk2 err
    err = 1;
  endif
endin
```

I replace this with the correct type checking function and got rid of it.  I also removed a deprecated single-char type member in `TEXT`. 

Added test.


kunstmusik (dismissed):

kunstmusik:
@vlazzarini The code change looks good but this branch now has a conflict with develop after I merged the other PR. Could you by chance update this branch with develop?

vlazzarini:
Will do!

vlazzarini:
> @vlazzarini The code change looks good but this branch now has a conflict with develop after I merged the other PR. Could you by chance update this branch with develop?

Done now. All ready.

kunstmusik (approved):
Thanks, looks good, merging now.

PR #2049, opened 2024-11-28
Fixed memory access after free issue on instr redefinition fail

vlazzarini:
Using address sanitizer, I picked up what looks like a longstanding issue where an instrument redefinition fail  was leading to access after free in `orcompact()`. This is fixed and a test has been added.

### DIFF
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -1367,6 +1367,14 @@ void insert_instrtxt(CSOUND *csound, INSTRTXT *instrtxt, int32 instrNum,
                      ENGINE_STATE *engineState, int32_t merge) {
   int32_t i;
 
+  /* redefinition not allowed in the same compilation */
+  if(UNLIKELY(engineState->instrtxtp[instrNum] != NULL) &&
+     !merge) {
+      synterr(csound, Str("instr %d redefined in code:\n"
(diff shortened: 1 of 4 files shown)


kunstmusik (approved):
Merging in, but I'll not that being unable to redefine an instrument in the same compilation has been something I wish wasn't a restriction. I realize this is what's made taking some of my live coding sessions and converting them into standalone files a problem as they would try to redefine an instrument in the same compilation pass and I'd have to use some hacks to work around that. Perhaps we can look at that matter separately.

vlazzarini:
This wasn't allowed before but I guess we can work on introducing it. Leave it with me, I can look at it, but if you could open a ticket it would help.

vlazzarini:
> Merging in, but I'll not that being unable to redefine an instrument in the same compilation has been something I wish wasn't a restriction. I realize this is what's made taking some of my live coding sessions and converting them into standalone files a problem as they would try to redefine an instrument in the same compilation pass and I'd have to use some hacks to work around that. Perhaps we can look at that matter separately.

I've got this now working on a branch, should I push it to origin and make a PR?

kunstmusik:
I think a PR would be great, would love to try it out!

vlazzarini:
Done.

PR #2050, opened 2024-12-01
Fixed bug in implicit global variable search

vlazzarini:
Changes to introduce global variables had a side-effect where explicit types starting with "g" were being searched for as implicit global variables. This was caused by still relying on the first letter for finding variables and arguments. 

This PR fixes the issue by removing character-based checks and using the typePools which hold variables in the different scopes. The first change was in the semantic analysis and the second in the compiler argument setting, both still using a check for 'g'. I also improved the explicit type test to check for this.


Engine/csound_orc_semantics.c
```diff
@@ -1517,6 +1502,7 @@ void add_arg(CSOUND* csound, char* varName, char* annotation,
       pool = find_global_annotation(varName, typeTable);
       type = csoundGetTypeWithVarTypeName(csound->typePool, annotation);
       typeArg = (void *) type;
+      printf("%s \n", varName);
```
kunstmusik: Think this is a leftover debug print statement?

vlazzarini: yes, thanks for spotting it

kunstmusik (approved):
Tested here with some CSDs and live coding experiments and all seemed to work. Merging now, thanks!

Issue #2042, opened 2024-11-25
Fix precedence so that SetOption() will always override .csound7rc


kunstmusik:
While reviewing PR #2035 I found that the tests failed due to having a .csound7rc file that contained `-odac -+rtaudio=pa_bl`. This ended up showing that API apps that call SetOption() but do not use CSD files will have values overridden by anything in .csound7rc. There is a call to checkOptions() in main.c:csoundStart() that is triggered in this scenario that causes the issue. 

The expected behavior is that .csound7rc has lower precedence than either CsOptions, commandline flags, or SetOption values (preferably in that order).

vlazzarini:
Here's a possible solution:

- in csoundSetOption() we call checkOptions() if csound->csd is zero , and set it so it is not called again (as csoundSetOption() may be called more than once)
- then we proceed to set the arguments so the precedence is the same as in csoundCompileArgs().

vlazzarini:
PR open so closing this.

PR #2051, opened 2024-12-01
changed precedence for csoundSetOption

vlazzarini:
This PR sets the precedence so that csoundSetOption always overrides .csound7rc, as requested in #2042


Top/main.c
```diff
@@ -242,405 +237,403 @@ int32_t csoundCompileArgs(CSOUND *csound, int32_t argc, const char **argv)
...
-    int32_t     n;
+  OPARMS *O = csound->oparms;
+  int32_t n;
 
-    /* if a CSD was not used, check options */
-    if (csound->csdname == NULL)
-          checkOptions(csound);
+  /* if a CSD was not used or options were not checked, check options */
+  if (csound->csdname == NULL && !csound->options_checked) {
+    checkOptions(csound);
```
kunstmusik, line 428: Should this set `csound->options_checked = 1` here?

kunstmusik: This was marked resolved but I didn't see a code change or a reply, unresolving for now.

vlazzarini: Maybe my comment did not come out for some reason.

It's not relevant as csoundSetOptions only runs before first compilation, never after. This code here can only be reached again after reset.

kunstmusik (approved):
Lot of formatting changes made this one a little hard to review, but looks good as far as I can see.

vlazzarini:
yep, my mistake there. I thought I'd try some formatting in this but it really messed my subsequent merging to and from develop. Will not make that mistake again.

PR #2052, opened 2024-12-01
Allow instrument redefinition in code

vlazzarini:
This PR introduces changes to allow instruments to be redefined in the code (single compilation). The default behaviour is to issue a syntax error and fail the compilation. This PR brings a command-line option that suppresses this and allows instruments to be multiply defined, 

```
--allow-redefinition
```

in this case the last definition seen by the compiler is the one that is kept.


kunstmusik:
I tried this but it doesn't seem to work. Attached a test CSD and livecode.orc file. If I don't turn on --alllow-redefinition, it doesn't give an error with instr P1, and if I turn it on it doesn't seem to evaluate the P1 instrument. Not sure if it has to do with #include processing?

[test_code.zip](https://github.com/user-attachments/files/18246601/test_code.zip)

vlazzarini:
what happens if you use current develop code, do you get an error?

vlazzarini:
it's to do with named instruments. It happens on develop as well.

EDIT: the second instrument definition is ignored. I also noticed that the parser seems to have trouble to set P1 as a instr def constant (it sets it to zero). Anything with P followed by a number is problematic.

vlazzarini:
I think I have this working now. There were a number of things that were wrong:

1 - In the original code (inherited from 6.x), named instruments could not be redefined in the same compilation but no syntax error or error messages. They just did not get redefined, which is confusing.

2 - There was no support for the `--allow-redefinition` for named instruments, it was not implemented in that case.

3 - The compiler was trapping arguments starting with capital P followed by a number (e.g. `P1`) as a p-field (that doesn't seem right, but it may be a established behaviour). So if you call your instrument `P1`, the variable `P1` is passed as an argument to schedule, but as a p-field and it doesn't work as expected.

So I fixed 1 & 2 above by extending what I implemented for numbered instruments to named instruments.

I fixed 3 by changing the order of argument trapping so any PN named global vars are searched for first. Possibly we could change the code so that only lower-case pN is trapped but that may change behaviour and break code. For now this is enough.

Also note that while doing this work I found a memory bug in strcat (from your test code) and put through a PR to fix.

Also: the ubuntu build on develop is broken. I have fixed it in  my PRs but it is annoying when I start a PR from develop and forget about it. We need to merge this build fix asap.

kunstmusik:
Re-tested with latest and it seems to work as expected both when using and not using --allow-redefinition. I've been using P1 as an instrument name for years so it seems like a regression. We can investigate further later and merge this now.

kunstmusik (approved):

vlazzarini:
The problem with P1 is that now instrument names are global instrDef constants. So the problem would not have arisen. I only noticed it existed because of this new feature.

If anyone else tried making a variable called P1 then there would be an issue. I have dealt with this more thoroughly in the latest PR #2069

PR #2053, opened 2024-12-02
Fixing bug in callback setting

vlazzarini:
This PR fixes a bug where some callbacks were not being set because of a typo in the code. This was leading to segfault in tests under Windows.


vlazzarini:
not needed since the fix had gone in already.

PR #2054, opened 2024-12-03
OPCODINFO back to public access

vlazzarini:
As the title says, OPCODEINFO is public again now.


vlazzarini:
Checking to see if it should be public. Let's wait.

PR #2058, opened 2024-12-05
Update ioscross build to only run on develop branch

nonameentername:


vlazzarini:
Should we enable the skipped tests to make sure they run OK before merging? That is to avoid having a failed build on develop and a red badge.

vlazzarini:
We may not need this PR as I can see now that all builds went ahead on develop as I merged an other PR.

nonameentername:
@vlazzarini ok, I'm going to go ahead and close this pr.  Not sure why the variables are not being used when creating pr's, but are being used on merge to development branch.

nonameentername:
> [!NOTE]  
> The variables are not being used when the pr is from a fork of the project.

In order for the variables to be populated the merge request needs to be from the same organization and not a fork.

vlazzarini:
thanks for clarifying.

PR #2060, opened 2024-12-06
Bugfix/float builds

vlazzarini:
A number of issues were detected in the 32-bit floats builds. This PR deals with these:

- variable defined outside the USE_DOUBLE guards unused in floats build.
- table sizes need to be limited to 2^24 size in floats build because there is not enough range to properly handle larger numbers in MYFLT. Since data in score and orchestra is represented using MYFLT, it is not possible to correctly handle integers larger than 2^24. Table sizes cannot be accurately set beyond 2^24.
- Some tests were failing because double constants were assumed.
- lrint was not being searched for correctly and it was not being found on MacOs.

I also introduced `var:i floatsize` to return the size of MYFLT so code can be written to depend on this.
CSD test for max table size was edit to take this into account. A new unit test was added to make sure
large table sizes can be set properly in doubles version.

Finally, I have set alternate MacOS and Linux builds to use 32-bit floats instead of doubles so we can catch any issues that may arise in the future.


CMakeLists.txt
```diff
@@ -597,10 +597,15 @@ if (NOT MATH_LIBRARY)
     set(MATH_LIBRARY "")
 endif()
 
+if(APPLE)
+include(CheckSymbolExists)
+check_symbol_exists(lrint math.h FOUND_LRINT)
```
kunstmusik, line 602: Any reason to change from HAVE_LRINT to FOUND_LRINT? All of our other checks use "HAVE_" as far as I understand.

tests/commandline/test.py
```diff
@@ -177,7 +177,6 @@ def runTest():
     ["test_plusname.csd", "test +Name for instr name"],
     ["testnewline.csd", "test newline in statements"],
     ["testmidichannels.csd", "test use of mapped multiport channels"],
-    ["test_max_table_len.csd", "test max table length"],
```
kunstmusik, line 180: Looks like this test was removed from the suite, was this intentional?

Engine/entry1.c
```diff
@@ -166,6 +166,7 @@ OENTRY opcodlst_1[] = {
   { "init.a", S(ASSIGNM),0,       "mmmmmmmmmmmmmmmmmmmmmmmm", "m", mainit },
   { "nstrnum",   S(IREF_NUM),0,    "i",    ":InstrDef;",   (SUBR) get_instr_num },
   { "init.instr", S(ASSIGN) ,0,  ":InstrDef;", ":InstrDef;", (SUBR) copyVarGenericInit},
+  { "floatsize", S(ASSIGN) ,0, "i", "", myflt_size },
```
kunstmusik, line 169: Just a note, my first reaction to this was why not a variable? But I think I'm 50/50 on it as an opcode and seems fine.

CMakeLists.txt
vlazzarini: It was the only I could make it work.

tests/commandline/test.py
vlazzarini: yes, I moved it to a unit test as I could make it test the difference between float build and double build better.

Engine/entry1.c
vlazzarini: I thought opcode was easier to implement.

tests/commandline/test.py
kunstmusik: Probably should delete the unused test csd then

kunstmusik: Maybe the current test_max_table_len.csd can be moved to the manual as an example to use for floatsize?

kunstmusik (approved):
Looks good, merging now, thanks!

vlazzarini:
thanks, it will be good to keep an eye on 32bit, we  often forget about it.

PR #2065, opened 2024-12-18
Sample accurate mode fix for RT events

vlazzarini:
While testing the sample accurate mode, I discovered that realtime events were not always correctly handled. A bug was found where events were sometimes shifted one kcycle forward because rounding instead of floor() was used (only for RT events). That was fixed and I added a test to check for correct insertion of sample-accurate event.

### DIFF
--- a/Engine/entry1.c
+++ b/Engine/entry1.c
@@ -134,6 +134,9 @@ OENTRY opcodlst_1[] = {
   { "midglobal",S(MIDGLOBAL),0,   "",     "Sm", midglobal, NULL, NULL, NULL},
   { "ihold",  S(LINK),0,          "",     "",     ihold, NULL, NULL, NULL  },
   { "turnoff",S(LINK),0,           "",     "",     NULL,   turnoff, NULL, NULL },
+  { "offsetsmps", S(AOP), 0, "i", "", (SUBR) sa_offset, NULL, NULL},
+  { "offsetsmps", S(AOP), 0, "k", "", NULL,(SUBR) sa_offset, NULL},
+  { "earlysmps", S(AOP), 0, "k", "", NULL,(SUBR) sa_early, NULL},
   { "str", S(IREF_NUM) ,0,  "S", ":InstrDef;", (SUBR) get_instr_name},
   /* VL: 10.2.22 this was thread  but with parser3 we need to make string assignment on threads 1 & 2 */
   {  "=.S",   S(STRCPY_OP),0,     "S",    "S",
--- a/Engine/entry1.c
+++ b/Engine/entry1.c
@@ -450,7 +450,7 @@ OENTRY opcodlst_1[] = {
   { "expsegb", S(EXXPSEG),0,       "k",    "iim",  xsgset_bkpt, kxpseg, NULL },
   { "expsegb.a", S(EXXPSEG),0,      "a",    "iim",  xsgset_bkpt, expseg },
   { "expsega",S(EXPSEG2),0,       "a",    "iim",  xsgset2, expseg2  },
-  { "exp=segba",S(EXPSEG2),0,       "a",    "iim",  xsgset2b, expseg2 },
+  { "expsegba",S(EXPSEG2),0,       "a",    "iim",  xsgset2b, expseg2 },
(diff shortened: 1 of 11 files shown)


vlazzarini:
Two other fixes:

- Adjusted vaget/vaset not to fail if index outside of sa bounds.
- sa parameters were not being propagated to UDOs, now they are.

Incidentally, ubuntu build was failing because of fltk, but fltk is not a dependency anymore; removed.

vlazzarini:
This is complete now, ready for review & merge
- fixed final issue with UDOs in generic local ksmps case
- improved CSD test.

Engine/entry1.c
```diff
@@ -134,6 +134,9 @@ OENTRY opcodlst_1[] = {
   { "midglobal",S(MIDGLOBAL),0,   "",     "Sm", midglobal, NULL, NULL, NULL},
   { "ihold",  S(LINK),0,          "",     "",     ihold, NULL, NULL, NULL  },
   { "turnoff",S(LINK),0,           "",     "",     NULL,   turnoff, NULL, NULL },
+  { "offsetsmps", S(AOP), 0, "i", "", (SUBR) sa_offset, NULL, NULL},
+  { "offsetsmps", S(AOP), 0, "k", "", NULL,(SUBR) sa_offset, NULL},
+  { "earlysmps", S(AOP), 0, "k", "", NULL,(SUBR) sa_early, NULL},
```
kunstmusik, line 138: Only question I have for this PR is an earlysmps should be added with i-rate output, to match the offsetsmps with i-rate output. I suppose it's possible someone might want to do init-time processing and the event both starts mid-buffer and ends before the end of the ksmps buffer (if there's a large ksmps size).

vlazzarini: the problem is that earlysmps will only return 0 at i-time. It is set at the last cycle of performance.

kunstmusik: Confused then here, so if an event both starts mid-buffer and ends before end of buffer (e.g., ksmps = 512, event starts 32 smps in, ends after 64 smps), there's no way to know earlysmps at init-time (in native code too)?  Is this a deeper problem that needs further internal work?

vlazzarini: What I am saying is that the  variable is set at the last cycle, before that it's zero. That's how it works. It does not make sense to have it set at init time (it's only meaningful at perf time anyway). 

The value is computed at the insertion, as offset is too. Possibly the misleading thing is to have offset as an i-time overload.

I don't think we should have earlysmps as i-time, that gives the wrong idea.

kunstmusik: i-time offsetsmps certainly gives me the impression both should be available. What I'm saying here is if the first cycle *IS* the last cycle, if the duration of the event is shorter than one ksmps buffer length.

vlazzarini: In that case, at perf-time earlysmps will be immediately non-zero, but it does not have a meaning at i-time.

kunstmusik: Is there a use case for having offsetsmps as init-rate then? Maybe best to get rid of that and get rid of any ambiguity and just force users to use both at perf-time.

vlazzarini: I agree, just k-time opcodes, removes any confusion.

vlazzarini: overlad removed.

Engine/entry1.c
```diff
@@ -450,7 +450,7 @@ OENTRY opcodlst_1[] = {
   { "expsegb", S(EXXPSEG),0,       "k",    "iim",  xsgset_bkpt, kxpseg, NULL },
   { "expsegb.a", S(EXXPSEG),0,      "a",    "iim",  xsgset_bkpt, expseg },
   { "expsega",S(EXPSEG2),0,       "a",    "iim",  xsgset2, expseg2  },
-  { "expsegba",S(EXPSEG2),0,       "a",    "iim",  xsgset2b, expseg2 },
+  { "exp=segba",S(EXPSEG2),0,       "a",    "iim",  xsgset2b, expseg2 },
```
kunstmusik: typo introduced here

vlazzarini: fixed.

kunstmusik (approved):

kunstmusik:
Looks good to me now, going to merge, thanks!

PR #2066, opened 2024-12-20
Replace OPDS by INSDS as variable context

vlazzarini:
This PR replaces OPDS * by INSDS * as the context for variables as it makes more sense. It also saves the context to the variable data structure.

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Install Dependencies
         run: |
           sudo apt-get update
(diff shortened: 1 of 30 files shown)


kunstmusik (approved):
Makes sense; approving but will wait to merge when I hear from @vlazzarini whether this should get merged before or after resolving/merging #2047.

vlazzarini:
> Makes sense; approving but will wait to merge when I hear from @vlazzarini whether this should get merged before or after resolving/merging #2047.

We can go ahead with this and I can merge it with the other branches I'm working on.

PR #2069, opened 2024-12-26
Feature/pfields

vlazzarini:
This PR improves the check for pfields by replacing `pnum()` by functions that use the type system.

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -50,7 +50,6 @@ char *remove_type_quoting(CSOUND *csound, const char *outype);
 /* from csound_orc_compile.c */
 extern int32_t argsRequired(char* arrayName);
 extern char** splitArgs(CSOUND* csound, char* argString);
-extern int32_t pnum(char*);
 OENTRIES* find_opcode2(CSOUND*, char*);
 char* resolve_opcode_get_outarg(CSOUND* csound,
                                 OENTRIES* entries, char* inArgTypes);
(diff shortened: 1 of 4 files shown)


vlazzarini:
Disabled the mingw/vcpkg build because it is failing to configure. This needs attention before it is put back online.

EDIT: it's been fixed and put back.

Engine/csound_orc_semantics.c
```diff
@@ -299,6 +298,25 @@ CS_VARIABLE* find_var_from_pools(CSOUND* csound, char* varName,
   return var;
 }
 
+/*
+  Check a symbol for pfield format (pN, PN)
+  and return the p-field num ( >= 0 ) 
+  else return -1  
+*/
+static int32_t IsPfield(CSOUND *csound, TYPE_TABLE* typeTable, char *s)
```
kunstmusik: Would prefer this to be renamed is_pfield to match function naming style in rest of codebase. In general the pascal/C# conventions of function naming with CapitalCamelCase has started to creep into our codebase and I think it's not good to mix styles. We could have one style designated for the API vs. internal coding, but just need some rule for this to make things clear.

kunstmusik: That said, we also have camelCase-named functions so there's already an inconsistency; my preference for snake_case is that it seems better in-line for C conventions.

vlazzarini: I am happy to follow any conventions that are set. Probably a job for a single PR to rename all internal functions according to an agreed style.

vlazzarini: I've renamed the functions. It makes sense to adopt ***_***() for new functions and then slowly rename others as we go along.  I have not paid attention to this but I think I have mostly using that format in my additions.

kunstmusik (approved):

PR #2071, opened 2024-12-30
Opcode Types and Operators

vlazzarini:
This PR introduces two types to manipulate opcodes and a number of opcodes that operate on them. The main principle behind the approach is that there are two steps in selecting and instantiating opcodes

1) Get a list of all overloads for a given name
2) Select a suitable opcode and instantiate

With an instantiated opcode, we can then operate on it, running init and perf passes by separate operators, or together by one single operator. Particularly useful is to be able to create arrays of instances and operate on them.

Below is a detailed description of the additions.

Opcode Reference Type
===

This type, named `OpcodeDef`, is designed to hold a reference to all existing overloads of an opcode. Opcodes can be either internal, external, or UDOs (essentially anything that is listed as an opcode in the system).

An `OpcodeDef` variable can be created by initialising it with a name:

```
ref:OpcodeDef init "oscili"
```

We can obtain a list of all overloads using `opcodeinfo`:

```
opcodeInfo ref
```
which prints

```
oscili: 10 overloads
(0)	oscili.a	out-types: a	in-types: kkjo
(1)	oscili.kk	out-types: k	in-types: kkjo
(2)	oscili.ka	out-types: a	in-types: kajo
(3)	oscili.ak	out-types: a	in-types: akjo
(4)	oscili.aa	out-types: a	in-types: aajo
(5)	oscili.aA	out-types: a	in-types: kki[]o
(6)	oscili.kkA	out-types: k	in-types: kki[]o
(7)	oscili.kaA	out-types: a	in-types: kai[]o
(8)	oscili.akA	out-types: a	in-types: aki[]o
(9)	oscili.aaA	out-types: a	in-types: aai[]o
```

Opcode Object Type
===

From an opcode reference, we can instantiate an opcode object, which is held by a variable of the type `Opcode`. For this, we have the `create` function, taking an opcode reference and an optional overload index (defaults to 0)

```
obj:Opcode create ref [,overload]
```

This instantiates the opcode and we can confirm its signature also using `opcodeinfo`:

```
opcodeInfo obj
```

which prints

```
oscili.a 	out-types: a	in-types: kkjo 
```
Opcode Operators
===

Besides the information opcodes discussed above, a number of other operators are provided. All operations are context-aware and checks are made for consistency in this regard.

Opcode Object Initialisation
-----

Once an opcode object is created, we can run an init pass on it, for example

```
a1  init obj, p4, p5
```

Type checking is performed to make sure the arguments are correct for the object signature. An init error is thrown if they are not. Type checking uses the rules as set out by the parser, but it is executed at run time rather than compile time.

The init opcode is stateless, so it can be run on different objects passed to it, in a loop for instance. 

Opcode Object Performance
-----

Equally, we can run a perf pass on an object, for example

```
a1  perf obj, p4, p5
```

Type checking is also performed to ensure no incorrect types have been passed, but only if arguments have changed from the init-time pass. This opcode is also stateless.

Both init and perf-time can be combined in a single opcode call using `run`

```
a1 run obj, p4, p5
```

Opcode Object Arrays
------

One of the most typical uses of opcode objects is to run several copies of the same opcode in a loop. For this, we would use an opcode object array, which can be created similarly to a single object, by passing a reference, number of opcodes in the array, and an optional overload index,

```
obj:Opcode[] create ref, num [,overload]
```

Here's an example

```
instr 1
 ival[] fillarray p5, p5*1.5
 asig[] init 2
 ref:OpcodeDef init "oscili"
 obj:Opcode[] create ref,2
 n:i = 0
 while n < 2 do
  asig[n] init obj[n], p4, ival[n]
  n += 1
 od
 m:k = 0
 while m < 2 do
  asig[m] perf obj[m], p4, ival[m]
  out asig[m]
  m += 1
 od
endin
```

Because loops are either run on init-time or on perf-time, we need two loops, and so the separate init and perf opcodes allow this use case. This type of code allows various different types of configurations and connections (serial, parallel) as it is all done in Csound code.

For straightforward applications (like in the example) where we only want to run several instances of the same opcode at one time, we can use a`run` overload that takes an opcode object array, as in this equivalent example

```
instr 2
 ival[] fillarray p5, p5*1.5
 asig[] init 2
 ref:OpcodeDef init "oscili"
 opcodeinfo ref
 obj:Opcode[] create ref, 2
 opcodeinfo obj[0]
 asig run obj, p4, ival, -1
 out asig[0] + asig[1]
endin
```

Object Deallocation
------

Normally, we do not need to be concerned with object deallocation, as variable memory is managed by Csound. However, in some cases it may be useful to set objects to be deleted at deinit time.

```
delete obj
```

Overloads for single and array objects are provided.

Note that some opcodes that hold resources such as file handles may have issues at deactivation time (on reset or at the end of performance) as Csound will attempt to access these and a access-after-free condition may arise. For this reason, it is best to be sure that the opcode dataspace can be disposed of, before using `delete`.


kunstmusik:
Haven't looked through this yet but could the names be changed to OpcodeDef and Opcode, to match InstrDef and Instr? Also some questions/comments:

1. Does perf really require any pfields beyond the Opcode object? I would have expected arguments to the struct to be connected at init (perhaps this will make more sense once I look at the implementation...)
2. Is there any attention to passing opcode objects between processing contexts? i.e., if created with an sr or ksmps that differs from where it is used. May need some warnings or other processing added for those situations. 
3. For printing, would be good to eventually move away from things like kkjo and showing the internal type string. (and perhaps for us to move away from using internal type strings altogether). Not something that is necessary immediately though. 
4. Perhaps we could look at adding opcode definitions to the global var pool as OpcodeDef so that one wouldn't have to use strings to initialize. (Similarly to how InstrDefs were added to the global var pool.) 

Otherwise great to see this implemented! Looking forward to looking at the code.

vlazzarini:
> Haven't looked through this yet but could the names be changed to OpcodeDef and Opcode, to match InstrDef and Instr? 

I thought about that long and hard, but opcodes are not always defined (like instruments), except for udos, so the terminology breaks down. I think ref seems better also to encompass the idea of overloads.

Since opcodes are classes I thought OpcodeObj was appropriate, although I had started off by calling it just Opcode.

EDIT1: changed these as requested.

> Also some questions/comments:
> 
> 1. Does perf really require any pfields beyond the Opcode object? I would have expected arguments to the struct to be connected at init (perhaps this will make more sense once I look at the implementation...)

it does because an important use case is to run it in a k-time loop. Without connecting args again, we had no way of making these k-time. I had originally designed to run without any args except for the opcode object, but when I started using it, I hit this issue.


> 2. Is there any attention to passing opcode objects between processing contexts? i.e., if created with an sr or ksmps that differs from where it is used. May need some warnings or other processing added for those situations.

There is. The opcodes do a context check and if there is incompatibility we get an error.

> 3. For printing, would be good to eventually move away from things like kkjo and showing the internal type string. (and perhaps for us to move away from using internal type strings altogether). Not something that is necessary immediately though.

I guess the format can be improved alright.


> 4. Perhaps we could look at adding opcode definitions to the global var pool as OpcodeDef so that one wouldn't have to use strings to initialize. (Similarly to how InstrDefs were added to the global var pool.)
> 

I started exactly with that idea. I wrote code to create readonly vars for all opcodes. I hit some parser issues because the vars had the same name as opcalls. I could make it work just about but not consistently. I gave it up so I could move on with the work. This also allowed me to work exclusively with runtime code in the engine without having to touch the compiler.

But I would be far happier to avoid having to use strings. Maybe using an underscore or another annotation? _oscili etc?

EDIT2: I tried this now and it works well.

> Otherwise great to see this implemented! Looking forward to looking at the code.

Thanks. I am hoping this will open up other possibilities.

gesellkammer:
First, I think this is a great idea, in general. And the approach, if somewhat bureaucratic, seems very flexible, which makes it better than other alternatives (like my own `poly` implementation)

Some questions:

1) Are user defined opcodes supported?

2) > asig run obj, p4, ival, -1

Would this create the `asig` array itself if not initialized before?, like

`asig[] run obj, p4, ival, -1`

How useful is it, though, to run an array of opcodes with the same arguments? Wouldn't it be better to allow array arguments and distribute the inputs across the instances?

3) I cannot think of a situation where one would like an opcode to outlive an event, but that might be just my lack of imagination. But if that is indeed the case, wouldn't it be better to delete the opcode instances once the event is finished, instead of needing a manual `delete`?

vlazzarini:
> First, I think this is a great idea, in general. And the approach, if somewhat bureaucratic, seems very flexible, which makes it better than other alternatives (like my own `poly` implementation)
> 

The approach follows what we've been doing for instruments, in an attempt to have some uniform way to deal with these things.

NB: as noted in my response to @kunstmusik there are now global OpcodeDef variables for all opcodes with an underscore following the opcode name (e.g. `_oscili`) . So there is no need to declare a variable to hold a definition (although this is still possible). For example, we can have

```
asig run create(_oscili), p4, p5
```

> Some questions:
> 
> 1. Are user defined opcodes supported?

yes, they are treated in the same way as C/C++ opcodes.

> 2. > asig run obj, p4, ival, -1
> 
> Would this create the `asig` array itself if not initialized before?, like
> 
> `asig[] run obj, p4, ival, -1`
> 

Not at the moment. The array needs to exist and match the size of the opcode object array. The idea is that we can have any variable there, array or not. We may want for instance to make a serial connection and for instance reuse the output as the input for the next iteration, etc. But I don't see a problem to create an array if it does not exist. 

EDIT1: I have pushed a change that allows this, and also resizes other existing arrays if necessary, so no mismatching size errors occur.

> How useful is it, though, to run an array of opcodes with the same arguments? Wouldn't it be better to allow array arguments and distribute the inputs across the instances?
> 

The run opcode can take any arguments, array or not. In the example above, the amplitude (and table) is the same, but the frequency is an array.

> 3. I cannot think of a situation where one would like an opcode to outlive an event, but that might be just my lack of imagination. But if that is indeed the case, wouldn't it be better to delete the opcode instances once the event is finished, instead of needing a manual `delete`?

Yes, you are right that it would seem to be the right thing to delete the opcode at the end. However, there is a problem. Currently some opcodes have state that is tied into lists handled by the engine. At reset or destroy, these are accessed to release any acquired resources. If we delete the opcode dataspace, we run into access after free issues. In some systems, these are tolerated, in others we may get a segfault. 

Since I use address sanitizer in my develop build I can track any such issues and that one was detected. So I was in two minds whether to offer a delete opcode at all. In the end I opted to do it but with a health warning.

Simple opcodes such as oscillators can be deleted without any problems, but opcodes that have state managed by the engine will be a problem. It would require a significant redesign to avoid it.

vlazzarini:
Further examples
===

A couple of typical examples of multiple opcode runs

Series connection
-----

```
instr 5
   obj:Opcode[] create _reson, 4
   sig:a rand linenr(p4,0.1,0.1,0.01)
   cmp:a = sig*0.5
   sig run obj, sig, p5, p5/p6, 1
   sig balance sig, cmp
   out sig
endin
```

Parallel connection
----

```
instr 6
   freq:i[] fillarray p5*0.75, p5, p5*1.333, p5*1.666
   obj:Opcode[] create _reson, lenarray(freq)
   src:a rand linenr(p4,0.1,0.1,0.01)
   sig:a[] run obj, src, freq, freq/p6, 2
   out sumarray(sig)/lenarray(sig)
endin
```

gesellkammer:
This example is really elegant:

```csound

instr 6
   freq:i[] fillarray p5*0.75, p5, p5*1.333, p5*1.666
   obj:Opcode[] create _reson, lenarray(freq)
   src:a rand linenr(p4,0.1,0.1,0.01)
   sig:a[] run obj, src, freq, freq/p6, 2
   out sumarray(sig)/lenarray(sig)
endin
```

Some questions:

* Does `_reson` mean the "reson" opcode? 
* I assume that the contents of the `freq` array are distributed across the instances of the filter. What would happen with an opcode which takes a 1D array as input? Would `run` expect a 2D array?
* In your new examples, I see that there is no `delete` anymore, is that correct?

vlazzarini:
> This example is really elegant:
> 
> ```csound
> instr 6
>    freq:i[] fillarray p5*0.75, p5, p5*1.333, p5*1.666
>    obj:Opcode[] create _reson, lenarray(freq)
>    src:a rand linenr(p4,0.1,0.1,0.01)
>    sig:a[] run obj, src, freq, freq/p6, 2
>    out sumarray(sig)/lenarray(sig)
> endin
> ```
> 
> Some questions:
> 
> * Does `_reson` mean the "reson" opcode?

_reson is a global var of OpcodeDef type, holding the reference to all reson overloads. It's equivalent to opocoderef("reson"). The underscore is to avoid parsing issues (and make it work without modifying the parser).


> * I assume that the contents of the `freq` array are distributed across the instances of the filter. What would happen with an opcode which takes a 1D array as input? Would `run` expect a 2D array?

At the moment  `run` can only take 1D arrays as args, so it does not handle opcodes with array inputs. To work with these, we need to employ init and perf. It may be possible to add support but I have not yet explored it.

> * In your new examples, I see that there is no `delete` anymore, is that correct?

The delete opcode exists but it's not needed in most cases, so I am not using it in the examples. Csound will reuse memory as usual and free it when it's possible to do so.

vlazzarini:
Access to opcode object parameters/arguments
====

Two perf-time opcodes are provided to access individual arguments of an initialised opcode object

`setp obj:Opcode, ndx:k, arg:*`

sets the `obj` input parameter/argument index `ndx` to `arg`. 
The argument type is checked against the opcode obj as it was initialised.

`arg:* getp obj:Opcode, ndx:k`

gets the `obj` output parameter/argument index `ndx`.

vlazzarini:
The work in #2091 now allows us to drop the underscore, so this code would be possible

```
// parallel connection
instr 6
   freq:i[] fillarray p5*0.75, p5, p5*1.333, p5*1.666
   obj:Opcode[] create reson, lenarray(freq)
   src:a rand linenr(p4,0.1,0.1,0.01)
   sig:a[] run obj, src, freq, freq/p6, 2
   out sumarray(sig)/lenarray(sig)
endin
```

if it is found to be better than using the underscore.

rorywalsh:
This is much nicer now, that underscore was a little awkward..

vlazzarini:
There are some issues in the parser to be resolved as some tests are failing, so may not be able to adopt this. The underscore is also good in that it signals the user we have a variable not a function, so it may be better after all. I'll see if I can figure out why these tests are failing.

NB: I figured these out. Here are the restrictions

- some opcodes have to be excluded (e.g. "instr", "xin" etc) which is sensible.
- variable names for opcodes with no inputs require an underscore to be added so the parser can figure these out.

Also since most opcodes are also shadowed by readonly variables, these cannot be re-used as variable names in the code. Possibly this is an argument for using underscores.

EDIT: inlining of array gets is not failing anymore.

vlazzarini:
Merged the #2091 branch into this and fixed the code and tests. This also does a couple of extra things

- removes the underscore in the opcode ref variable names,

```
instr 1
   opcodeinfo oscili
   oobj:Opcode create oscili
   opcodeinfo oobj

   a1 run oobj, p4, p5, p6
      out a1
endin
```



- excludes various opcodes from the process of readonly var creation, including opcodes with no inputs (if we find a fix to this, we can put these back).

kunstmusik:
Did some basic testing of the opcode type and looks good. One issue found when working with opcode arrays, but this is a more general issue of needing to have arrays of complex datatypes work as references rather than values and should be handled in a separate PR. Going to merge this in now.

kunstmusik (approved):

PR #2078, opened 2025-01-09
fixed problem with multiple whitespace on csoundSetOption

vlazzarini:
Multiple whitespaces were causing problems for csoundSetOption, this PR fixes it now.

### DIFF
--- a/Top/argdecode.c
+++ b/Top/argdecode.c
@@ -1580,6 +1580,18 @@ end:
 
 void checkOptions(CSOUND *csound);
 
+char *unquote_arg(CSOUND *csound, char *arg) {
+  char *out = cs_strdup(csound, arg);
+  char *op = out;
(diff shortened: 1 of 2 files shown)


scottsitar:
Will this fix the issue mentioned here: https://github.com/csound/csound/pull/1900 ?

I did that a while ago when there was a problem mentioned on the mailing list, but if this is a better solution, I'm happy to close my older PR.  I hadn't rebased it against the new develop branch anyhow.

vlazzarini:
@scottsitar no, yours is a different issue in a another part of the code. 

Here what I am doing is fixing the API call to behave exactly as the command line so that multiple options can be passed correctly.

rorywalsh (approved):
@gesellkammer has confirmed this fix works so I'm going to merge.

PR #2080, opened 2025-01-11
solves issue 1414: midi smf0 to smd1

gleb812:


vlazzarini (approved):
Thanks, all looks right to me.

Issue #2081, opened 2025-01-12
Regression: error with strtol


kunstmusik:
[test.zip](https://github.com/user-attachments/files/18387738/test.zip)

I haven't had a chance to diagnose, but I recently built from develop branch and am getting an error after almost 10 seconds of running the attached live code project:

INIT ERROR in instr 61 (opcode hexbeat) line 309: strtol: invalid format
 from file livecode.orc (2)
 from file test.csd (1)
ibeatPat        strtol  #S9
                   T  9.875 - note deleted.  i61 had 1 init errors
           T 10.000 TT 10.000 M:  0.11585  0.13353

There's no error to start then an error repeatedly happens after this. Seems to be a regression as this was working prior to updating to latest develop. (Don't recall when the last time I built and installed to system, but would guess within the past month this changed.)

scottsitar:
Just a quick observation.  I stuck in some debug message lines, and it seems like strcat might be the issue.  I print out the two arguments and the result.  The invocation of instrument P1 behaves normally, but in the invocation of instrument S1, I start seeing:

```
INSIDE strcat: 0x
INSIDE strcat: a
LEAVING strcat: 0xaa
```

So, it looks like there might be a null terminator or size calculation issue.  They keep growing as the score progresses:

```
INSIDE strcat: 0x
INSIDE strcat: 0
LEAVING strcat: 0xcaca0a0
```

for example, and then eventually the errors you're seeing are caused by an invalid format for strtol showing up in the buffer.  For me, it happens to be channel names appearing at the end of the buffer:

```
INSIDE strcat: 0x
INSIDE strcat: a
LEAVING strcat: 0xaaaaaaa0a0a0a0aaaaaaaaa0a0a0a0aaaaaaaaa0a0a0a0aaaaaaaaa0a0a0a0aaaaaaaaa0a0a0a0S2.cuta
```

After that, strtol is failing (correctly) at this check:

https://github.com/csound/csound/blob/develop/OOps/str_ops.c#L696

since there is still data at the end of the buffer after it has used up all valid characters.

scottsitar:
I think I found the culprit, and put up a quick PR.

vlazzarini:
I came across a memory bug in that area sometime last year. The code was using the alloc size of the destination minus one and then it lead to access past the end (and was caught by the address sanitizer), #2068. I fixed it by changing it for the source size minus one. Thanks for catching it.

kunstmusik:
Fixed in #2082 thanks!

PR #2082, opened 2025-01-12
Fix initial memcpy in strcat operator

scottsitar:
We were doing a memcpy of the total allocated size minus 1, presumably because we knew that the string would be null terminated and we didn't want to copy the null.  However, strcat requires both of its arguments to be null terminated, so if we don't copy over the null terminator from str1, then we are only guaranteed that our first argument to strcat will be null terminated in the case where we re-Calloc'ed the buffer.  Otherwise, we might run into danger.

With this change, the attached code in github issue #2081  seems to run without errors.


kunstmusik:
Thanks @scottsitar, this looks to have fixed #2081. Much appreciated!

kunstmusik (approved):

PR #2085, opened 2025-01-14
Fix atsa bug

vlazzarini:
Two left-over if statements were stopping audio data from being read/written.


kunstmusik (approved):
Took me a moment to see how SndfileRead uses MYFLT underneath the hood to determine which sound file read/write function to use. Going to merge this in.

PR #2086, opened 2025-01-14
Add methods to the PerformanceThread to access more functionality of the API 

gesellkammer:
This PR adds three methods to the CsoundPerformanceThread to give access to the API:

* CompileOrc: schedules a compilation action
* EvalCode: calls csoundEvalCode within the performance thread, calls a given callback with the result of the evaluation
* RequestCallback: calls a callback from the performance thread, allowing to have access to the API within the context of the performance thread

Exposes all these methods to the API as csoundPerformanceThreadCompileOrc, csoundPerformanceThreadEvalCode and csoundPerformanceThreadRequestCallback

These methods were tested within the context of multithreaded c++, resulting in better and more consistent latency when sending code to the API. Additional tests were done in python code accessing the csound API via libcsound. This resulted in better performance and latency than directly accessing the API (csoundCompileOrc, csoundEvalCode, etc.) or using the process callback within the performance thread.

### DIFF
--- a/Top/csPerfThread.cpp
+++ b/Top/csPerfThread.cpp
@@ -482,6 +482,99 @@ public:
     ~CsPerfThreadMsg_SetScoreOffsetSeconds() {}
 };
 
+
+#define _PERFTHREAD_COMPILE_BUFSIZE 1024
+
+class CsPerfThreadMsg_CompileOrc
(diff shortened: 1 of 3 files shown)


vlazzarini (approved):
all looks ok to me.

PR #2090, opened 2025-01-20
Remove unused semantic transformation

scottsitar:
In the branch that tried to process an already processed T_OPCALL, there was a transformation that we tried to do if the first identifer on the line was recognized as an opcode.  However, this seems to now be unused after the fix for recognizing opcodes with a parentisized first argument vs a function start.  Simply removing this transformation, all tests still passed.

I've added a test to ensure that the original report is now fixed, and that we can assign to a variable with the same name as an opcode using the non-functional syntax.


scottsitar:
I found an issue with this in that a statement like `print fmax:i` will still fail to parse.  Closing to investigate further.

Issue #1964, opened 2024-09-25
Parser still confusing opcode and variable names


vlazzarini:
The new parser is supposed to tell variable names from opcodes, but there are some issues still.

For example, let's say we have a variable declared as `fmax:i`, which is also the name of an opcode. If we declare it by assigning a value,

```
fmax:i = p5
```

it works as expected. However, if we use `init` as the means of a first declaration,

```
fmax:i init p5
```

then it doesn't:

```
error:  Unable to find opcode entry for 'fmax' with matching argument types:

Found:
  (null) fmax S

Candidates:
  i[] fmax i[]i[]
  k[] fmax k[]k[]
  i[] fmax i[]i
  k[] fmax k[]k
  i[] fmax i[]i[]
  k[] fmax k[]k[]
  i[] fmax i[]i
  k[] fmax k[]k
```

It is also curious that it finds `(null) fmax S`, when there is no S type in the expression at all (p5, another variable or a literal yield the same error message).

If the variable name is changed in such a way that an opcode with that name does not exist, the code compiles correctly.

kunstmusik:
Interesting, I think this is probably a byproduct of the disambiguation process trying to function more like earlier parsers where opcode names took precedence. Might be possible to just change that precedence and consult var table first. Will need to research further to see if doing that will remain backwards compatible.

vlazzarini:
@scottsitar maybe we could use your flex/bison expertise to tackle this one?
I checked today and this is still failing to parse.

scottsitar:
@vlazzarini Sure, I can look into it.

vlazzarini:
thanks.

scottsitar:
From an initial look, I think the issue might not be flex/bison building the parse tree, but rather the semantic analysis afterwards.  For the following simple instrument:

```
instr 1
    fmax:i init p4
endin
```

I get this parse tree, which seems reasonable to me:

```
<tree addresses="(0x600003120310 : 0x0)" type="288" name="INSTR_TOKEN" loc="7:/Users/audiodev/issues/1964/test.csd">
 <treeleft addresses="(0x600003120130 : 0x6000000205d0)" type="299" name="INTEGER_TOKEN" value="1" loc="6:/Users/audiodev/issues/1964/test.csd">
 </treeleft>
 <treeright addresses="(0x6000031202b0 : 0x0)" type="284" name="T_OPCALL" varname="init" loc="8:/Users/audiodev/issues/1964/test.csd">
[...]

Issue #2061, opened 2024-12-12
Pass by reference issue 


vlazzarini:
There is an issue with pass by reference where k-rate inputs are not being passed correctly

```
opcode Test(var:k):(k,k)
 b:k = var*2
 xout var, b
endop
```

run as

```
instr 1
k0 = 1
k1, k2  Test k0
printk2 k1
printk2 k2
endin
```

gives

```
instr 1:	    0.00000
instr 1:	    0.00000
```

however,

```
instr 1
k1 = 1
k1, k2  Test k1
printk2 k1
printk2 k2
endin
```

gives the right answer

```
instr 1:	    1.00000
instr 1:	    2.00000
```

but swapping the outputs

```
instr 1
k1 = 1
k2, k1  Test k1
printk2 k1
printk2 k2
endin
```

doesn't, we have zeros again.
NB: this might be a general issue with other types, but I only tested k-vars.

NB2: it doesn't seem to happen with single outvars. The opcode below works as expected.

```
opcode Test(var:k):k
xout var*2
endop
```

vlazzarini:
So I check these a bit more and it seems to be comprehensively broken for multiple arguments with k, a, or i types.  I'm focusing on udo.c:

```
static void handle_pass_by_ref(CSOUND* csound, UOPCODE* p, INSDS* lcurip)
```

which may be the source of the problems. I noticed some oddities in terms of the
arg pointers stored in hash table and then retrieved but even if I fix these, it's still broken so I suspect the problem is wider. I need to understand the code better I think

vlazzarini:
One of the issues is related to the storing in hash tables. Stored data gets overwritten if a variable is used for instance in the arguments (xin) and in xout. The input pointer is set to the output pointer.  

If I fix it so that existing names are not stored by xout, the problem partly goes away, but we cannot set an input arg variable to xout (the pointer is not set), and we cannot use the same variable for two separate outputs. In fact, we can't use the same variable for more than one output without this fix either, as the hash entry is overwritten.

I think the solution is to think of a different means of storing the xout pointers. Xin is fine since the var names have to be unique anyway, so we can guarantee that each entry is unique.

vlazzarini:
So here is my proposal:

- with pass-by-reference semantics it is not possible to connect inputs to outputs directly, as they are references to distinct variables.
- it is not possible to connect the same output variable to more than opcode outputs, unless the actual UDO outputs are the same variable (so the reference is the same).

[...]

PR #2091, opened 2025-01-20
Fix handling when opcode shadows variable

scottsitar:
Since the fix for differentiating between opcodes with a parenthesized first argument and function starts, there was a transformation in `convert_statement_to_opcall()` that was no longer getting called in its intended situation.  However, it was getting called and performing an invalid transformation on other trees.  It has been removed.

Also, in cases where we have a line of the form `ident1 ident2`, we were not checking whether ident2 was a variable and respecting that, but instead just checking whether it was an opcode.  There is now a check for looking through the variable pools.

Tests were added to the ambigous opcall csdtest to cover these cases.

This addresses the issue in #1964 as well as one other case found during testing.

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -1790,20 +1790,6 @@ TREE* convert_statement_to_opcall(CSOUND* csound, TREE* root, TYPE_TABLE* typeTa
   }
 
   if (root->value != NULL) {
-    /* xout exp(0) */
-    if (root->left != NULL &&
-        root->left->type == T_IDENT &&
-        find_opcode(csound, root->left->value->lexeme) != NULL) {
-      TREE* top = root->left;
-      root->left = NULL;
-      top->right = root;
-      top->next = root->next;
-      root->next = NULL;
-      top->type = T_OPCALL;
-      top->right->type = T_FUNCTION;
-      return top;
-    }
-
     /* Already processed T_OPCALL, return as-is */
     return root;
   }
@@ -1887,6 +1873,7 @@ TREE* convert_statement_to_opcall(CSOUND* csound, TREE* root, TYPE_TABLE* typeTa
   if (leftCount == 1 && rightCount == 1) {
     TREE* newTop;
     if(root->right->type == T_IDENT &&
+       find_var_from_pools(csound, root->right->value->lexeme, root->right->value->lexeme, typeTable) == NULL &&
        find_opcode(csound, root->right->value->lexeme) != NULL) {
       newTop = root->right;
       newTop->type = T_OPCALL;
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -330,32 +330,39 @@ char* get_arg_type2(CSOUND* csound, TREE* tree, TYPE_TABLE* typeTable)
     TREE* nodeToCheck = tree;
 
     if (tree->type == T_ARRAY) {
-      varBaseName = tree->left->value->lexeme;
-      var = find_var_from_pools(csound, varBaseName, varBaseName, typeTable);
-
-      if (var == NULL) {
+      if (tree->left->type == T_FUNCTION) {
         char *fnReturn;
-        if (tree->left->type == T_FUNCTION &&
-            (fnReturn = get_arg_type2(csound, tree->left, typeTable)) &&
+        if ((fnReturn = get_arg_type2(csound, tree->left, typeTable)) &&
             *fnReturn == '[') {
           return cs_strdup(csound, &fnReturn[1]);
         } else {
+          synterr(csound,
+                  Str("non-array type for function %s line %d\n"),
+                  tree->left->value->lexeme, tree->line);
+          do_baktrace(csound, tree->locn);
+          return NULL;
+        }
+      } else {
+        varBaseName = tree->left->value->lexeme;
+        var = find_var_from_pools(csound, varBaseName, varBaseName, typeTable);
+
+        if (var == NULL) {
           synterr(csound,
                   Str("unable to find array operator for var %s line %d\n"),
                   varBaseName, tree->line);
           do_baktrace(csound, tree->locn);
           return NULL;
+        } else {
+          if (var->varType == &CS_VAR_TYPE_ARRAY) {
+            return cs_strdup(csound, var->subType->varTypeName);
+          } else if (var->varType == &CS_VAR_TYPE_A) {
+            return cs_strdup(csound, "k");
+          }
+          synterr(csound,
+                  Str("invalid array type %s line %d\n"),
+                  var->varType->varTypeName, tree->line);
+          return NULL;
         }
-      } else {
-        if (var->varType == &CS_VAR_TYPE_ARRAY) {
-          return cs_strdup(csound, var->subType->varTypeName);
-        } else if (var->varType == &CS_VAR_TYPE_A) {
-          return cs_strdup(csound, "k");
-        }
-        synterr(csound,
-                Str("invalid array type %s line %d\n"),
-                var->varType->varTypeName, tree->line);
-        return NULL;
       }
     }
 
--- a/tests/commandline/test_ambiguous_opcall.csd
+++ b/tests/commandline/test_ambiguous_opcall.csd
@@ -10,6 +10,12 @@ nchnls = 2
 
 chn_k "bla",2
 
+opcode toStrArray, S[], S
+  StringIn xin
+  SArray[] fillarray StringIn
+  xout SArray
+endop
+
 instr 1
  // from github issue 1697
  // could be parsed as either of the following, but we want the first:
@@ -26,9 +32,17 @@ instr 2
  print fmax:i
 endin
 
+instr 3
+  // found while testing github issue 1964:
+  // variable shadowing udo that returns array - we can tell the difference
+  toStrArray:i init 1
+  String = toStrArray("Inline string-array get")[0]
+  prints "%s\n", String
+endin
 </CsInstruments>
 <CsScore>
 i 1 0 1
 i 2 0 1
+i 3 0 1
 </CsScore>
 </CsoundSynthesizer>
--- a/tests/commandline/test_ambiguous_opcall.csd
+++ b/tests/commandline/test_ambiguous_opcall.csd
@@ -18,8 +18,17 @@ instr 1
  chnset k(1), "bla"
 endin
 
+instr 2
+ // from github issue 1964
+ // make sure we can use a variable name which shadows an opcode
+ fmax:i init p4
+ print fmax
+ print fmax:i
+endin
+
 </CsInstruments>
 <CsScore>
 i 1 0 1
+i 2 0 1
 </CsScore>
 </CsoundSynthesizer>


vlazzarini:
Thanks for this. I tested these with a few opcode names and it seems to work well. Even some perverse code like this 

```
instr 1

 cpspch:i = p5
 ampdb:i = p4
 oscili:a = oscili(ampdb(ampdb),cpspch(cpspch))
     out oscili

endin
```

I wonder if it will actually facilitate what I am trying to do with #2071  - allowing us to drop the initial underscore. I think this is brilliant, thanks.

Btw, there are about 700 tests under ../tests/soak. Maybe they could be used to test this? I am not sure yet of the status of these as I don't normally run them.

scottsitar:
Ah thanks!  When I was getting my environment initially setup, I only focused on being able to locally build what was being built in the github actions.  So, I haven't tried running the soak tests yet, but I'll give it a go against the develop branch to see what it gives as a baseline.

vlazzarini:
I've tested this PR merged with #2071 and yes this now allows me to set up readonly variables for all opcodes, and this code is now possible

```
instr 5
   obj:Opcode[] create reson, 4
   sig:a rand linenr(p4,0.1,0.1,0.01)
   comp:a = sig*0.5
   sig run obj, sig, p5, p5/p6, 1
   sig balance sig, comp
   out sig
endin
```

vlazzarini:
I found an issue with an edge case. This is from arrays/array_get_inline.csd

```
opcode toStrArray, S[], S
  StringIn xin
  SArray[] fillarray StringIn
  xout SArray
endop

instr 1
  String = toStrArray("Inline string-array get")[0]
  prints "%s\n", String
endin
```

Now if we (perversely) add a variable with the opcode name

```
instr 1
  toStrArray:i init 1
  String = toStrArray("Inline string-array get")[0]
  prints "%s\n", String
endin
```

the code fails to compile
```
error:  invalid array type i line 28

Could not parse type for argument
```

line 28 is the inline string assignment.

I spotted this when testing your code with my opcode type PR (see comments in #2071)

scottsitar:
Oh, nice one!  I'll take a look to see what's going on.

vlazzarini:
Another one I noted in #2071 . Opcodes with no inputs raise an issue:

```
instr 1
 in:a init 0
 a1 in
endin
```

does not compile

```
error:  Unable to find opcode with name: a1

``

**Comment by scottsitar** [comment]
> Another one I noted in #2071 . Opcodes with no inputs raise an issue:
> 
> ```
> instr 1
>  in:a init 0
>  a1 in
> endin
> ```
> 
> does not compile
> 
> ```
> error:  Unable to find opcode with name: a1
> 
> Parsing failed due to syntax errors
> ``
> ```

I'm not sure about this one - this is actually what I would expect given my current mental model of parsing.  The way the current disambiguation works is that it uses the shape of the statement (number of out args, number of in args, whether somthing is an opcode/variable) to decide how to parse it.  Once it's made that decision, if there's an error, then so be it.  This sounds like we need to try harder?  If there are multiple ways to parse something, then try each in turn (in some kind of priority ordering) until we either find one that works or run out of options?  Is that how this particular example would be expected to work?

**Comment by vlazzarini** [comment]
This one is hard alright. 

The context I was experimenting with  is one where "in" is the name of an opcode definition reference  that we can pass as an argument to another opcode.

obj:Opcode create in

which works, but then we can't use it as an opcode, as we get the error.

We should be able to figure out that 

asig in 

is a var followed by an opcode, and not an opcode followed by a var.

Maybe I should try to tackle it in the context where it is arising, because in that case the variable is of a very specific type and we can check for that. I was able to tackle another issue by doing that, maybe I might find the solution as well.

So maybe we leave this one for the moment.

NB: I also should note that if my PR goes in, we will have a situation where most opcode names will be taken by readonly vars and will not be available to users for naming vars in their code (which was the original motivation for this PR)

**Comment by scottsitar** [comment]
Ah, thanks for the details.  I'll focus on the array issue, then.  The next few evenings are busy, but I think I've figured out how to fix it, so I'll hopefully be able to update this PR by the weekend.  It comes down to this:

https://github.com/csound/csound/blob/develop/Engine/csound_orc_semantics.c#L332

When we're doing type checking and we've got an array, we take the identifier from the left tree and look it up in the pools, no matter what the left tree happens to be.  I'm not sure this makes sense for T_FUNCTION trees?  I think it is safe to check the left tree for type T_FUNCTION first (essentially doing the `var == NULL` block first for functions), and then go hunting in the type table with the rest of the logic there.

**Comment by scottsitar** [comment]
I've pushed up another commit with a fix for the array variable shadowing issue you found, and I've added that as a test to the ambiguous opcall csd test.  All seems fine with it, best I can tell.

**Comment by vlazzarini** [comment]
Great, I tested this with my local branch where I have this merged with the feature/opcodetype branch and it works perfectly.

**Review by vlazzarini — COMMENTED** [review]
It all seems good to me. I tested the code with an internal opcode as well and it's good

```
printk2 fillarray(1,2,3)[0]
```

So I am happy to approve this.

**Review by vlazzarini — APPROVED** [review]

**Comment by vlazzarini** [comment]
@kunstmusik let me know if you are happy with this so we can merge it.

**Comment by kunstmusik** [comment]
Catching up with this, I'm wondering if this isn't opening up a can of worms. Feels like similar issues in design as Lisp-1 vs Lisp-2 in terms of allowing the same symbols to map to different things. Adding in scope, I would expect this:

```
instr 1
 in:a init 0
 a1 in
endin
```

to have locally defined "in" to be a variable of type a, and the second line to be treated as a failure. If we're thinking of symbols and their associations, I could imagine interpreting "in" normally as a globally defined symbol mapping to an opcode, and any symbol could be locally shadowed with a mapping to a var. 

Having it parse as both an opcode and a var in the same instr context takes us down a route more of a Lisp-2, which I think might make things more complicated in the long run. 

For example, in Javascript we get:

```js
a = () => { return 10 }
b = () => {
  a = 3
  return a + 4
}
b() // returns 7
```

and within b you can not both call a() and use the a=3 variable.

kunstmusik:
(accidentally closed, meant to just leave a comment)

scottsitar:
@kunstmusik That makes sense to me from a programming language point of view, for sure.  My assumption going in here was that we would try to disambiguate where we could to help out, but should we be more strict/predictable in the scoping-style rules that are in play here?  So, if a symbol has been defined as a variable in an opcode/instrument, then it can only be used as such for that block?

vlazzarini:
Not sure I completely understand @kunstmusik 's point because we can already use a symbol as a local variable and as a (global) opcode name, as enabled by this PR, in the same scope

```
instr 1
oscili:a init 0
oscili oscili p4, p5
   out oscili
```

So that part is good. 

Except in the cases where the opcode does not take any parameters as in the case of `in`. That one seems to be a hard one to parse, possibly because it's hard to disambiguate such a line as `in in` without the args on the left (but I may be wrong in this).

scottsitar:
If I'm understanding right, I think the idea is that we don't want to allow people to use a symbol as both a local variable and an opcode in the same instrument/udo.  So, in an instrument like this:

```
instr 1
   in:a init 0
   a1 in
endin
```

The second line would be a compilation error with the message "a1 is not an opcode" or something like that.  Since "in" is already a variable, we wouldn't even try to interpret it as otherwise.

What the right decision is, I'm not sure, but that's my interpretation of it.

vlazzarini:
I see, that seems limiting, and would not allow us to use the opcode names for opcode ref vars.

kunstmusik:
@scottsitar yes that is correct in regards to interpretation. It was my expectation for language design that there would be one symbol table, not multiple tables on context. 

@vlazzarini could you explain how this would be a limitation for "opcode names for opcode ref vars."? I'm not sure I see where that'd be a problem. 

FWIW, besides Javascript, seems to be same convention in Python, C, C++, Rust, Go, etc.

vlazzarini:
If you read the comments in #2071, you will see that this PR here is allowing me to use opcode names for opcode ref readonly global variables. I've tested it and it works correctly (with the exception of opcodes with no args).

Without this PR, it's not possible, so the ultimate aim is to use it for that application. If we can't allow a variable and an opcode with the same name coexist in the same scope, then we have to add an underscore or something like that instead.

Btw what @scottsitar is saying already happens in this PR, if we use `in` as a variable, we get an error message exactly like that, `a1 is not an opcode`.

However, if we use an opcode with left hand side arguments, there is no error and everything is parsed correctly.

scottsitar:
To muddy the waters, I can think of languages that behave both ways.

As an example of the more strict scoping/context rules, there's something like C++.  If you define a global variable and then shadow it with a local variable, you can still access the global variable with the scope resolution operator `::`, but by default, you get the local variable.  That seems similar to the underscore prefix here.

However, there are also languages like Perl which are very, very context dependent, and the same identifier can be treated very differently based on the context in which it's found (scalar, list, or hash).  That seems similar to what this PR would allow - letting the compiler choose what it thinks is best.

I'm not advocating for either approach.  I just wanted to point out a few examples from either side of the fence to show that a reasonable argument could be made either way.  Well, depending on how "reasonable" you think Perl syntax is, haha!  Although, in my defence, I've tried to pick languages that were equally "write only" :)

kunstmusik:
@vlazzarini I took a look at comments in #2071 but I think I'll have to try out that branch further and look deeper as to why having opcodes defined as vars causes problems. I wonder if it wouldn't be possible to remove adding opcode defs to the var pool and instead have a more generic lookup_symbol() that handles looking through vars, then instr and opcode defs. The order of lookup looks good in 2071 but having the def in the var pool and in the opcode list might be an issue. 

I'll jump over to 2071 first and come back here after I try things out. It might well be that I'm not seeing things fully but I think I need to write some csound code testing out the opcode instance features to understand the problem further.

vlazzarini:
Sounds good to me. The branch is up-to-date with develop, so you can go ahead and explore it.

kunstmusik:
Took me a while to find some time to look at this and #2071. I tried this on its own and saw that:

```csound
instr 1
   in:a init 0
   a1 in
endin
```

failed as I would have expected. 

I also tried #2071 on its own, but also with this merged in, and both worked. 

However, when I tried changing the way #2071 worked to add opcode defs to the global var pool using the clean name without underscores.  I found the same as @vlazzarini that it fails when there's opcodes with 0 in-args. I get a lot of failures with our test suite with lines like:

```csound
kCycle     timeinstk
```

and

```csound
  asig in
``` 

failing. 

I'm wondering how to proceed here. I want to make sure that code like this is not valid:

```
instr 1
oscili:a init 0
oscili oscili p4, p5
   out oscili
endin
```

but also do want to use "oscili" rather than "_oscili".  

@vlazzarini would you rather 

1. merge this into #2071 and continue there with changes to get non-underscore names working or
2. merge #2071 and this into develop and start another branch that to make non-underscore names work

I'd prefer option 1 and just getting the opcode type work cleaned up there before merging to develop. If that works, @scottsitar can retarget this PR to merge into the #2071, we can merge, then continue on there.

vlazzarini:
I also prefer option 1 - so that we have clean and tested merge into develop later. What I propose is this
- I already have this branch merged locally with #2071. I can update that branch and push it.
- We can close this PR 
- We then continue this discussion there.

How does that sound to you @scottsitar and @kunstmusik ? Let me know.

kunstmusik:
Sounds good to me, the merge from this branch into #2071 should preserve @scottsitar 's commit message and history. If you could go ahead with that @vlazzarini, we can add a note to #2071 to this PR for the benefit of future triaging of anything and then close.

scottsitar:
Sounds good to me.  Thanks for all the help working through this one!  I've started looking a little bit into #2061 , so I can focus on that one now.  It'll be my first time through the runtime code and how all the parameters get passed around, which will be good.

vlazzarini:
this is done, so closing now.

PR #2094, opened 2025-02-05
Update build images to use gh container registry

nonameentername:
Github artifacts expire after some period of time.

This pr updates build images to use gh container registry to store docker images instead of artifacts.


vlazzarini:
ios osxcross and android builds disabled in this PR. Can we check that they work before merging?

nonameentername:
I'm closing this pr and opening another one so that osxcross and android builds can execute.

PR #2096, opened 2025-02-05
Update ioscross build image to use the correct secret

nonameentername:
This pr updates the ioscross build image Dockerfile to use the new id for the secret.


nonameentername:
The ioscross container image build is failing because the incorrect key is being used after updating the build image workflows.

This PR fixes the issue.

[GitHub Action Run](https://github.com/csound/csound/actions/runs/13160137704/job/36726517979)

vlazzarini (approved):

PR #2100, opened 2025-02-17
added deinit to OSCbundle

vlazzarini:
This is to fix the issue of sockets being left open, which may cause the problem of failing to open sockets after a while.


MMarcoll:
Thanks! This seems to be working now! :)

vlazzarini:
@rorywalsh could you review this?

rorywalsh (approved):

PR #2108, opened 2025-03-05
Updating Csound for Android

vlazzarini:
This PR updates the Android code, build, scripts,. and Android Studio projects for Csound 7. The Android directory has also been cleaned of old projects and scripts. 

In particular, the plugins projects are not relevant anymore since Csound now builds all of its core opcodes into the library on Android.

Gradle has been updated to 8.5. It may be updated further, but that is beyond the scope of this PR. Also the AudioTrack class is deprecated so we should think about removing that code from CsoundObj.java. It's probably something for the Java devs to decide on.

I should also note that there is an Android build in github actions, but I am guessing this only builds a JNI without any bindings. It could potentially be expanded to include the SWIG bindings as well. Another possible future development is to move away from SWIG to hand-written bindings,  but we need someone to contribute and test these on Android.

For the moment, this PR should allow us to build Csound 7-based Android apps.


vlazzarini:
I tested this now on the emulator and the example apps are working.

rorywalsh:
This looks good. I built it here and ran it on my phone without any issues. Thanks for this.  

[update] looks like the audio input is not right, seems the permissions are not correct...

kunstmusik (approved):

kunstmusik (dismissed):
Jumped the gun on approving the PR; will stay out from here and leave it @rorywalsh 😅

rorywalsh (approved):
This is building and running fine now. Access to audio input must be granted by the end-user when they launch the sample app. This could be updated to ask for permission only in the activities that need audio input, but I think it's work well enough as an example.

PR #2109, opened 2025-03-06
Update CsoundObj on iOS for 7.0

vlazzarini:
This is only a minor fix. CsoundStart() needs to be explicitly called after the first compilation for the engine to start.


rorywalsh (approved):
I see no issues here. Thanks for this.

Issue #2111, opened 2025-03-08
Csound iOS SwiftExamples crashes on iOS device


Megal:
I'm experience a crash when trying to run iOS swift example from develop.

commit: https://github.com/csound/csound/commit/f528965667ad6eeecdd24ea07872a40958ef13ce
artifact: https://github.com/csound/csound/actions/runs/13726924626/artifacts/2712894617
XCode: 16.2
CPU: M1 Pro
macOS: 15.3.1
iOS: 16.0

Steps to reproduce:
1. Download artifact, unarchive, open Csound iOS SwiftExamples.xcodeproj
2. Build - compilation error libs are missing
3. Fix it by removing libs from project and adding libSndfileiOS.xcframework and CsoundiOS.xcframework
4. Compilation now successful
5. Crash when clicking on switch on first test view controller.

Crased at CsoundObj.m:369:36
Following backtrace: 
```
(lldb) bt
* thread #16, name = 'AURemoteIO::IOThread', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x0000000105479a28 Csound iOS SwiftExamples.debug.dylib`Csound_Render(inRefCon=0x0000000281b48710, ioActionFlags=0x000000016b7ce3dc, inTimeStamp=0x000000016b7ce390, dump=0, inNumberFrames=941, ioData=0x000000010525c520) at CsoundObj.m:369:36
    frame #1: 0x0000000210fc2eac libEmbeddedSystemAUs.dylib`ausdk::AUInputElement::PullInput(unsigned int&, AudioTimeStamp const&, unsigned int, unsigned int) + 176
    frame #2: 0x0000000210fa9bec libEmbeddedSystemAUs.dylib`AUInputFormatConverter2::InputProc(OpaqueAudioConverter*, unsigned int*, AudioBufferList*, AudioStreamPacketDescription**, void*) + 160
[...]

PR #2112, opened 2025-03-08
Fixed/Updated Csound for iOS build with source code reorganisation

vlazzarini:
Csound for iOS needed a lot of updating as examples were not working with old code, libs and headers. Now updated to use the xcframeworks for csound and libsndfile

https://github.com/csound/csound/pull/2111


vlazzarini:
Found some other API issues and fixed them. I compiled CsoundObj.m and now it is error free.
I think the problem in the issue was that CsoundObj.m was not completely updated for the new API.

vlazzarini:
Artefact tested on XCode 16 and iOS 18 emulator.

rorywalsh (approved):
Thanks for this Victor. I just tested this here. I was able to build and deploy to my iPad without any issues. Audio input/output works, and the interaction between Csound the UI works without a problem.

PR #2115, opened 2025-03-13
fixed pconvolve

vlazzarini:
pconvolve had a bug preventing it from accessing IR channels > 1 in mono output mode. Fixed.


kunstmusik (approved):
Tested and fix is good, thanks!

PR #2116, opened 2025-03-13
Fixed longstanding access-after-free bug

vlazzarini:
With the use of address sanitizer we have identified a longstanding access after free issue in the allocation of named instruments, present in the code since 6.x (and verified). Code tried to set three pointers to NULL, but the memory for these pointers had already been freed a few lines before (csound_orc_compile.c:1161). The solution was to use an `else` branch to avoid trying to set the pointers if the memory had been freed.


kunstmusik (commented):
Did some live coding to test and did not get a crash. Change otherwise looks good to me, will merge now, thanks!

kunstmusik (approved):

Issue #2119, opened 2025-03-21
if parsing error


gesellkammer:
In the following example I would expect the line "if changed..." to throw a parsing error on that line due to a missing "== 1". csound fails with a "Parsing failed due to no input!", which is unexpected in this context. Moreover, there is no indication as to where the error originated. 

```csound
instr 1
  kval = 1
  if changed(kval) then
    println "changed: %d", kval
  endif
endin
```

vlazzarini:
Improved diagnostics in #2120. @kunstmusik it would be good to have the parse create a synthetic expression to allow for this type of case to be acceptable. But that's a different issue.

PR #2120, opened 2025-03-22
Conditionals

vlazzarini:
This improves the diagnostics of failed conditionals in if/then and while/until/do (https://github.com/csound/csound/issues/2119).
It also fixes an access-after-free issue caught by the address sanitizer while reviewing the issue.

### DIFF
--- a/Opcodes/emugens/emugens.c
+++ b/Opcodes/emugens/emugens.c
@@ -2819,8 +2819,12 @@ int32_t printsk_init(CSOUND *csound, PRINTLN *p) {
                                              p->strseg.data, maxSegmentSize);
         p->strseg.size = maxSegmentSize;
         p->allocatedBuf = 1;
-        csound->RegisterResetCallback(csound, p,
-                                      (int32_t(*)(CSOUND*, void*))(println_reset));
+        /* VL 22.03.25 this callback is causing access-after-free as 
+           it tries to access p->buf.data that is already freed by
+           Csound on reset. Caught by address sanitizer.
+         */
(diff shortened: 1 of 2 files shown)


Opcodes/emugens/emugens.c
```diff
@@ -2819,8 +2819,12 @@ int32_t printsk_init(CSOUND *csound, PRINTLN *p) {
                                              p->strseg.data, maxSegmentSize);
         p->strseg.size = maxSegmentSize;
         p->allocatedBuf = 1;
-        csound->RegisterResetCallback(csound, p,
-                                      (int32_t(*)(CSOUND*, void*))(println_reset));
+        /* VL 22.03.25 this callback is causing access-after-free as
```
kunstmusik: Think if this is going to get commented out, might as well delete this block of commented code and remove println_reset altogether.

vlazzarini: I commented it out because it's not our (core devs) code. The author might want to re-use it or whatever without having to go into the git history.

kunstmusik: @gesellkammer could you take a look at this and comment on whether it's fine to remove? Seems to me that everything that is in println_reset is already cleared by standard memory cleaning in csoundReset and thus the problem occured in trying to free a second time.

gesellkammer: yes, it is fine to remove it

kunstmusik: Thanks for confirming, I'll go ahead and remove then merge this PR.

kunstmusik (approved):

PR #2123, opened 2025-03-29
GEN53 enhancements and NP2 API fix

vlazzarini:
This PR adds two improvements to GEN53

- The requirement of a power-of-two is lifted, since support for other FFT sizes is present in Csound.
- The requirement for table size to be specifically set in relation to the source table is lifted, the size is dynamically adjusted if needed (deferred size also supported now).

NB: as part of this I realised the np2 API interface was broken and this has been fixed now.


rorywalsh (approved):
These looks good. I'll go ahead and merge.

Issue #1989, opened 2024-10-09
oversampling not compatible with "new UDO syntax"


ketchupok:
Csnd 7.0 beta: Oversampling within UDOs requires the "oversampling ifactor" statement after the UDO definition, but before xin, see example below, which is not possible to do with the new UDO syntax.

Error caused: 
"new alloc for instr SineSound2:
INIT ERROR in instr 2 (opcode SinOscO2) line 26: can't set sr after xin"

```
<CsoundSynthesizer>
<CsOptions>
-o out2.wav
</CsOptions>
<CsInstruments>

sr = 44100
ksmps = 32
nchnls = 2
0dbfs = 1


opcode SinOscO, a, k ; this works!
oversample 2
kamp xin
kfreq expseg 3000, 10, 50000
asine = poscil(kamp, kfreq)
afilt = butlp(asine, 10000)
xout afilt
endop


opcode SinOscO2(kamp):a ; this not!
oversample 2
kfreq expseg 3000, 10, 50000
asine = poscil(kamp, kfreq)
afilt = butlp(asine, 10000)
xout afilt
endop

instr SineSound1
asin = SinOscO(0.4)
out asin, asin
endin

instr SineSound2
asin = SinOscO2(0.4)
out asin, asin
endin

</CsInstruments>
<CsScore>
i "SineSound" 0 10
i "SineSound2" 0 10
</CsScore>
</CsoundSynthesizer>
```

kunstmusik:
This doesn't seem to be too tricky to solve, new-style UDO's insert an xin statement. We can add a check for oversample and if found insert after that.

gesellkammer:
I was meaning to review this PR but there are a couple of things I don't completely understand about the oversampling mechanism in general (although I do understand that this PR is only about making the oversample opcode compatible with new style udos). 

What happens when an opcode with oversample outputs via "out", "outch", etc., or any other opcode accessing global state, for that matter?

vlazzarini:
It doesn't work. I have not tested but it could even segfault (which is bad so needs fixing). I actually should look at this and give an error if these opcodes are used. The same goes for local ksmps - I think it won't work. Global variables and channels are also not allowed, although channels might be supported in the future. With local SR we need converters to be in place for each signal coming in and out of the opcode (audio and control), and at the moment, these are only fitted to xin/xout.

PR #2124, opened 2025-04-04
New-style UDO oversampling support

vlazzarini:
This UDO adds support for oversampling to new style UDO. Previously, we had a
restriction to `oversample` and `undersample`, that they had to run before `xin`.
This caused a problem in newstyle UDOs, since `xin` is implicit and always runs
first. 

This PR adjusts the code to remove that restriction and adds a test to the suite.


Engine/insert.c
```diff
@@ -1628,6 +1628,11 @@ int32_t setksmpsset(CSOUND *csound, SETKSMPS *p)
   return OK;
 }
 
+/* from udo.c */
+int32_t set_inbufs(CSOUND *csound,
```
kunstmusik: Seems like this should be moved to udo.h

kunstmusik:
I'm not sure about this solution as a whole. My original thought in trying to solve this was that we'd rewrite the tree to move things like oversampleset, setksmps, etc. to the start of the opcode list for the instrument/UDO. The rationale for that was to avoid any init-order errors, such as having some opcode init at one SR or ksmps, then others at different rates/block sizes. In my head, I was treating these as something of a special-case "directive" for the instrument. 

I think calling set_inbufs will work but only if oversampleset is at the top. I tried this in test_oversample.csd:

```
opcode Test(k1:k,a1:a):a
a2 oscili 1, 440
oversample 2,3
a2 = a1*k1*a2
  xout a2
endop
``` 

and get different results when commenting out `oversample 2,3`.

vlazzarini:
I  see your point. It is currently a problem with all UDOs that both setksmps and oversample/undersample need to be before any opcodes or it won't work correctly. I was aware of this limitation.

What I was thinking with this PR was to allow new-style udos to work within this limitation, then as a second step, we would do the parser changes and fix it for all cases.

My idea is that at least by removing the error we would at least be able to use local srs in new style udos.

I also see this branch has a conflict. I'll fix this locally and push changes.

Engine/insert.c
vlazzarini: I think this was already done in develop so OK after merge.

vlazzarini:
I should add that this PR fixes #1989

gesellkammer (approved):
Even if there are still things to clear out about how oversampling works in general, being able to use this feature within the new syntax or udos is needed and also helpful for testing.

PR #2130, opened 2025-04-19
unschedule opcode

vlazzarini:
This PR adds an opcode to remove future events from the realtime list,

```
unschedule p1, p2, p3, ...
```
where p1 etc are used to match the event requested. If there are more than one matching event, only the first one found is removed.

The PR also fixes a bug in two overloads of schedule, which used to insert one extra kcycle to the start time due to the use of inputMessageInternal() without flushing.


rorywalsh:
Just a few questions here. I assume this works with fractional p1s, and events scheduled from the event opcodes, as well as schedule? 

`If there are more than one matching event, only the first one found is removed.`

Should there also be a way to remove all matching events? Maybe `unscheduleall`, or introduce a flag as the first input argument to specify if all, or just the first event, should be removed.

vlazzarini:
yes, I guess that can be done. I'll see to it and report back.

vlazzarini:
Ok, there's an `unscheduleall()`  that removes all matching events. I don't think this will be used very often as it is more normal to schedule events with different arguments, so that only one event would actually match. But no harm.

rorywalsh (approved):

PR #2138, opened 2025-05-08
Fix events for pfields > 2000. This PR is just a reference as to what needs to be changed

gesellkammer:


vlazzarini:
I've tackled this with a root-and-branch revision. Now pfields are dynamic so the limit is removed.

https://github.com/csound/csound/pull/2140

PR #2140, opened 2025-05-12
Dynamic pfields

vlazzarini:
This PR introduces dynamic pfields for Csound events. Previously any pfield arrays larger than `PMAX` were pushed into an 'extra' field, which was not ideal. This PR also removes any limitation to pfield array size in events.
However, opcodes are still limited by `VARGMAX`.

This PR does the following

- EVTBLK pfields is now just a pointer to an externally-allocated array of pfields
- the 'extra' element is removed.
- Memory for score events pfields is allocated in `rdscor()` and retained until reset.
- Memory for realtime events pfields is allocated in `copy_evtblk()` (musmon.c) and recovered after event is removed from list.
- `insert_score_event_at_sample()` now takes a pfield array as input.
- New function `insert_score_args_at_sample()` takes pfield as an array of argument pointers.
- API functions and opcodes have been updated to use this and to minimise data copying.
- CSD tests added.
- Unit tests added.


gesellkammer:
I tested this PR with my own plugins, it is working great for me and makes scheduling much easier.

vlazzarini:
Thanks a lot! If you have any ideas for further tests, that would be great. Given how these aspects are crucial to Csound operation, it would be good to be sure they work correctly so we have no surprises later.

gesellkammer:
I get a segmentation fault whenever I try to read a soundfile with ftgen or an f statement in the orchestra.

Simplest example which results in a segfault (replace foo.wav with any path to an existing soundfile):

```csound

<CsoundSynthesizer>
<CsOptions>
-odac 
</CsOptions>

<CsInstruments>
sr     = 44100
ksmps  = 64
nchnls = 2
0dbfs  = 1

itab ftgen 2, 0, 0, -1, "foo.wav", 0, 0, 0

</CsInstruments>

<CsScore>
; f 2 0 0 -1 "foo.wav" 0 0 0

; f0 3600

</CsScore>
</CsoundSynthesizer>

```

vlazzarini:
should be fixed now.

gesellkammer:
Thanks, for now all seems to work

gesellkammer:
I would really appreciate a decision on this PR since it implies somewhat extensive changes for code relying on scheduling events with large amounts of data.

vlazzarini:
Sorry...pressed a button by mistake. It's out of my hands, I'm afraid. If you want to make a review I can give permissions.

gesellkammer:
Yes, I can make a review

Engine/fgens.c
```diff
@@ -118,8 +119,10 @@ int32_t create_function_table(CSOUND *csound, FUNC **ftpp, const EVTBLK *evtblkp
     msg_enabled = csound->oparms->msglevel & 7;
     memset(&ff, '\0', sizeof(ff)); /* for Valgrind */
     ff.csound = csound;
-    memcpy((char*) &(ff.e), (char*) evtblkp,
-           (size_t) ((char*) &(evtblkp->p[2]) - (char*) evtblkp));
+  
+    memcpy(&(ff.e), evtblkp, sizeof(EVTBLK));
+    ff.e.p = (MYFLT *) csound->Calloc(csound, sizeof(MYFLT) * (PMAX+1));
```
gesellkammer: Why allocate first and then possibly reallocate in 166 if we already know the size?.

Engine/fgens.c
```diff
@@ -157,20 +160,12 @@ int32_t create_function_table(CSOUND *csound, FUNC **ftpp, const EVTBLK *evtblkp
...
-        (MYFLT*)csound->Malloc(csound,sizeof(MYFLT) * (evtblkp->c.extra[0]+1));
-      memcpy(ff.e.c.extra, evtblkp->c.extra,
-             sizeof(MYFLT) * (evtblkp->c.extra[0]+1));
-    }
-    else
-      memcpy(&(ff.e.p[2]), &(evtblkp->p[2]),
+    int msize = PMAX;
+    while (UNLIKELY(ff.e.pcnt>msize)) {
+        msize += PMAX;
+        ff.e.p = csound->ReAlloc(csound, ff.e.p, sizeof(MYFLT) * msize);
```
gesellkammer: Independently of the comment to the first allocation, this realloc could just be outside the loop. If msize needs to be a multiple of PMAX and higher than pcnt then msize = (msize + PMAX - 1) / PMAX * PMAX and the loop is not necessary.

OOps/schedule.c
```diff
@@ -319,44 +316,52 @@ int32_t instance_opcode_Instr(CSOUND *csound, LINEVENT2 *p)
...
-    pp.argno = pfields->sizes[0] + 1;
-    for (i=1; i < pp.argno ; i++) {
-      pp.args[i] = args+i-1;
-    }
-    pp.flag = 1;
-    return event_opcode_init(csound, &pp, 0, 'i');
+    pp.opcod = 'i';
+    pp.pcnt = pfields->sizes[0];
+    insert_score_event_at_sample(csound, &pp, args, csound->icurTimeSamples);
+    csound->Free(csound, pp.p);
```
gesellkammer: What is this Free actually freeing if the pp.p pointer cannot be anything but 0? I could not find any modification to it down the call stack (insert_score_event_at_sample declares its param as const)

Engine/fgens.c
vlazzarini: we don't know the size yet at 122 and we need the memory to copy data there. In the unlikely case it exceeds the original minimum, we reallocate.

Engine/fgens.c
vlazzarini: yes, you are right. We can simplify it by checking against PMAX and then just reallocating however many bytes we need to.

OOps/schedule.c
vlazzarini: Ah, thanks for spotting it, it's a leftover from earlier revision.

Engine/fgens.c
vlazzarini: actually you are right, we always know the count then. Changed the code to allocate it here.

Engine/fgens.c
```diff
@@ -109,6 +109,8 @@ int32_t create_function_table(CSOUND *csound, FUNC **ftpp, const EVTBLK *evtblkp
     FUNC    *ftp;
     FGDATA  ff;
     MYFLT   flen;
+
+    printf("%d \n", evtblkp->pcnt);
```
gesellkammer: Looks great to me, thanks! Just this printf and I think it's good to go.

vlazzarini: will remove.

gesellkammer (approved):

PR #2141, opened 2025-05-15
MIDI file opcodes

vlazzarini:
This PR enhances the support for MIDI files in the engine

- multiple concurrent MIDI file playback
- MIDI file opcodes for opening and playing files

Changes
---
- The `-F` flag is now a MIDI file input switch if used with no filename
- The `miditempo()` opcode can take a MIDI file id optional argument 

New Opcodes
-----

MIDI files can be opened with

```
id:i  = midifileopen(name:S [, port:i])
```

`port` defaults to 0. The `id` returned by this opcode can be used
to play etc. the MIDI file. Zero ID is reserved for any file passed to `-F`.

MIDI playback:

```
midifileplay(id:o)
midifilepause(id:o)
midifilemute(id:o)
midifilerewind(id:o)
midifiletempo(tempo:[i,k], id:o)
midifilepos(pos:[i,k],id:o)
```

These are mostly self-explanatory - `midifilemute()` toggle-mutes the playback as
opposed to pausing it. Once the MIDI file playback reaches the end, it's put on pause. The `midifiletempo()` opcode sets the playback tempo in bpm (`tempo` >= 0)
or as an absolute scaling factor (`tempo` < 0). The `id` parameters are optional, defaulting to zero (-F midifile).

```
len:i = midifilelen(id:o)
```

gives the MIDI file length in seconds.

MIDI file playback opcodes are non-op in MIDI-triggered instruments (so they can be safely used anywhere). For example,

```
<CsoundSynthesizer>
<CsOptions> 
-odac 
-F ; required to enable midi file input
</CsOptions>
<CsInstruments>
0dbfs=1

// open MIDI file using port 1 (channels 17-32)
mf@global:i = midifileopen("catherine.mid", 1)   

instr 17
 midifileplay(mf)
 amp:i = ampmidi(0.5)
 cps:i = cpsmidi() 
 sig:a = vco2(amp, cps)
 env:a = madsr(0.001, 0.1, 0.5, 0.1)
   out(env*sig*0.1)
endin

schedule(17,1,0)     // starts playback
event_i("e",0,midifilelen(mf)) // closes Csound
</CsInstruments>
</CsoundSynthesizer>
```

The operation of `-F filename` and `-T` is unchanged. A MIDI file opened through this means can be used alongside other files opened by the code.


rorywalsh (commented):
Thanks Victor. It's nice to finally have some midi file reading opcodes in Csound, after almost 40 years! The `miditempo` opcode has been updated in a recent update to this PR. I think this will cause the example .csd file to fail on 
`print miditempo(gimf)`
or maybe I've got this wrong? Where originally it simply reported the tempo, now it's used to set the tempo. Can one read multiple MIDI files in a single orc?

vlazzarini:
yes, you are right, the tempo reported was wrong due to inverted scaling. Fixed now.

rorywalsh (approved):
This looks good Victor. Thanks.

PR #2143, opened 2025-05-20
Complex Array Operations

vlazzarini:
This PR adds several opcodes for complex type arrays. Complex arithmetic cxpressions involving both complex arrays, real and complex scalars are implemented, as well as operators to get real, imag, abs, arg, convert to/from polar, and perform FFT. 

Arrays can now be used as a means to process analytic signals - array to asig conversions, as well complex array to asig, are supported. Versions of both `hilbert` and `hilbert2` can generate complex vectors, for example,

```
instr 1
 Sig:Complex[] = hilbert(diskin("fox.wav", 1,0,1))
 Mod:Complex[] = oscili(0.5,100,-1,0.25), oscili(0.5,100,-1)
 Ssb:Complex[] = Mod * Sig
   out real(Ssb)
endin
```

PR #2147, opened 2025-05-27
Add test for testCompileCSD

nonameentername:


vlazzarini:
That's because the test is wrong. The API call should be

```
int32_t result = csoundCompileCSD(csound, instrument,1,0);
```

I copied it without noticing the error. I had tested with a file and then forgot to change the parameter.

btw the test was already added to https://github.com/csound/csound/pull/2143

vlazzarini:
I understand the issue now and I have diagnosed it. Will push a PR soon, then let you know. Thanks for spotting it.

vlazzarini:
I've fixed this in a PR #2149 - asked you to review.

Notice that if you have an i1 0 -1 in the CSD to start with, Csound will exit immediately as it has a negative duration in the score.  While this works as an RT event when Csound is already running, in a score it doesn't, unless you add an f0 statement. That was an issue with your test - I modified for i1 0 1000 so it can keep Csound open for performKsmps().

PR #2157, opened 2025-06-19
Feature/optimisations

vlazzarini:
This PR introduces a few optimisations - mostly for floats builds but also an important change to pvsynth that leads to a significant speed up on pvsynth. Some updates to Java code on Android were also made.


rorywalsh (approved):
I've tested and I see some nice gains, especially on the Daisy. I'm happy to merge.

PR #2167, opened 2025-07-01
Pipewire backend

vlazzarini:
This branch introduces a backend module for pipewire 

```
-+rtaudio=pw
```

or `rtpw`, `pipewire`.

Not fully tested as it was developed on a VM with dodgy pipewire audio, but it seems to work correctly. The intention here is to add the module and then it can be further tested on native linux devices.

NB: also made some changes to fix warnings on Linux/gcc.


gesellkammer:
I tested this on linux with pipewire 1.2.4. In general it works fine, very similar performance to the jack backend. --use-system-sr does not seem to work.

vlazzarini:
yes, that needs to be implemented, as well as dev listing. Although, I think pipewire accepts arbitrary sample rates and just resamples (unlike jack).

vlazzarini:
Implemented system sr. It is only meaningful for input, since pipewire takes any sr from a source stream, so we can output at any sr. For input, yes it is important to match the source input stream sr, otherwise the audio is corrupted.

gesellkammer:
When I call csound with

```
csound -iadc -+rtaudio=pw --get-system-sr
```

I don't get any output from csound. Is this expected?

----

When called with

```
csound -odac -+rtaudio=pw --get-system-sr
```
I get 44100, which seems ok since pipewire does not force a particular rate and will resample. I have not seen a way to call the pipewire api to query the hardware clock since it can be configured to switch rate itself

----

When called with
```
csound -idac -odac -+rtaudio=pw --get-system-sr
```
I get 48000 as system sr and a message pointing to a rate mismatch:

```
WARNING: reading 256-byte blks of unknown from adc (RAW)
rtpw: mismatched input sampling rate,
system-sr 48000.0 csound-sr 44100.0
```

But there seems to be no actual mismatch. When running a .csd with the given command (`csound -idac -odac -+rtaudio=pw --use-system-sr mycsd.csd`) the sr is actually 48000, confirmed both via `prints "sr = %d", sr` and via `pw-top`

**NB**: As a side note, I get a segfault if csound is called with `-iadc` but no `-odac`. This is not specific to pw but also under jack

```csound -iadc -+rtaudio=pw  --use-system-sr xxx.csd```

Here is an example file which will list all sinks (outputs from the point of view of csound): 

To be compiled via: `gcc listsinks.c -o listsinks $(pkg-config --cflags --libs libpipewire-0.3)`

listsinks.c
```c

#include <pipewire/pipewire.h>
#include <spa/param/audio/format-utils.h>
#include <spa/param/props.h>
#include <stdio.h>
#include <string.h>

struct sink_data {
    struct pw_main_loop *loop;
    struct pw_context *context;
    struct pw_core *core;
    struct pw_registry *registry;
    struct spa_hook registry_listener;
    int done;
};

struct sink_info {
    uint32_t id;
    char name[256];
    char description[512];
    uint32_t n_channels;
    bool is_default;
};

static void registry_event_global(void *data, uint32_t id, uint32_t permissions,
                                  const char *type, uint32_t version,
                                  const struct spa_dict *props)
{
    struct sink_data *sink_data = data;
    const char *media_class;
    const char *node_name;
    const char *node_description;
    const char *is_default;
    
    // Check if this is an audio sink node
    if (strcmp(type, PW_TYPE_INTERFACE_Node) == 0) {
        if (props) {
            media_class = spa_dict_lookup(props, PW_KEY_MEDIA_CLASS);
            if (media_class && strcmp(media_class, "Audio/Sink") == 0) {
                node_name = spa_dict_lookup(props, PW_KEY_NODE_NAME);
                node_description = spa_dict_lookup(props, PW_KEY_NODE_DESCRIPTION);
                
                printf("Sink Node Found:\n");
                printf("  ID: %u\n", id);
                printf("  Name: %s\n", node_name ? node_name : "Unknown");
                printf("  Description: %s\n", node_description ? node_description : "No description");
                printf("  Media Class: %s\n", media_class);
                printf("---\n");
            }
        }
    }
}

static void registry_event_global_remove(void *data, uint32_t id)
{
    printf("Node removed: %u\n", id);
}

static const struct pw_registry_events registry_events = {
    PW_VERSION_REGISTRY_EVENTS,
    .global = registry_event_global,
    .global_remove = registry_event_global_remove,
};

static void core_event_done(void *data, uint32_t id, int seq)
{
    struct sink_data *sink_data = data;
    sink_data->done = 1;
    pw_main_loop_quit(sink_data->loop);
}

static void core_event_error(void *data, uint32_t id, int seq, int res, const char *message)
{
    struct sink_data *sink_data = data;
    printf("Core error: %s\n", message);
    sink_data->done = 1;
    pw_main_loop_quit(sink_data->loop);
}

static const struct pw_core_events core_events = {
    PW_VERSION_CORE_EVENTS,
    .done = core_event_done,
    .error = core_event_error,
};

int query_pipewire_sinks(void)
{
    struct sink_data data = {0};
    struct spa_hook core_listener;
    int ret = 0;
    
    // Initialize PipeWire
    pw_init(NULL, NULL);
    
    // Create main loop
    data.loop = pw_main_loop_new(NULL);
    if (!data.loop) {
        fprintf(stderr, "Failed to create main loop\n");
        ret = -1;
        goto cleanup;
    }
    
    // Create context
    data.context = pw_context_new(pw_main_loop_get_loop(data.loop), NULL, 0);
    if (!data.context) {
        fprintf(stderr, "Failed to create context\n");
        ret = -1;
        goto cleanup;
    }
    
    // Connect to PipeWire daemon
    data.core = pw_context_connect(data.context, NULL, 0);
    if (!data.core) {
        fprintf(stderr, "Failed to connect to PipeWire\n");
        ret = -1;
        goto cleanup;
    }
    
    // Add core event listener
    pw_core_add_listener(data.core, &core_listener, &core_events, &data);
    
    // Get registry
    data.registry = pw_core_get_registry(data.core, PW_VERSION_REGISTRY, 0);
    if (!data.registry) {
        fprintf(stderr, "Failed to get registry\n");
        ret = -1;
        goto cleanup;
    }
    
    // Add registry event listener
    pw_registry_add_listener(data.registry, &data.registry_listener, &registry_events, &data);
    
    printf("Querying PipeWire sink nodes...\n");
    printf("===================================\n");
    
    // Run main loop until done
    pw_main_loop_run(data.loop);
    
cleanup:
    // Clean up resources
    if (data.registry) {
        spa_hook_remove(&data.registry_listener);
        pw_proxy_destroy((struct pw_proxy*)data.registry);
    }
    if (data.core) {
        spa_hook_remove(&core_listener);
        pw_core_disconnect(data.core);
    }
    if (data.context)
        pw_context_destroy(data.context);
    if (data.loop)
        pw_main_loop_destroy(data.loop);
    
    pw_deinit();
    return ret;
}

// Example usage
int main(int argc, char *argv[])
{
    printf("PipeWire Sink Node Query\n");
    printf("========================\n\n");
    
    int result = query_pipewire_sinks();
    
    if (result == 0) {
        printf("\nQuery completed successfully.\n");
    } else {
        printf("\nQuery failed with error code: %d\n", result);
    }
    
    return result;
}

```

vlazzarini:
segfault is bad, I think we have not worked out that users may want only input. Needs to be fixed.
The mismatch may be issued because the sampling rate was originally 44100 then pushed to 48000. 

I'll check.

Thanks for the code, is it yours or did it come from somewhere else?

vlazzarini:
Ok here's the status at the moment:

1) `--get-system-sr` returns the system SR and exits (as per the `--help` message)
2) `--use-system-sr` works with `-iadc` only if `-odac` is not given. This is only going to work correctly if the backend module sets the system sampling rate on input (I don't know if they all do, but that's out of scope for this PR).

If this is working, I would like to get this merged in. The device listing can be added later.

vlazzarini:
Output node listing added. Number of channels still not available.
All device listing for output is implemented now.

gesellkammer (approved):

PR #2169, opened 2025-07-03
for loop 

vlazzarini:
This PR fixes the for-in-loop introduced earlier on in the new parser so that it can work with explicit types (as well as implicit types); a bug preventing the remainder of the code after the loop from executing was also fixed.

This was not documented at all - I have now added a commented example and can write a man page when this PR is merged.

Basically the for-in loop has the following syntax

```
for var in array do
 ...
od
```

where `var` and `array` can be either of i or k-types. The type of the loop (init or perf) is determined as follows

1) if `var` does not exist, it is created to match the type of the array variable. The loop type is determined by the array variable.
2) if `var` exists, it determines the type of the loop, regardless of the array type.

For example

```
/* case 1
   loop var is not declared
   loop type follows array type
   (i-type in this case)
*/
instr 1
for j in [1,2,3] do
 print j
od
endin
```

and

```
/* case 2
   loop var is declared
   loop type follows var type
   regardless of array type
*/
instr 2
j:k init 0
for j in [1,2,3] do
 printk2 j
od
endin
```

The converse case - an i-type loop var and a k-type array also runs only at i-time.

### DIFF
--- a/Engine/csound_orc.y
+++ b/Engine/csound_orc.y
@@ -534,7 +534,7 @@ for_in : FOR_TOKEN identifier in expr DO_TOKEN statement_list OD_TOKEN
           $3->right = $6;
           $$ = make_node(csound,LINE,LOCN, FOR_TOKEN, $2, $3);
         }
-        | FOR_TOKEN identifier ',' identifier in expr DO_TOKEN statement_list OD_TOKEN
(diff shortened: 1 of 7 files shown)


joachimheintz:
i tested it successfully.
Great to see this important improvement.

joachimheintz (approved):

PR #2170, opened 2025-07-04
improvements to coreaudio backend

vlazzarini:
This PR provides some improvements to the coreaudio backend:

- relinquishing the CPU while waiting for samples to be sent/received
- fixing the circular buffer to avoid races
- reducing the default buffer sizes from (-B4096, -b1024 *excessive*!) to (128, 64).

The fixes to the circular buffer allow very low latency. I was able to run Csound
with (32,16).


kunstmusik:
I tried this branch locally and found the 128,64 gave buffer underruns. Tested using macbook pro speakers. Suggest using (1024,256) or (512, 128).

vlazzarini:
did you test with auhal or portaudio?

kunstmusik:
I tested with both and both gave unusable sound quality and buffer underruns (with pa_bl)

vlazzarini:
pa_bl is not usable at this level, pa_cb is better. 

 I think auhal and pa_cb can work ok at least with 256,64. Max latency seems to be given by -B, so that is what we would like to minimise. I think 1024 is too much ~46ms roundtrip at 44.1.

Not sure why you are getting poor performance. As I noted I could run down to -B32 here.

kunstmusik:
Neither pa_cb or auhal work for me at 256,64. The audio is distorted in a
way I can't describe. 256,128 does work though I got some glitch listening
with pa_cb at that rate (been using trapped for all of these). I'm sure low
settings would work great on my Scarlett device, but on-board audio for
Macbook doesn't work. Are you testing on-board or with hardware audio
interface?

vlazzarini:
Oh, I see what's happening. It's to do with ksmps - possibly because it's too large (100). I got the same results here; setting it to 64 or lower fixes it. 

It's an interesting question now. Should we tie them together?

kunstmusik:
Ah, interesting. I think having more conservative defaults like 1024,256
might be the safest path for users for out of the box experience. Users can
always add flags to their .csound7rc for user-wide tuning. FWIW, I
generally use higher settings by default so I don't have to fuss with
things when I switch to Bluetooth, and set lower settings on projects that
require lower latency. I think it's just hard to find a good generalized
solution.

vlazzarini:
OTOH, some user may start Csound, get horrendous latency and go out saying that Csound can't do realtime properly. I'm tired of hearing that too. 
So I am not sure really.

kunstmusik:
Yes, that's the other side of the problem definitely and that's annoying.

It doesn't solve the problem, but maybe we do a check if ksmps doesn't
align with buffer sizes, we print a warning of some sort to inform the user
they're using non optimal settings.

vlazzarini:
I'll try to think of a plan.

vlazzarini:
I've pushed a commit now forcing -b to be at least ksmps. If it's less than ksmps, it's clamped to ksmps and -B is set to twice that size. I tested with trapped and it does the job.

kunstmusik:
I'm doing some testing and the clamp seems to work, but with my live coding
system I get some bad glitching with the defaults being set so low. Will
try to experiment further to see what's a sweet spot for values.

vlazzarini:
ok, just make sure tests are for auhal. The pa code is not as performant but I have some changes coming up that'll make it better.

vlazzarini:
Is it possible to get a decision on this? This PR has essential improvements to rtauhal that should be released soon. Particularly with https://github.com/csound/csound/pull/2206, it is important to get these in the code base.

kunstmusik:
Tried the latest here with pa_bl, pa_cb, and auhal and using my live code system. I moved my .csound7rc out of the way to ensure defaults were used. 

* pa_bl: Default 64 samp buffer gave highly distorted audio. (I have used pa_bl for years as a default and this was unusaable.)
* pa_cb: ran, audio crackled every once in a while. Unsure what this is about, could be my system needs a reboot. 
* auhal: ran, but audio was lower in pitch. Project has sr=48000, and it sounds like the project rendered in 48000 but output in 44100. 

The auhal pitch is of largeest concern, followed by pa_bl distortion.

vlazzarini:
blocking needs a much higher latency OK. I'll fix the sr issue in rtauhal.
Maybe we can settle for 512/128?

EDIT: tested here and I am not getting pitch change at all. When changing the sampling rate, sometimes audio may not start immediately - or sometimes it is interrupted and resumed.  A sine wave test plays always 440 Hz.

vlazzarini:
I have changed the SR setting code - now it makes more than one attempt, this improves the performance,

- no interruptions in audio at the start when the device SR is changed.
- no segfaults if the device SR is changed to a high SR (e.g. 96k) from a low one (e.g 48k).

As I indicated there was no change in pitch, the SR was set correctly, so you probably had another issue. However, I noticed those two things above and fixed them.

I have also settled the defaults to 512/128. There is no point on testing on portaudio blocking as I keep saying - block IO always depends on large buffers and so it will not be good. It's not the default for portaudio either so users can try it at their own risk. It's basically a fallback but really we should be thinking of removing it. 

I hope we can move on now.

kunstmusik:
The SR change has it now that auhal is now correctly playing back and is not pitched down here. (Pitch was definitely a problem prior to this.) Checking and pa_bl and pa_cb runs around 35% cpu while auhal shows 99-100% CPU in activity monitor. I get the least glitches with pa_bl at the moment but have to restart my computer and diagnose that separately. 

With auhal, I have to hit ctrl-c twice to exit csound. Seems some signal handling is not working correctly there compared to portaudio.

vlazzarini:
I cannot reproduce the exit issue - one ctrl-c closes the Csound command here in all situations, I can't provoke one that doesn't. If you have a test case, then I can try it, otherwise I can't do anything. If you can show that these changes have caused an issue on exit, then I can work on it, otherwise we are stuck.

I am not sure why CPU is 100%, I do not get it here at all anymore. The reason we had 100% is because Csound would not relinquish the CPU while waiting for coreaudio. This is changed so that we give it back as we wait. This PR have improved that significantly (that is why I want it merged asap), now I get similar measurements here with pa_cb.

Tests here indicate the best latency is auhal, then pa_cb. Pa_bl is dreadful - I don't recommend it.

I would really want to see this merged asap - the 100% issue although generally benign is not ideal.

EDIT: increased sleep time to 100 micros - yielding more CPU reduces stats.

kunstmusik:
I pulled the latest and rebuilt. ctrl-c stopped auhal, so I think that's okay. However, I am consistently getting the problem of pitch with auhal. Tested using trapped.csd using:

```
./csound ../examples/trapped.csd -+rtaudio=auhal --sample-rate=48000
```

Also finding pa_cb to crackle here and there. Have been testing with the attached zip. 

[test.zip](https://github.com/user-attachments/files/21397901/test.zip)

FWIW, regarding the auhal problem, my audio setting for "Audio MIDI Setup" in operating system is set to 48000:

image

kunstmusik:
I uploaded a video of the audio problem with 48000 SR vs 44100 SR with trapped and auhal here:

https://youtu.be/q9WVmMDUQcg

kunstmusik:
I think could ignore the audio/midi setup panel, seems to be changing to whatever SR csound runs at.

vlazzarini:
I can reproduce it here now. It is ok with 1 channel, but shows the problem with more channels. I think I have something to work with now.

vlazzarini:
I found the issue; I made a change in the latest commits that caused the problem - setting one of the format elements by multiplying by nchnls. I thought it was wrong but it turns out it was right.

Tested here now and is fixed.

kunstmusik:
Things are working well now with auhal. I still get clicks with pa_cb but I don't think it's related to this PR. I think we're good to merge this and I can look at pa_cb in a separate PR.

kunstmusik (approved):

vlazzarini:
Thanks, there's a PR to improve things on pa_cb for *nix

PR #2172, opened 2025-07-05
fixed k[] inarg with i[] var

vlazzarini:
This PR fixes an issue where i[] could not be used with k[] in-args.


kunstmusik:
I wonder if there isn't a different way to do this, but I tested it and it fixes the issue. Think it's better to merge and potentially refactor later.

kunstmusik (approved):
Thanks for the quick fix!

PR #2173, opened 2025-07-05
Shorthand for genarray

vlazzarini:
To complement the shorthand for `fillarray()` (eg `[1,2,3]`). I have added a short hand for `genarray`:

```
arr:i = [start ... end, incr]
```
for `genarray(start,end,incr)`
 
For example,

```
for j in [1 ... 10] do
 print j
od
```


hlolli (approved):
bravo, love it

PR #2174, opened 2025-07-05
For loop (2nd form)

vlazzarini:
This PR completes the missing parts of the 2nd form of the for-in loop. In this form, a second argument is given to hold the index variable,

```
instr 1
for j, i in [2,4,6] do
 print j, i
od
endin
```
prints

```
instr 1:	j = 2.000	i = 0.000
instr 1:	j = 4.000	i = 1.000
instr 1:	j = 6.000	i = 2.000
```


vlazzarini:
Added type operators as requested.

We can use 

```
typ:S = typeof(var:)
```

to get the type name

and

```
var1 === var2  
```

for type equality.

joachimheintz (commented):
i tried 
  for el,indx in ["foo","bar"] do
    print el,indx
  od
and get this error message:
error:  only i- or k-type arrays allowed

is this expected?

hlolli:
Looks good to me for the changes to for-loop 💚 

As for the triple equality grammar for type equality, I'm not sure if that's the correct choice just to not be confused with what is common in other languages. If this is equal type and value, it makes sense. If this is just a shorthand for typeof(var:) == typeof(var:) I wouldn't make a new grammar for it. The types are first-class objects (or interned singletons) so it should be sufficient to do simple equals operand. That's just my opinion, if there's a popular demand for this feature then I'm ok with that.

vlazzarini:
> i tried for el,indx in ["foo","bar"] do print el,indx od and get this error message: error: only i- or k-type arrays allowed
> 
> is this expected?

yes, for loops currently only take numeric values.

vlazzarini:
> Looks good to me for the changes to for-loop 💚
> 
> As for the triple equality grammar for type equality, I'm not sure if that's the correct choice just to not be confused with what is common in other languages. If this is equal type and value, it makes sense. If this is just a shorthand for typeof(var:) == typeof(var:) I wouldn't make a new grammar for it. The types are first-class objects (or interned singletons) so it should be sufficient to do simple equals operand. That's just my opinion, if there's a popular demand for this feature then I'm ok with that.

I think we need some different sign because it's not possible to overload `==` directly. It's not so much a shorthand because it's a fast comparison, it does not need to do string comparison. I can keep `===` for type AND value, but it would also be nice to have a shorthand for `typecheck()` which is what it is currently. 

Note for instance that we can't use `typecheck()` directly: https://github.com/csound/csound/issues/2175


What about `=t` ?

joachimheintz (approved):

PR #2176, opened 2025-07-07
Type equality operator

vlazzarini:
It was pointed out that `===` could be misinterpreted as both type and value equality, so this PR changes the operator
to `=t `

```
if var1 =t var2 then
```

I checked and this does not break any existing code.

Type equality can also be done with `typecheck(var1, var2)`


hlolli (approved):
looks good, nice :)

PR #2177, opened 2025-07-07
feat(test): add opcode assert for testing

hlolli:
rebase of https://github.com/csound/csound/pull/1663

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -767,6 +767,7 @@ set(libcsound_SRCS
     OOps/schedule.c
     OOps/sndinfUG.c
     OOps/str_ops.c
+    OOps/test_ops.c
     OOps/ugens1.c
     OOps/ugens2.c
     OOps/ugens3.c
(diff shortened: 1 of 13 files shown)


vlazzarini:
@hlolli looks good, but why do we need the command-line flag?

Also: probably need to add the new source file to wasm and Android builds, since they don't use CMake.

NB: funny that the rebase did not solve the conflicts

hlolli:
@vlazzarini sorry I see that I hadn't in fact resolved the merge conflicts doh.

I was exploring the possibility to have some sort of test runner. This can be marked as "beta" feature the command line flag. The idea is this. Users add assert into their csound code, but it's just empty, so that you can run your instruments fine with assert opcodes in the csound code. It would be unnecessary to always enable the assert opcodes, for optimisations reasons (and indeed we can optimise later).

I'll fix conflicts and follow up.

I can add to wasm and android builds altough I should get done with the CMake wasm task soon too.

vlazzarini:
It's a very good idea and feature, should be very useful. Also thanks for adding the unit test.

hlolli:
closing in favor of https://github.com/csound/csound/pull/2190

Issue #2062, opened 2024-12-12
functional syntax with multiple outputs and opcode type specifiers


rorywalsh:
I've noticed some interesting behaviour when using the new functional syntax. I'm testing with a custom opcode developed using plugin.h. Here's what I see:

```
k1, k2 = cabbageGetValue:kk("toggle") // works
k1, k2 = cabbageGetValue:k("toggle") // doesn't work but no error k1 and k2 are never updated
k1, k2 = cabbageGetValue("toggle") // works as expect
```

Victor's noted that a UDO that outputs multiple args will cause an error if `var = myUdo:k()` is used. There seems to be some inconsistency in the parsing?

vlazzarini:
I tested this with overloaded UDOs and I can confirm the behaviour:

```
opcode Test,kk,k
var:k xin
b:k = var*2
xout var, b
endop

opcode Test,k,k
var:k xin
xout var*2
endop
```

`k2, k1  = Test(1)` works and we have `k2=1,k1=2` , `k2, k1  = Test:k(1)` doesn't, we get 0 and 0, and `k2, k1  = Test:kk(1)` also works. 

Running with -v we see why

```
Test args:	k2	k1	1
printk2 args:	k2	0	0
```

so the single-output opcode is selected correctly (as per `:k`)  but passed the arguments k2 and k1 as its out and in args, respectively, and 1 is ignored. So k1 ends up as 0 and the result is 0. Surprisingly, we don't have an undef input argument error here.

If we set `k1=1` to 1 at the top, we get 2 in `k2`.

Now this is user error clearly, but I wonder if there is a way to warn the user of this fact.

rorywalsh:
A warning would be great. it took me some time to track down the problem.

vlazzarini:
Fixed now in https://github.com/csound/csound/pull/2184

Issue #1043, opened 2018-10-09
Request for missing "opcode:x[]" syntax


maurocsound:
The "hint" system to make the user choice which version of an overloaded opcode he want to use inside an expression, actually, doesn't work with array types so, something like that:

  `kArr[] = genarray:k[](1, 128)`

doesn't work.

vlazzarini:
https://github.com/csound/csound/pull/2184 addresses this.

PR #2184, opened 2025-07-11
Array-type annotation for opcodes

vlazzarini:
This PR addresses 2 isues:

1) https://github.com/csound/csound/issues/1043 by allowing opcodes to be annotated with array-types (using type[]), e.g.

```
arr:k[] = genarray:k[](1,10)
```

2) https://github.com/csound/csound/issues/2062 where wrong overloads are picked up based on annotation error. Where there is a discrepancy, we print a warning and go ahead with the correct selection. This also fixes a bug where for instance an oscillator k-rate might be used to fill an audio rate var, producing output with zeros.


rorywalsh (dismissed):
Thanks for this Victor 👍

rorywalsh (approved):

Issue #1641, opened 2022-10-20
OSCraw does not receive arrays via OSC


Oeyvind:
If I send an array with OSCsend, OSCraw will produce an array showing the message type to be "b" (bundle?), and then just empty strings for the rest of the array (nothing is copied I guess).

```
<CsoundSynthesizer>
<CsOptions>
-odac
</CsOptions>
<CsInstruments>

sr = 48000
ksmps = 128
0dbfs = 1
nchnls = 2

instr    1
  kval line 440, p3, 880
  kSend[] init 2
  kSend[0] = kval
  ksend metro 10
  OSCsend ksend+1, "127.0.0.1", 9999, "/foo", "A", kSend

  Smess[] init 8
  nxtmsg:
    Smess[], klen OSCraw 9999
    if klen > 0 then
      printarray Smess
      kgoto nxtmsg
    endif
endin

</CsInstruments>
<CsScore>
i1 0 10
</CsScore>
</CsoundSynthesizer>

```

vlazzarini:
Looking at the code, it seems to try to get the first number in the array as its size

```
          } else if (c == 'b') {
            len = *((uint32_t *) buf);
            byteswap((char*)&len,4);
            len = ceil((len)/4.)*4;
            if (len > str[n].size) {
              str[n].data = csound->ReAlloc(csound, str[n].data, len+1);
              str[n].size  = len+1;
            }
            strNcpy(str[n].data, buf, len+1);
            //str[n].data[len] = '\0';
            buf += len;
          }
```

Not sure whether this is the right thing, but we need to know how long the array is, and I don't know if there is another way.

kunstmusik:
@vlazzarini @Oeyvind : Checking to see if this is still an issue; if so I will schedule for 7.0.0.

Oeyvind:
@kunstmusik @vlazzarini : Yes I think it is still an issue. I haven't trioed anythng more with this, as I understood from @vlazzarini answer that sending arrays would not work with the current version of the opcode. It is possible that I misunderstood the reply(?). I haven't looked into the details of sending arrays over OSC in other languages. If it is possible in other languages, it would be good if we can adpot a similar strategy/syntax.
* did just a quick search: ..., and it seems python-osc does not support sending arrays (https://stackoverflow.com/questions/42427812/python-osc-message-sending-to-supercollider), but that supercollider might support it in its communication between SC client and SC server
[...]

PR #2186, opened 2025-07-11
OSCRaw - array type

vlazzarini:
This PR addresses https://github.com/csound/csound/issues/1641 - OSCRaw now receives arrays. It only works correctly with matching float versions at both ends of the connection.


rorywalsh (approved):
Thanks Victor. It might be worth updating the manual to mention that 'it only works correctly with matching float versions at both ends of the connection.'

Issue #2002, opened 2024-10-17
Handling of += etc


vlazzarini:
I noticed that the ##addin, ##subin opcodes that we have in entry1.c
do not seem to be used at all. I would have expected them to be reserved for the += and -= operators, but
those just seem to be further expanded into + and - expressions. 

Then I checked to see how they were introduced and it was John who did it

https://github.com/csound/csound/commit/2709c99b851a17b5476903bcc50af83f43e12446

to handle adding/subbing to global vars.

This seems to have fallen by the wayside. I see that csound_orc_semantics.c still references it but
I don’t think the code is ever reached.

Question is:

1. this seems to have been introduced to be more efficient and also threadsafe for PARCS. Should
we look to reinstate these opcodes?

2. Does it make sense then to handle all the += with special opcodes to make it all optimised?

3. Should we then also implement overloads for the case where += is used in an expression and
produces an output?

vlazzarini:
Done in https://github.com/csound/csound/pull/2189

PR #2189, opened 2025-07-12
Feature/opassign

vlazzarini:
This PR addresses https://github.com/csound/csound/issues/2002 by adding op-assignment operators and adjusting the parser/compiler to use them.


tests/commandline/test_opassign.csd
```diff
@@ -0,0 +1,32 @@
...
+</CsOptions>
+<CsInstruments>
+
+opcode Assert,i,ii
+ i1,i2 xin
+ if i1 != i2  then
+  exitnow(-1)
+ endif
+ xout i1
+endop
```
hlolli, line 13: 😄 nice

hlolli (approved):
I wonder if there's dead code that can be cleaned up after this. But I remember the vectoral math operators sometimes being annoying. Happy to see them improved. Also would be nice to have more array used in the unit tests as well if you have time to add them.

Issue #2093, opened 2025-02-05
Additional opcodes for Instr instance handling


kunstmusik:
Following up from #2047, look at adding further opcodes for working with instr instances:

* is_playing(instrInstance) - detects if an instance is playing
* is_releasing(instrInstance) - to detect if an instance is turned off but releasing

Ideally, user code have parity with internal scheduler code so that users can write their own voice allocation and handling code.

vlazzarini:
addressed by https://github.com/csound/csound/pull/2194

PR #2194, opened 2025-07-13
Instance check opcodes

vlazzarini:
This PR addresses https://github.com/csound/csound/issues/2093, adding opcodes to check for active and releasing instances.


gesellkammer:
"isreleasing" seems to do somewhat the same as the "release" opcode. Could these be unified?

vlazzarini:
yes, you are right - I forgot about that. We probably don't need a new name, just overload release to take an instance parameter, what do you reckon?

vlazzarini:
I have now changed names to `release(inst:Instr)` and `active(inst:Instr)` for consistency. Also added `res:B = release()` as an overload to the existing `res:k = release()`. Also placed all opcodes together in the same file.

gesellkammer:
I did like the previous names better... Particularly "release" sounds
confusingly imperative, like asking the instr to stop. Would it be an idea
to keep the previous names and make the old "release" an alias?

vlazzarini ***@***.***> schrieb am Fr., 25. Juli 2025, 15:35:

vlazzarini:
that's also possible.

vlazzarini:
ok, so now we have various flavours of `isreleasing()`, and `isactive()` gets its name back,

```
instr 1
 xtratim 0.1
 actv:b = isactive(this)
 if actv then
   prints "this is running\n"
 else
   exitnow(-1)
 endif

 rels:B = isreleasing(this)
 if rels then
   printks "releasing this\n", 1
   printk  0.1, times:k()
 endif
 
 rels:B = isreleasing()
 if rels then
   printks "releasing - boolean\n", 1
   printk  0.1, times:k()
 endif

 rels1:k = isreleasing()
 if rels1 > 0 then
   printks "releasing - original\n", 1
   printk  0.1, times:k()
 endif
endin
```

gesellkammer (approved):

Issue #2175, opened 2025-07-07
boolean function


vlazzarini:
Currently the parser won't allow a boolean function where a boolean expression is expected.

For example, let's say we have a function returning type b or B,`boolfunc(x)`, the expression

```
if boolfunc(x) then
```

won't parse, but we already have

```
var:b = boolfunc(x)
if var then
```

which works.

PR #2196, opened 2025-07-13
Enabled boolean functions to work in conditionals

vlazzarini:
This addresses https://github.com/csound/csound/issues/2175, allowing a function returning a boolean to be used in conditional expressions,

```
opcode Test(i:i,j:i):b
 xout (i == j)
endop

instr 1
if Test(1,1) then
 prints "passed\n"
endif
endin
```


vlazzarini:
Hi @hlolli  - if you have a chance to review this one, should be straightforward and then we can call this part of the parser closed.

rorywalsh (approved):

PR #2208, opened 2025-07-21
Improvements to portaudio backend

vlazzarini:
Portaudio backend re-write, tested on MacOS - needs Windows testing.


vlazzarini:
@stekyne would you be able to check that audio input and output work correctly on Windows?

stekyne:
@vlazzarini Built it there and tested it. Audio output and input working fine 👍

.github/workflows/csound_builds.yml
```diff
@@ -445,15 +443,13 @@ jobs:
           iscc /o. installer\\windows\\csound7_x64_github.iss
 
       - name: Upload installer
-        if: github.ref == 'refs/heads/develop'
```
stekyne, line 448: Just FYI, I added these conditions because we were getting storage/data usage warnings from Github. If we generate installers and zips for every branch, it adds up quickly.

stekyne (dismissed):

vlazzarini:
Thanks, I removed those so anyone could test the artefacts but was planning to put them back before merging. I'll reinstate them now.

stekyne (approved):

PR #2209, opened 2025-07-22
Improved portaudio (*nix platforms)

vlazzarini:
The portaudio code was modified to use csound->Sleep() because of Windows, but we can get better responsiveness using usleep on the other platforms (linux/macos). This PR restores usleep() to them.


kunstmusik (approved):
Tested here on macos, I'm now getting less crackles with pa_cb and my live code testing  at smaller buffer sizes.

PR #2214, opened 2025-07-29
optional types for newstyle udos

vlazzarini:
This PR adds optional types as per internal opcodes to newstyle UDOS:

```
opcode myop(var1:i,var2:o,var3:p,var4:q,var5:v,var6:j,var7:h):void
print var1,var2,var3,var4,var5,var6,var7
endop

instr 1
myop(1)
endin
```

prints

```
UDO myop:	var1 = 1.000	var2 = 0.000	var3 = 1.000	var4 = 10.000	var5 = 0.500	var6 = -1.000	var7 = 127.000
```

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -1536,6 +1536,25 @@ CS_VAR_POOL *find_global_annotation(char *varName, TYPE_TABLE* typeTable) {
   return pool;
 }
 
+// on new-type UDOS type-annotations can
+// be used for optional types
+// this checks and converts it to i or k type names
+char *check_optional_type(CSOUND *csound, char *name) {
+    if(is_in_optional_arg(name)) {
+      char *t = (char*)OPTIONAL_IN_TYPES[0];
+      char *o, str[2] =  {0};
+      for (int i = 0; t != NULL; i += 2) {
+        o = (char*)OPTIONAL_IN_TYPES[i + 1];
+        if (strcmp(t, name) == 0) {
+          str[0] = *o;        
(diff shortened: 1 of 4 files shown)


joachimheintz (commented):
"i tested also with default k-variables OPVJ and it works"

joachimheintz (approved):

kunstmusik:
Late to this, but I had always thought to not support the older optional types and to instead introduce a proper syntax for it, like:

```
opcode MyOp(ival1:i, ival2:i = 17, ival3:i = 41) 

endop
```

I suppose this work does not get in the way of the above being implemented, but thought it worth mentioning.

vlazzarini:
@kunstmusik 
I agree that this is desirable.

I started investigating this in a branch. I got it to parse and store the init values but then I hit a stumbling block. The semantic analysis does not have a mechanism for optional arguments beyond the codes (o, p etc), so while it is straightforward to read optional values in that way, it is not possible to call an opcode without passing arguments to it. 

This is because we parse UDOs like opcodes, so call like `myop()` looks for an opcode with no arguments or with coded-optional arguments. If the type is 'i', then it will expect an argument and give a synterr.

A solution would be to transform 'i' etc into an internal code that then is understood by the opcode resolution. That transformation needs to happen when the UDO is first parsed. The internal code could also contain the literal used to initialise the value. Then the values read could be put into synthesized constants like we do with 'o' etc

This would work for i and k, and possibly S. I don't know if we could entertain the idea of supporting other types, it may be hard (starting by the semantics).

In the absence of this, I think the baseline is to support what the classic-style UDO supports.

Issue #2221, opened 2025-08-03
string array segfaults with negative index


joachimheintz:
i think we want to avoid any segfault so i am reporting this.
```
instr 1
  mylist:S[] = ["bla","blo","blu"]
  puts(mylist[-1],1)
endin
```
returns:
```
new alloc (instance 1) for instr 1:
Calloc failed: memory allocate failure for 93982474977712 
csound command: Segmentation fault

```

PR #2222, opened 2025-08-03
Protecting get/set array for negative indices

vlazzarini:
Dealing with  #2221  - error on negative indices.


joachimheintz (commented):
tested and works

joachimheintz (approved):

Issue #2225, opened 2025-08-09
feature request: extend for-loop to string arrays


joachimheintz:
it is such a pleasure to work with the for-loop.
is it difficult to extend it to
`for s in ["a","b","c"] do` ?
this would be fantastic.
if this is not possible now, we should improve the error message.  currently in
`error:  only i- or k-type arrays allowed`
there is no line number provided, so normal users will not understand what's going wrong.

joachimheintz:
fantastic.
should i review and merge Misc fixes (PR #2219) first?
or vice versa?

PR #2226, opened 2025-08-10
Generalising for loop

vlazzarini:
This PR addresses  #2225 by generalising for-loops so they can handle any other types of arrays (beyond i and k). The rules are

- if a loop variable exists, it needs to match the loop array type, except for i and k types which may be converted automatically (and the loop type matches the loop variable type)
- k, a, and Complex arrays generate perf-time loops, all other types generate init-time loops.
- loop index variable if given will be either i or k depending on the type of loop.

### DIFF
--- a/Engine/csound_orc_expressions.c
+++ b/Engine/csound_orc_expressions.c
@@ -1566,14 +1566,15 @@ TREE* expand_until_statement(CSOUND* csound, TREE* current,
 TREE* expand_for_statement(CSOUND* csound, TREE* current, TYPE_TABLE* typeTable,
                            char* arrayArgType) {
 
-  CS_TYPE *iType = (CS_TYPE *)&CS_VAR_TYPE_I;
-  CS_TYPE *kType = (CS_TYPE *)&CS_VAR_TYPE_K;
-  CS_TYPE *aType = (CS_TYPE *)&CS_VAR_TYPE_A;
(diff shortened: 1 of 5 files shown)


joachimheintz (commented):
i tested with strings and all works perfectly.

is it also meant to work with a 2-dimensional array?
i tried this:
instr 1
  arr2d:i[] init 2,3
  arr2d = [1,2,3,4,5,6] //result is [[1,2,3],[4,5,6]]
  printarray(getrow(arr2d,0))
  printarray(getrow(arr2d,1))
  for el:i[] in arr2d do
    printarray(el)
  od
endin
but error message is:
error: syntax error, unexpected T_IDENT, expecting IN_TOKEN or ','  (token "i") from file /home/jh/Desktop/test.csd (1)
 line 15:
>>>  for el:i <<<

i also tried without assigning a type to el:
instr 1
  arr2d:i[] init 2,3
  arr2d = [1,2,3,4,5,6] //result is [[1,2,3],[4,5,6]]
  printarray(getrow(arr2d,0))
  printarray(getrow(arr2d,1))
  for el in arr2d do
    puts("hello",1)
  od
endin
but then:
PerfError in wrong mode 1
PERF ERROR in instr 1 (opcode ##array_get.x) line 0: Array dimension 1 out of range for dimensions 2
   note aborted

vlazzarini:
no, the for loop cannot handle 2D arrays, we would need a special case for that.

joachimheintz (approved):

Issue #2228, opened 2025-08-10
add audio variables to fillarray


joachimheintz:
`chnls:a[] = [poscil(.2,400),poscil(.2,500)]`
returns:
```
error:  Unable to find opcode entry for 'fillarray' with matching argument types:
Found:
  a[] fillarray aa
```
can this added to fillarray?

vlazzarini:
fillarray (and []) are i-time only. While we could add an overload, it won't run a perf-time, just initialisation. 

I think it will be confusing to have it.

joachimheintz:
true that --- i forgot about arguments for fillarray being i-time only.

from a user's perspective something like
myrandompair:k[] = [random:k(1,2),random:k(2,3)]
and
mytwosocillators:a[] = [poscil:a(.2,400),poscil:a(.2,500)]
would feel consistent, rather than the awkward
myrandompair:k[] init 2
myrandompair[0] = random:k(1,2)
myrandompair[1] = random:k(2,3)

would this be an enhancement in future, or is there another way which i 
overlooked?

vlazzarini:
I think @kunstmusik meant the [ ] syntax to be for array initialisation. That's what I see in the code, and I think it makes sense. 

I don't know if we ought to extend it to performance time, but we can't change the behaviour of fillarray() now. 

It may be possible to use another construct and opcode.

kunstmusik:
I think we would need a different syntax and set of perf-time opcodes to accommodate this. The code:

```
myrandompair:k[] init 2
myrandompair[0] = random:k(1,2)
myrandompair[1] = random:k(2,3)
```

Is consistent with what one would have to do in C, to allocate an array and then, at time of function call, fill the contents of the array with the outputs of other function calls. 

We would need some kind of syntax differentiator to determine above like code, maybe something like:

```
mytwosocillators:a[] = ~[poscil:a(.2,400),poscil:a(.2,500)]

or

mytwosocillators:a[] = ^[poscil:a(.2,400),poscil:a(.2,500)]

```

which would be syntax-sugar for something like the first code example.

vlazzarini:
I like that. I think it should be straightforward to introduce an opcode like fillarray to do the lot, init and perf, then add the syntactic sugar. I'll have a go at it and see. I'll find another name so it does not get confused, maybe fillarrayk (like genarrayk).

vlazzarini:
[...]

PR #2231, opened 2025-08-12
additions to fillarray / genarray

vlazzarini:
This PR does two things:

1 - adds an audio-array overload to fillarray, also making sure array initialisation happens at i-time for this and the k-rate overload (#2228)
2 - adds an overload to genarray corresponding to the existing genarray_i so that the syntactic sugar is translated into optmised code running at i-time only if input is only i-time.


joachimheintz (commented):
the audio array fetaure works, but i think your branch did not contain the for-extensions for strings.
how to handle this without messing things up?

joachimheintz (commented):
looking at 
{ "genarray.ki", sizeof(TABGEN),0, "k[]", "iip", (SUBR)tabgen, NULL, NULL},
and 
{ "genarray_i", sizeof(TABGEN),0, "k[]", "iip", (SUBR) tabgen, NULL, NULL},  
i am wondering whether we can consider genarray_i as deprecated now?

vlazzarini:
git should merge everything fine, otherwise we would have a conflict.
I'd say we should review deprecation separately.

joachimheintz (commented):
ok i have some extensions to the for-in tests.  will send this later in a separate branch.

joachimheintz (approved):

Issue #2235, opened 2025-08-16
fractional numbers for named instruments in CS7 only working as string


joachimheintz:
```
instr MyInstr
  print(p1)
endin
schedule("MyInstr.1",0,0)
```
works but not
```
instr MyInstr
  print(p1)
endin
schedule(MyInstr.1,0,0)
```
can we allow this?

vlazzarini:
No, because MyInstr is a variable (read-only), so the semantics don't allow, but

schedule(instrnum(MyInstr)+0.1, ...)

should work.

joachimheintz:
i see, thanks.

i cannot say about the internals, but from the user's point of view i 
would like to do:

instr Call
   for instance in [1 ... 9] do
     schedule(MyInstr.instance,0,0)
   od
endin
schedule(Call,0,0)

instr MyInstr
   print(p1)
endin

it works this way (except using nstrnum as you mentioned):

instr Call
   for instance in [1 ... 9] do
     schedule(sprintf("MyInstr.%d",instance),0,0)
   od
endin
schedule(Call,0,0)

instr MyInstr
   print(p1)
endin

this is not too bad, of course, but i like the new way we can work with 
instrument names as own entities.

vlazzarini:
We don't have '.' semantics for InstrDef types, it would require a language redesign. The '.' semantics is for user-defined structures.

I have added an overload to + to handle instances as syntactic sugar for instrnum(Instr) + fraction. I think that's about all we can do.

PR #2237, opened 2025-08-17
Overload + for InstrDef

vlazzarini:
This PR adds an overload to + to handle the case of
fractional instruments with InstrDef objects (#2235):

```
instr MyInstr
print p1
endin
schedule(MyInstr + 0.5,0,1)
```


joachimheintz (commented):
tested, works --- thanks

joachimheintz (approved):

Issue #2236, opened 2025-08-17
tablew not accepting global audio


joachimheintz:
giTable = ftgen(0,0,60*sr,2,0)

instr 1
  gaSig = poscil:a(.2,400)
  tablew(gaSig,line:a(0,10,10*sr),giTable)
endin
schedule(1,0,1)

returns:

INIT ERROR in instr 1 (opcode tablew.aa) line 15: table: index type inconsistent with output
 from file /home/jh/Desktop/test2.csd (1)
	tablew.aa	gaSig	#a1	giTable	0	0	0	
		   T  0.000 - note deleted.  i1 had 1 init errors

it works with a local audio variable:

instr 1
  gaSig = poscil:a(.2,400)
  aSig = gaSig
  tablew(aSig,line:a(0,10,10*sr),giTable)
endin
schedule(1,0,1)

it also works with tabw instead of tablew.

vlazzarini:
I cannot reproduce it here.

```
--Csound version 7.0 (double samples) Aug 17 2025
...
ftable 101:	2646000 points, scalemax 0.000
orch now loaded
audio buffered in 128 sample-frame blocks
not writing to sound disk
SECTION 1:
new alloc (instance 1) for instr 1:
B  0.000 ..  2.000 T  2.000 TT  2.000 M:  0.00000  0.00000
End of Performance inactive allocs returned to freespace
		   overall amps:  0.00000  0.00000
	   overall samples out of range:        0        0
0 errors in performance
```

joachimheintz:
which code did you try?

vlazzarini:
This 

```
instr 1
gaSig = poscil:a(.2,400)
tablew(gaSig,line:a(0,10,10*sr),giTable)
endin
```

joachimheintz:
strange.  i tried again but no change.
my version is:
--Csound version 7.0 (double samples) Aug 14 2025
[commit: dd222c5a04e24f6cdb029ec8ce5984475138353f]
(debian 12)

i will pull from latest develop to see whether it makes a difference

vlazzarini:
I see a problem now, which only materialises with ksmps > 10. Global variable arguments are not being correctly checked for types. I don't have a fix yet.

UPDATE: I have diagnosed it, but no fix yet.

joachimheintz:
ah good it made some sense.
yes my ksmps was 128.

PR #2238, opened 2025-08-17
Global var memory allocation

vlazzarini:
Global variables were not properly allocated causing their type info to be lost (#2236).
This PR fixes the problem.


kunstmusik (approved):
Unsure how to test this but the code looks simple enough. Approving and merging.

joachimheintz:
yes works now!
i am wondering why tabw worked but not tablew.
we can use this as test (looks ridiculous but was failing before)

sr = 44100
ksmps = 128
nchnls = 2
0dbfs = 1

giTable = ftgen(0,0,60*sr,2,0)

instr 1
gaSig = poscil:a(.2,400)
tablew(gaSig,line:a(0,10,10*sr),giTable)
endin

vlazzarini:
there's a type check in tablew, that's what was failing.

PR #2241, opened 2025-08-19
Variable redefinition

vlazzarini:
This PR implements support for variable redefinition, with the following principles:

- Explicit-type only (it does not make sense for implicit types) - non-array tyoes only.
- Variables can be redeclared with the same name of an existing variable but with a new type
- If in the same pool (local or global), the new variable of a different type replaces the previous variable
- if in different pool a local variable always shadows the global definition. 
- A global variable cannot shadow a local var - it is a synterr to redefine it as global of a different type (in the same context) 
- redeclaring it as global with the same type (in the same context) is a non-op with a warning.
- A redeclaration of a variable with the same type in the same pool is a non-op (silent), reuses storage

Examples:

```
test@global:i init 1 // variable declared

instr 1
 test:k init 0   // replaced by new definition - shadowed by a local
 test:a oscili 1000, 440 // local var replaced by new definition
 out oscili(0.5,test+p4)
 prints typeof(test) // 'a': type is always of the last definition - the first is gone
 prints "\n"
endin

instr 2
 print test    // prints 1 - the global var contents
 prints typeof(test) // 'i'
 prints "\n"
endin

num@global:i init 1   // global declaration
num@global:i init 2   // just initialisation of the same

instr 3
num:i init 0   // local var shadows global var
print num     // prints 0
endin

instr 4
num:i init 0 
num@global:i init 2 // WARNING: @global annotation ignored for variable num
endin

instr 5
num:i init 0 
num@global:k init 2 // error:  global variable num:k cannot shadow local variable num:i
endin

num@global:k init 0 // ok: different context, redefining global

instr 6
oscili:i init 0 // redefinition of 'oscili' variable as i-type
print oscili   // prints 0
endin
```


joachimheintz:
i think this is a very clear solution.  it also gives us more 
flexibility in general; in particular to use a local variable of the 
same name as a global, in this case ignoring the global one.

my suggestion is to ALWAYS throw a warning if such a redifinition is 
done.  this is a big help for the user who might have overlooked this 
conflict.

your example works for me; in the current form it is not compiling 
because the line in instr 5 throws an error.  i think that's what you 
wanted to show.

what is not working: i do not get a warning for instrument 4.

vlazzarini:
The warning should be there - it's at compilation so before any performance messages:

```
csound test.csd                          
UnifiedCSD:  test.csd
WARNING: @global annotation ignored for variable num
Elapsed time at end of orchestra compile: real: 0.005s, CPU: 0.005s
```

joachimheintz:
ah yes i had -m128 as option.
if possible, i would prefer the warning at the same output position as 
the warning for seed(0) or similar.

vlazzarini:
Not easy, because it's a compile-time warning, the code does not run at perf time. We would need to introduce a new check or a new messaging system, all of which does not seem to be worth it (that kind of redesign would consume a lot of coding time/attention and we don't have anyone else but me working on it at the moment).

joachimheintz:
that's fair.
if you agree, you may add the warnings for each redifinition.
i am testing it now and will report any issues.

vlazzarini:
I think that may be too much. Since there are all the opcodes as variables, if we use say `play` etc, we'll get too many warnings. I prefer to use warnings for behaviour that may be confusing like declaring a global variable but not getting one. Maybe we can give a warning for redefinition within the same scope - that's probably a good compromise, and it will help catch programming mistakes.

joachimheintz:
yes i think that is a good compromise.

joachimheintz:
ok found a bug.
looks like global variables are not global any more:

instr 1
   ***@***.***:k = random:k(1,2) * random:k(1,2)
   printk(1,Amp)
endin
instr 2
   printk(1,Amp,10)
endin
schedule(1,0,-1)
schedule(2,0,-1)

instr 1:	time     0.00000:     2.43143
instr 2:	time     0.00000:               0.00000
instr 1:	time     1.00136:     2.21672
instr 2:	time     1.00136:               0.00000
instr 1:	time     2.00272:     1.26931
instr 2:	time     2.00272:               0.00000
instr 1:	time     3.00408:     2.53184
instr 2:	time     3.00408:               0.00000
instr 1:	time     4.00544:     1.67895
instr 2:	time     4.00544:               0.00000

joachimheintz:
just to note that global audio works.
and global i-variables, too.

vlazzarini:
Your code does not make sense, can you rewrite it?

EDIT: I can reproduce it. It has to do with expressions on the left hand side. If you first declare the variable and then assign it, it works. Or if there is no expression, just assignment, at the declaration. I'll look for a fix. Good catch.

vlazzarini:
This is fixed now - the `@global` was being prematurely stripped - which became a problem as we allowed local variables to shadow global vars.

joachimheintz:
i tested quite a lot and all worked.
should me merge?

vlazzarini:
Happy for it to be merged. If there are issues, we'll fix them. Note that all of this only affects explicit types, so we're not concerned with backwards compatibility - but we're setting the rules that we'll adhere to when 7 is out.

joachimheintz (approved):

PR #2271, opened 2025-09-12
Add additional condition to ftndx assignment.

rorywalsh:


vlazzarini:
can you add 
if: false

to the jobs that are failing?

Like this
https://github.com/csound/csound/blob/f33cfff669ac8e7e5e798ea79418dda5781a4433/.github/workflows/csound_builds.yml#L338

https://github.com/csound/csound/blob/f33cfff669ac8e7e5e798ea79418dda5781a4433/.github/workflows/csound_builds.yml#L373

vlazzarini (approved):

PR #2275, opened 2025-09-13
Fixed #2252

### LINKED ISSUE #2252: custom types starting with i,k, or a..
It seems that using custom types that start with i,k or a is problematic as the parser stops reading the type once it finds a valid one. Consider the following, which results in an error:

```
struct iMyType val0:i, val1:i

opcode myop(inval:i):(iMyType)
a:iMyType init 1, 2
xout a 
endop

instr 1
    b:iMyType = myop(1)
    print(b.val0)
    print(b.val1)
endif
```

If we change `iMyType` to `MyType` it works fine. Another side-effect of this is the following code doesn't return an error, even though the type has not been declared:

```
random:iAmNotSureAboutThis(0, possibleTables)
```

Csound simply assumes `random` should run at i-rate. How difficult would it be to support this?

vlazzarini:
Multi-char outtypes are now disambiguated from single-char types, fixing #2252.

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -872,6 +872,7 @@ int32_t check_in_arg(char* found, char* required) {
     return 1;
   }
 
+
   if (*required == '.' || *required == '?' || *required == '*') {
     return 1;
   }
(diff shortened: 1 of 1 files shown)


rorywalsh (approved):
thanks Victor 👍

PR #2277, opened 2025-09-15
Adjusting version printing and help message

vlazzarini:
This PR adjusts the order of version printing, reinstates `--print-version` and documents it in the help message.

### DIFF
--- a/Top/argdecode.c
+++ b/Top/argdecode.c
@@ -343,7 +343,7 @@ static const char *longUsageList[] = {
     Str_noop("--limiter[=num]         include clipping in audio output"),
     Str_noop("--vbr                   set MPEG encoding to variable bitrate"),
     Str_noop("--suppress-version      do not print version details"),
-    Str_noop("--print-version         print version and exit"),
+    Str_noop("--print-version         always print version details"),
     Str_noop("--run-unit-tests        enable assertion opcodes and report test failures"),
     Str_noop("                          (assertions are ignored by default)"),
     Str_noop("--help                  long help"),
@@ -890,9 +890,7 @@ static int32_t decode_long(CSOUND *csound, char *s, int32_t argc, char **argv) {
     csound->print_version = 0;
     return 1;
   } else if (!(strcmp(s, "print-version"))) {
-     print_csound_version(csound);
-     csound->print_version = 0;
-     csound->info_message_request = 1;
+     csound->print_version = 1;
      csound->LongJmp(csound, 0);
      return 1;
   } else if (!(strncmp(s, "logfile=", 8))) {
@@ -1083,6 +1081,7 @@ static int32_t decode_long(CSOUND *csound, char *s, int32_t argc, char **argv) {
     return 1;
   } else if (!(strcmp(s, "version"))) {
     print_csound_version(csound);
+    csound->info_message_request = 1;
     csound->LongJmp(csound, 0);
   } else if (!(strcmp(s, "help"))) {
     longusage(csound);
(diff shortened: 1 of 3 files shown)


Top/argdecode.c
```diff
@@ -342,8 +342,9 @@ static const char *longUsageList[] = {
         "--aft-zero              set aftertouch to zero, not 127 (default)"),
     Str_noop("--limiter[=num]         include clipping in audio output"),
     Str_noop("--vbr                   set MPEG encoding to variable bitrate"),
-    Str_noop("--suppress-version      do not print version details")
-    Str_noop("--run-unit-tests         enable assertion opcodes and report test failures"),
+    Str_noop("--suppress-version      do not print version details"),
+    Str_noop("--print-version         print version and exit"),
```
gesellkammer: But if --version already prints the version and exits, what does --print-version do differently?

gesellkammer (dismissed):

Top/argdecode.c
vlazzarini:
```
csound --version
--Csound version 7.0 beta (double samples) Sep 15 2025
[commit: 22771beae0e0a3f24584dc6a3acaa15e99b8a4cc]
		   overall amps:      0.0
0 errors in performance
--Csound version 7.0 beta (double samples) Sep 15 2025
[commit: 22771beae0e0a3f24584dc6a3acaa15e99b8a4cc]
```

```
csound --print-version
--Csound version 7.0 beta (double samples) Sep 15 2025
[commit: 22771beae0e0a3f24584dc6a3acaa15e99b8a4cc]
```

vlazzarini: I've fixed --version to be like the --print-version above. Kept --print-version to force printing even if the Csound object has been reset. By default, printing only happens after csoundCreate() is called and then it's suppressed in subsequent resets/restarts.

vlazzarini:
@gesellkammer this needs a new review.

Top/argdecode.c
hlolli: how about one --version  and that be what is --print-version here above? It seems starting performance shouldn't be something that happens when you only want to know the version of a cli application?

gesellkammer (approved):

Top/argdecode.c
vlazzarini: yes, --version is now just printing and exiting. All other spurious messages are suppressed.

joachimheintz:
i get now an error for any csd i want to run in csoundqt:

unknown long option: '--print_version'
...failing to parse options

vlazzarini:
CsoundQT probably needs to be re-built and re-linked to Csound.

joachimheintz:
isn't it because the option '--print_version' does not exist any more?
rebuilding did not help.
i wonder whether the old behaviour could be left as alias, for backwards 
compatibility.
perhaps csoundqt is not the only application which uses it.

vlazzarini:
the option exists. You can try it in the command line using the develop branch

joachimheintz:
there is --print-version but not --print_version.
it also gives an error on commandline.
is this a typo?  was it --print_version before?

vlazzarini:
it could have been, but the long options all use `-` so this was out of line. I guess CsoundQT sources need to be adjusted to use the existing flags.

I also think Csound should not exit when it encounters an unknown option. I'll change that.

joachimheintz:
yes it definitely way with underscore; see 
https://csound.com/docs/manual/CommandFlags.html

i again suggest to keep the old version as alternative, for backwards 
compatibility.

gesellkammer:
I adapted csoundqt to the new flags in csound7 already yesterday, as soon
as the PR was adopted to develop (see commit
ae746badbc2d8e3356c0d3d90c63ba4926344ceb).

PR #2289, opened 2025-09-25
Implementing #2288

### LINKED ISSUE #2288: can also a loop variable shadow an opcode name?
this works because the local variable shadows the opcode name:
instr 1
  count:i = 5
  print(count)
endin

but this does not:
instr 2
for partial,count in [1,2.78,5.18,8.16,11.66,15.64,19.99] do
  chnset(partial,sprintf("partial %d",count+1))
od
endin

is it possible to extend the shadowing rules to this case?

vlazzarini:
Adding shadowing in for loops.


joachimheintz (approved):
works now --- thanks!

Issue #2291, opened 2025-09-27
UDO input arguments not considered to be local


joachimheintz:
opcode udo1(arr:k):()
endop
opcode udo2(arr:i):()
endop

returns:
warning: Replacing previous definition arr:k by arr:i, line 16

which is wrong because there is nothing to replace.

and for arrays it throws an error:

opcode udo1(arr:k[]):()
endop
opcode udo2(arr:i[]):()
endop

error: arr:i[] - type mismatch for existing array variable arr:k[]

PR #2293, opened 2025-09-27
Fixed issue #2291

vlazzarini:
Local type pool was created too late and so args were searched for in wrong location (#2291)


kunstmusik (approved):
Did not do extensive testing but code makes sense to me; approving and merging.

PR #2296, opened 2025-10-04
setting windows defaults

vlazzarini:
Setting more up-to-date defaults for windows buffer sizes.


rorywalsh (approved):
Thanks Victor, I can't believe these were so large.

Issue #2287, opened 2025-09-24
diskin output type not recognized automatically


joachimheintz:
instr 1
  sound:a = diskin("fox.wav")*random:k(0,1)
endin
schedule(1,0,1)

returns:
PERF ERROR in instr 1 (opcode =.generic) line 13: Opcode given variables with two different types: a : [

it works for
  sound:a = diskin("fox.wav")
and also for 
  sound:a = diskin:a("fox.wav")*random:k(0,1)

but i guess it should choose the ':a' type of diskin because of the result and despite the multiplication?

vlazzarini:
yes that is because the expression is parsed as

#a0[] = diskin()
sound:a = #a0*random:k()

I don't think the parser is clever enough to do otherwise. You need to give hints in this case. @kunstmusik should be able to say more.

NB: diskin:a[] has higher precedence than diskin:a in the opcode list, so that can be seen as the 'default'.

joachimheintz:
thinking loudly this becoming more clever would mean that the parser 
decides between 'a' and 'a[]' depending on the left hand side result, as 
it does in the case of
sound:a = diskin("fox")
before applying the multiplication?

certainly nothing very important; just to sketch a future enhancement.

kunstmusik:
Yes I think it just needs to have a type hint. It's ambiguous otherwise, and Csound needs a more advanced handling of types to deduce something like this. It was always the plan to get explicit types working for Csound 7 and implicit type deduction for Csound 8.

joachimheintz:
sounds like a good plan!

i wonder how we can describe this borderline for the users:
in csound 6 we always needed the type specification, like
    aSine = poscil:a(.2,400)

now this is not any more necessary; we can say
    sine:a = poscil(.2,400)
    vibrato:k = poscil(10,3)
    sound:a = diskin("fox.wav")
    sound:a[] = diskin("stereo.wav")
but not
    sound:a = diskin("fox.wav") * ampdb(-6)

where exactly is the borderline?

does it work for inline extensions, like
    new:k = wrap(chnget:k("central pitch")+random:k(7,9),50,85) ?
my guess is that
- wrap is assigned to :k because it is on top level, and new:k requires it
- for chnget:k and random:k we must specify it

is this right?  (running with --verbose seems to confirm.)

vlazzarini:
There is no actionable issue here, so I am closing this. The discussion can be carried on in  one of the email lists.

PR #2299, opened 2025-10-05
Sumarray aliases

vlazzarini:
This PR introduces '=' and i(), k(), a() as aliases to sumarray. It allows code to be parsed with fewer errors. For example, the issue discussed in #2287 can now be parsed without errors in a semantically correct way.


kunstmusik:
I think in general these might be a bit perilous, especially '='.  Would rather see people write code like:

```
sound:a = sum(diskin("fox.wav")*random:k(0,1))
```

which expresses the operations more clearly to me.

gogins:
Agreed.


Michael Gogins
Irreducible Productions

vlazzarini:
They can still do, not sum, but sumarray(), if they wish. In my opinion, it's quite clear that the assignment implies sumarray() and is preferable to have it than the error in that issue.

joachimheintz:
i checked out the /feature/reduce branch and can confirm that this line 
works now:
	sound:a = diskin("fox.wav")*random:k(0,1)
so it looks like a useful improvement to me.

but i wonder whether this is the right way to go.
if i get it right, then
	diskin("fox.wav")*random:k(0,1)
is an audio array with one element, so sumarray flattens the audio array 
to an audio signal.

i have two doubts here:
1. i wonder why
	sound:a = sumarray(diskin("fox.wav")*random:k(0,1))
is not working.  should it not do the same?
2. i think the clean way to make
	sound:a = diskin("fox.wav")*random:k(0,1)
work, would be to let diskin choose the proper version (a not a[]).
a normal user will not know that there is a hidden sumarray operation here.

vlazzarini:
1. should work, = is an alias. I need to check.
2. is not possible because the array output overload has a precedence over audio signal output. I don't think we can change that because it may break people's code. 

NB: just checked 1. Congratulations, you found a bug in the semantic analysis. The type is being stripped of `[]` and so it comes out as `:a;` (which is meaningless here). It's a nasty one. I don't have any viable solution for it currently.

vlazzarini:
Found a fix for the bug - in PR #2301.

joachimheintz:
as to the second item:
i did not mean to change the precedence, but to enable the parser to 
choose the appropriate version of (in this case) diskin, in the same way 
as it is currently working for
	sound:a = diskin("fox.wav")
choosing the .a version and
	sound:a[] = diskin("fox.wav")
choosing the .a[] version.
this is what steven mentioned as a plan for csound 8.

it might be better to encourage users for now to use the type specifiers 
if they want to be sure the appropriate version is selected, so in this case
	sound:a = diskin:a("fox.wav")*random:k(0,1)

if we can provide in the case of
	sound:a = diskin("fox.wav")*random:k(0,1)
  a good error message, it would be very helpful.

PR #2300, opened 2025-10-05
A-var multiple assignment

vlazzarini:
Unlike i and k types, multi assignment had not been implemented for a-types. This PR now adds the missing code. Test3 has been expanded to account for this.

### DIFF
--- a/OOps/aops.c
+++ b/OOps/aops.c
@@ -231,6 +231,29 @@ int32_t mainit(CSOUND *csound, ASSIGNM *p)
   return OK;
 }
 
+int32_t mainit2(CSOUND *csound, ASSIGNM *p)
+{
+  uint32_t nargs = p->INOCOUNT;
+  uint32_t nouts = p->OUTOCOUNT;
+  uint32_t offset = p->h.insdshead->ksmps_offset;
+  uint32_t early  = p->h.insdshead->ksmps_no_end;
+  uint32_t i, n, nsmps = CS_KSMPS;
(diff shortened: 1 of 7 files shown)


rorywalsh (approved):

OOps/aops.c
```diff
@@ -231,6 +231,29 @@ int32_t mainit(CSOUND *csound, ASSIGNM *p)
...
+  MYFLT *aa;
+  early = nsmps - early;      /* Bit at end to ignore */
+  if (UNLIKELY(nargs != nouts))
+    return csound->InitError(csound,
+                             Str("Out and in numbers not matching in "
+                                 "assignment (%d,%d)"),nouts, nargs);
+  for (i=0; i<nargs; i++) {
+    aa = p->a[i];
+    MYFLT *r =p->r[i];
+    for (n = 0; n < nsmps; n++)
```
gesellkammer, line 250: Maybe split the loop to make it more efficient?

for (n = 0; n < offset; n++)
  r[n] = FL(0.0);               // maybe even memset?
for (n = offset; n <= early; n++)
  r[n] = aa[n];                 
for (n = early + 1; n < nsmps; n++)
  r[n] = FL(0.0);

gesellkammer (approved):

OOps/aops.c
vlazzarini: that's the original code, I never touched it. Sure, it can be improved I guess.

PR #2301, opened 2025-10-07
Bug fix: array expressions inlined in opcall

vlazzarini:
A bug involving array expressions inlined in an opcall prevented code such as this to
be successfully parsed,

```
sig:i = sumarray(fillarray(1,2,3)*2)
```

because the resulting type of `fillarray(1,2,3)*2` was not found correctly. This PR fixes the issue and adds a test.

### DIFF
--- a/Engine/auxfd.c
+++ b/Engine/auxfd.c
@@ -31,56 +31,56 @@ static CS_NOINLINE void fdchprint(CSOUND *, INSDS *);
 
 void auxalloc(CSOUND *csound, size_t nbytes, AUXCH *auxchp)
 {
-    if (auxchp->auxp != NULL) {
-      /* if allocd with same size, just clear to zero */
-      if (nbytes == (size_t)auxchp->size) {
-        memset(auxchp->auxp, 0, nbytes);
-        return;
-      }
-      else {
-        void  *tmp = auxchp->auxp;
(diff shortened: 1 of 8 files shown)


joachimheintz (commented):
tested and works, and the error messages are much more helpful now.
for the line
	sound:a = diskin("fox.wav")*random:k(0,1)
i get:
	PERF ERROR in instr 1 (opcode =) line 14: 
	cannot handle different types (a, array)
	sound = #a6[]

i wonder whether we could add the type specifier to the variable name, so in this case
	sound:a = #a6[]
to make the connection between the diagnosis "different types (a, array)" and the variable "sound" as "a" more obvious for the user?

vlazzarini:
I'll take a look but I think the type annotation  is long gone by that stage in performance. The messages come from two different parts in the code and the second does have direct access to type information.

vlazzarini:
~~I just checked and there is no type information available when that second line is printed, only the argument name. So unless there is a major change in the way argument data is stored, it's not possible to do this. I do not think it warrants that change, particularly at this stage of development.~~

There might be a way actually.
UPDATE: it's done now.

joachimheintz (commented):
great --- you found a way!
ready to merge?

vlazzarini:
I think it's ready.

joachimheintz (approved):

PR #2303, opened 2025-10-10
Fix for pass-by-ref UDO arguments

### LINKED ISSUE #2290: set value to array not working in new style UDO
this UDO should set all array elements to newval:
opcode arrset(arr:k[],newval:k):()
  printk(0,newval)
  for val,i in arr do
    printk(0,arr[i])
    arr[i] = newval
    printk(0,arr[i])
  od
endop

but instead, all elements are set to zero:
instr 1
  arr:k[] = [1 ... 10]
  printarray(arr)
  arrset(arr,random:k(0,1))
  printarray(arr)
  turnoff
endin
schedule(1,0,1)

the code works without an UDO as it should:
instr 1
  arr:k[] = [1 ... 10]
  printarray(arr)
  newval:k = random(0,1)
  for val,i in arr do
    printk(0,arr[i])
    arr[i] = newval
    printk(0,arr[i])
  od
  printarray(arr)
  turnoff
endin
schedule(1,0,1)

even a simple assignment (without a loop) does not work:
opcode arrset2(arr:k[],newval:k):()
  printk(0,newval)
  arr[0] = newval
  printk(0,arr[0])
endop
instr 2
  arr:k[] = [1]
  printarray(arr)
  arrset2(arr,random:k(0,1))
  printarray(arr)
  turnoff
endin
schedule(2,0,1)
returns:
 1.0000 
UDO arrset2:	time     0.00000:     0.88287
UDO arrset2:	time     0.00000:     0.00000
 0.0000

vlazzarini:
This PR fixes the issues raised in #2290 where input args were not connected properly into opcodes of multiple input args.

### DIFF
--- a/tests/commandline/udo/test_args_in.csd
+++ b/tests/commandline/udo/test_args_in.csd
@@ -1,0 +1,23 @@
+<CsoundSynthesizer>
+<CsOptions>
+-n 
+</CsOptions>
+<CsInstruments>
+0dbfs=1
+opcode Test(sig:k,fn:i):(a)
+a1 flooper2 sig,1,0, 3, 0.1,fn
+xout a1
+endop
+
+i1 ftgen 1, 0, 0, 1, "../fox.wav", 0, 0, 1
+instr 1
+a1 Test 0.5,1
+out a1
+endin
+
+</CsInstruments>
+<CsScore>
+i1 0 1
+</CsScore>
+</CsoundSynthesizer>
+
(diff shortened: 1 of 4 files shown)


tests/commandline/udo/test_args_in.csd
```diff
@@ -0,0 +1,23 @@
...
+0dbfs=1
+opcode Test(sig:k,fn:i):(a)
+a1 flooper2 sig,1,0, 3, 0.1,fn
+xout a1
+endop
+
+i1 ftgen 1, 0, 0, 1, "../fox.wav", 0, 0, 1
+instr 1
+a1 Test 0.5,1
+out a1
```
hlolli, line 15: would it be possible to assert something here, like the amplitude ?

hlolli (approved):
I was checking if I had this fixed (I felt I encountered this issue), it's not.

tests/commandline/udo/test_args_in.csd
vlazzarini: yeh, we can improve it for sure

PR #2308, opened 2025-10-11
i(k) again

vlazzarini:
Matching csound 6.18 behaviour.


rorywalsh (approved):
Thanks for this Victor, it was tripping up a few older .csd files when running them with Csound 7.

PR #2310, opened 2025-10-12
Implement break/continue for until and while loops

scottsitar:
There are a few pieces to this:
- add the tokens to the lexer and parser
- in expand_until(), keep track of the targets for break and continue statements for later conversion into gotos
- in verify_tree(), maintain a stack of active loops (to properly handle nested loops)
- when we expand an until/while loop, push onto the stack
- when we encounter the ending label of a loop, pop off the stack
- if we encounter a break/continue statement when there are no active loops, throw an error

### DIFF
--- a/tests/commandline/test_continue_outside_loop_fails.csd
+++ b/tests/commandline/test_continue_outside_loop_fails.csd
@@ -1,0 +1,19 @@
+<CsoundSynthesizer>
+<CsInstruments>
+
+instr 1
+    a:i = 1
+    b:i = 2
+    c:i = 3
+
+    while (c>0) do
+        c = c - 1
+    od
+    continue
+endin
+
+</CsInstruments>
+<CsScore>
+i1 0 1
+</CsScore>
+</CsoundSynthesizer>
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -198,6 +198,12 @@ SYMBOL          [\[\]+\-*/%\^\?:.,!]
 "enduntil"      { *lvalp = make_token(csound, yytext);
                   (*lvalp)->type = OD_TOKEN;
                   return OD_TOKEN; }
+"break"         { *lvalp = make_token(csound, yytext);
+                  (*lvalp)->type = BREAK_TOKEN;
+                  return BREAK_TOKEN; }
+"continue"      { *lvalp = make_token(csound, yytext);
+                  (*lvalp)->type = CONTINUE_TOKEN;
+                  return CONTINUE_TOKEN; }
 "switch"        { *lvalp = make_token(csound, yytext);
                   (*lvalp)->type = SWITCH_TOKEN;
                   return SWITCH_TOKEN; }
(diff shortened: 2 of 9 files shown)


scottsitar:
This implements break and continue just for until/while loops.  If the approach looks good, I can finish the implementation on for loops (expand_for() was a little bigger than expand_until(), so I started with the easier one), and then shore it up with some tests.

vlazzarini:
Thanks very much! It's great to see this.

It's looking good, very readable. I am wondering now about `switch`, which currently has an implicit break (see 'test_switch.csd'). Would it not be better to change it so that it allows fall-through and introduce break to it? @hlolli you are the author, what do you think?

vlazzarini:
I think this looks good but we need a csd test. Could you add one with both init-time and perf-time loops?

scottsitar:
For sure!  I was going to finish implementing these statements in for loops and add some tests before asking to merge.  For the switch statements, I like the way that golang does it - every case has an implicit break at the end, but if you want to fall through, there's an explicit "fallthrough" statement you can put.  It goes against the "C grain", though, so I can definitely understand why people would prefer the other way as well.

scottsitar:
I've pushed up another commit to finish the implementation with for loops as well.  This also adds tests for init and perf rate loops, as well as some negative tests for break/continue statements found outside of loops.  Apart from the question about switch statements, I think this is ready for review now.

scottsitar:
Should I submit a PR against the manual for opcode reference pages for break and continue?

vlazzarini:
> Should I submit a PR against the manual for opcode reference pages for break and continue?

that'd be super

vlazzarini (approved):

tests/commandline/test_continue_outside_loop_fails.csd
```diff
@@ -0,0 +1,19 @@
...
+
+instr 1
+    a:i = 1
+    b:i = 2
+    c:i = 3
+
+    while (c>0) do
+        c = c - 1
+    od
+    continue
```
hlolli, line 12: the continue is here outside of the while scope, is that intentional?

hlolli:
(I've been with cold last few days) I like it a lot, well done, clean code. My only comment is on the test where the continue token is outside a scope. I wonder if we should leave a helpful error message if the tokens break and continue are misplaced.

tests/commandline/test_continue_outside_loop_fails.csd
vlazzarini: that leads to a fail, so it's just a test to make sure it exits cleanly with an error.

scottsitar:
Thanks @vlazzarini and @hlolli for your kind comments, and for the merge!

Yes, as Victor mentioned, the continue outside of the loop was intentional for that test, as it's making sure it's a parsing failure.  There's a similar test for having a break outside loops.  The break happens before a loop and the continue happens after, and this was meant to test that there was no funny business with cleaning up the stack of active loops.

PR #2313, opened 2025-10-15
run: fixed bug in arg mem setting

vlazzarini:
This fixes a bug in setting up the argument memory for array copies in run.


rorywalsh (approved):
Thanks Victor, these look good.

PR #2314, opened 2025-10-17
ci: fix vcpkg build on windows

hlolli:


vlazzarini (approved):

stekyne:
The need for this PR is a good example of why I don't like these custom ports for VCPKG in our repo

vlazzarini:
Can we replace them?

stekyne:
For Portaudio, I need to check if the current latest vcpkg version will support building the ASIO SDK component by default. If it does, then I can remove this overlay port and this issue wouldn't have come up. For the other ports, @nonameentername would need to clarify if he can merge his changes to the ports upstream.

If we keep these ports in our repo, we will end up running into these issue as the ports/packages decay over time. This case is a common example where URLs become stale or removed entirely as they are hosted externally to VCPKG. Even if the dependency version doesn't change, vcpkg maintainers keeps the ports updated to use the correct URLs/data to build the port. We need to leverage that to avoid having to manually resolve build issues with VCPKG.

Ideally, we just need to pull the latest VCPKG changes periodically to keep it all working and up to date.

vlazzarini:
I am also wondering about ASIO. With Wasapi and WDM, is it still relevant? I don't know much about the Windows audio ecosystem, but it may be worth asking the question.

rorywalsh:
I think ASIO is still the only show in town for low latency audio on
Windows, at least in my experience.

stekyne:
Yeah WASAPI is a good default and has decent latency but afaik ASIO is still the best.

PR #2319, opened 2025-10-21
Fix for pcount

### LINKED ISSUE #2318: pcount reporting wrong number of p-fields
instr 1
  schedule(2,0,1)
  schedule(3,1,1)
endin
schedule(1,0,3)

instr 2
  schedule(4,0,1,2,3)
endin

instr 3
  schedule(4,0,1,2)
endin

instr 4
  print pcount()
endin

returns (with -m128):
instr 4:	#i0 = 5.000
instr 4:	#i0 = 5.000

but the second one has actually only four p-fields, not five.

this wrong outputs only happens when schedule is called via another schedule, and only if the number of p-fields is less than in the previous instrument call.

vlazzarini:
This PR fixes #2318.


joachimheintz (approved):
tested and works --- thanks for the quick fix!

PR #2325, opened 2025-10-23
Miscellaneous

vlazzarini:
This PR 

- fixes OSCsend so that it recognises "localhost" as an address
- fixes a memory bug in the OSC message retrieval where the order of checks was wrong
- adds a test for the udp server OSC support
- suppress spurious 33m] characters on output when no msg attributes are used.


rorywalsh (approved):
I'll be happy to see no more of those '[m' chars clogging up my console output!

rorywalsh:
The merge failed with:

[FAILED TESTS]

[FAIL] - Test 142: test OSC in udp server (test_osc_server.csd)
	Return Code: 1	Expected: 0
	
Strange considering all tests had passed prior to merging?

PR #2333, opened 2025-10-26
fixed size calculation in array scalar set etc

### LINKED ISSUE #2332: Check if multidim array size calculations are correct for tabmax, tabmin and tabsum
According to this coderabbitai review https://github.com/csound/csound/pull/2213#discussion_r2462811183 big if true if we are calculating the multidimensional array sizes incorrectly. We should be able to verify this claim by testing it in csd. I'd suggest testing it, if true, make a failing regression test in command line tests (or add it as a case to existing one).

vlazzarini:
Fix for #2332 - scalarset

### DIFF
--- a/OOps/array_ops.c
+++ b/OOps/array_ops.c
@@ -3526,10 +3526,10 @@ int32_t array2asig_perf(CSOUND *csound, ARR2A *p) {
 
 int32_t scalarset(CSOUND *csound, TABCOPY *p) {
   IGN(csound);
-  uint32_t siz = 0 , dim = p->tab->dimensions, i;
+  uint32_t siz = p->tab->sizes[0], dim = p->tab->dimensions, i;
   MYFLT val = *p->kfn;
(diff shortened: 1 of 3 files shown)


vlazzarini:
and tabmin, tabmax, tabsum

hlolli (approved):
nice!

OOps/array_ops.c
```diff
@@ -2772,13 +2772,13 @@ int32_t tabqset1(CSOUND *csound, TABQUERY1 *p)
 int32_t tabmax(CSOUND *csound, TABQUERY *p)
 {
   ARRAYDAT *t = p->tab;
-  int32_t i, size = 0, pos = 0;;
+  int32_t i, size = t->sizes[0], pos = 0;;
```
hlolli, line 2775: nit-pick: double semicolon needed?

vlazzarini: yep not needed. was it there before?

PR #2336, opened 2025-10-26
sequ overload with start point

### LINKED ISSUE #2200: Feature request: can we have a `kstart` for `sequ`?
Having random mutations mode and all is fun but a basic sequencer function in many-a-real-life sequencers (see e.g. all Korg products) is to have the ability to change the starting point of the sequence. 

`sequ` only lets me change the end of the sequence (with `klen`).

vlazzarini:
This PR introduces a new version of sequ with a start point

```
kres = sequ(kstart, irhythm[], iinstr[], idata[], kbpm, klen [, kmode] [, kstep] \
            [, kreset] [, kverbose])
kres = sequ(kstart, irhythm[], iinstr[], idata[][], kbpm, klen [, kmode] [, kstep] \
            [, kreset] [, kverbose])
```
closes https://github.com/csound/csound/issues/2200

### DIFF
--- a/Opcodes/sequencer.c
+++ b/Opcodes/sequencer.c
@@ -61,6 +61,40 @@ typedef struct {
   int32_t         seq[128];
 } SEQ;
 
+typedef struct {
+  OPDS        h;
+  MYFLT       *res;           /*  state */
+  MYFLT       *kstart;        /* kstart */
+  ARRAYDAT    *riff;          /* initial note row */
+  ARRAYDAT    *instr;         /* renderers for each note */
+  ARRAYDAT    *data;          /* extra data for pitch info */
+  MYFLT       *kbpm;          /* speed of sequence */
+  MYFLT       *klen;          /* Length of sequece to use */
+  MYFLT       *mode;          /* Mode; -1 backward,
+                                 0 loop frward;
+                                 +ve mutate
+                                 -1 backward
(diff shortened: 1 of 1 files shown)


vlazzarini:
man page edit will follow PR merging.

rorywalsh:
I'm just wondering if this shouldn't be added as a new `sequ2` opcode rather than overloading the existing one?

vlazzarini:
I prefer not to proliferate the number of opcodes, since now we have the overloading mechanism in place.

rorywalsh:
My first thought was a version 2 (which I have never liked), simply because that falls in line with how things have been done in the past and for better or worse has been a pretty consistant approach.

rorywalsh (approved):

vlazzarini:
I see, but we did not have a consistent overloading mechanism back then, now we do, so I thought we may as well promote it.

rorywalsh:
I wonder if we might now go back and overload all those other version 2 opcodes, marking them as deprecated. There was also talk of moving toward a consistent naming pattern for all opcodes. This is probably part of a larger discussion, but consistency leads to clarity. Anyhow, this PR is merged now. We can take up this discussion in another thread. 👍

vlazzarini:
I think that is a good idea to explore.

PR #2337, opened 2025-10-28
Fix for backwards compatibility in ternary expressions

vlazzarini:
This fixes an issue reported where the expression

`k2 = k0 > 0 ? k0: -1`

gives a parsing error.

### DIFF
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -330,7 +330,7 @@ SYMBOL          [\[\]+\-*/%\^\?:.,!]
             }
 }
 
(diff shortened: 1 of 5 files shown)


gesellkammer (dismissed):

kunstmusik:
Just to note, seems like this will fix the issue, but root cause seems to me that LABEL_TOKEN shouldn't be happening anywhere except at the start of lines. Would be fine to merge this in now but we should really look at the underlying problem in the future.

vlazzarini:
Ok - fixed it in the lexer. Can we have another review @kunstmusik  @gesellkammer

gesellkammer (approved):

PR #2342, opened 2025-11-02
DSSI plugin: Fix pointer type error in GetInputArgCnt() and GetOutput…

### LINKED ISSUE #2341: [PATCH] Csound 7: fixes for DSSI plugin
Trying to compile the "libdssi4cs.so" plugin on Linux, I ran into a couple of problems. The first thing was four similar compiler errors about incompatible pointer types.

Changing GetOutputArgCnt(p) to GetOutputArgCnt(&p->h) etc. solved that problem and the plugin would compile, but would segfault immediately when trying to load a LADSPA plugin.

Running under GDB showed that dssiinit() was trying to access DSSIPlugin->ksmps before the DSSIPlugin object was created. Moving the offending line down to some place after the initialization solved that problem and now it seems to be working, at least for the "zita-reverbs" ladspa plugin.

I should probably mention that I am completely unfamiliar with with the csound code base so everything here was just a wild guess.

[dssi-argc.patch](https://github.com/user-attachments/files/23280326/dssi-argc.patch)

yetanothergeek:
As requested in issue #2341 this PR fixes the first problem (compiler error)

Hope to have a PR for the second problem (segfault in dssiinit) on the way soon.

(BTW this is my first time ever creating a PR, please let  me know if I did anything wrong.)

### DIFF
--- a/Opcodes/dssi4cs/src/dssi4cs.c
+++ b/Opcodes/dssi4cs/src/dssi4cs.c
@@ -261,7 +261,6 @@ int32_t dssiinit(CSOUND * csound, DSSIINIT * p)
     DSSI4CS_PLUGIN *DSSIPlugin =
         (DSSI4CS_PLUGIN *) csound->QueryGlobalVariable(csound, "$DSSI4CS");
     CS_TYPE* argType = GetTypeForArg(p->iplugin);
-    DSSIPlugin->ksmps = Ksmps;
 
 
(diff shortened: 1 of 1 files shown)


yetanothergeek:
Still not sure how this stuff works but from here it looks like the second commit (to fix the segfault in dssiinit) was automatically added to this PR. I thought I would have to open a new PR for that one, but I guess not.

vlazzarini (approved):

PR #2349, opened 2025-11-03
Updated plugin registration for 7.0

### LINKED ISSUE #2347: limit() of array has no effect
Example:
```
<CsoundSynthesizer>
<CsInstruments>

instr 1
 iArray1[] fillarray 1, 1, 1
 iArray2[]=limit(iArray1, 7, 7)
 printarray(iArray2)
endin

</CsInstruments>
<CsScore>
i1 0 1
</CsScore>
</CsoundSynthesizer>

```

If I run this under csound 6.18.1, the output is as expected:
```
7.0000  7.0000  7.0000
```
But under the latest git develop branch, the output is:
```
0.0000  0.0000  0.0000
```

vlazzarini:
Plugin registration had not been updated to 7.0 conventions. This PR fixes the issue.

Closes #2347

### DIFF
--- a/include/plugin.h
+++ b/include/plugin.h
@@ -1172,15 +1172,20 @@ template <typename T>
 int32_t plugin(Csound *csound, const char *name, const char *oargs,
            const char *iargs, uint32_t thr, uint32_t flags = 0) {
   CSOUND *cs = (CSOUND *)csound;
-  if(thr == thread::ia || thr == thread::a) {
+  SUBR initf = NULL, perf = NULL;
+  if(thr == thread::i ||
+     thr == thread::ik ||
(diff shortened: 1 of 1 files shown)


rorywalsh (commented):
I'm assume it's fine to pass NULL as the init function parameter when dealing with exclusively a or k rate threads? In the previous code the init function was always passed as an argument to AppendOpcode()

vlazzarini:
it's not only OK, but it's also the only way now to guarantee the code not to run at init, perf or deinit since we removed the redundant `thread` attribute in OENTRY.

rorywalsh (approved):
Great, thanks for clarifying.

PR #2351, opened 2025-11-03
Deprecation support

vlazzarini:
This PR adds support to handle deprecation

- OENTRY field
- Option to trigger compile error
- Checks for deprecation in semantic analysis
- Examples of deprecated opcodes in entry.c

For example,

```
instr 1
 asig,asig ins
 outs asig, asig
endin
```

gives

```
Compiling Csound code...
warning: opcode ins is deprecated, line 10
warning: opcode outs is deprecated, line 12
	 ...done
```

but with `--error-deprecated`

```
error: opcode ins is deprecated, line 10
 aa ins 
```


rorywalsh (approved):
This looks good Victor. Frontend devs will need to rebuild due to the ABI changing, but better now than after a major release. It would be nice to offer more insightful messages, but I guess that's not possible without some serious work. The manual can best deal with suggestions for more up to date opcodes to use.

vlazzarini:
Not sure if it breaks ABI actually as it's added at the end, but maybe it does. We have not released anything so it's good time to do it.

kunstmusik:
I like this implementation and think this will provide good value. One thing is internally we could split out the OENTRY's for deprecated opcodes to a second list and have an add_deprecated_opcodes function that's called as part of Csound initialization. I think that'd help for code organization.

PR #2352, opened 2025-11-04
Adding aliases for opcodes with underscore and uppercase chars. Also …

rorywalsh:
…created alias for oscbnk to avoid confusion woth osc opcodes


Engine/entry.c
```diff
@@ -649,6 +649,7 @@ OENTRY opcodlst_1[] = {
   { "=.L", sizeof(TABCPY), 0, ".[]", ".[]", (SUBR)tabcopy2, (SUBR)tabcopy2 },
   { "tabgen", sizeof(TABGEN), _QQ, "k[]", "iip", (SUBR) tabgen, NULL    },
   { "tabmap_i", sizeof(TABMAP), _QQ, "k[]", "k[]S", (SUBR) tabmap_set   },
```
vlazzarini, line 651: will you mark the old names as deprecated?

rorywalsh: yes, I completely forgot about that! leave it with me. 

[edit] but we're not really making these opcode deprecated are we. They are still valid, only with a different name. Wouldn't marking `OSCsend`, for example, as deprecated leave users wondering what they should be using instead?

joachimheintz:
just to note that from current develop i get:
warning: opcode directory is deprecated

is "directory" replaced, or is this a mistake?

rorywalsh:
It shouldn't be, let me check..

rorywalsh:
Can you test ftsamplebank Joachim, it uses the same csound->AppendOpcode()
function as directory. Sorry, I don't have the option of building from dev
here at the moment.

vlazzarini:
Yes, from current develop you will get that, but we fixed it in this branch.

vlazzarini:
@rorywalsh I've added a function to the module API to mark an opcode as deprecated.  This should be useful for opcodes registered using csound->AppendOpcodes:

```
csound->Deprecate(CSOUND *csound, char *name, char *oargs, char *inargs, int32_t deprec)
```

I've also added an extra parameter to csnd::Plugin<T>() to mark an opcode as deprecated.

joachimheintz:
thanks rory.  it was just a warning, and i could use the directory 
opcode as usual.  i was only surprised that / whether it was deprecated.

joachimheintz:
i think it is this:

commit b70422e0f83d6f8079bb372bc433084b6925eee2
Merge: f0a5d1d85 681f75c4a
Author: Rory Walsh ***@***.***>
Date:   Mon Nov 3 20:05:16 2025 +0000

     Merge pull request #2351 from csound/feature/deprec

     Deprecation support

vlazzarini:
yes, this PR fixes it. If you run the code from this branch, there will not be any warnings from directory

vlazzarini (approved):

PR #2353, opened 2025-11-05
Fix for linevent_open

vlazzarini:
A longstanding bug relating to line events was caught.

The function `linevent_open()` was called before `init0()` , leading to opening issues for `-L` when realtime events were placed in instrument 0. This PR fixes the issue by moving the call to before `init0()`.


kunstmusik (approved):
Looks good to me.

PR #2360, opened 2025-11-10
Fix for mfree locks

vlazzarini:
In debug builds, locking/unlocking was unbalanced leading to an early exit on MacOS. This PR fixes it.


hlolli (approved):
the prelude to this mess was a segfault I was chasing, sorry about this

kunstmusik:
+1 looks good to me

PR #2366, opened 2025-11-16
Fix performance regression

kunstmusik:
fix(memalloc): removed pointer validation in mfree for release build path


vlazzarini (approved):
trapped and tests are running as before (in Release build).

PR #2368, opened 2025-11-18
WebAudio: Fix microphone input on prod build

kunstmusik:
Production build had an issue where optimization caused an issue that prevented enableAudioInput() (e.g., microphone input) from working. 

I built a new wasm-bin with latest from develop, as well as locally tested with build:prod version of @csound/browser. Went ahead and released new versions of both wasm-bin and browser packages.


rorywalsh (approved):
Thanks Steven!

PR #2369, opened 2025-11-18
updating erroneous line number

rorywalsh:
The used before defined error was showing the wrong line number.

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -999,7 +999,7 @@ char* get_arg_type2(CSOUND* csound, TREE* tree, TYPE_TABLE* typeTable)
         char *res = create_array_arg_type(csound, var);
         if (res==NULL) {        /* **REVIEW** this double syntax error */
           synterr(csound, Str("Array of unknown type\n"));
-          csoundMessage(csound, Str("Line: %d\n"), tree->line-1);
+          csoundMessage(csound, Str("Line: %d\n"), tree->line);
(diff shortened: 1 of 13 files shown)


vlazzarini:
looks ok but
1) why was this deliberately offset by 1?
2) anywhere else we need to fix?

I find that line number reporting is a bit of a hit and miss. It would be good if it were fixed once and for all.

rorywalsh:
It looks like it was deliberately offset, but I can't think why, maybe
something to do with error reporting in CsoundQT. @Joachim Heintz
***@***.***> would you know? I can live with them all being off by
1, but other line numbers are not. Don't merge yet. I might go and touch up
all of the error/warnings so they are a little more consistent in terms of
formatting. It would make adding error reporting to IDEs that support
Csound much simpler.

kunstmusik:
I did an inquiry with Gemini 3 and it found another location that had incorrect line number. (Pushed a commit for that.) I also asked for an analysis about why the -1 was there and it didn't have definitive information, but guessed:

```
1. A carry-over from a previous parser implementation where line numbers were 0-based.
2. An attempt to correct an off-by-one error that existed at that time but is no longer relevant.
```

It also validated that removing the -1 was correct.

rorywalsh:
Thanks Steven, er, AI 🤣 Btw, is there a test .csd there that will trigger all possible errors? I would like to do something with the formatting.

vlazzarini (approved):
I think we can move with this and if other fixes are required we can open another PR

PR #2370, opened 2025-11-20
Fix: WebMidi

kunstmusik:
* Use bracket notation for MIDI and worker methods to prevent minification issues with production build
* Packaged already released for @csound/browser 7.0.0-beta13 to npm


hlolli (approved):
Closure happy

PR #2375, opened 2025-12-01
Correct FSF mailing address

RobinsAviary:
I noticed that in 2024 the free software foundation moved from a physical office to being fully remote. As such, the mailing address has changed as well. I got the new address and information about its update from here: https://en.wikipedia.org/wiki/Free_Software_Foundation
And the new address is also listed on their website: https://www.fsf.org/about/contact/mailing

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,8 +35,8 @@ GNU Lesser General Public License for more details.
 
 You should have received a copy of the GNU Lesser General Public
 License along with this software; if not, write to the Free Software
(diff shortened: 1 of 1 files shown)


DaveSeidel:
Does FSF really use AI extensively? That's news to me. Certainly Csound
does not use it at all (except as a code review aid during development).

vlazzarini:
We don't use AI extensively, but this looks like a good task for it. I've opened a PR to update the addresses. @RobinsAviary thanks for bringing this to our attention.

Issue #1707, opened 2023-04-24
Csound breaks init-time of K opcode outputs between 6.14 and 6.18


hartmans:
Hi.  I upgraded to Debian's csound 6.18.1 and my orchestra broke.  After a bit of diagnosis, I came up with a fairly simple reproducing test case.
My understanding of the opcode docs is that a K output parameter should be initialized both at i-time and at k-time.
So I ought to be able to do
```
sr=44100
opcode inittest,K,0
kinit init 1
xout kinit
endop
instr 1
kinit inittest
if i(kinit) != 1 then
prints "fail\n"
prints "kinit is %d\n", i(kinit)
endif
endin
```
With a score liuke
```
i 1 0 0.1
```
and not see it print failed.
That's what I get with csound 6.14:
```
root@07e5e53f388e:/# csound -onull /tmp/init.orc /tmp/init.sco
0dBFS level = 32768.0
--Csound version 6.14 (double samples) 2020-11-20
[commit: none]
libsndfile-1.0.31
orchname:  /tmp/init.orc
rtaudio: ALSA module enabled
rtmidi: ALSA Raw MIDI module enabled
Elapsed time at end of orchestra compile: real: 0.001s, CPU: 0.001s
sorting score ...
        ... done
Elapsed time at end of score sort: real: 0.001s, CPU: 0.001s
graphics suppressed, ascii substituted
0dBFS level = 32768.0
orch now loaded
audio buffered in 256 sample-frame blocks
writing 512-byte blks of shorts to null (WAV)
SECTION 1:
new alloc for instr 1:
B  0.000 ..  0.100 T  0.100 TT  0.100 M:      0.0
Score finished in csoundPerform().
inactive allocs returned to freespace
end of score.              overall amps:      0.0
           overall samples out of range:        0
0 errors in performance
Elapsed time at end of performance: real: 0.001s, CPU: 0.001s
256 512 sample blks of shorts written to null (WAV)
```
But with 6.18.1:
```
orchname:  /tmp/init.orc
Elapsed time at end of orchestra compile: real: 0.001s, CPU: 0.001s
sorting score ...
        ... done
Elapsed time at end of score sort: real: 0.001s, CPU: 0.001s
--Csound version 6.18 (double samples) 2022-11-24
[commit: none]
libsndfile-1.2.0
graphics suppressed, ascii substituted
sr = 44100.0, kr = 4410.000, ksmps = 10
0dBFS level = 32768.0, A4 tuning = 440.0
[...]

PR #2377, opened 2025-12-03
fix: restore K-type support for k-rate with initialization in polymorphic input types for old-style UDOs

kunstmusik:
Old-style UDOs support "K" type that supposedly copied values at init-time. AI reported that:

The K type handling was added to csound6 in April 2023 (commits 6329e6830 and 4124c3560 by Victor Lazzarini to fix issue #1707)

This fix uses POLY_IN_TYPES to fix the issue.

### DIFF
--- a/Engine/csound_standard_types.c
+++ b/Engine/csound_standard_types.c
@@ -675,6 +675,7 @@ const char* POLY_IN_TYPES[] = {
     "U", "Sikcpr",
     "i", "cpri",
     "k", "cprki",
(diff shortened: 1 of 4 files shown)


vlazzarini (changes requested):
All looks good but can we also have a csdtest?

kunstmusik:
@vlazzarini I added a test_K_type.csd

vlazzarini (approved):
Thanks

PR #2378, opened 2025-12-03
Crash in xoutset - classic style udo

vlazzarini:
This PR fixes a crash in xoutset for classic-style UDOs by reverting recent changes. A test is updated to check for this regression.

It also reverts other unwarranted changes in set_inbufs and prevents reallocation in array copy if there is enough allocated memory already.


Engine/csound_standard_types.c
```diff
@@ -252,7 +252,9 @@ static void array_copy_value(CSOUND* csound, const CS_TYPE* cstype, void* dest,
         }
         aDest->data = cs->Calloc(cs, aSrc->arrayMemberSize * arrayNumMembers);
         aDest->allocated = aSrc->arrayMemberSize * arrayNumMembers;
-    }
+    } else if(arrayNumMembers < array_get_num_members(aDest)) // set sizes, don't reallocate
```
kunstmusik: @vlazzarini I think this should be <=, so it also sets the size in the case when array num members match.

vlazzarini: If they match, then sizes should match too, shouldn't they? I can change this, not a big deal.

kunstmusik: I think there's a subtle edge case, if it goes from something like:

array[4][2]

to:

array[2][4]

The number of members is the same but the sizes are different.

vlazzarini: done.

joachimheintz:
i could figure out now the cause of the error which i get.
it seems to be related to for-loop of string arrays; nothing to do with 
udos.

instr 1
   arr:S[] = ["a", "b", "ccc"]
   for Sample,i in arr do
     print(i)
   od
   turnoff
endin

new alloc for instr 1:
instr 1:	i = 0.000
instr 1:	i = 1.000
instr 1:	i = 2.000
PERF ERROR in instr 1 (opcode ##array_get) line 0: Array index 3 out of 
range (0,2) for dimension 1
Sample:S ##array_get x__synthetic_301[]:S[],i__synthetic_300:i
...event aborted

vlazzarini:
that's something different. Has your original crash been fixed?

joachimheintz:
yes; i thought i replied last night with the report, but perhaps i was 
using a wrong address.
ah yes, i see now ...

so last night i sent the message below to nowhere.  and now i figured 
out the reason for this error message.  it is different, yes.

"yes i did and i think we are on the right track.
it is running now, stable and with normal performance.

i do get strange perf errors which i did not get before:
PERF ERROR in instr 24 (opcode ##array_get) line 0: Array index 3 out of 
range (0,2) for dimension 1
Sample:S ##array_get x__synthetic_409[]:S[],i__synthetic_408:i
...event aborted

any idea?
i don't think it is a programming mistake on my side but i can look 
tomorrow if you don't find something suspicious in the source code. "

vlazzarini:
that may be related to the other change. I'll revert it and you can test it again.

joachimheintz:
unfortunately the simple test which i sent has the same error:
(btw strange warning about loop_lt which is not used at all)

--Csound version 7.0 (double samples) Dec  4 2025
[commit: 16c53752d3d91e729f630fd71a0e582c63edaf98]
Compiling Csound code...
warning: opcode loop_lt.i has been renamed (uppercase to lowercase / 
underscores removed), line 0
	 ...done
(...)
new alloc for instr 1:
instr 1:	i = 0.000
instr 1:	i = 1.000
instr 1:	i = 2.000
PERF ERROR in instr 1 (opcode ##array_get) line 0: Array index 3 out of 
range (0,2) for dimension 1
Sample:S ##array_get x__synthetic_301[]:S[],i__synthetic_300:i
...event aborted
B  0.000 ..  1.000 T  1.000 TT  1.000 M:  0.00000  0.00000
end of Performance
inactive allocs returned to freespace
		   overall amps:  0.00000  0.00000
	   overall samples out of range:        0        0
1 errors in performance

vlazzarini:
leave it with me

edit: it's unrelated to my changes. I am getting this behaviour on the develop branch. It seems something else was wrong.

joachimheintz:
sure --- there is no rush and i am relieved that the main issue is solved.

vlazzarini:
Fixed now. The problem was that ##array_get is called at perf-time for S vars, and by then the loop counter was out of range. The fix, to avoid backwards compat issues, is to have a version of ##array_get that only runs at i-time for S vars. The test was not picking it because the instr was run with p3=0. Now it's been changed to catch it.

joachimheintz:
yes ---- both the simple test and my own project csd now run without 
issues =)
shall we approve and merge?

vlazzarini:
I guess @kunstmusik is reviewing, let's hear from him.

kunstmusik:
Code looks good, but there's a deeper concern about for loops and defaulting to array_geti. Is this going to be the way moving forward for any datatype that it defaults to i-rate for loop? If so, we may need some kind of specific syntax to be able to for k-rate for-loops.

vlazzarini:
It's a specific problem with S because all other types are either perf or k-rate with their array_get operations (as far as I can see). Another solution is to try to change array_get into two overloads which figure out whether it is a k or i index, as it uses "m" ~~(which I think matches both)~~. I'll try that. My worry was change in behaviour, but I guess if the index is i then the perf time get is always the same as i-time.

EDIT: I checked and the situation is this

- S ##array_get() takes i-time indices, but it runs at both i and perf time. Changing this will break any code that modifies the string array at perf time. Since ##array_set() has a similar behaviour for S arrays, then really we have no alternative here. We can't change the opcode, can't have an overload, need a new name.
- The loops have to be set for either perf or init time, exclusively, so S vars need to be placed in one or the other, it makes sense to place them at i-time. 
- The code choses which depending on the array type, currently, a, k and Complex are perf-time only, others are all i-time (including I guess all UDTs).
- ##array_get() for all types except i, a, k, S, and Complex run only at i time if the index given is i-type or only at perf time if the index is k-type.
- For a, k, S, and Complex if the index is i, they run both at i and perf time; for i, it runs only at i time.
- Given this, S-types ##array_get() can run in two ways: given an i-type index, it runs at i and perf time; given a k-type index, it runs at perf time only.
- The type of the index is synthesized in the code, i for i-time loops and k for perf-time loops.

So as far as I can see with the current solution we are covered for everything except in the case of S-types, which we handle as an exception. Strings were always defective and I wished someone had designed them better when they were introduced. We keep having to handle these with special code, it is a real pain.

What we are not considering but it is out of scope for here is how to handle UDT arrays that have perf-time members. At the moment, they are not going to run at perf-time ever.

vlazzarini:
Can we have a decision on this? It looks like this bug has caught others and we need to act.

kunstmusik:
@vlazzarini to clarify, when you mention index, are you speaking of syntax like so for forcing perf-time:

```
 for Sample, kndx in arr do
     print(i)
   od
```

I think that's a reasonable syntax to differentiate between init-time enhanced for-loops and perf-time. For strings, I'm wondering if we couldn't use array_geti and array_getk for specific versions of array getting, and leave array_get as "m" type, which would mostly get used outside of enhanced for loops. (Thinking of "then vs. ithen vs kthen" as something of a precedent for specification of rates for if-elseif-else branches).

vlazzarini:
We don't have such syntax. The type in the case of for-in is implied by the array type (a, k and Complex have k indices, all others have i indices), if the var doesn't already exist. The internal synthetic index also follows thar rule.

The problem is not there. The problem is that S-type array_get runs at both i and perf time. 

The internal synthetic index var is incremented past the array bounds after the init pass and then at the first round of perf-pass it tries to access the array and fails. This happens in any for-in loop and is to do with the internal index not the explicit count var is the culprit.

This issue only happens with S arrays.  This is solved by using the special i-time only array_get.

kunstmusik:
I'm proposing we use this kind of syntax to disambiguate what rate to use with data types generically. It's the only thing so far that's made any sense to me as a rule. No types outside of i, k, a, b, and B have an inherent rate to them. Adding the limitation of indx or kndx for the second arg in for-in loops gives us the opportunity to clearly state exactly what rate we want for any data type when using for-in loops. If it's kndx, it should always get set to 0 each per-pass. Doing this should then work with any arrays.

vlazzarini:
That is not going to solve the current problem. We are already making a clear disambiguation between i and k loops. The problem is that the S array_get will run on both i time and perf time while we need it to run *only* at i time. Using an i-time var to distinguish the pass does not fix this problem. The problem persists because of the way array_get is defined in entry.c.

I propose we merge the fix and then discuss any additions to the for-in loop syntax as a separate issue.

Note that the basic for-in does not have an explicit counter var and so it needs to have a rule regarding action time. All of this is beyond the scope of this PR.

kunstmusik:
I'm okay with merging this in as a stop-gap but the bigger problem needs addressing for both for-in and general array_get (and array_geti/array_getk). We should probably stop using 'm' and could probably get around all of this by differentiating based on the index (i or k) rather than anything related to the array type. (i.e., have a generic array_get with "." and ".[]i" and another with "." and ".[]k".) Same would go for array_set to differentiate on the index.

I'll approve for now but let's prioritize sorting this out.

kunstmusik (approved):

vlazzarini:
I thought about that too, but then worried about backwards compatibility.

PR #2386, opened 2025-12-12
fixed opcode listing to exclude deprecated

vlazzarini:
This PR fixes opcode listing so it understands the new deprecation code.


rorywalsh (approved):
Thanks Victor, look good. 👍

PR #2387, opened 2025-12-19
Check for dlfcn.h with cmake

rrrapha:
This change makes this part of the code work on OpenBSD (and probably other BSDs as well).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -658,7 +658,7 @@ set(HEADERS_TO_CHECK
     unistd.h io.h fcntl.h stdint.h
     sys/time.h sys/types.h termios.h
     values.h winsock.h sys/socket.h
-    dirent.h inttypes.h execinfo.h)
+    dirent.h inttypes.h execinfo.h dlfcn.h)
 
 foreach(header ${HEADERS_TO_CHECK})
     # Convert to uppercase and replace [./] with _
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -676,7 +676,7 @@ if (HAVE_EXECINFO_H)
     add_definitions("-DHAVE_EXECINFO")
 endif()
 
-if (HAVE_DLFCN_H)
+if (HAVE_DLFCN_H OR ANDROID)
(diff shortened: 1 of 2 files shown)


vlazzarini:
@rrrapha thanks for this - would you know/be able to add a BSD build to our actions (https://github.com/csound/csound/blob/develop/.github/workflows/csound_builds.yml) so we can keep track of builds for that platform in CI?

vlazzarini:
@rrrapha looks like the changes to `csmodule.c` guards triggered errors in a couple of the builds.

rrrapha:
> @rrrapha looks like the changes to `csmodule.c` guards triggered errors in a couple of the builds.

Yes, I guess the check has to be the other way round, like this:

```
#if defined(HAVE_DLFCN_H)
#include <dlfcn.h>
#elif defined(WIN32)
#include <windows.h>
#endif
```

.. and the same condition on line 1034 has to be reversed as well.

I will commit such a change in a minute.

vlazzarini:
It doesn't seem to do the trick. Maybe restoring the original guards and making one just for BSD?

rrrapha:
> It doesn't seem to do the trick. Maybe restoring the original guards and making one just for BSD?

I would love to know why this is not working!

Of course, the easy solution is to add someting like:
`|| defined(__OpenBSD__) || defined(__FreeBSD__) || defined(__NetBSD__) || ..  `.

Using `defined(__BSD__)` would be better, but for this, another header has to be included (`sys/param.h`).

I will make a suggestion next week ..

vlazzarini:
It seems the Android build fails because it can't find dlerror, and the VCPKG fails because it can't load shared objects (plugins) during the test phase. Beyond that I can't say I understand it either.

vlazzarini:
I am afraid that won't fix Android as it doesn't use CMake. It needs to be fixed in the csmodule.c guards (basically restoring them for Android).

I set the CI running to confirm it.

rrrapha:
Okay, this approach won't work in this case. I will open a separate pull request with a different solution.

PR #2393, opened 2025-12-23
Avoid %n format specifier

rrrapha:
The `%n` format specifier for `printf` and friends is not portable.

It has been disabled in OpenBSD because of its security implications:
https://man.openbsd.org/printf.3

It also doesn't work with MSVC for the same reason:
https://learn.microsoft.com/en-us/cpp/c-runtime-library/format-specification-syntax-printf-and-wprintf-functions

The compiler-warnings in OpenBSD disappear with this change, but I don't really know how to test this ..


vlazzarini:
Looks OK; the error is because the C compiler does not like the variable declaration in the switch block - wrapping the code with { } should fix it.
Thanks.

vlazzarini (approved):

PR #2394, opened 2025-12-23
Misc fixes

vlazzarini:
This PR:

- fixes tree printing for assignment expressions
-  moves opcode def readonly var creation from compile_tree to orcparse function. This allows the orcparse function to be called independently in the correct way.


rorywalsh (approved):

joachimheintz:
great to see this active now:

instr 1
   if "S" == "S" then
     puts("Hurra!",1)
   endif
endin

SECTION 1:
Hurra!
end of Performance

=)

PR #2396, opened 2025-12-26
Fix python shebang lines

rrrapha:
Using `env` seems to be more portable (I hope CI can handle it too). On OpenBSD, for example, the python executable is not located in `/usr/bin`.


vlazzarini (approved):
Looks good thanks.

PR #2398, opened 2025-12-27
fixed event opcode so it can take a minimum of 2 args

### LINKED ISSUE #2397: event "e" not working with only one argument
instr 1
  event("e",1)
endin

throws 
error: insufficient required arguments for opcode event on line 12:

it works for event_i, but currently event requires one additional arugument.
would be good to have the same requirements as for event_i.

vlazzarini:
Closes #2397


joachimheintz (approved):
works now, thanks

Issue #2402, opened 2025-12-29
ServerTests.testServer hang


rrrapha:
When running `make test` on OpenBSD, test #64 (ServerTests.testServer) hangs forever:
Here is the output of the hanging test:

```
x1$ ./tests/c/unittests --gtest_filter=ServerTests.testServer
Running main() from /usr/obj/ports/gtest-1.15.2/googletest-1.15.2/googletest/src/gtest_main.cc
Note: Google Test filter = ServerTests.testServer
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ServerTests
[ RUN      ] ServerTests.testServer
--Csound version 7.0 (double samples) Dec 29 2025
[commit: 45b655eec6a0a24a1d737ff6253ef05f2ac5911f]
using callback interface
using libsndfile-1.2.2
graphics suppressed, ascii substituted
sr = 44100.0, kr = 4410.000, ksmps = 10
0dBFS level = 32768.0, A4 tuning = 440.0
audio buffered in 256 sample-frame blocks
PortAudio V19.7.0-devel, revision 147dd722548358763a8b649b3e4b41dfffbcfbb6
   0: dac0 (default [sndio, 128 in, 128 out])
PortAudio: selected output device 'default'
writing 256 sample blks of 64-bit floats to dac
SECTION 1:
```

Surprisingly, if I build csound with `-DREQUIRE_PTHREADS=OFF`, the test passes.

Is this expected?

vlazzarini:
No, I would not expect it unless you don't have pthreads. Generally I would think it needs some other threading lib in place of it - I think the only current case is on Windows - for the server to work.

Are you sure you can run the server with that setting?

EDIT: It doesn't look like you have a server running in that test. Here's what you should see:

```
victor@firebird debug % tests/c/unittests --gtest_filter=ServerTests.testServer
Running main() from /Users/victor/src/googletest/googletest/src/gtest_main.cc
Note: Google Test filter = ServerTests.testServer
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from ServerTests
[ RUN      ] ServerTests.testServer
Reading options from $HOME/.csound7rc
--Csound version 7.0 (double samples) Dec 29 2025
[commit: 45b655eec6a0a24a1d737ff6253ef05f2ac5911f]
using libsndfile-1.2.2
UDP server started on port 44100
graphics suppressed, ascii substituted
[...]

PR #2404, opened 2025-12-29
Server errors if thread cannot be started

vlazzarini:
In #2402 there is a report of a test error because of the server thread not starting (failing silently). This PR makes it return an error if the thread is not started.


vlazzarini:
Also added C11 threads support.

rorywalsh (approved):

PR #2405, opened 2025-12-30
Simplify CMakeLists.txt by adding definitions for headers in the loop

### LINKED ISSUE #2402: ServerTests.testServer hang

rrrapha:
This also (partly) resolves #2402 by fixing a typo (`HAVE_SYS_SOCKETS_H` -> `HAVE_SYS_SOCKET_H`).


CMakeLists.txt
```diff
@@ -1469,33 +1470,10 @@ endif()
 
 # Pass flags according to system capabilities
 
-if(HAVE_WINSOCK_H OR HAVE_SYS_SOCKETS_H)
```
vlazzarini: Are these defined elsewhere? You are removing all of them, are they not needed?

vlazzarini, line 1472: I checked this again and while it looks OK, there is a slight change of behaviour. In the previous script the flags below were added to the libcsound_CFLAGS list and in this change, they are added as definitions to the complete project. Probably does not hurt but it needs to be confirmed.

vlazzarini: I had another think and this is probably OK and maybe an improvement as the tokens are made available everywhere. Let's go with and if we see issues, we'll rethink.

vlazzarini (approved):

PR #2409, opened 2026-01-02
removed extraneous HAVE_INTTYPES_H definition

vlazzarini:
With the latest changes in token definition, the compilation of mpadec was broken because of an extraneous `HAVE_INTTYPES_H` definition. This fixes the problem.


vlazzarini:
Superseded by #2410

PR #2410, opened 2026-01-02
Fixed array expressions.

vlazzarini:
This PR fixes an issue reported by @gesellkammer where the type of array in an expression was mangled.

```
kA[] fillarray 1, 2, 3
kB[] fillarray 2, 4, 6
kC[] = (kA * kB) * 0.5

error:

error: opcode '##mul' for expression with arg types :k[;c not found, line 17
SEMERR: opcode '##mul' for expression with arg types :k[;c not found, line 17
```
The above code now works and has been added to a test.

### DIFF
--- a/Opcodes/OSC.c
+++ b/Opcodes/OSC.c
@@ -1178,8 +1178,8 @@ static OENTRY localops[] = {
     { "OSCcount", S(OSCcount), 0,  "k", "",
     (SUBR)OSCcounter, (SUBR)OSCcounter, NULL, NULL, 2 },
   /* aliases */
-  { "OSCsend_lo", S(OSCSEND), 0,  "", "kSkSN",
-    (SUBR)osc_send_set, (SUBR)osc_send, (SUBR) oscsend_deinit, NULL },
+  { "oscsendlo", S(OSCSEND), 0,  "", "kSkSN",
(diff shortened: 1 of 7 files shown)


vlazzarini:
@hlolli I ended up restoring the old code for this. I am not sure what motivated the changes but the issue addressed here was that one of the checks was reversed - but the comment above it seemed to agree with the faulty code. I left your code commented out and added some comments. All tests are passing, so I am suspecting restoring the old code is the best way forward.

Opcodes/OSC.c
```diff
@@ -1178,8 +1178,8 @@ static OENTRY localops[] = {
     { "OSCcount", S(OSCcount), 0,  "k", "",
     (SUBR)OSCcounter, (SUBR)OSCcounter, NULL, NULL, 2 },
   /* aliases */
-  { "OSCsend_lo", S(OSCSEND), 0,  "", "kSkSN",
```
kunstmusik, line 1181: checking here as I don't use the OSC opcodes much, is this okay to change the name of the opcode here?

vlazzarini: yes, that's an alias. Rory's work on normalising names, this one was left over.

kunstmusik (approved):
I double checked to print out the kC[] values at perf-time to make sure it was calculating the correct values which it was.

PR #1778, opened 2023-10-31
io: Don't delete an already existing file: juste overwrite it.

lkoenig:
This allows system monitoring the file to get notified when the file change.

PR #2414, opened 2026-01-06
File overwrite warning fixed

vlazzarini:
#1778 introduced a warning message replacing removal of existing files. However this message was always printed if the file existed or not. This gave the wrong impression that files were overwritten when they did not exist in the first place.

This PR fixes the problem by restoring the original code and adding a warning that is only triggered if the file exists, correcting the issue introduced by that PR.

### DIFF
--- a/Engine/filesys.c
+++ b/Engine/filesys.c
@@ -428,9 +428,9 @@ static FILE *csoundFindFile_Std(CSOUND *csound, char **fullName,
 
 static void overwrite_warning(CSOUND *csound, const char *name) {
   int32_t fd;
-  fd = open(name, O_RDONLY | O_CREAT | O_EXCL);
+  fd = open(name, WR_OPTS | O_EXCL);
   if(fd == -1)
-    csoundWarning(csound, "file %s exits...\n...will be overwritten", name);
+    csoundWarning(csound, "file %s exists...\n...will be overwritten", name);
   else 
     close(fd);
 }
--- a/Engine/filesys.c
+++ b/Engine/filesys.c
@@ -426,6 +426,15 @@ static FILE *csoundFindFile_Std(CSOUND *csound, char **fullName,
     return (FILE*) NULL;
 }
 
+static void overwrite_warning(CSOUND *csound, const char *name) {
+  int32_t fd;
+  fd = open(name, O_RDONLY | O_CREAT | O_EXCL);
+  if(fd == -1)
+    csoundWarning(csound, "file %s exits...\n...will be overwritten", name);
+  else 
+    close(fd);
+}
+
 static int32_t csoundFindFile_Fd(CSOUND *csound, char **fullName,
                              const char *filename, int32_t write_mode,
                              const char *envList)
@@ -451,6 +460,7 @@ static int32_t csoundFindFile_Fd(CSOUND *csound, char **fullName,
     }
     else if (csoundIsNameFullpath(name)) {
       /* if write and full path: */
+      overwrite_warning(csound, name);
       fd = open(name, WR_OPTS);
       if (fd >= 0)
         *fullName = name;
@@ -480,6 +490,7 @@ static int32_t csoundFindFile_Fd(CSOUND *csound, char **fullName,
     }
     /* if write mode, try current directory last */
     if (write_mode) {
+      overwrite_warning(csound, name);
       fd = open(name, WR_OPTS);
       if (fd >= 0) {
         *fullName = name;
@@ -613,8 +624,7 @@ char *csoundFindInputFile(CSOUND *csound,
  */
 char *csoundFindOutputFile(CSOUND *csound,
                            const char *filename,
-                           const char *envList)
-{
+                           const char *envList) {
     char  *name_found;
     int32_t   fd;
 
@@ -623,10 +633,6 @@ char *csoundFindOutputFile(CSOUND *csound,
     fd = csoundFindFile_Fd(csound, &name_found, filename, 1, envList);
     if (fd >= 0) {
       close(fd);
-      csound->Warning(csound,
-                      Str("output file %s already exists...\n"
-                                  "\t...will be overwritten"),
-                      filename);
     }
     return name_found;
 }
--- a/Engine/filesys.c
+++ b/Engine/filesys.c
@@ -58,8 +58,7 @@
 
 
 /** Check if file name is valid, and copy with converting pathname delimiters */
-char *csoundConvertPathname(CSOUND *csound, const char *filename)
-{
+static char *convert_path_name(CSOUND *csound, const char *filename) {
     char  *name;
     int32_t   i = 0;
 
@@ -88,7 +87,7 @@ char *csoundConvertPathname(CSOUND *csound, const char *filename)
 }
 
 /**  Check if name is a full pathname for the platform we are running on. */
-int32_t csoundIsNameFullpath(const char *name)
+static int32_t is_name_full_path(const char *name)
 {
 #ifdef WIN32
     if (isalpha(name[0]) && name[1] == ':') return 1;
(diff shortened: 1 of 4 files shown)


Engine/filesys.c
```diff
@@ -426,6 +426,15 @@ static FILE *csoundFindFile_Std(CSOUND *csound, char **fullName,
     return (FILE*) NULL;
 }
 
+static void overwrite_warning(CSOUND *csound, const char *name) {
+  int32_t fd;
+  fd = open(name, O_RDONLY | O_CREAT | O_EXCL);
+  if(fd == -1)
+    csoundWarning(csound, "file %s exits...\n...will be overwritten", name);
```
kunstmusik: typo: shoudl be `file %s exists`

kunstmusik:
@vlazzarini: I saw the behavior now of removing files as part of finding the file which seemed strange to have as part of API. (Seems like user of that funciton should be using find then determining to remove or not). Also saw the changes to mpadec.h which seemed unrelated to the PR.  I saw there were some other recent changes regarding HAVE_INTTYPES_H but didn't follow those closely; could you double-check this is correct here?


I also ran an AI review of the changes locally and it gave:

 ---
  Critical Issues 🔴

  1. Race Condition + Side Effects in overwrite_warning() (Engine/filesys.c:429-436)

  static void overwrite_warning(CSOUND *csound, const char *name) {
    int32_t fd;
    fd = open(name, O_RDONLY | O_CREAT | O_EXCL);  // PROBLEMATIC
    if(fd == -1)
      csoundWarning(csound, "file %s exits...\n...will be overwritten", name);
    else
      close(fd);
  }

  Problems:
  - Creates zero-byte files: When file doesn't exist, O_CREAT creates an empty file that's immediately closed
  - TOCTOU race condition: Gap between this check and actual open(name, WR_OPTS) on lines 464/494
  - Wrong check for existence: Using file creation to test existence is incorrect

  Recommendation:
  static void overwrite_warning(CSOUND *csound, const char *name) {
    struct stat st;
    if (stat(name, &st) == 0 && S_ISREG(st.st_mode))
      csoundWarning(csound, Str("file %s exists...\n...will be overwritten"), name);
  }

  ---
  2. BREAKING CHANGE: Silent File Deletion (Engine/filesys.c:636-637)

  Before (develop branch):
  csound->Warning(csound, Str("output file %s already exists...\n\t...will be overwritten"), filename);

  After (this PR):
  if(remove(name_found)<0)
    csound->DebugMsg(csound, Str("Remove failed\n"));

  Problems:
  - ⚠️ Major behavioral change: Function now deletes files instead of just warning
  - Data loss risk: Silently removes user files without explicit consent
  - API contract violation: csoundFindOutputFile() should find locations, not delete files
  - Only debug message on failure: Users won't know files were deleted

  Question: Is this intentional? The commit message says "warning fixed" but this changes behavior from warning to deleting. If
  deletion is needed, it should be a separate, explicit operation, not hidden in a "find" function.

  ---
  Medium Issues 🟡

  3. Typo in Warning Message (Engine/filesys.c:433)

  csoundWarning(csound, "file %s exits...\n...will be overwritten", name);

  Should be: "file %s **exists**" (not "exits")

  ---
  4. Inconsistent Warning Coverage

  The overwrite_warning() is only called in 2 of 3 places where output files are opened:
  - ✅ Line 463: Full path write mode
  - ✅ Line 493: Current directory write mode
  - ❌ Missing: Files opened via search paths (lines 474-489)

  The original warning in csoundFindOutputFile() covered all cases. This version has inconsistent coverage.

  ---
  5. libmpadec Config Changes (InOut/libmpadec/mpadec_config.h:33-35)

  Removed:
  #undef OSS                          // Line 33
  #define HAVE_INTTYPES_H             // Line 35

  Impact:
  - Windows: OSS no longer explicitly undefined (likely harmless but unclear why)
  - Non-Windows: HAVE_INTTYPES_H no longer defined (could affect type handling)

  Question: These changes seem unrelated to the "file overwrite" feature. What's the rationale?

  ---
  Positive Aspects ✅

  - Code formatting improvement (brace placement on line 627)
  - Intent to warn users before overwriting is user-friendly

  ---
  Recommendations

  1. Fix overwrite_warning(): Use stat() instead of open() with O_CREAT
  2. Reconsider file deletion: Either:
    - Revert to warning-only behavior, OR
    - Make deletion explicit and opt-in with clear documentation
  3. Fix typo: "exits" → "exists"
  4. Ensure consistent warning coverage: Cover all file opening paths
  5. Clarify libmpadec changes: Either revert if unrelated or add justification
  6. Add tests for file overwrite scenarios

  ---
  Testing Checklist

  Before merging:
  - No zero-byte files created
  - Warning shows correctly when files exist
  - Files aren't unexpectedly deleted (if deletion kept)
  - Works with concurrent processes
  - Cross-platform testing (Windows, macOS, Linux)

  ---
  Status: 🔴 DO NOT MERGE

  This PR has critical issues that must be addressed:
  1. Race condition and file creation side-effect
  2. Potentially unintended breaking change (silent file deletion)
  3. Inconsistent warning coverage

  The changes need significant revision before merging.

vlazzarini:
This is all wrong:

Testing Checklist

Before merging:

No zero-byte files created

- zero byte files were always created, this is why the warning message was wrong.

Warning shows correctly when files exist
Files aren't unexpectedly deleted (if deletion kept)

- warning showed if file existed or if did not. This is the whole point of the PR.

The concurrency issue needs to be investigated but these two points are rubbish.

EDIT: the concurrency issue is misleading, as the code does attempt to stop overwriting (as before) only issues a warning, which in 6.x was never issued anyway. The problem with stat is that it does not exist on all platforms (I checked) so the suggested code is a non-starter.

~There are problems with unit tests, which I will look at.~ Looked at these, I had the wrong mode set up so it failed because of that, let's see if it passes.

kunstmusik:
API-wise it doesn't make sense to me to remove() a file when trying to "find" it:

```
if(remove(name_found)<0)
csound->DebugMsg(csound, Str("Remove failed\n"));
```

why is this here? Maybe it's better to change the name of the function then as it's confusing, or at minimum update the function docs to explain it will remove any files if a suitable path is found but a file exists. 

The other issue related to 0-byte file is due to using `O_CREAT`. That might actually be the source of the issue as to why it'd trigger a warning of an existing file as it's going to create one.

vlazzarini:
> API-wise it doesn't make sense to me to remove() a file when trying to "find" it:
> 
> ```
> if(remove(name_found)<0)
> csound->DebugMsg(csound, Str("Remove failed\n"));
> ```
> 
> why is this here? Maybe it's better to change the name of the function then as it's confusing, or at minimum update the function docs to explain it will remove any files if a suitable path is found but a file exists.
> 
I don't know for sure, I think this code was there for > 20 years. It may be because the file is created (with 0 bytes!) by the call used to check if the output file can be written.

> The other issue related to 0-byte file is due to using `O_CREAT`. That might actually be the source of the issue as to why it'd trigger a warning of an existing file as it's going to create one.

I don't think it's an issue. The code always did that anyway. I made a mistake in my code using 
O_RDONLY, it should have been O_WRONLY.

kunstmusik:
I guess the removal behavior just needs documentation then if it's been
used that way for a long time. It's a little awkward in terms of the "find"
name which implies no write operations but can live with it if it's clearly
denoted.

vlazzarini:
Do you mean documenting in the source code? 

I am not particularly married to the idea of using remove() there, it is how it was so I just wanted to restore behaviour. The file is created and then removed. It could be that if Csound fails to continue before the file is open again, there will be a file with zero bytes left over.

kunstmusik:
Yes I was thinking of documentation in the source code. I think it'd be
fine if we restored the behavior in this PR to get us back to a good
baseline and we can look at another PR for further changes. I think you
already removed O_CREAT in a58e2c8d368a8765279db4a7d1dfb96f87f3c3ab so we
might already be in a better position than before as it shouldn't create a
0-byte file now. Changing the behavior to not remove will take some digging
around the codebase to update calling code and things and probably better
to just leave it for now and get this merged.

kunstmusik:
I merged in #2410 and it looks like there's a conflict now. I'm unsure if the `#undef OSS` is correct or not. @vlazzarini could you check this conflict?

vlazzarini:
will do

kunstmusik:
@vlazzarini double checking here, latest changes seem to reset things and rely on removing O_CREAT as the primary fix. Name changes for filesys.h seem okay. Anything else I'm missing? Otherwise can approve and merge.

vlazzarini:
> @vlazzarini double checking here, latest changes seem to reset things and rely on removing O_CREAT as the primary fix. Name changes for filesys.h seem okay. Anything else I'm missing? Otherwise can approve and merge.

Changes were to do with using write options instead of read options which was an error on my part at the outset. I think that resolved the problem.

kunstmusik:
Sounds good, let's run with this.

kunstmusik (approved):

PR #2416, opened 2026-01-07
Removed old manual from Windows installer

vlazzarini:
The Windows installers from CI were carrying the old 6.x manual. This has now been removed from the installer. It is better to refer to the up-to-date new manual online, which can be updated more dynamically.

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -502,13 +502,6 @@ jobs:
       - name: Run unit test framework
         run: cmake --build build --target RUN_TESTS --config Release
 
-      - name: Acquire Csound manual
-#        if: github.ref == 'refs/heads/develop'
-        shell: powershell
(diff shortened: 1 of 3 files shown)


rorywalsh:
I think this is a reasonable modification although I think some frontends might depend on the manual being installed with Csound? @gesellkammer should know if this is the case with CsoundQT, I'm happy for Cabbage users to manually download the manual if they need a local copy.

gesellkammer:
We have an option to download the manual from csoundqt or to point csoundqt to the place where the user downloaded it. In any case, I think it is a good thing to decouple the manual from csound itself.

kunstmusik:
I'm including the manual with Blue (used to have user option to set a
manual directory/URL, but removed that; may put that back in). Either way,
no issues here if the manual is removed.

rorywalsh (approved):

PR #2418, opened 2026-01-08
Code reorganisation

vlazzarini:
This PR re-organises part of the core engine codebase:

1) Function names are normalised according to agreed patterns:
   (a) API functions: csound->*** mapped to csound*** (some of these are made public if in host API,   not all are)
   (b) other functions use ***_***() patterns (whenever possible)
2) local functions marked static.
3) Obsolete and unused code removed.
4) Code moved to be grouped into related units. In particular Top/csound.c was split into separate source files as it had become too large and unwieldy.

Some ancillary changes to build scripts to fix issues with the Debug configuration (as it included optimisations when it should not) and the macosx brew build custom configuration tightened to pick up errors.

In general, this PR did not touch the actual code behaviour except for csoundEvalCode() where the return result in case of failure was clarified.


vlazzarini:
superseded by #2421

PR #2421, opened 2026-01-09
Code reorganisation and JS API fixes

### LINKED ISSUE #2419: Wasm out of sync with Csound 7 API
A look at the file

https://github.com/csound/csound/blob/develop/wasm/src/exports.json

and 

https://github.com/csound/csound/blob/develop/wasm/browser/src/libcsound.js

seems to indicate that the wasm build is out of sync with the Csound 7 API (in /include).

There are many functions used in both files that have either changed or have been removed.
Some new functions are also not figuring there. 

This needs to be reviewed.

vlazzarini:
This PR re-organises part of the core engine codebase and adjust the JS code (libcsound.wasm and browser) to match the Csound 7 API.

1) Function names are normalised according to agreed patterns:
   (a) API functions: csound->*** mapped to csound*** (some of these are made public if in host API,   not all are)
   (b) other functions use xxx_xxx() patterns (whenever possible)
2) local functions marked static.
3) Obsolete and unused code removed.
4) Code moved to be grouped into related units. In particular Top/csound.c was split into separate source files as it had become too large and unwieldy.
5) JS API exports and browser imports are fixed.

Some ancillary changes to build scripts to fix issues with the Debug configuration (as it included optimisations when it should not) and the macosx brew build custom configuration tightened to pick up errors.

In general, this PR did not touch the actual code behaviour except for csoundEvalCode() where the return result in case of failure was clarified.

In the case of JS, one of the tests had to be disabled because of the API changes. One small change removing an obsolete API call was made.

Closes #2419


vlazzarini:
@kunstmusik ready for review

kunstmusik:
---

### Code review

No issues found. Checked for bugs and CLAUDE.md compliance.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>
---

kunstmusik:
(FYI: Tried out a new Claude Code code review tool locally and it posted a message automatically which I deleted 😅. Still reviewing here.)

vlazzarini:
I guess it helps out with such a big PR.

hlolli:
I take it that the code style guide here is, camelCase for exported functons and snake_case for internal csound function. For for example mmalloc it's nice to have consistency as it changes names. Also preventing symbols from clashing. I like this refactor. It's big but I see mostly the same kind of no-op changes. Plus one from me. But I await Steven's more thorough review :)

vlazzarini:
> I take it that the code style guide here is, camelCase for exported functons and snake_case for internal csound function. For for example mmalloc it's nice to have consistency as it changes names. Also preventing symbols from clashing. I like this refactor. It's big but I see mostly the same kind of no-op changes. Plus one from me. But I await Steven's more thorough review :)

camelCase for API functions but not all are exported to the host API. Some are only accessible through the Csound object, for example csoundMalloc() is not public but can be got via csound->Malloc() if csoundCore.h is included. Hosts cannot include it though.

If we find that we need to export any of these, we only need to add it to an API header.

kunstmusik:
I flagged two changes for WASM that needs some attention:

1. replacement of csoundStop() with a dummy fn
2. removal of csoundPerform()

For #1, I'd prefer it just be removed and any existing calls to it be removed. csound.stop() is still important for webaudio csound browser library to exist as it does other things to stop csound performance. If we're going to remove it we should do it completely and remove it from not only WASM but browser code, but elsewhere in the library (rthaiku.c, csound.c). 

For #2, I have to check what is going on in regards to a book example that renders to disk ahead of time. I was calling csound.perform() which I believe was calling csoundPerform(), and awaiting that to complete before post processing the rendered audio. I think it's probably fine and I may need to update the example code and do some text revision in how this is explained. I'll need to do some local builds and testing to verify, may take a day or two to test all this.

.github/workflows/csound_builds.yml
```diff
@@ -212,7 +212,7 @@ jobs:
           brew install bison flex jack googletest
 
       - name: Configure build
-        run: cmake -B build -DBUILD_TESTS=1 -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake"
+        run: cmake -B build -DCUSTOM_CMAKE="./platform/osx/custom-osx.cmake"
```
kunstmusik, line 215: Why is BUILD_TESTS=1 removed here?

kunstmusik:
Quick update, #2 is fine to remove csoundPerform and csound.perform(), ahead-of-time render using webWorker version of browser Csound works fine using a renderEnded event listener. I did this testing using the last published version of @csound/browser package, so it should be fine for this version here too. 

@vlazzarini regarding #1, what do you think about removing csoundStop completely from wasm (but careful to preserve csound.stop() in the CsoundObj API), and removal of the function in rthaiku.c and csound.c?

kunstmusik:
Note: Was doing a few AI reviews earlier with different models, one raised a concern about removing PUBLIC from .c files for Windows, but I think it was ultimately a false positive. One interesting thing that did come up from the session was this recommendation:

```
Recommendation: Run a Windows build and dump exports (e.g., with dumpbin /EXPORTS csound.dll) to confirm the functions like csoundSetPlayopenCallback are actually exported. This validates the build rather than assuming failure.
```

I wasn't aware of dumpbin, so I thought I'd note that here.

.github/workflows/csound_builds.yml
vlazzarini: it's already set in custom-osx.cmake.

vlazzarini:
> @vlazzarini regarding #1, what do you think about removing csoundStop completely from wasm (but careful to preserve csound.stop() in the CsoundObj API), and removal of the function in rthaiku.c and csound.c?

the problem with wasm is that the code depends on the existence of a "csoundStop" in order to generate/bind a csound.stop(). I ran into that problem and instead of making changes to the browser code, I just decided to add a dummy call. If you or @hlolli  know a better work-around I'm happy to go with it. 

The plan was to remove it altogether if it hasn't gone yet from the Csound library.

kunstmusik:
@vlazzarini I'll take a look at removing it from the WASM/browser side

vlazzarini:
> @vlazzarini I'll take a look at removing it from the WASM/browser side

Ok, the issue arises in the initialize function in the vanilla.main.js, sab.main.js and worklet.main.js files. There is a switch case that is never reached if csoundStop is not in the API list. Due to this, stop() never gets bound to the object. I guess a solution is to define csoundStop() similarly to csoundPause() and csoundResume().

I did not try it because I don't have the toolchain locally and I did not want to do trial-and-error on CI again.

kunstmusik:
@vlazzarini : I spent some time looking at removing stop from the browser build and it's not very pleasant. Overall, I came to the conclusion that we would benefit from a bigger rewrite of the wasm/browser codebase. I suppose it's harmless right now to have the no-op so will leave that for this PR and continue on with rest of review.

@hlolli: let's schedule some time to meet and go through the wasm/browser codebase to refactor.

H/csound_orc_semantics.h
```diff
@@ -42,7 +42,7 @@ OENTRIES* find_opcode2(CSOUND*, char*);
 char* resolve_opcode_get_outarg(CSOUND* csound,
                                 OENTRIES* entries, char* inArgTypes);
 void do_baktrace(CSOUND *csound, uint64_t files);
-char* get_arg_string_from_tree(CSOUND* csound, TREE* tree,
+char* csoundGetString_from_tree(CSOUND* csound, TREE* tree,
```
kunstmusik: Unsure about this function rename, as it's not an API function is it? It's also got a weird mix of camelCase and snakeCase, and the function name feels less descriptive than something like csoundGetArgStringFromTree might be.

vlazzarini: yes, that's an artefact from sed. It should not have changed. I'll revert that one.

kunstmusik:
@vlazzarini: I see that Top/init_static_modules.c was removed from the WASM NIX build. I see it is still included in Android.mk.  Seems to me csmodule.c is now the source-of-truth for static module initialization, and that we can:

1. remove init_static_modules.c from the repo
2. Add INIT_STATIC_MODULES=1 to the Android (and iOS?) builds
3. Update Top/README.md

Thoughts?

Engine/rdscor.c
```diff
@@ -24,7 +24,7 @@
 #include "corfile.h"
 #include "udo.h"
 
-char* get_arg_string(CSOUND *csound, MYFLT p)
+char* csoundGetString(CSOUND *csound, MYFLT p)
```
kunstmusik: Note, this name is pretty generic (vs. something like csoundGetArgString), but it *does* match with the GetString in the CSOUND_ struct. Suppose its fine to keep going with this but thought I'd note this here.

Engine/musmon.c
```diff
@@ -1376,9 +1373,9 @@ static int32_t insert_event_node(CSOUND *csound, EVTNODE *e, int64_t time_ofs) {
                  csound->engineState.instrtxtp[i] == NULL)) {
       if (i > INT32_MAX-10)
         csoundErrorMsg(csound, "%s",
-                      Str("insert_score_event(): invalid named instrument\n"));
+                      Str("csoundEvent_(): invalid named instrument\n"));
```
kunstmusik: @vlazzarini: Double-chcking here that this message was intended to change from "insert_score_event()" to "csoundEvent_()". Seems like a another mistaken find/replace issue but want confirmation.

Engine/entry.c
```diff
@@ -1172,8 +1172,8 @@ const OENTRY opcodlst_1[] = {
   { "tigoto", S(GOTO),0,          "",     "l",    tigoto                  },
   { "tival",  S(EVAL),0,          "i",    "",     tival                   },
   { "print",  S(PRINTV),WR,       "",     "m",    printv                  },
-  { "display.k",S(DSPLAY),0,      "",     "kioo", dspset, kdsplay,NULL    },
-  { "display.a",S(DSPLAY),0,      "",     "aioo", dspset ,dsplay    },
+  { "csoundDisplay.k",S(DSPLAY),0,      "",     "kioo", dspset, kdsplay,NULL    },
```
kunstmusik: @vlazzarini looks like another mistaken find/replace issue as this changes the opcode name

Engine/insert.c
```diff
@@ -1887,7 +1887,7 @@ static INSDS *instantiate(CSOUND *csound, int32_t insno, int32_t link)
     setup_opcode_argpp(csound, opds, ttp, ep, ip, tp, lclbas, lcloffbas, opMemStart);
 
   }
-  /* display instantiated instrument */
+  /* csoundDisplay instantiated instrument */
```
kunstmusik: This comment change seems awkward, but also the comment itself isn't very clear. (I'm not sure how it applies to the code below...)

include/csound.h
```diff
@@ -176,7 +176,7 @@ extern "C" {
     /* output soundfile format */
     int32_t outformat;
     /* sample size */
-    int32_t     sfsampsize;
+    int32_t     csoundSndfileSampleSize;
```
kunstmusik: @vlazzarini this rename one looks like a find/replace error as it's on a field.

InOut/rtpw.c
```diff
@@ -421,7 +421,7 @@ struct sink_data {
   struct spa_hook registry_listener;
   CSOUND *csound;
   int no_devs;
-  int display;
+  int csoundDisplay;
```
kunstmusik: @vlazzarini another field rename that looks suspcious

InOut/rtpw.c
```diff
@@ -435,7 +435,7 @@ static void registry_event_global(void *data, uint32_t id, uint32_t permissions,
   const char *node_description;
   CSOUND *csound = sink_data->csound;
   struct sink_info *sinks = sink_data->sinks;
-  int32_t display = sink_data->display;
+  int32_t csoundDisplay = sink_data->csoundDisplay;
```
kunstmusik: @vlazzarini another field rename that looks suspcious

InOut/rtpw.c
```diff
@@ -499,12 +499,12 @@ static const struct pw_core_events core_events = {
...
 };
 
-int32_t query_pipewire_sinks(CSOUND *csound, CS_AUDIODEVICE *list, int32_t display)
+int32_t query_pipewire_sinks(CSOUND *csound, CS_AUDIODEVICE *list, int32_t csoundDisplay)
 {
   struct sink_data data = {0};
   struct spa_hook core_listener;
   data.csound = csound;
-  data.display = display;
+  data.csoundDisplay = csoundDisplay;
```
kunstmusik: @vlazzarini another field rename that looks suspcious

InOut/winascii.c
```diff
@@ -41,7 +41,7 @@ void KillAscii(CSOUND *csound, WINDAT *wdptr)
     wdptr->windid = 0;          /* just to make out that it's dead */
 }
 
-/* display an n-pnt float array using simple ascii chars */
+/* csoundDisplay an n-pnt float array using simple ascii chars */
```
kunstmusik: @vlazzarini another rename that looks suspcious

InOut/window.c
```diff
@@ -31,14 +31,14 @@ extern void MakeAscii(CSOUND *, WINDAT *, const char *);
 extern void DrawAscii(CSOUND *, WINDAT *);
 extern void KillAscii(CSOUND *, WINDAT *);
 
-/* somewhere to invoke for no display */
+/* somewhere to invoke for no csoundDisplay */
```
kunstmusik: @vlazzarini seeing this as a general issue, wonder if it's worth grepping for csoundDisplay and adjusting case-by-case as it seems "display" has been replaced in a number of places that looks unrelated to PR goals

kunstmusik:
Saw that some leftover Yield callback code was cleaned up. I asked AI to check if there was anything remaining for yield and it mentioned:

Remaining cleanup candidates
If you want to finish the cleanup, the remaining spots are:

1. Remove or deprecate the API declarations in include/csound_threads.h and H/prototyp.h (if backward compatibility isn’t required).
2. Remove the stale comment in csoundCore.h. (  /* VL: internal yield callback */ )

vlazzarini:
> @vlazzarini: I see that Top/init_static_modules.c was removed from the WASM NIX build. I see it is still included in Android.mk. Seems to me csmodule.c is now the source-of-truth for static module initialization, and that we can:
> 
> 1. remove init_static_modules.c from the repo
> 2. Add INIT_STATIC_MODULES=1 to the Android (and iOS?) builds
> 3. Update Top/README.md
> 
> Thoughts?

Sounds good, and I've done 1 and 3 now locally. Not sure about 2, do we need it? I grepped the sources and could not find a mention of that token.

Engine/musmon.c
vlazzarini: not quite a mistake but having function names on user-oriented error message is not helpful either. Changed it to be more user-friendly.

InOut/window.c
vlazzarini: yeh, the display() name change was a total mess. I did a full search and got things fixed now.

Engine/rdscor.c
vlazzarini: we could change both function names at this pre-release stage if you feel we should.

kunstmusik:
> > 1. remove init_static_modules.c from the repo
> > 2. Add INIT_STATIC_MODULES=1 to the Android (and iOS?) builds
> > 3. Update Top/README.md
> > 
> > Thoughts?
> 
> Sounds good, and I've done 1 and 3 now locally. Not sure about 2, do we need it? I grepped the sources and could not find a mention of that token.

Looks like INIT_STATIC_MODULES was used previously in CS6. I removed it from the wasm build.

Engine/rdscor.c
kunstmusik: How about GetStringFromArg and csoundGetStringFromArg?  Seems a bit better named.

vlazzarini: that sounds ok to me.

vlazzarini: done now.

util/pvanal.c
```diff
@@ -490,8 +490,6 @@ static int32_t pvxanal(CSOUND *csound, SOUNDIN *p, SNDFILE *fd, const char *fnam
         for (k = 0; k < chans; k++) {
           frame = frame_c[k];
           chanbuf = inbuf_c[k];
```
kunstmusik, line 492: I see removal of csound->CheckEvents and longjmp here and in other utility files. That changes behavior; what's the rationale for this change?

tests/c/csound_option_parser_test.cpp
```diff
@@ -25,7 +25,7 @@ class OptionParserTests : public ::testing::Test {
     {
         csoundSetGlobalEnv ("OPCODE6DIR64", "../../");
         csound = csoundCreate (NULL, NULL);
-        csoundCreateMessageBuffer (csound, 0);
+        //csoundCreateMessageBuffer (csound, 0);
```
kunstmusik: I see this commented out, wondering why so, and whether this is a mistake or line should be completely removed.

kunstmusik:
@vlazzarini: I tried running unit test (i.e., make test) locally and it hung on the first test for me:

```
Test project /Users/stevenyi/work/csound/cs7-prs/build
      Start  1: ChannelTests.ControlChannelParams
```

just hangs there. Does that do the same for you?

kunstmusik:
Note: The change from ->RandSeed1 to ->RandSeed31 is a bit curious. I haven't ever touched that area of code and looking at it it's a bit odd.

Opcodes/arrays.c
```diff
@@ -105,10 +105,10 @@ static int32_t init_fft_complex(CSOUND *csound, FFT *p) {
   tabinit(csound, p->out, N, p->h.insdshead);
   size_t bytes = (size_t)N * 2u * sizeof(MYFLT);
   csound->AuxAlloc(csound, bytes, &p->mem);
-  if(p->out->arrayType == csound->GetType(csound, "Complex")
-     && p->in->arrayType == csound->GetType(csound, "Complex"))
+  if(p->out->arrayType == csound->csoundGetType(csound, "Complex")
```
kunstmusik: @vlazzarini: This one looks like a find/replace issue. Expecting csound->GetType.

Engine/rdscor.c
```diff
@@ -64,7 +64,7 @@ char* get_arg_string(CSOUND *csound, MYFLT p)
     return ss;
 }
 
-char* get_arg_string_from_evt(CSOUND *csound, MYFLT p, EVTBLK *evt)
+char* get_string_from_evt(CSOUND *csound, MYFLT p, EVTBLK *evt)
```
kunstmusik: @vlazzarini I found the previous name get_arg_string_from_evt a little more self-documenting.

util/pvanal.c
vlazzarini: CheckEvents was  csoundYield originally. This was removed because it was not needed and never used in 20 years. There is no real change in behaviour.

tests/c/csound_option_parser_test.cpp
vlazzarini: I think it was a leftover from playing around with the test, it can go.

vlazzarini:
> @vlazzarini: I tried running unit test (i.e., make test) locally and it hung on the first test for me:
> 
> ```
> Test project /Users/stevenyi/work/csound/cs7-prs/build
>       Start  1: ChannelTests.ControlChannelParams
> ```
> 
> just hangs there. Does that do the same for you?

not here or in CI.

vlazzarini:
> RandSeed31

I guess I thought it would match Rand31() as it seems to be used solely with that function

tests/c/csound_option_parser_test.cpp
vlazzarini: Reinstated the line.

Opcodes/arrays.c
vlazzarini: yes, that was wrong. Fixed now.

Engine/rdscor.c
vlazzarini: changed back.

kunstmusik:
I think were good to go now. I'm investigating why the unit tests are hanging here but if I can't figure it out quickly I think since they're running on CI and on @vlazzarini computer we'll be okay to merge.

kunstmusik:
I figured out there was some issue with OPCODE7DIR64 here; setting it to build directory got the tests to run just fine. Will assume my environment got into a weird place from testing. Going to approve and merge now.

kunstmusik:
Sorry, just remembered I wanted to do a local build of WebAudio Csound and run some tests against that build with some webapps to ensure compatibility. Will report back when that is complete.

vlazzarini:
> I figured out there was some issue with OPCODE7DIR64 here; setting it to build directory got the tests to run just fine. Will assume my environment got into a weird place from testing. Going to approve and merge now.

I had a look at this as it was strange. So I think your issue is arising from this: there is a token `-D_CSOUND_RELEASE_` that needs to be set for the opcode dir default to the installed place. If it is not set, it defaults to the current directory. I think this was added by default to the Debug build earlier but I removed it because it did not seem to agree with a Debug build. The token is only used for that purpose.

The token name is not helpful. So I will do this: add an option to set default dir and make it ON by default, changing the token name for a more helpful one. I'll make a separate PR for this.

kunstmusik:
I did local testing of WebAudio csound with book examples and my live coding site and all worked. I went ahead and published a new version of @csound/wasm-bin and @csound/browser. WIll approve and merge now.

kunstmusik (approved):

PR #2423, opened 2026-01-11
feat: add min and max values as builtin macros, with tests

hlolli:
Got this idea from the mailinglist from Philipp, similar concept exists in js

```node
> Number.MAX_VALUE
1.7976931348623157e+308
> Number.MIN_VALUE
5e-324
```

### DIFF
--- a/Engine/csound_pre.lex
+++ b/Engine/csound_pre.lex
@@ -26,9 +26,19 @@
 #include <string.h>
 #include <stdlib.h>
 #include <ctype.h>
+#include <float.h>
 #include "csoundCore.h"
 #include "corfile.h"
 #include <inttypes.h>
+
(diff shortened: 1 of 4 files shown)


hlolli:
@coderabbitai review

vlazzarini (approved):

hlolli:
@vlazzarini I could turn off the python code formatter. It creates noisy diffs in .py files. Not sure if we should enforce it or I just turn it off my side. Or run it once every blue moon. Not a big deal from my side, sometimes make things more readable, most often just no difference.

vlazzarini:
I think it's not a big deal. I first misinterpreted as adding some newlines before args but it was just the way it appeared on a small screen.

PR #2427, opened 2026-01-20
fix: UDO array argument handling and add tests for implied types

kunstmusik:
* also adds comments about internal/external format to relevant locations in code

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -1570,39 +1570,6 @@ char* convert_internal_to_external(CSOUND* csound, char* arg) {
   // now remove any : or ; leftover in typename
   type = remove_type_quoting(csound, arg);
 
-  /** This is doubtful code leading to garbled types
-      - restoring previous (dbc1ce2b1)
-
-    NB: changing the check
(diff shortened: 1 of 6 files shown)


Engine/csound_orc_semantics.c
```diff
@@ -1557,9 +1573,9 @@ char* convert_internal_to_external(CSOUND* csound, char* arg) {
   /** This is doubtful code leading to garbled types
```
vlazzarini: Should we remove this comment? I always meant for it to be temporary. If the reverted code is good, then we should perhaps?

kunstmusik: Seems good to remove considering it's dead code now

kunstmusik: I went ahead and removed the commented out code. Let me know if everything else is good and I will do another version bump for wasm and browser release and commit to this PR.

vlazzarini (approved):

kunstmusik:
@vlazzarini I had wanted to do a release of the wasm and browser packages as part of this PR (see my other comment). I will setup a small separate PR to do version bump and release.

PR #2428, opened 2026-01-20
New options

vlazzarini:
This PR adds a couple of new options 

```
--code=string
```
supplies Csound code as a string

```
--events=string
```

supplies events as a string

For example, this plays a 440 tone

```
csound --code="out(oscili(p4,p5))" --events="i0 0 1 1000 440" -odac
```

These options may be used to turn Csound into a Swiss-knife for various audio processing tasks. For example, taking a chunk from an audio file and writing another

```
csound --code=$'out(linen(diskin:a("pianoc2.wav",1,3),.01,p3,0.01))' --events="i0 0 2" -o edited.aif
```


rorywalsh (approved):
This looks neat.

PR #2432, opened 2026-01-21
Fix for score offset

### LINKED ISSUE #2431: segfault in setscorepos()
I am getting a segfault in setscorepos(). It looks like line ~165 of Top/csound_events.c is the culprit: it tries to assign to evt.p[1] but evt.p is NULL.

```
<CsoundSynthesizer>
<CsInstruments>
instr 1
  setscorepos(1)
endin
</CsInstruments>
<CsScore>
i 1 0 2
</CsScore>
</CsoundSynthesizer>
```

I am not familiar with the code but maybe adding this +line would fix it?
```
  evt.pcnt = 3;
+ evt.p = pfields;
  evt.p[1] = evt.p[0] = FL(0.0);
```

vlazzarini:
This PR fixes an issue in the score offset code, with a csd and unit tests added.
Closes #2431

### DIFF
--- a/tests/commandline/test_setscorepos.csd
+++ b/tests/commandline/test_setscorepos.csd
@@ -1,0 +1,23 @@
+<CsoundSynthesizer>
+<CsOptions>
+-n
+</CsOptions>
+<CsInstruments>
+gi1 init 0
+instr 1
+  prints "Top...\n"
+  setscorepos(gi1)
+endin
+instr 2
+  gi1 = 3
+  prints "Rewinding...\n"
+  rewindscore()
+endin
+</CsInstruments>
+<CsScore>
+i 1 0 1
+i 2 1 0 
+</CsScore>
+</CsoundSynthesizer>
+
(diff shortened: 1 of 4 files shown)


tests/commandline/test_setscorepos.csd
```diff
@@ -0,0 +1,23 @@
...
+gi1 init 0
+instr 1
+  prints "Top...\n"
+  setscorepos(gi1)
+endin
+instr 2
+  gi1 = 3
+  prints "Rewinding...\n"
+  rewindscore()
+endin
```
hlolli, line 16: would it be possible to add assertion (using simple if-else) about the expected scorepos when i2 is called?

vlazzarini: Good idea. I've added code that fails if setscorepos() and/or rewindscore() doesn't work.

hlolli (approved):
nice!

PR #2436, opened 2026-01-24
oscbnk entry in OENTRY corrected for new plugin interface

fggp:
Needed for all oscbnk syntax to work

### DIFF
--- a/Opcodes/oscbnk.c
+++ b/Opcodes/oscbnk.c
@@ -3001,7 +3001,7 @@ static int32_t vco2(CSOUND *csound, VCO2 *p)
   static const OENTRY localops[] =
     {
   { "oscbnk", sizeof(OSCBNK), TR, "a", "kkkkiikkkkikkkkkkikooooooo",
-    (SUBR) oscbnkset, (SUBR) oscbnk },
+    (SUBR) oscbnkset, (SUBR) oscbnk, NULL, 2 },
    { "oscilbank",     sizeof(OSCBNK),     TR,  "a",  "kkkkiikkkkikkkkkkikooooooo",
      (SUBR) oscbnkset, (SUBR) oscbnk                }, /* alias */
    { "grain2",     sizeof(GRAIN2),     TR,      "a",    "kkkikiooo",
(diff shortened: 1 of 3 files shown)


Opcodes/oscbnk.c
```diff
@@ -3001,7 +3001,7 @@ static int32_t vco2(CSOUND *csound, VCO2 *p)
   static const OENTRY localops[] =
     {
   { "oscbnk", sizeof(OSCBNK), TR, "a", "kkkkiikkkkikkkkkkikooooooo",
-    (SUBR) oscbnkset, NULL, (SUBR) oscbnk, NULL, 2 },
+    (SUBR) oscbnkset, (SUBR) oscbnk },
```
vlazzarini: needs a 

, NULL, 2}

at the end of the entry - the new name is "oscilbank" so we need to mark it as the old version.

vlazzarini (changes requested):

Opcodes/oscbnk.c
fggp: Done

Opcodes/oscbnk.c
```diff
@@ -3001,7 +3001,7 @@ static int32_t vco2(CSOUND *csound, VCO2 *p)
   static const OENTRY localops[] =
     {
   { "oscbnk", sizeof(OSCBNK), TR, "a", "kkkkiikkkkikkkkkkikooooooo",
-    (SUBR) oscbnkset, NULL, (SUBR) oscbnk, NULL, 2 },
+    (SUBR) oscbnkset, (SUBR) oscbnk, NULL, 2 },
```
vlazzarini: It needs 2 NULLS (for deinit and for opcodeinfo) before the deprecated field.

vlazzarini (changes requested):

Opcodes/oscbnk.c
fggp: Seems good now

fggp (commented):
Typo in opcode name

fggp:
osclisten 3rd form needed entries

vlazzarini (approved):
Thanks.

PR #2437, opened 2026-01-26
Various Fixes for live coding

kunstmusik:
I went through multiple years of live code practice sessions with my csound-live-code system to test Csound 7. I uncovered a number of issues that required various fixes which are all here together in this PR. I have already released the latest here as beta builds for csound/wasm and csound/browser and have been testing in browser with some live code examples and things seem stable now. 

- Fix UDO argument type matching by ensuring [inargs_check](cci:1://file:///Users/stevenyi/work/csound/cs7-fix-new-style-udo/Engine/csound_orc_compile.c:1607:0-1623:1) validates full string length matches, avoiding mis-detection with numeric constants.  
- Prevent WASM long‑run crashes by freeing hash table bucket arrays on resize/teardown (fixes memory leak).  
- Remove unused hash table from pass‑by‑ref UDO rewiring to simplify UDO init/perf setup.  
- Fix parser/compile edge cases: spacing after `(` in control flow, bitwise operator precedence/token typo, interleaved instrument/UDO compilation with optional args, and null line access in error reporting.  
- Add/adjust UDO performance execution stub and initialization checks for stability.
- Fixed operator precedence for bitwise operators and corrected token typo


Engine/udo.c
```diff
@@ -608,12 +608,36 @@ int32_t useropcdset(CSOUND *csound, UOPCODE *p)
...
+ */
 int32_t useropcd(CSOUND *csound, UOPCODE *p)
-{  if (UNLIKELY(p->h.nxtp))
-    return csoundPerfError(csound, &(p->h), Str("%s: not initialised"),
-                           p->h.optext->t.opcod);
-  else
+{
+  /* If init wasn't run yet (ip is NULL or init_done not set), just return OK.
+     This allows UDOs skipped at i-time to safely exist in the perf chain. */
+  if (p->ip == NULL || !p->ip->init_done)
```
vlazzarini: Should this not issue a warning?

Engine/csound_orc.y
```diff
@@ -115,14 +115,14 @@
 /* Precedence Rules */
 %right '?'
 %left S_AND S_OR
+%left S_LT S_GT S_LE S_GE S_EQ S_NEQ '=' // VL 6.8.25 for backwards compat
```
vlazzarini: Is this now the same precedence as in 6.x?
(That is no change of behaviour?)

Engine/udo.c
kunstmusik: It's a complicated one, apparently UDOs have always linked into the perf chain even if they don't have any perf-time opcodes. I was getting PERF errors with an init-only UDO that didn't happen in CS6 with one of my live code examples where the init-time UDO was called within a ternary conditional. Unsure why this arose now and dug in for a while and ultimately landed on just taking this solution and moving on. Perhaps the issues is elsewhere in the conditional not skipping at both init and perf time?

Engine/csound_orc.y
kunstmusik: Good question, looks like the main fix was the S_BITSHIFT_RRIGHT typo. Working on restoring the precedence to CS6 style and updating the csd test. Will commit shortly.

kunstmusik:
@vlazzarini : operator precedence is now reverted back to match csound 6. Out of curiosity, I asked for an analysis from AI to compare vs C:

Here’s the **current Csound precedence** (low → high, based on [csound_orc.y]
 
`?` → `S_AND/S_OR` → `|` → `&` → comparisons (`< <= > >= == != =`) → shifts (`<< >>`) → `+ -` → `* / %` → `^` (power) → `#` (bitwise XOR) → unary (`S_UNOT`, `S_UMINUS`, `S_UPLUS`).  
@Engine/csound_orc.y#115-128

**C precedence (low → high)**:  
`?:` → `||` → `&&` → `|` → `^` → `&` → `== !=` → `< <= > >=` → `<< >>` → `+ -` → `* / %` → unary.

### Key differences vs C
1) **Shifts vs comparisons**
- **Csound:** shifts bind tighter than comparisons.  
  `1 << 1 < 2` → `(1 << 1) < 2`
- **C:** comparisons bind tighter than shifts.  
  `1 << 1 < 2` → `1 << (1 < 2)`

2) **Bitwise XOR operator**
- **Csound:** `#` (XOR) is very high precedence (above `^` power and arithmetic).  
  `1 & 2 # 3` → `1 & (2 # 3)`
- **C:** XOR is `^` and sits **between** `&` and `|`.  
  `1 & 2 ^ 3` → `(1 & 2) ^ 3`

3) **`^` is different**
- **Csound `^` = power**, not XOR; there’s no direct C analog.

Everything else (logical `&&/||` vs bitwise) aligns broadly with C’s ordering.

Engine/csound_orc.y
kunstmusik: Restored the precedence and updated the test; matches Csound 6 again.  Thanks for spotting that.

Engine/udo.c
vlazzarini: were you able to pin down an exact case of where this happened so we could compare to Csound 6?

Edit: I saw your test for this, and I think your hunch was right, it's the conditional that's wrong. See this line

```
tick:i = (itick <= 0) ? now_tick() : itick
```

It's running at both i and perf-time, but it should only run at i-time. The UDO has only i-time code, this single line creates a perf-time routine that runs and then fails because the UDO is not initialised. Replace it with an if-then

```
  if itick <= 0 then
   tick:i = now_tick();
  else
   tick:i = itick
  endif
```
and the problem goes away, the UDO is now i-time only.

kunstmusik: it's odd as I would expect the ternary to compile down in the same way as the if-then; I had also tried with if-then earlier and had the same issue but maybe something else had changed. I'll investigate a little bit further here. If I can't find a solution and if I can switch to if-then for now, I think we can open an issue for ternary operator separately.

kunstmusik: @vlazzarini took a long while to try out different things, but I went with rolling back and then modifying to set ->perf to NULL for init-only UDOs. This gets the right behavior for the ternary operators and seems to be working without regression. Let me know what you think; if it looks okay I'll add one more commit to this PR for another webaudio csound release.

vlazzarini: It looks okay, but otoh it also seems a bit heavy-handed, having to go through all opcodes, check for udos, check for perf.

~~I wonder why we need this now, since before all we needed was to check if there were any functions in the perf chain at run time. Maybe it's because we got rid of "thread" in OENTRY? Although that was always set to 7 in entry1.c in Csound 6. I also wonder why there is a difference between conditional expressions and if/then/else.~~ Never mind, a lot has changed, we need to keep improving the code.

~~I think we can go ahead and this can be revisited in the future with a lighter approach.~~

EDIT: your solution is better than the current state in 6.x because we now don't need to run stubs in perf-time unless there's an error. I think we should definitely set the perf function to NULL in these cases.

However, I think it can be improved by only checking UDOs as they are compiled. So here's my proposal:

1) Add a field to `OPCODEINFO` to hold the correct OENTRY. This is needed because the OENTRY in the `ip` field is stale

```
  typedef struct opcodinfo {
    int32    instno;
    char    *name, *intypes, *outtypes;
    int16   inchns, outchns;
    bool newStyle;
    bool passByRef;
    CS_VAR_POOL* out_arg_pool;
    CS_VAR_POOL* in_arg_pool;
    struct instr *ip;
    OENTRY *oentry;
    struct opcodinfo *prv;
  } OPCODINFO;
```

2) Set this field when the UDO is defined in symbtab.c: add_udo_definition()

```
      newopc = csound_find_internal_oentry(csound, &tmpEntry);
      inm->oentry = newopc;
```


3) Move the checking code to `case UDO_TOKEN:`

```
    if(!udo_has_perf_opcodes(instrtxt)) {
       OPTXT *optxt = (OPTXT *) instrtxt;
       TEXT *ttp = &optxt->t;
       opinfo->oentry->perf = NULL;
    } 
      
    /* Handle Inserting into CSOUND here by checking id's (name or
     * numbered) and using new insert_instrtxt?
     */
    break;
```

Also possibly update the opcodeinfo->ip->t.oentry to ip->oentry here (because that's stale).

I tested this here and it works.

kunstmusik: That sounds great, I was looking for a better solution last night but it was a long day and I hit my limit. If you could, please push the change to this branch. I'll test it out and release a webaudio build and then we can merge this.

vlazzarini: done.

kunstmusik: Added a commit after reviewing with AI of a possible scenario when redefining UDOs that could trigger a NULL deref. Found an intermittent problem that I'll open up a separate issue in the tracker, then will release a new webaudio build and push that commit. That should let us close out this PR.

kunstmusik:
@vlazzarini I think everything is set now, I've gone ahead and published another webaudio build with the latest here. If you could approve and merge, would appreciate it!

vlazzarini:
Shall we make a beta release? Last one was November

kunstmusik:
Beta release seems like a good idea, there's a number of changes that would benefit from wider testing

vlazzarini:
I've just commented out one line - to see if fixes your intermittent issue. Was it there before I made the changes today?

kunstmusik:
I hadn't tried recompilation of the same UDO like this with CS7 until today so I'm not sure when it would have started. I'll try out the commit shortly.

vlazzarini:
I'm pretty sure it is to do with setting the instrument number in opcodeinfo, as that should not be zero. It could also be caused by a race condition as this is done on merge.

kunstmusik:
that did not fix it here and I was still able to reproduce the INIT error. I'll do a short session with AI to see if there's anything it finds for possible causes.

vlazzarini:
> that did not fix it here and I was still able to reproduce the INIT error. I'll do a short session with AI to see if there's anything it finds for possible causes.

OK point out that it looks like the opcodeinfo does not have instno set - it's zero. It should be > 200.

vlazzarini:
I'll go ahead and merge. We can carry the discussion over in the issue.

vlazzarini (dismissed):

kunstmusik:
Ah I have a fix one moment

kunstmusik:
@vlazzarini: AI summary of the change:

Summary: when a UDO is redefined, we now carry over the previous instno into the new OPCODINFO before swapping oentry->useropinfo. During the recompile window, any runtime UDO calls continue to resolve the same (valid) instrument number instead of 0, so the init path won’t hit “Broken redefinition of UDO 0.” After merge, the UDO still uses the updated definition, but without the transient zero-ID gap.

kunstmusik:
Ugh, just tried a smoke test and found a problem:

```
opcode test():i
    ival = 3
    xout ival 
endop

print test()
```

When livecoding, this was giving me 0. Then when I restarted and tried again, it gave me 3, but then changing the ival and re-evaluating did not update and still gave 3. There's a problem here, going to have to diagnose this.

vlazzarini:
~3 is also very low. It should start at 200.~ sorry I didn't read your post properly. I think the problem may be sharing of csound->opcodeInfo, may be an idea to put this in engineState and do the merge (which is threadsafe afaik).

kunstmusik:
Trying out some things here around ENGINE_STATE and merging

kunstmusik:
Took a while but I think I have an update that works with ENGINE_STATE merging and fixes the issue. I have an additional csd test added that uses compilestr to test versions of UDOs are updating correctly. I'm still checking a few things but should have a commit ready tomorrow.

vlazzarini:
Great. Again, not sure what changed to bring that race condition on, but the way that linked list was handled was surely a vulnerability.

kunstmusik:
@vlazzarini I pushed commits for fixing up redefinitions of UDOs during live coding. Did not end up going down path of ENGINE_STATE and merging as there were some complications on when UDOs were being parsed and resolved and when ENGINE_STATE was being created. I think for now this should tide us over and we can revisit to do merging at another time.

vlazzarini:
> @vlazzarini I pushed commits for fixing up redefinitions of UDOs during live coding. Did not end up going down path of ENGINE_STATE and merging as there were some complications on when UDOs were being parsed and resolved and when ENGINE_STATE was being created. I think for now this should tide us over and we can revisit to do merging at another time.

Ok, everything checks out, just trying to understand the locks you put in. If the compilation mode is asynchronous, are they needed?  I thought the code in threadsafe.c would ensure things separate compilations are done in sequential order and in between performKsmps calls.

Actually, just to check are you using csoundCompileOrc with async set to 1?

vlazzarini (approved):

kunstmusik:
The locks were recommended from AI and looked harmless to me to protect against concurrent modifications. I was testing with UDP so I assume that's async 1. I'll go ahead and merge now. I want to spend a little time doing some performance measurements and can return to this after that.

PR #2440, opened 2026-01-27
fixing access of global UDT vars

### LINKED ISSUE #2439: Accessing fields of UDT global var gives error
As described in the docs, this works:

````
struct MyType val0:i, val1:i

instr 1
  testVal:MyType init 8, 88
  print(testVal.val0)
  print(testVal.val1)
endin

````

But if I declare "testVal" as a global, it fails:
````
struct MyType val0:i, val1:i
testVal@global:MyType init 8, 88

instr 1
  print(testVal.val0)
  print(testVal.val1)
endin

````
The error is:
````
Undeclared local variable 'testVal.val0' used before declaration

````

As a workaround, I can point a local var at the global and it works:

````
struct MyType val0:i, val1:i
testVal@global:MyType init 8, 88

instr 1
  tmpVal:MyType=testVal
  print(tmpVal.val0)
  print(tmpVal.val1)
endin
````

### PR DESCRIPTION
**Pull request by vlazzarini**
Accessing global UDT vars directly was not implemented. This PR fixes the issue,
closes #2439

### CODE REVIEW

**Comment by vlazzarini** [comment]
Also fixed a couple of issues related to compilation failures leading to segfaults; improved error messaging (silenced duplicate syntax error msgs).

**Comment by vlazzarini** [comment]
The code for UDT references was misplaced in variable initialisation and affecting the creation of independent struct vars. This was moved to instr instantiation to fix the issue.

**Comment by vlazzarini** [comment]
On second thoughts, I removed the code to handle structure references. 
My reasoning:

- structure initialisation semantics through reference diverged from all other types, where initialisation is all by copy.
- the original code implementing this was inappropriate and caused a bug (see #2439).
- we do not have the means in the parser/compiler to handle references more widely (beyond UDO args) in a proper manner.

Therefore I am removing my earlier fix, which did not appropriately handle this as it replaced all struct assignment code with references. This did not allow copying of structures at all, which was wrong. I am removing the test for referencing in structures, and we can revisit this if/when reference types are introduced.

**Comment by kunstmusik** [comment]
Took an initial look and am a little concerned about two things:

1. allocation of sizeof TREE vs. ORCTOKEN: I'd like to understand a little better where it goes off and wonder if this is fixing an issue but hiding the root cause

2. The struct changes: if I understand it correctly, it seems right to me that we copy on assignment as this is the case for other data types (i.e., `asig2 = asig1` would be two different values). I need to dive a little deeper to make sure I understand the change.


FWIW: Requested a review from Codex and got the following (will also be reviewing here shortly to see how much of this is valid):

**Scope of impact**
- Parser/semantic compile path changes for arg resolution (local/global/struct paths) and error reporting: csound_orc_compile.c, insert.c, symbtab.c, new_orc_parser.c, csound_orc_semantics.c.
- AST/token memory handling touched (allocation sizing + tree deletion): csound_orc.lex, csound_orc_semantics.c, csound_orc_optimize.c.
- UDT struct initialization and global struct access paths: csound_orc_semantics.c, insert.c.

**Possible regressions**
- In `create_arg`, the final fallback now calls `setup_arg_for_var_name()` but errors only if it returns non-null. That appears inverted and could allow missing variables to slip through with a null `argPtr`, causing later NULL derefs or misclassification of locals. See csound_orc_compile.c.  
- `new_token()` now allocates `sizeof(TREE)` instead of `sizeof(ORCTOKEN)`. This avoids list traversal issues, but it relies on token nodes being safely treated as `TREE` for list chaining. If any code assumes token memory layout is compact, this is a semantic change (memory footprint + structure assumptions). See csound_orc.lex.
- Error output formatting changed (syntax errors now prefixed differently and backtrace ordering adjusted). This could affect tooling that parses error lines. See symbtab.c, csound_orc_semantics.c, new_orc_parser.c.

**Sufficient test coverage**
- Added CLI tests for compilestr failure cleanliness and global struct var access: test_fail_compilestr.csd, test_global_struct_var.csd, wired into the suite: test.py.
- Existing CI runs and JUnit pass, but coverage does **not** explicitly exercise global UDT vars passed as opcode args or struct-path dereference through `ARG_GLOBAL` paths in `setup_opcode_argpp`. That behavior was adjusted in insert.c.

**Additional concerns**
- The removed struct-reference test reduces regression coverage around assignment-by-reference semantics. Current tests only check direct reads, not aliasing or copy behavior. See remaining struct tests in test_structs.csd.
- The `delete_tree` refactor is safer but now zeroes pointers; if any code retains references into a tree after deletion, those will now be null and may surface latent lifecycle bugs. See csound_orc_semantics.c.

**Comment by vlazzarini** [comment]
Test for issue with ORCTOKEN allocation

```
i1=0
instr
  nonsense()
endin
```

run this on develop built with ASA. 
Note that this issue is not particularly linked to the fixes in the PR but I caught it while working on them.

The error messaging is not a regression, it's an improvement, we had duplication that was getting in the way, and messaging is also more uniformly formatted now.

I don't think the changes in check_args() should cause any regression. They were needed because global UDT variable args weren't picked up.

The struct reference thing imho was just wrong from the start and should not be implemented. As I said we do not have the infrastructure to handle it correctly, even if we were ok with the break in semantics.

**Comment by kunstmusik** [comment]
I tried that code with both this PR branch and with latest from develop and both segfault pretty hard where I have to use `killall -9 csound` to exit csound. Investigating now.

**Comment by vlazzarini** [comment]
No segfault from this branch here, I'm afraid. See 

```
victor@MacBookPro debug % csound test.csd
Reading CSD file:  test.csd
--Csound version 7.0 beta (double samples) Jan 29 2026
[commit: d1ec0fb68a528cc6654ebba20823d6fe95bda2fb]
Compiling Csound code...

syntax error, unexpected T_IDENTB (token "nonsense"),  line 7
>>>nonsense <<<
from file test.csd (1)
No number following instr

syntax error, unexpected ENDIN_TOKEN, expecting end of file (token "endin"),  line 9
>>>endin <<<
from file test.csd (1)
Parsing failed due to 1 syntax error
Stopping on parser failure
	 ... failed to compile code.
		   overall amps:      0.0
1 errors in performance
--Csound version 7.0 beta (double samples) Jan 29 2026
[commit: d1ec0fb68a528cc6654ebba20823d6fe95bda2fb]
no sound written to disk
```
are you sure you tested this and not develop? This is before the revert to ORCTOKEN.

EDIT: tested with your new changes and it's good. Probably the best solution.

**Comment by kunstmusik** [comment]
Yes, was using the PR branch. I've reverted the change to allocate TREE and put back as ORCTOKEN, then went through and found a class of issues where on "error" rules, the $$ value was still getting dereferenced but it might have been garbage data. Setting $$ to NULL seems to have fixed all of those. 

Could you try the latest changes @vlazzarini ?

**Comment by vlazzarini** [comment]
> Yes, was using the PR branch. I've reverted the change to allocate TREE and put back as ORCTOKEN, then went through and found a class of issues where on "error" rules, the $$ value was still getting dereferenced but it might have been garbage data. Setting $$ to NULL seems to have fixed all of those.
> 
> Could you try the latest changes @vlazzarini ?

works here, can you think of a more comprehensive test?

**Comment by vlazzarini** [comment]
One question I have is whether that data was freed somewhere or will we have leaks on error.

**Comment by kunstmusik** [comment]
Adding a number of tests now. Unsure about free but I haven't seen any crashes as of yet with the new code and testing error paths. Should be done shortly.

**Comment by vlazzarini** [comment]
I think your fix is great. I can look later at memory usage running failed compilations to see if there is an issue.

File: Engine/csound_orc_semantics.c [hunk 1]
```diff
@@ -4081,30 +4065,34 @@ TREE* make_opcall_from_func_start(CSOUND *csound, int32_t line, uint64_t locn,
   return left;
 }
 
-void delete_tree(CSOUND *csound, TREE *l)
+static void delete_tree(CSOUND *csound, TREE **l)
```
kunstmusik, line 4068: Finding this change a little strange, at least the code gets a little more awkward with (*l) everywhere. Is the change here  primarily to NULL out the passed in TREE **l?

vlazzarini: I had lots of access-after-free here. So I needed to null the pointers across the full tree. The only way is to dereference the pointers.

vlazzarini: but this could have been a side effect of the garbage issues. For example, that first compilation fail gave me this on develop

```
==40006==ERROR: AddressSanitizer: heap-use-after-free on address 0x607000007128 at pc 0x0001020fcd28 bp 0x00016f3c48b0 sp 0x00016f3c48a8
READ of size 8 at 0x607000007128 thread T0
    #0 0x0001020fcd24 in delete_tree csound_orc_semantics.c:4092
    #1 0x0001020fd054 in csoundDeleteTree csound_orc_semantics.c:4107
    #2 0x00010213abb0 in csoundParseOrc new_orc_parser.c:339
    #3 0x000102130838 in csound_compile_orc csound_orc_compile.c:2232
    #4 0x000101ca7680 in csoundCompile main.c:326
    #5 0x000100a3a3e0 in main csound_main.c:21
```

vlazzarini: Reverted the code and got the issue straight away again with

```
i1 = 0
instr 
123 = 1
endin
```

```
==41887==ERROR: AddressSanitizer: heap-use-after-free on address 0x60b0000e5388 at pc 0x0001017a0590 bp 0x00016fd507b0 sp 0x00016fd507a8
READ of size 8 at 0x60b0000e5388 thread T0
    #0 0x0001017a058c in delete_tree csound_orc_semantics.c:4075
    #1 0x0001017a07a8 in delete_tree csound_orc_semantics.c:4081
    #2 0x0001017a0500 in csoundDeleteTree csound_orc_semantics.c:4089
    #3 0x0001017de22c in csoundParseOrc new_orc_parser.c:339
    #4 0x0001017d4038 in csound_compile_orc csound_orc_compile.c:2230
    #5 0x00010134b680 in csoundCompile main.c:326
    #6 0x0001000ae3e0 in main csound_main.c:210
    #7 0x000186915d50  (<unknown module>)
```

your new tests all fail. So we better keep the changes.

kunstmusik: Sounds good to move forward with this. 👍

kunstmusik (dismissed):

kunstmusik:
Went ahead and approved. @vlazzarini if you test memory and it looks okay, then feel free to move forward and merge this.

Ah, saw that the review was just dismissed while I was typing this. Let me know if there's something else to look at.

Engine/csound_orc_semantics.c
vlazzarini: improved the readability - it was ugly alright. Can you approve again?

vlazzarini:
> Went ahead and approved. @vlazzarini if you test memory and it looks okay, then feel free to move forward and merge this.
> 
> Ah, saw that the review was just dismissed while I was typing this. Let me know if there's something else to look at.

I just made that code look less awkward.

kunstmusik (approved):

vlazzarini:
I'll wait for CI confirmation, then merge, thanks.

PR #2444, opened 2026-01-29
Typo in opcode entry

fggp:
The new alias name for opcode genarray_i was missing the 'i' letter: genarray instead of genarrayi


rorywalsh (approved):
Thanks @fggp This was my fault, let's hope that's the last of them.

PR #2445, opened 2026-01-29
midi default instr

vlazzarini:
This PR allows Csound to default midi input to instruments starting from 0.


hlolli (approved):
nice feature

PR #2455, opened 2026-02-03
Fix for newstyle pass-by-copy

vlazzarini:
The following code produced garbage

```
opcode testVco(iFreq:i):a
  setksmps 1
  aSig vco2 1, iFreq
  xout aSig
endop
```

because it ran as pass-by-ref when it should have been pass-by-copy. The test for this was put *before* the init-pass, so setksmps changed the ksmps *after* the test passed.
Relocated the test to after running either setksmps, oversample or undersample so that it can check for local ksmps or sr.

Added test for this.

### DIFF
--- a/Engine/udo.c
+++ b/Engine/udo.c
@@ -494,6 +494,23 @@ int32_t useropcdset(CSOUND *csound, UOPCODE *p)
     memcpy(&(lcurip->p1), &(parent_ip->p1), 3 * sizeof(CS_VAR_MEM));
   }
 
+   // check for setksmps or over/undersample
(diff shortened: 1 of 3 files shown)


kunstmusik:
@vlazzarini I went through and found there was a bug in the check during the init-pass that it needed && to skip over setksmps/oversample/undersample. I modified that and extended the tests to check for oversample/undersample too. Pushed two changes. Could you check those commits? I'll mark as approved and you can merge if those two are good or you can dismiss the review and let me know.

kunstmusik (approved):

PR #2456, opened 2026-02-03
Fix for UDO pops

### LINKED ISSUE #2453: Pops when using connect and udo
I hit a bug that causes pops when using connect and udos. Tested with beta 12

I have a minimal test case. If the udo usage is commented, clicks do not happen. Also if `test` instr uses the `out` opcode directly the bug does not happen either. Moving the `output` instr after the `test` instr also works around the bug.

``` csound
<CsoundSynthesizer>
<CsInstruments>
nchnls = 2
0dbfs = 1

connect "test", "out", "output", "in"

alwayson "output"

instr output
  aSig inleta "in"
  out aSig
endin

opcode testVco, a, i
  iFreq xin
  aSig vco2 1, iFreq
  xout aSig
endop

instr test
  iAmp = p4
  iFreq = p5
  aSig testVco iFreq
  ;aSig vco2 1, iFreq
  kAmpEnv linsegr 0, .1, iAmp, .1, 0
  aSig *= kAmpEnv
  outleta "out", aSig
endin

</CsInstruments>
<CsScore>

i"test" 0 2 [1/8] 220
i. + . . [220*4/3]
i. + . . [220*3/2]
e 7

</CsScore>
</CsoundSynthesizer>
```

vlazzarini:
Cleared pass-by-copy udo outputs to fix pops on instance reuse.
Closes #2453

### DIFF
--- a/Engine/udo.c
+++ b/Engine/udo.c
@@ -751,6 +751,7 @@ int32_t xoutset(CSOUND *csound, XOUT *p)
       current->varType->copyValue(csound, current->varType, out, in, p->h.insdshead);
     }
     else if (csoundGetTypeForArg(out) == &CS_VAR_TYPE_A) {
+      memset(in, 0, sizeof(MYFLT)*CS_KSMPS); // clear output
       if(CS_ESR != parent_sr) {
         if((udo->cvt_out[k++] = src_init(csound, p->h.insdshead->out_cvt,
                                          parent_sr/CS_ESR, CS_KSMPS)) == 0)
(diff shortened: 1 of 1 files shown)


vlazzarini:
I found the root cause of the regression: all outputs were being copied at i-time and that's wrong for a- and k-sig vars. So I added a check instead of clearing variables and now the code behaves as before these changes and exactly like 6.x (we cannot change behaviour).

vlazzarini:
@kunstmusik this is ready for your review

kunstmusik (approved):
Only thing I can think of is adding to the comment that this is CS6 behavior we need to maintain for backwards compatibility, but I think we'll be able to look that up if there's any questions

vlazzarini:
I also think the behaviour is formally correct, k- and a-vars should not be changed at init time.  Amongst other things, it leads to problems as reported in the issue. Out of order evaluation should not lead to pops.

One associated thing I am not sure now, is the behaviour of K-type outargs - these have init time copying but I am not sure the code is preserving that.

I remember dealing with this at some point but then I think my code disappeared in these latest changes.

I'll merge this and then check separately if that is being preserved.

PR #2457, opened 2026-02-07
fix: support for multiple and empty case statements

hlolli:


kunstmusik (approved):
Mentioned to @hlolli there's an issue when there's a switch with no cases and only a default; he said he would address in a separate PR. Approving and merging this one now.

PR #2458, opened 2026-02-07
Fixed double closing of files in mp3len

vlazzarini:
The code for mp3len etc was closing the mp3 files twice (once in the opcode and once in the mp3 uninit function).  This causes no problem on the desktop but crashes wasm.

https://editor.p5js.org/vlazzarini/sketches/L_gIW4Wwz

This PR should fix the problem.

@hlolli @kunstmusik could you make a new release for browser once this is merged?

### DIFF
--- a/Opcodes/mp3in.c
+++ b/Opcodes/mp3in.c
@@ -299,11 +299,9 @@ int32_t mp3len_(CSOUND *csound, MP3LEN *p, int32_t stringname)
   }
   if (UNLIKELY((r = mp3dec_get_info(mpa, &mpainfo, MPADEC_INFO_STREAM)) !=
                MP3DEC_RETCODE_OK)) {
-    fclose(f);
     mp3dec_uninit(mpa);
     return csound->InitError(csound, "%s", mp3dec_error(r));
   }
-  fclose(f);
   if(!strcmp(GetOpcodeName(&p->h), "mp3len"))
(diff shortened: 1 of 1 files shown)


hlolli:
verified no remaining raw fclose() calls in mp3in.c, cleanup is done through FileClose

... unrelated and take this with grain of salt ...
Pre-existing file-handle leak on mp3dec_init_file() error paths: mp3dec_uninit() and mp3dec_reset() only call FileClose when MP3DEC_FLAG_INITIALIZED is set, but several mp3dec_init_file() failures happen before that flag is set.

Close guard: mp3dec.c (line 209) and mp3dec.c (line 225)
Early failure before init complete: mp3dec.c (line 98)

hlolli (approved):

vlazzarini:
> verified no remaining raw fclose() calls in mp3in.c, cleanup is done through FileClose
> 
> ... unrelated and take this with grain of salt ... Pre-existing file-handle leak on mp3dec_init_file() error paths: mp3dec_uninit() and mp3dec_reset() only call FileClose when MP3DEC_FLAG_INITIALIZED is set, but several mp3dec_init_file() failures happen before that flag is set.
> 
> Close guard: mp3dec.c (line 209) and mp3dec.c (line 225) Early failure before init complete: mp3dec.c (line 98)

yep, I think that's right, we can reinstate FileClose for those failures.

PR #2460, opened 2026-02-07
Fixed regression with udo K-type args

vlazzarini:
The K-type used in old-style UDOs experienced regression (not caught by tests), this PR fixes the issue and tightens the tests.


kunstmusik (approved):
Looks good to me. I'm not sure what happens for K[] inputs but perhaps something for another PR.

vlazzarini:
> Looks good to me. I'm not sure what happens for K[] inputs but perhaps something for another PR.

I don't think we had that type in Csound 6.x so we can ignore it or implement it later.

PR #2462, opened 2026-02-08
Fixed segfault in structs with Opcode types

### LINKED ISSUE #2461: Opcode type in UDT?
Is it legal to use the Opcode type inside a UDT? I'm getting a segfault:
```
<CsoundSynthesizer>
<CsInstruments>

struct MyType val0:i, val1:Opcode

instr 1
  testVal:MyType init 16, create(sqrt)
endin

</CsInstruments>
<CsScore>
i1 0 1
</CsScore>
</CsoundSynthesizer>
```

vlazzarini:
Structs using Opcode types were causing a segfault because the context was not passed at initialisation. Test added. Closes #2461


kunstmusik (approved):
Looks good to me!

PR #2463, opened 2026-02-09
Fix for GEN49 bugs on deferred size allocation

vlazzarini:
GEN49 had a couple of bugs on deferred-size allocation, the first leading to a segfault, the second getting the wrong size in samples to allocate the table. This PR fixes both and adds a test.


kunstmusik:
Did some review together with AI. Code changes make sense to me to make consistent with mp3len, but there was one concern that was brought up:

```
Potential Concerns
Integer precision in duration: The duration field is computed as (frames * frame_samples + freq/2) / freq (integer seconds, rounded). Multiplying back by frequency can lose sub-second precision. For a file whose true sample count doesn't land on a whole-second boundary, this could slightly over- or under-allocate. However:

The decode loop in gen49raw has a bounds check (if (p >= flen)) preventing buffer overruns on under-allocation
The key goal is consistency with mp3len * mp3sr, which this achieves
Both mpainfo.frames and mpainfo.duration are zero when no tag info is available, so neither formula would work in that case
```

Mainly that mp3len code looks to use integer for seconds. That's a separate matter so I'm going to approve and merge and this can be looked at in a separate PR.

kunstmusik:
Did some review together with AI. Code changes make sense to me to make consistent with mp3len, but there was one concern that was brought up:

```
Potential Concerns
Integer precision in duration: The duration field is computed as (frames * frame_samples + freq/2) / freq (integer seconds, rounded). Multiplying back by frequency can lose sub-second precision. For a file whose true sample count doesn't land on a whole-second boundary, this could slightly over- or under-allocate. However:

The decode loop in gen49raw has a bounds check (if (p >= flen)) preventing buffer overruns on under-allocation
The key goal is consistency with mp3len * mp3sr, which this achieves
Both mpainfo.frames and mpainfo.duration are zero when no tag info is available, so neither formula would work in that case
```

Mainly that mp3len code looks to use integer for seconds. That's a separate matter so I'm going to approve and merge and this can be looked at in a separate PR.

kunstmusik (approved):

PR #2469, opened 2026-02-12
feat: fully initialize sigaction

hlolli:
Based on totally unrelated finding from coderabbitai https://github.com/csound/csound/pull/2302#discussion_r2800462169

Why this matters:

- sigaction is like a settings form for signal handling.
- Before, only one field was filled and the rest could contain random garbage from memory.
- Random settings can make Ctrl+C/signal behavior unreliable (ignore signals, restart unexpectedly, etc.).
- Initializing it prevents unpredictable behavior and makes signal handling consistent.


kunstmusik (approved):
Looks good to me

PR #2470, opened 2026-02-12
lint: prevent scope leak in macro expansion

hlolli:
Another finding from the rabbit.

In current code, the main call site passes a variable named tval, so the mistake was hidden. Using tval makes it fragile: it can fail to compile or use the wrong value depending on local variable names. Also, on many builds HAVE_C99 is enabled, so the code uses the other branch (llrint) and never hits this buggy fallback.


kunstmusik (approved):
Looks good to me, I see failing builds but I think this branch does not have the vckpg fix, so I will go ahead and merge

PR #2471, opened 2026-02-13
Fix for expr in conditionals

### LINKED ISSUE #2466: Syntax error using inline expression as UDO argument
Just ran across an error in what I think should be valid code.
Here is a simple example:

````
<CsoundSynthesizer>
<CsInstruments>

// True if A plus B equals C
opcode ApBeC(A:i,B:i,C:i):(b)
  xout A+B==C
endop

instr 1

  tf:b=false
  tf=ApBeC(1+0, 2, 3) // OK

  if (1+1==2) && ApBeC(1+0, 2, 3) goto end // OK

  if ApBeC(   1,   2,   3 ) goto end // OK
  if ApBeC(   1, 2+0,   3 ) goto end // OK
  if ApBeC(   1,   2, 3+0 ) goto end // OK
  if ApBeC( 1+0,   2,   3 ) goto end // FAIL (syntax error)

  end:
endin

</CsInstruments>
<CsScore>
i1 0 1 0
</CsScore>
</CsoundSynthesizer>
````
So it appears that the problem only happens when the UDO call is the FIRST item in a conditional expression, AND the inline expression is the FIRST argument to the UDO.

The error is:
````
warning: error: opcode ApBeC with output type (null) not found, line 19

syntax error, Unable to find opcode entry for 'ApBeC'
  with matching argument types, line 19
Found:
  i ApBeC i

Candidates:
  b ApBeC iii
````

vlazzarini:
A bug in the parser arising in expressions of the type `if boolfunc(expr, ...) then` is fixed by this PR. Closes #2466.


kunstmusik:
@vlazzarini: I did a code review with Codex and it found another issue, so iterated with it to get a code change and additional test. Could you review the new code?

vlazzarini:
> @vlazzarini: I did a code review with Codex and it found another issue, so iterated with it to get a code change and additional test. Could you review the new code?

Thanks. I was actually wondering about that part of the code, that it would probably need a similar fix, but I am treading water here so kept my changes narrow. Good that you picked it up and confirmed it with a test.

kunstmusik (approved):

PR #2472, opened 2026-02-13
fix: empty default switch clause and harden the switch unit tests

hlolli:


hlolli:
@coderabbitai review

vlazzarini (dismissed):
All good for this one. 
One other thing I'd like to remind you to have a look at some point is the rules for i-time and perf-time.

 I think these should be like for if ... then ... endif.

- if test is i-time, conditional execution is both at i and perf
- if the test is perf-time, conditions are bypassed and all branches execute to guarantee proper initialisation, then conditional execution resumes at perf-time.

hlolli:
@vlazzarini makes sense, I'll try to add two more tests for this point specifically

hlolli:
@vlazzarini the test turned out to reveal a problem, without the different opcode expansion in Engine/csound_orc_expressions.c it would print

```
assert error, i-time switch executed wrong branch at perf-time
```

igoto => goto
cigoto => cggoto

this means that the switch statement
iCond = 2
switch iCond

was just incorrect. Not sure about the magic of initialising the k-rates. It was incorrect before (as if case where iCond = 1 was hit), now it's correct.

vlazzarini:
tested the init/perf rules and they appear to be the same as if ... then ... endif, that's all good now. However, I spotted this:

```
switch ktest

case 1
 prints "case 1\n"
 printks "casek %d\n", 0, ktest

default
 prints "default\n"
 printks "default %d\n", 0, ktest
endsw
```

giving an error

```
syntax error, unexpected NEWLINE, expecting CASE_TOKEN or DEFAULT_TOKEN (token "
"),  line 10
>>> <<<
```
so I added a rule to allow a newline before the first case.

vlazzarini (approved):

hlolli:
thanks, great teamwork @vlazzarini

PR #2473, opened 2026-02-14
dev: allow darwin systems to find opcode dir from uninstalled builds

hlolli:
As I'm developing locally, the plugin loader will be default try to find CsoundLib64 in `CsoundLib64.framework/Versions/7.0/Resources/Opcodes64/CsoundLib64` but that path only exists after cmake installation. In the build folder, this file is located at a different place under `CsoundLib64.framework/Versions/7.0/CsoundLib64`. With this change it can also locate the corresponding Frameworks directory without OPCODE7DIR. Which makes it little less frictionless to iterate quickly on changes while developing.

### DIFF
--- a/Top/csmodule.c
+++ b/Top/csmodule.c
@@ -529,12 +529,92 @@ static int32_t check_opcode_deny(CSOUND * csound, const char *fname)
 }
 
 
-static int32_t _dir_exists(char *path) {
+static int32_t _dir_exists(const char *path) {
   // returns 1 if path is a directory and it exists
   struct stat s;
   int32_t err = stat(path, &s);
   return (err == 0 && S_ISDIR(s.st_mode)) ? 1 : 0;
 }
+
+#if defined(__MACH__) && (TARGET_OS_IPHONE == 0)
+#ifdef USE_DOUBLE
+#define CS_OPCODE_SUBDIR "Opcodes64"
+#else
+#define CS_OPCODE_SUBDIR "Opcodes"
+#endif
+
+static int32_t append_existing_dir(char *dst, size_t dst_len, const char *dir)
+{
+  size_t len;
+  if (UNLIKELY(dir == NULL || dir[0] == '\0'))
+    return 0;
+  if (UNLIKELY(!_dir_exists(dir)))
+    return 0;
+  len = strlen(dst);
+  if (len > 0) {
+    if (UNLIKELY((len + 1) >= dst_len))
+      return 0;
+    dst[len++] = ':';
+    dst[len] = '\0';
+  }
+  if (UNLIKELY((len + strlen(dir)) >= dst_len))
+    return 0;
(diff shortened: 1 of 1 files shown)


vlazzarini:
Not sure if the paths you quoted are right -
install plugin dir does not have CsoundLib64 in it, that is the name of the Csound dynamic library.

Aside from that - can't you solve this with a CMake option? Or other changes to cmake config?

Search paths in cmodule.c are already very complicated and I think
adding to that just makes it that harder to debug if there is a problem or to modify if we need to.

If the problem is the location of plugin dir, I think the fix is over specified. Let's try to find a simpler solution.

NB: for what's worth I'm used to always run make install. I have CMAKE_INSTALL_PREFIX set to my home dir, so no need for sudo

kunstmusik:
For local builds I usually cd into the build folder and  use `export
OPCODE7DIR64=$pwd` and then do testing.

hlolli:
I knew about OPCODE7DIR64 but I was just annoyed needing to add it, CMAKE_INSTALL_PREFIX to the build directory I didn't consider, if I have this inconveinience annoying me more I'll look rather into CMake configuraiton. No big deal.

vlazzarini:
one think we could do is to

1) have cmake build the framework structure exactly as installed (but in the build directory) to make things uniform.
2) have cmake set the default plugin did to the correct place in the build dir, with some CMake option that you can set and forget.

That way all annoyances will be gone.

hlolli:
I realise that I was probably misled by the warning.

While the portaudio didn't work and `CsoundLib64.framework/Versions/7.0/Resources/Opcodes64` didn't exist, it wasn't really what it was complaining about.

I was looking for the plugins directory portaudio and not the Framework directory, I think the framework stuff was working just fine and CsoundLib64 was found fine.

The problem was, it wasn't finding .dylib files, specifically in my case librtpa.dylib for portaudio.

The fix is very very simple

```
set(CS_OPCODE_DIR "${CMAKE_CURRENT_BINARY_DIR}")
``` 

If I set this in my Custom.cmake I don't have this warning and I get portaudio loaded correctly.

If I didn't set this myself, CS_OPCODE_DIR would defaults to empty string.
As seen on line 149 in CmakeLists.txt `set(CS_OPCODE_DIR "" CACHE PATH "opcodedir path")`

And Csound then falls back to PLUGIN_INSTALL_DIR logic.

CS_OPCODE_DIR becomes CS_DEFAULT_PLUGINDIR macro, this is set by default to framework Resources/Opcodes64 install-style path.

I think we can't modify this without in CMake without breaking the installed csound. The CS_OPCODE_DIR works perfectly for developers and we could recommend it as approach. And perhaps make the warning about the opcode folder not found be a little better. I have a feeling this path that is printed is the last one of fallback paths the module loader used to look for the plugins, making the false assumption that it was looking for the framework binary.

PR #2474, opened 2026-02-16
fixed regression in oversampling input

vlazzarini:
There was a regression in UDO oversampling of inputs leading to a memory exception. Fixed it now.


rorywalsh (approved):
Thanks for this Victor.

Issue #2467, opened 2026-02-12
@csound/browser : only one instance outputing data in multi-instance scenario


jeromew:
I am trying to setup several instance of single-threaded csound processor of @csound/browser **on the same audio context.**
they all have `-odac`

something around the lines of 

```
const csound1 = Csound({ audioContext: ac, autoConnect: false })
await csound1.setOption('-m0d'); 
await csound1.setOption('--sample-accurate');
await csound1.setOption('-odac');
await csound1.compileCsdText(csd);
await csound1.compileOrc(presetsOrc);
await csound1.start();
```
and the same of csound2 & csound3.

I connect their `getNode()` worklet to a destination.

but it seems that only the last csoundN created can output data to the web audio pipeline (?)

i wonder if I am missing an option to make the dac not exclusive or something.

I tried setting `odac1` `odac2` `odac3` but it does not change the result

do you see a reason that could explain this inside csound ?

note: I have tested with 6.18.7 and 7.0.0-beta20 and observe the same behavior

kunstmusik:
@jeromew: This should be supported so this seems like a bug. -odac should be sufficient for all instances. 

I think @hlolli and I need to take a look to see what is going on.

jeromew:
thanks.
i'll keep looking also. it could also be a subtle bug on my side that i still don't understand.
i'll keep looking & try to put up a minimal example.

hlolli:
This could be csound instance isolation issue, can you give a full example for me to test. My initial suspect is the html5 AudioContext is just disconnecting the older csound signals. The name of -odac shouldn't matter either, each csound instance only communicates with the webaudio it creates (or is passed) and there's nothing else in that isolated context. But I don't rule out that some global callbacks are causing one instance to disrupt another, that would be a bug.

jeromew:
@kunstmusik, @hlolli I found a minimal example showing the problem

```js
 import { Csound } from 'https://esm.sh/@csound/browser@7.0.0-beta20'
//import { Csound } from 'https://esm.sh/@csound/browser@6.18.7'

 const code = `
instr 1
  out linenr(oscili(0dbfs*p4,p5),0.01,0.5,0.01)
endin
`;

const ac = new AudioContext();

async function getCsoundInstance(ctx) {
  const c = await Csound({ audioContext: ctx, autoConnect: false });
  await c.setOption("-odac");
  await c.compileOrc(code);
  await c.start();
  return c;
}

const csound1 = await getCsoundInstance(ac);
[...]

PR #2475, opened 2026-02-16
Fix: multiple webaudio nodes

kunstmusik:
Moves state from global to instance for single-thread workers; fixes issue with multiple csound instances in webaudio (issue #2467)


hlolli (approved):
many improvements and cleanup in scope leaks and good use of WeakMap

PR #2480, opened 2026-02-17
Fix: global state webaudio

kunstmusik:
This PR improves wasm/browser instance isolation and shared-context lifecycle safety in the WebAudio paths.

It addresses three areas:

1. Prevents AudioContext ownership bugs when multiple Csound instances share a user-provided context.
2. Removes module-global wasm host state that could leak across instances in shared worklet scopes.
3. Cleans up retained node mappings in the worklet worker.

### DIFF
--- a/Engine/csound_orc_expressions.c
+++ b/Engine/csound_orc_expressions.c
@@ -336,6 +336,13 @@ static TREE *create_cond_expression(CSOUND *csound,
     last = last->next;
   }
 
+  if(last->left == NULL) {
+    csound->Message(csound,
+                    "missing boolean expression in " 
+                    "conditional expression, line %d\n", root->line-1);
+    return NULL;
+  }
+
   if (left[0]=='S' || right[0]=='S') {
     type = (last->left->value->lexeme[1]=='B') ?2 : 1;
     eq = (last->left->value->lexeme[1]=='B') ?"#=.S" : "=.S";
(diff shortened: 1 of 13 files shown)


hlolli (approved):

jeromew:
+1 for not closing AudioContext when it is an external AudioContext.

there is something else that I think would be nice in this `terminateInstance` case.

currently, from my cursory reading, the `process` function of the audio worklets never return false, which means that the worklet will never be discarded by the audioContext and thus probably never garbage collected.

cf https://github.com/csound/csound/blob/develop/wasm/browser/src/workers/worklet.singlethread.worker.js#L251

there should probably be a state check inside process to check that the instance is terminated and return false in this case.

https://developer.mozilla.org/en-US/docs/Web/API/AudioWorkletProcessor/process

> **Returning true forces the Web Audio API to keep the node alive**, while returning false allows the browser to terminate the node if it is neither generating new audio data nor receiving data through its inputs that it is processing.

kunstmusik:
FYI: I merged latest from develop and released beta23, then saw comment from @jeromew. Looking at terminateInstance and returning false now.

kunstmusik:
@jeromew : I've put in a change to return false once terminateInstance is called. (Also has another fix for worker version related to shared audiocontext issues.) This is published as 7.0.0-beta25 now. 

@hlolli: This is ready for review/merge.

hlolli (approved):

PR #2484, opened 2026-02-18
Multiport MIDI output

### LINKED ISSUE #1750: feature request: several midi ports

vlazzarini:
Multiport MIDI output implemented in the engine and deployed with portmidi backend. 

- All midi out opcodes now have an extra optional parameter setting the MIDI output port.
- Portmidi can open single output devices as before or all output devices with `-Qa` or `-Qm` (as for input). The first sends messages to all open ports, ignoring port number, the second maps messages to ports using the optional parameter in opcodes.

Additionally, two module API functions were added

`SendMidiMsg()` - opcodes can send midi messages directly using this.
`GetMidiOutPort()` - backends can retrieve the midi outport for a message.

Closes #1750  

EDIT: also reinstates commented out timing precision code.


rorywalsh (approved):
Thanks Victor, if you get a chance can you update the manual pages for the opcodes in question so the changes are visible to users.

PR #2485, opened 2026-02-20
Exponential interp in adsynt2

### LINKED ISSUE #1521: adsynt2 cubic
would it make any improvement for adsynt2 to have an option for cubic interpolation as well?
The opcode already sounds much better than adsynt due to linear interpolation, but perhaps it would sound better still with the cubic option?

vlazzarini:
This PR adds an option to set exponential interpolation for amplitudes in adsynt2.
Closes #1521

### DIFF
--- a/Android/CsoundAndroid/jni/version.h
+++ b/Android/CsoundAndroid/jni/version.h
@@ -46,4 +46,3 @@
 #endif
 
 #endif /* CSOUND_VERSION_H */
-
--- a/Opcodes/gab/gab.c
+++ b/Opcodes/gab/gab.c
@@ -484,6 +484,8 @@ static int32_t adsynt2_set(CSOUND *csound,ADSYNT2 *p)
   if (p->pamp.auxp==NULL ||
       p->pamp.size < (uint32_t)(sizeof(MYFLT)*p->count))
     csound->AuxAlloc(csound, sizeof(MYFLT)*p->count, &p->pamp);
+  
+  if(*p->iphs >= 0) {
   // linear
   if(!*p->interp) memset(p->pamp.auxp, 0, sizeof(MYFLT)*p->count);
   else { // expon
@@ -491,6 +493,7 @@ static int32_t adsynt2_set(CSOUND *csound,ADSYNT2 *p)
     for(count = 0;count<p->count;count++) {
       pamp[count] = 0.0001*csound->Get0dBFS(csound);
     }
+  }
   }
   return OK;
 }
@@ -508,6 +511,7 @@ static int32_t adsynt2(CSOUND *csound,ADSYNT2 *p)
     uint32_t early  = p->h.insdshead->ksmps_no_end;
     uint32_t n, nsmps = CS_KSMPS;
     MYFLT odbfs = csound->Get0dBFS(csound);
+    int32_t interp = (int32_t) *p->interp;
 
   /* I believe this can never happen as InitError will remove instance */
   /* The check should be on p->amptp and p->freqtp  -- JPff            */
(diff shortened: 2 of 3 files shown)


vlazzarini:
@kunstmusik this one is ready to go

kunstmusik:
Ran code reviews with Opus and Codex and both picked up on the same two issues. The reply from Opus below which had a little more information to it:


## PR Review: Exponential interp in adsynt2

**Summary:** Adds an optional `interp` parameter to `adsynt2` to select between linear (default, `0`) and exponential (`non-zero`) amplitude interpolation. CI is green. The feature itself is straightforward but there are a few issues.

### 1. Regression: `iphs < 0` no longer preserves previous amplitudes

**Severity: High**

The original code in `adsynt2_set` had:

```c
if (p->pamp.auxp==NULL || p->pamp.size < ...)
    csound->AuxAlloc(csound, sizeof(MYFLT)*p->count, &p->pamp);
else if (*p->iphs >= 0)        /* AuxAlloc clear anyway */
    memset(p->pamp.auxp, 0, sizeof(MYFLT)*p->count);
```

The `else if (*p->iphs >= 0)` guard meant that when `iphs < 0`, the previous amplitudes in `pamp` were **preserved** across reinit — consistent with how negative `iphs` skips phase reinitialization a few lines above.

The new code (gab.c) **unconditionally** overwrites `pamp`:

```c
if(!*p->interp) memset(p->pamp.auxp, 0, sizeof(MYFLT)*p->count);
else { ... pamp[count] = 0.0001*csound->Get0dBFS(csound); }
```

This is a **behavioral regression** for both linear and exponential modes. Existing instruments using `iphs = -1` to preserve state across tied notes or reinits will now get their previous amplitudes clobbered to zero, causing audible clicks/discontinuities.

**Fix:** Wrap the new initialization in the same `iphs` guard, e.g.:
```c
if (*p->iphs >= 0 || p->pamp.auxp == NULL) {
    if(!*p->interp) memset(p->pamp.auxp, 0, sizeof(MYFLT)*p->count);
    else { /* expon init */ }
}
```

### 2. Division by zero / NaN in exponential mode when amplitude reaches zero

**Severity: High**

In the performance routine (gab.c):

```c
amp = amp >= 0 ? amp : odbfs*0.0001;
ampIncr = pow(amp/amp2, CS_ONEDKSMPS);
```

If the target amplitude `amp` (from `amptbl[c] * amp0`) is **exactly zero** — a very common case (e.g., turning off a partial):

1. `amp >= 0` is true, so `amp` stays `0`
2. `ampIncr = pow(0/amp2, ...) = 0`
3. In the inner loop: `amp2 *= 0 = 0`
4. `prevAmp[c] = amp` stores `0`
5. **Next k-period**: `amp2 = prevAmp[c] = 0`, so `ampIncr = pow(new_amp / 0, ...)` → **Inf/NaN**, corrupting the output permanently

**Fix:** Clamp `amp` to a minimum positive floor (not just negative values), and also clamp `amp2`:
```c
amp = amp > FL(0.0) ? amp : odbfs * FL(0.0001);
amp2 = amp2 > FL(0.0) ? amp2 : odbfs * FL(0.0001);
ampIncr = pow(amp/amp2, CS_ONEDKSMPS);
```

### 3. Minor: opcode signature change is backward-compatible (OK)

The change from `"kkiiiio"` to `"kkiiiioo"` adds `interp` as a second optional param. When omitted, `*p->interp` defaults to 0 (linear), preserving existing behavior. This is correct.

### 4. Minor: per-sample branch on `*p->interp`

Inside the inner sample loop (gab.c):
```c
if(!*p->interp)
    amp2 += ampIncr;
else 
    amp2 *= ampIncr;
```

This dereferences `p->interp` every sample. Since `interp` is i-rate, it would be slightly cleaner to cache it in a local variable at the top of the perf function (like `floatph` already is). Very minor — the compiler may optimize this anyway.

### 5. Cosmetic: inconsistent style

Minor: `if(!*p->interp)` lacks space after `if`, and there are mixed brace styles. Not a blocker but inconsistent with surrounding code.

---

**Bottom line:** Issues #1 and #2 should be fixed before merge. The `iphs < 0` regression will silently break existing instruments, and the zero-amplitude division-by-zero will produce NaN output in normal usage of exponential mode. 

Completed: *Write review summary* (3/3)

vlazzarini:
yes, I was thinking of 1 alright. The other one is a typo, I meant to have excluded 0.

vlazzarini:
should be good now.

kunstmusik (approved):

kunstmusik:
Looks good to me, merging.

PR #2486, opened 2026-02-20
wasm: fix unlink and filename overwrites issues

### LINKED ISSUE #2045: WASM fs: unlink not working, writeFile can create duplicate files
In the WASM implementation, fs.unlink does currently not delete files, and creating a file with fs.writeFile in an attempt to overwrite it results in a duplicate with the same name. Incidentally file appending works OK, so it seems existing data may not be modified. The following are examples of how to recreate:

    // unlink does not remove files
    let file = "/deleteme";
    await csound.fs.writeFile(file, new Uint8Array([0]));
    await csound.fs.unlink(file);
    console.log(await csound.fs.readdir("/"));
    // ["deleteme"]

    // appearance of duplicate files when writing same path
    for (var x = 0; x < 6; x++) csound.fs.writeFile("/test", new Uint8Array([99 + x]));
    console.log(await csound.fs.readdir("/"));
    // ["test", "test", "test", "test", "test", "test"]
    // using readFile on "/test" shows the content is 99 - the first file written appears to be the one read

hlolli:
Fixes #2045

### DIFF
--- a/wasm/browser/src/filesystem/wasi.js
+++ b/wasm/browser/src/filesystem/wasi.js
@@ -1164,7 +1164,7 @@ WASI.prototype.sock_shutdown = function () {
 
 WASI.prototype.findBuffers = function (filePath /* string */) {
   const maybeFd = this.findEntry(filePath);
-  return maybeFd && maybeFd.buffers;
+  return maybeFd?.buffers;
 };
 
 // fs api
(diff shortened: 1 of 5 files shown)


wasm/browser/src/filesystem/wasi.js
```diff
@@ -1151,7 +1163,7 @@ WASI.prototype.sock_shutdown = function () {
 // helpers
 
 WASI.prototype.findBuffers = function (filePath /* string */) {
-  const maybeFd = Object.values(this.fd).find(({ path }) => path === filePath);
+  const maybeFd = this.findEntry(filePath);
   return maybeFd && maybeFd.buffers;
```
kunstmusik: Maybe could be `return maybeFd?.buffers`

hlolli: done

kunstmusik (approved):
Published new wasm-bin and browser releases.

PR #2490, opened 2026-02-21
Adding column ranges to parse parameters

### LINKED ISSUE #545: Enhancing parser output with source ranges
Csound reports lines where syntax errors occur, but seemingly not columns, even when using the `--verbose` option. It may be helpful if Csound reported information about source files as line/column ranges, perhaps along the lines of of Clang’s [`-fdiagnostics-print-source-range-info` flag](http://clang.llvm.org/docs/UsersManual.html#opt-fdiagnostics-print-source-range-info).

vlazzarini:
This PR adds column ranges to parse parameters in the lexer.

 - csound_orcerror() modified to use this for better diagnostics
 - ~future enhancements may add this to the TREE (through make_node() etc)~ column ranges added to ORCTOKEN
 - verify_tree() modified to use this for better diagnostics - demonstrates the use in the parser.
 - this is now available for other synterr() applications (beyond the scope of the PR)

NB: synthetic tokens do not set line ranges, as we would expect them not to raise syntax errors themselves.

closes #545

### DIFF
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -53,12 +53,7 @@ ORCTOKEN *make_label(CSOUND *, char *s);
 #define YY_EXTRA_TYPE  PARSE_PARM *
 #define PARM    yyget_extra(yyscanner)
 
-/* #define YY_INPUT(buf,result,max_size)  {\ */
-/*     result = get_next_char(buf, max_size, yyg); \ */
-/*     if ( UNLIKELY( result <= 0  )) \ */
-/*       result = YY_NULL; \ */
-/*     } */
-
+#define YY_USER_ACTION PARM->first_column = yycolumn; PARM->last_column = yycolumn + yyleng - 1; yycolumn += yyleng;
 #define YY_USER_INIT
 
 struct yyguts_t;
@@ -571,9 +566,9 @@ ERSTR           "}R"
 {FILE}          { BEGIN(src); }
 
 <src>{
-  [ \t]*     /* eat the whitespace */
+  [ \t]*     /* eat the whitespace */ {yycolumn += yyleng;}
   {FNAME}    { PARM->locn = atoll(yytext); }
-  "\n"       { BEGIN(INITIAL); }
+  "\n"       { BEGIN(INITIAL);}
 }
 
 .               {
@@ -589,14 +584,16 @@ ERSTR           "}R"
 
 <ignorenewline>{
   "\n" {
+    yycolumn = 1;
     csound_orcset_lineno(1+csound_orcget_lineno(yyscanner),
                                          yyscanner);
   }
 }
 
-<INITIAL>"\n"            { csound_orcset_lineno(1+csound_orcget_lineno(yyscanner),
+<INITIAL>"\n" { yycolumn = 1;
+                csound_orcset_lineno(1+csound_orcget_lineno(yyscanner),
                                        yyscanner);
-                  return NEWLINE; }
+                return NEWLINE; }
 
 %%
 
@@ -729,3 +726,16 @@ uint64_t csound_orcget_ilocn(void *yyscanner)
 //    struct yyguts_t *yyg  = (struct yyguts_t*)yyscanner;
     return PARM->ilocn;
 }
+
+uint32_t csound_orcget_first_column(void *yyscanner)
+{
+  //   struct yyguts_t *yyg  = (struct yyguts_t*)yyscanner;
+    return PARM->first_column;
+}
+
+uint32_t csound_orcget_last_column(void *yyscanner)
+{
+  //   struct yyguts_t *yyg  = (struct yyguts_t*)yyscanner;
+    return PARM->last_column;
+}
+
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -309,6 +309,7 @@ ERSTR           "}R"
   }
 
   "\n"  { /* The next two should be one case but I cannot get that to work */
+           yycolumn = 1;
            if (PARM->xstrptr+2>=PARM->xstrmax) {
                PARM->xstrbuff = (char *)realloc(PARM->xstrbuff,
                                                        PARM->xstrmax+=80);
@@ -557,7 +558,9 @@ ERSTR           "}R"
 {WHITE}         { }
 
 {SLINE}         { BEGIN(sline); }
-<sline>{INTGR}   { csound_orcset_lineno(atoi(yytext), yyscanner); }
+<sline>{INTGR}   { csound_orcset_lineno(atoi(yytext), yyscanner);
+                  yycolumn = 0; /* reset for new source line;
+                                    0 accounts for trailing space in #sline format */ }
 <sline>[ \t]*   { BEGIN(INITIAL);}
 {LINE}          { BEGIN(line); }
 
@@ -570,7 +573,7 @@ ERSTR           "}R"
 {FILE}          { BEGIN(src); }
 
 <src>{
-  [ \t]*     /* eat the whitespace */ 
+  [ \t]*     /* eat the whitespace */
   {FNAME}    { PARM->locn = atoll(yytext); }
   "\n"       { BEGIN(INITIAL); yycolumn = 1;}
 }
@@ -622,7 +625,7 @@ ORCTOKEN *lookup_token(CSOUND *csound, char *s, void *yyscanner)
     if (csound->parserNamedInstrFlag == 1) {
         return ans;
     }
-    ans->type = type; 
+    ans->type = type;
     return ans;
 }
 
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -597,6 +597,11 @@ ERSTR           "}R"
 
 %%
 
+#if defined(_WIN32) || defined(_WIN64)
+# define strtok_r strtok_s
+#endif
+
+
 ORCTOKEN *lookup_token(CSOUND *csound, char *s, void *yyscanner)
 {
     int32_t type = T_IDENT;
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -597,11 +597,6 @@ ERSTR           "}R"
 
 %%
 
-#if defined(_WIN32) || defined(_WIN64)
-# define strtok_r strtok_s
-#endif
-
-
 ORCTOKEN *lookup_token(CSOUND *csound, char *s, void *yyscanner)
 {
     int32_t type = T_IDENT;
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -40,13 +40,13 @@
 #include "filesys.h"
 YYSTYPE *yylval_param;
 YYLTYPE *yylloc_param;
-ORCTOKEN *make_string(CSOUND *, char *);
-extern ORCTOKEN *lookup_token(CSOUND *, char *, void *);
-ORCTOKEN *new_token(CSOUND *csound, int32_t type);
-ORCTOKEN *make_int(CSOUND *, char *);
-ORCTOKEN *make_num(CSOUND *, char *);
-ORCTOKEN *make_token(CSOUND *, char *s);
-ORCTOKEN *make_label(CSOUND *, char *s);
+ ORCTOKEN *make_string(CSOUND *, char *, void*);
+ ORCTOKEN *lookup_token(CSOUND *, char *, void *);
+ ORCTOKEN *new_token(CSOUND *csound, int32_t type, void *yyscanner);
+ ORCTOKEN *make_int(CSOUND *, char *, void*);
+ ORCTOKEN *make_num(CSOUND *, char *,  void*);
+ ORCTOKEN *make_token(CSOUND *, char *s,  void*);
+ ORCTOKEN *make_label(CSOUND *, char *s,  void*);
 #define namedInstrFlag csound->parserNamedInstrFlag
 #include "parse_param.h"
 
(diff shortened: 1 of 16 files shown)


kunstmusik:
Ran this by Codex and Opus and they found some similar issues. I posted the Opus report here. The must-fix and should-fix look valid to me, as well as the note about "source line echo".


## PR #2490 Full Review: "Adding column ranges to parse parameters"

### Overview

This PR (5 commits on `feature/columns`) adds column-range tracking to the Csound orchestra parser so error messages can report `line N, columns X-Y` instead of just line numbers. The changes span the lexer, parser, token structures, error messages, and CSD file reading.

**Build:** Compiles cleanly on macOS. **Tests:** All 78 tests pass.

---

### Commit-by-Commit Summary

| Commit | Description | Files |
|--------|-------------|-------|
| `ffca3587e` | Add `first_column`/`last_column` to `PARSE_PARM`; track via `YY_USER_ACTION`; use in `csound_orcerror` | 4 files |
| `d2882eaf2` | Add `\n` to CSD error messages; adjust unknown-tag logic | 2 files |
| `9cb2fca19` | **Core change:** Add columns to `ORCTOKEN`; update all token-creating functions to accept `yyscanner`; move `lookup_token` to lexer; propagate column info into error messages | 10 files |
| `3564952f3` | Move Windows `strtok_r` define to csound_orc_compile.c | 1 file |
| `743fff546` | Move Windows `strtok_r` define to sysdep.h globally | 1 file |

---

### Issues

#### Bug: Double-counting `yycolumn` in `<src>` whitespace rule

csound_orc.lex:
```
<src>{
  [ \t]*     /* eat the whitespace */ {yycolumn += yyleng;}
```

`YY_USER_ACTION` already does `yycolumn += yyleng` before every rule action. The explicit `{yycolumn += yyleng;}` in the `<src>` whitespace action **double-counts** the column advance. This inflates `yycolumn` for tokens following whitespace in `<src>` mode. Should be just a comment with no action body, matching the `<line>` whitespace rule on csound_orc.lex.

#### Bug: Missing `yycolumn = 1` reset in newline-consuming rules

Only the `<INITIAL>` and `<ignorenewline>` newline rules reset `yycolumn = 1`. These do not:

| Rule | Line | State | Impact |
|------|------|-------|--------|
| `{CONT}` (line continuation `\<newline>`) | csound_orc.lex | INITIAL | **High** — common in user code; column will be wrong on continued lines |
| `"\n"` in `<udoarg>` | csound_orc.lex | udoarg | Medium — affects UDO arg definitions |
| `"\n"` in `<line>` | csound_orc.lex | line | Low |
| `"\n"` in `<src>` | csound_orc.lex | src | Low |
| `"\n"` in `<xstr>`/`<rstr>` | csound_orc.lex/csound_orc.lex | xstr/rstr | Low — multi-line strings |

At minimum, `{CONT}` and `<udoarg>` should add `yycolumn = 1;`.

#### Bug/Logic: Inverted condition or misleading comment in one_file.c

one_file.c:
```c
if(result > 0) // reading failed before here
    csoundMessage(csound, Str("unknown CSD tag: %s"), buffer);
else
    csoundMessage(csound, "\n");
```

`result` starts as `TRUE` (1) and becomes `FALSE` (0) on failure. So `result > 0` means **success**, but the comment says "reading failed before here." The behavior is also questionable: on failure, an unknown tag is swallowed and replaced with just `"\n"`. Additionally, the `if` body is not indented relative to the `if` — reads as if at the same level.

#### Information regression: Source line echo removed from `csound_orcerror`

The old `csound_orcerror` printed the actual source line of code between `>>>` and `<<<` markers. The new version only shows `line N, columns X,Y`. This is a meaningful usability regression for command-line users who relied on seeing the offending code in the error output. Consider keeping the source echo alongside the new column numbers.

#### Missing `\n` in several new error messages

Several of the new column-enhanced error messages in csound_orc_semantics.c don't end with `\n`:

- csound_orc_semantics.c: `"columns %d-%d"` — no trailing newline
- csound_orc_semantics.c: `"columns %d-%d"` — no trailing newline  
- csound_orc_semantics.c: `"columns %d-%d"` — no trailing newline
- csound_orc_semantics.c: `"columns %d-%d"` — no trailing newline
- csound_orc_semantics.c: `"columns %d-%d"` — no trailing newline
- csound_orc_semantics.c: `"columns %d-%d"` — no trailing newline

Some of these are followed by `csoundMessage` calls that handle their own newlines, but the final `"\nLine: %d\n columns %d-%d"` message at csound_orc_semantics.c has no trailing newline before `do_baktrace`.

#### Style: Redundant `strtok_r` defines remain

After moving the `strtok_r` define to sysdep.h, the local definitions in symbtab.c and csound_orc_compile.c are now redundant (and could lead to "macro redefined" warnings on Windows). They should be removed.

#### Style: Inconsistent indentation in new functions

- `make_token` and `make_string` have their body indented at 2 spaces ([line 637](Engine/csound_orc.lex#L637)), while surrounding functions use 4:
  ```c
  ORCTOKEN *make_token(CSOUND *csound, char *s, void *yyscanner)
  {
    ORCTOKEN *ans = new_token(csound, STRING_TOKEN, yyscanner);  // 2-space indent
      ans->lexeme = csoundStrdup(csound, s);                      // 4-space indent
  ```

- `new_token` body mixes 1-space and 4-space indentation ([line 631](Engine/csound_orc.lex#L631)):
  ```c
  if(yyscanner) {
   ans->first_column = PARM->first_column;    // 1-space indent
   ans->last_column = PARM->last_column;
  }
  ```

#### Style: Trailing whitespace

- csound_compiler.h: `last_column;    ` (trailing spaces)
- parse_param.h: `last_column;  ` (trailing spaces)
- csound_orc_semantics.c: `last_column;      ` (trailing spaces)

#### Style: Stray blank lines

- csound_orc.y: extra blank line between `%lex-param` and `%token`
- csound_messages.c: three whitespace-only additions (noise)
- csound_orc.lex: trailing blank line at end of file

#### Minor: "never reached?" comment is incorrect

csound_orc.lex:
```c
ans->type = type; // never reached?
return ans;
```
This IS reached when `parserNamedInstrFlag != 1` (the normal case). The comment should be removed.

---

### Positive Aspects

- **Well-structured approach**: Column info flows from lexer → `PARSE_PARM` → `ORCTOKEN` → error messages. The `YY_USER_ACTION` approach is the canonical flex pattern.
- **Thorough API update**: All `make_token`/`make_int`/`make_num`/`make_string`/`make_label`/`new_token` callers correctly updated. Grammar actions from the parser pass `NULL` (no scanner context) which is properly handled.
- **`copy_node`/`copy_node_shallow`** properly preserve column info when cloning trees.
- **Windows `strtok_r` consolidation** into sysdep.h is a good cleanup (just needs the redundant defines removed).
- **All 78 tests pass**. No regressions detected.
- **ABI change to `ORCTOKEN`** is acceptable since this is pre-release (v7 beta).

---

### Summary of Recommended Actions

| Priority | Issue | Fix |
|----------|-------|-----|
| **Must fix** | Double-counting `yycolumn` in `<src>` whitespace | Remove `{yycolumn += yyleng;}` |
| **Must fix** | Missing `yycolumn = 1` in `{CONT}` rule | Add reset |
| **Should fix** | Inverted comment / questionable logic in one_file.c unknown-tag handling | Clarify intent |
| **Should fix** | Missing `\n` in new column error messages | Add trailing newlines |
| **Should fix** | Redundant `strtok_r` defines in `symbtab.c` and `csound_orc_compile.c` | Remove |
| **Should fix** | Missing `yycolumn = 1` in `<udoarg>` newline | Add reset |
| **Nice to have** | Inconsistent indentation in new functions | Normalize to 4 spaces |
| **Nice to have** | Consider restoring source-line echo alongside columns | Re-add `>>>...<<<` display |
| **Nice to have** | Trailing whitespace, stray blank lines, incorrect comment | Clean up |

vlazzarini:
- The one_file.c diagnostics is just plain wrong. It needs to jump the message because of some error (result > 0), so that it does not report that there was a wrong tag. Logic is not reversed. I meant it to be like that. Changed the comment.

- Also the newlines are already added by synterr - removed so we don't have double newlines.

- the offending  code is already reported by the error message, no need to repeat it.

I'll take care of the changes in the lexer code now.

kunstmusik:
I think the one_file.c diagnostic is getting confused as it's not particularly clear code. Seems to me we don't even need the check for result > 0, since if you get to that point it's always going to be an unknown tag and we'd want to report that even if there's previous failures, yes?

vlazzarini:
> I think the one_file.c diagnostic is getting confused as it's not particularly clear code. Seems to me we don't even need the check for result > 0, since if you get to that point it's always going to be an unknown tag and we'd want to report that even if there's previous failures, yes?

No, that's not always the case. For example, if the parser picks up an early error like a missing end quote on a string, the original code fails to recognise that and prints that we found an unknown tag. We need to check to see if there was such an error so it does not print that. We also need to print a newline because early errors like that do not add a newline when csoundErrorMsg() is called by the lexer. 

OTOH, when there is an unknown tag, result is zero and we get a printout.

vlazzarini:
Actually now I see that error is not coming from the lexer, but from the CSD reader, but the solution is still holding.

vlazzarini:
I've added the source of the CSD error to improve the diagnostics plus the missing newline now.

vlazzarini:
found another missing newline in error message. We'll need to do an audit at some point.

kunstmusik:
The problem with the existing code is that once there's an error, result will always be 0 and unknown tags won't print. It's a minor issue in the end as most people won't hit that case but something to take care of eventually.

vlazzarini:
> The problem with the existing code is that once there's an error, result will always be 0 and unknown tags won't print. It's a minor issue in the end as most people won't hit that case but something to take care of eventually.

It's better than printing the wrong message when an error is reached, which was the case originally (and without a newline, which made it doubly confusing). The error in question is in CSD reading, not actual compilation.

kunstmusik:
Not arguing which is better, just that there's a problem in this area of the code.

vlazzarini:
The nature of error detection is such that if an invalid is found before the CSD read error, it will trigger the message. If an error is found before, then the reading stops there. There are only two errors - unmatched " and lack of a closing tag - for CsInstruments and CsScore. I wonder if the the first shouldn't be ignored and passed on to the compiler, but it's probably an issue for a different PR.
EDIT: there's code for that, it's ifdef'd out in favour of the current state of things. We can adopt it. It just delegates the error checking to the compiler.

vlazzarini:
On balance, I think we should just pass onwards all the text inside the CsInstruments and CsScore tags without interpretation, and let the compiler and score reader deal with it. The error messages are better, with line numbers etc. So I have changed the code to use the simpler reading code instead of the one that does some error checking. That also allows me to remove the check for errors in the unknown tag message.

kunstmusik:
Sounds like a good plan, I'm currently looking at generating some unit tests (c tests) to check column numbering. I'll probably finish up tomorrow and double-check before pushing anything (probably tomorrow).

kunstmusik:
I did a re-review with Opus. i think we're in good shape now. Opus listed some remaining issues that I'll post below for the sake of having it mentioned, but I think they're not blockers. (Though, 4a noting the comma vs dash separator is interesting.) I'm going to go ahead and approve. @vlazzarini LMK what you think about 4a; if you're fine with it feel free to merge.


### 4. Remaining Issues

#### 4a. Minor: Inconsistent column format in error messages

`verify_opcode` (semantics) uses **`columns %d-%d`** (dash separator):
```c
synterr(csound, Str("unable to find opcode with name: %s, line %d,"
                     " columns %d-%d"), ...);
```

`csound_orcerror` (parser) uses **`columns %d,%d`** (comma separator):
```c
csound->ErrorMsg(csound, Str(" line %d, columns %d,%d\n"), line, column1, column2);
```

**Recommendation:** Standardize on one format (prefer `columns %d-%d` for readability, or `%d:%d-%d:%d` for line:col range).

#### 4b. Minor: `do_at` signature change

The `do_at` function was changed from taking `struct yyguts_t *yyg` to `void *yyscanner, char *yytex`. The extra `yytex` parameter was needed because `yytext` is a macro that expands via `yyg`, which is no longer in scope with abstracted `void *`. This is fine, but the parameter name `yytex` is easy to confuse with `yytext` — consider renaming to `text` for clarity. Very minor.

#### 4c. Cosmetic: Trailing whitespace / blank lines in csound_messages.c

Three blank-line additions in csound_messages.c appear to be accidental whitespace changes (no functional change).

#### 4d. Note: `<xstr>` and `<rstr>` string tokens report closing-delimiter columns

For multi-line strings (`{{ ... }}` and `R{ ... }R`), the `YY_USER_ACTION` fires on the closing `}}` or `}R` token, so the column range reported is that of the **closing delimiter**, not the opening. This is acceptable behavior — the token is produced at the closing delimiter — but worth documenting.

#### 4e. Note: Quoted string column tracking

For regular `"..."` strings, flex's `input()` function is used for character-by-character scanning, which bypasses `YY_USER_ACTION`. The reported columns reflect only the opening `"` position (first_column = last_column = position of `"`). Newline resets still work correctly for subsequent tokens. This is a known limitation of flex's `input()` approach and not easily fixable without major restructuring.

#### 4f. gab.c and `gab.h` changes

These changes (removing the `interp` parameter from `adsynt2` and simplifying the opcode to linear-only interpolation) appear **unrelated to column tracking**. The `adsynt2` opcode's argument string is changed from `kkiiiioo` to `kkiiiio` (7→6 optional args), and the exponential interpolation code path is removed. This should be in a separate PR or at least called out in the PR description.

kunstmusik (approved):

vlazzarini:
I think 4a should be normalised to the dash alright, but it's minor. I am concerned by 4f - I don't remember touching those files. If they have been touched it's a regression. Let me check now.

vlazzarini:
I think there's some hallucination going on: gab.c and gab.h are not part of the changes in this PR.

kunstmusik:
Yeah I think the gab.c/gab.h were just seeing that there were differences but not accounting for not merging latest from develop into this branch. It should be perfectly fine.

Issue #2492, opened 2026-02-23
extend shorthands for slicearray


joachimheintz:
this works currently:
  arr:i[] = [1 ... 10]
  sec1:i[] = arr[0:5]
  sec2:i[] = arr[5:9]

it would be great to extend this to
  sec1:i[] = arr[:5]
  sec2:i[] = arr[5:]

PR #2494, opened 2026-02-23
Enhancements to slices and overloads

### LINKED ISSUE #2493: overload print
i suggest to overload print (if possible) to print also i-rate strings and arrays:
print("my string") would be the same as puts("my string",1), and 
print([1 ... 10]) would be the same as printarray([1 ... 10])

vlazzarini:
This PR enhances the slicearray shorthand by allowing the [:b] and [a:] syntaxes. It also adds `print` overloads for strings and arrays. Closes #2493 and #2492.


vlazzarini:
This is ready now - let's try to close it.

rorywalsh:
@joachimheintz can you test this?

joachimheintz:
yes will do tomorrow --- occupied with other things these days ...

joachimheintz (commented):
tested and works.
i have two comments / suggestions:
1. print("string") should also add a newline
2. can we do the same overload for printk?
  printk(1,kvar)
  printk(1,"string")
  printk(1,karray)
would print once a second kvar, "string", and the karray.

i will approve anyway in case you do not want to do this now, or in case it needs more discussion.

joachimheintz (approved):

PR #2496, opened 2026-02-24
SRC to use type system

### LINKED ISSUE #1869: Complete Local SR to work with UDTs and type system
Local SR work in #1844 only works with a- and k- types. This should be further iterated upon to work generically via the type system so that it will work with arrays of a- and k- as well as UDTs. Will work with @vlazzarini to work through this.

vlazzarini:
This PR refactors the sample rate conversion objects to use the type system directly:

- SRC units are set up for each input/output argument.
- Using the type system they check for argument type
- If arguments are audio/control signals or audio/control arrays, a converter is provided
- Otherwise, SRC is bypassed.

In the UDO, xinset/xoutset take care of set up if local sampling rate is used. The relevant
opcode perf code then calls the SRC function, passing the converters - for each perf-time
argument (in/out). This takes care of conversion or just copy (in the case of bypass).

UDTs raise an error - they are not supported.

A test for arrays has been added.

closes #1869


kunstmusik:
Didn't have a chance to look took carefully yet but ran it through Opus for a review and got the following (take it with grain of salt; will do a more in-depth review tomorrow):

## PR #2496 Review — "SRC to use type system"

All CI checks pass, the project builds locally, and all 91 tests pass. The refactoring goal (one SRC converter per argument using the type system, with bypass for non-k/a types) is clean and well-structured. However, I found several issues:

---

### Bug 1 (Regression): Missing array index in `src_linear_deinit` — double-free / memory leak

srconvert.c:

```c
for(int i = 0; i < pp->ncvt; i++) {
   csound->Free(csound, pp->dat->bufferin);   // <-- should be pp->dat[i].bufferin
   csound->Free(csound, pp->dat->data);       // <-- should be pp->dat[i].data
}
```

When `ncvt > 1` (e.g., arrays with multiple elements), this frees `dat[0]`'s fields N times (double-free after the first iteration) and **leaks** `dat[1..N-1]`'s `bufferin` and `data`. This will be triggered by the new array test (`Test1` opcode with a 2-element a-rate array) when `mode >= 4` (linear converter). **This is the most critical issue.**

---

### Bug 2: Memory leak on partial `src_new` failure in SRC version of `src_init`

srconvert.c:

```c
else {
  // init fail - free memory
  csound->Free(csound, pp->dat);
  csound->Free(csound, pp);
  return NULL;
}
```

If `src_new()` fails at iteration `i > 0`, the already-initialized `dat[0..i-1]` entries (their `bufferin`, `bufferout`, SRC state, and SRC struct) are leaked. The cleanup should iterate over the successfully initialized entries and call `src_delete`/`Free` on each.

---

### Minor: Typo in header declaration

srconvert.h: Parameter name `csoubnd` should be `csound`:

```c
int32_t src_convert(CSOUND *csoubnd, SR_CONVERTER *p, MYFLT *in, MYFLT *out);
//                          ^^^^^^^ typo
```

Doesn't affect compilation (it's just a declaration), but confusing for readers.

---

### Minor efficiency note: Unnecessary converter allocation for i-rate args

In `set_inbufs` and `xoutset`, converters are now allocated for **every** argument when `esr != parent_sr`, including i-rate variables. The `src_init` function creates a bypass converter (`ncvt = 0`) for these, allocating an `SR_CONVERTER` struct that is never used at perf-time (i-rate args are skipped in `useropcd_pass_by_copy`). Adding a type check before calling `src_init` for i-rate/b-type variables would avoid these unnecessary allocations. Not a correctness issue.

---

### Design observation: Converter deinitialization relies on contiguity

The cleanup in insert.c uses `else break` on the first NULL pointer to stop iterating the `cvt_in`/`cvt_out` arrays. This works now because `src_init` always returns a valid (non-NULL) converter for all supported types (bypass mode for non-k/a). But if `src_init` were ever to return NULL without error (e.g., for a future type), this would silently skip freeing later converters. The current code is correct but fragile.

vlazzarini:
1 and 2 are valid alright. The "inefficiency" maybe not, setting a unit for each argument is better than skipping and for little price, as we don't leave room for mistakes. I don't think the last comment applies, this code can handle any new types  via bypass, and I don't see where insert.c is involved, the code changes are all in udo.c and those names are used in that file. I'll check it anyway.

vlazzarini:
These should be fixed now - plus a couple of other minor bugs

vlazzarini:
Found a bug in the linear oversampling code - wrong deref levels. Fixed it, I think it all should be good now.

kunstmusik:
WIll review later today, but I wanted to mention an idea that might make this easier is to get the type system involved so that type definitions have resample methods per-type. That'd make it a little easier to deal with UDTs I think if they can tree walk down their members to resample. What do you think @vlazzarini?

vlazzarini:
I guess that's another valid way to go about this, but I think it may just move the complexity elsewhere, in the case of UDTs. 

Arrays are straightforward, single subtype, flat implementation, no issue. In my head UDTs are not that easy. Anyway, see what you think.

vlazzarini:
@kunstmusik could we move this forward? If the current approach is acceptable, we could go for it and then later on refactor if necessary.

hlolli (approved):

kunstmusik (approved):
Apologies for the delay, I'm away on a trip at the moment. I went through this a few days ago and it looked good to me for what it was, but ultimately I think we do need to involve the type system to make over/under sampling support reflected in any new types. That said, I think this moves the needle forward and correxrs things so we should merge now and address issues in another pr.

Issue #407, opened 2014-12-16
Add UGen API 


kunstmusik:
The UGen API should build upon work done in "Extending Aura with Csound Opcodes" from ICMC 2014.  The work should allow instantiation and running of opcodes on their own.  This work will allow host systems to use opcodes on their own outside of the Csound engine. The Csound engine should also be extended to allow for running of arbitrary UGen graphs.  This will allow users to assemble graphs at runtime without pre-compiling them into instrument instances.  

The Opcode list system should also be extended so that library users can inspect what opcodes are available, for use with generating or wrapping UGens to make them more idiomatic to the target language.

gogins:
---------- Forwarded message ----------
From: "Allen Downey" <downey@allendowney.com>
Date: Jan 14, 2015 6:41 PM
Subject: [music-dsp] I am working on a new DSP textbook using Python.
Comments are welcome!
To: <music-dsp@music.columbia.edu>
Cc:

I am developing a textbook for a computational (as opposed to mathematical)
approach to DSP, with emphasis on applications -- especially sound/music
processing.  People on this list might like this example from Chapter 9:

http://nbviewer.ipython.org/github/AllenDowney/ThinkDSP/blob/master/code/chap09preview.ipynb

I have a draft of the first 9 chapters, working on one or two more. I am
publishing excepts and the supporting IPython notebooks in my blog, here:

http://thinkdsp.blogspot.com

Of if you want to go straight to the book, it is here:

http://think-dsp.com

Comments (and corrections) are welcome!
--
dupswapdrop -- the music-dsp mailing list and website:
subscription info, FAQ, source code archive, list archive, book reviews,
dsp links
http://music.columbia.edu/cmc/music-dsp
http://music.columbia.edu/mailman/listinfo/music-dsp

kunstmusik:
@gogins: Was this meant to be emailed to the Csound User's mailing list?

gogins:
Yes, I forwarded it.

Best,
Mike

kunstmusik:
I'm deleting it and the above messages from this Github issue thread then.

kunstmusik:
Completed in #2498

PR #2498, opened 2026-02-25
Implementation of UGen API

kunstmusik:
This PR implements the UGen API (https://github.com/csound/csound/issues/407) to allow for using opcodes as individual Unit Generator objects that can be arbitrarily assembled into processing graphs. 

Initial work includes:

- Introduced a new API for creating and managing Csound Opcodes as unit generators (UGens) in `ugen.h`.
- Added functions for UGen factory creation, deletion, and opcode listing.
- Implemented argument handling for UGen inputs and outputs, including setting and getting values by pointer and by value.
- Developed a UGen graph API for managing connections between UGen instances.
- Created a comprehensive set of unit tests in `ugen_test.cpp` to validate the functionality of the UGen API, covering factory operations, UGen creation, argument handling, initialization, performance, and context management.
- Ensured memory layout integrity and stability during multiple cycles of UGen operations.

Currently has limitations on S and F types and needs review for API design and implementation.


vlazzarini:
A welcome development, I'd like the idea to be promoted. I can see this being useful in a context with glue languages, like Python and
JS. Maybe for a collection of webaudio nodes?

I had a quick glance at the API and it seems to have the things I'd expect, but I'll need to have a closer look. One stylistical note I should make, it departs from the norm we've been adopting, snake_case for internal functions and camelCase() for exported functions. It's not a big deal but I thought I should note it now.

gogins:
I think this is potentially extremely useful. If you had a C++ API you
could use SWIG or something else to expose it to Python, and you could use
Emscripten Embind to expose it to JavaScript. And of course just use it in
C++.


Michael Gogins
Irreducible Productions

rorywalsh:
This looks very impressive Steven. Can we already do things like:

```
    osc1 = factory.new_ugen("oscili", "a", "kkjo")
    osc2 = factory.new_ugen("oscili", "k", "kkjo")

    osc2.set_input_value(1, osc1)
```

kunstmusik:
> This looks very impressive Steven. Can we already do things like:
> 
> ```
>     osc1 = factory.new_ugen("oscili", "a", "kkjo")
>     osc2 = factory.new_ugen("oscili", "k", "kkjo")
> 
>     osc2.set_input_value(1, osc1)
> ```

It's a little more low-level to connect ugens together, one has to create a MYFLT and set it as as the output and input pointers. I've generated an example that demonstrates that in https://github.com/csound/csound/pull/2498/changes/e9564af672a060fccb5a86dcd55447c1f997672e

We could look at other API options whether at the C level or binding level (i.e., ctcsound)

kunstmusik:
@vlazzarini Thanks for pointing out the camelCase issue, I've updated the UGEN API to adhere to that in https://github.com/csound/csound/pull/2498/changes/cd4300e9fc7f2e5d89c91b568e0869e6b3a04d3a

Regarding bindings, ctcsound.py covers Python and is being used in the ugen_example.py file which is acting as somewhat of a test as well as guide to usage. WebAudio Csound will need to be updated to expose the ugen functions.  I don't think there's an active node.js binding for Csound 7, but whenever one is done it should be straightforward to do something like what is done in ctcsound.py.

kunstmusik:
FYI: I've updated the API to use a UGEN_VAR wrapper for intput and output values. This allows caching of var handles for reading/writing, but also works with convenience methods for setting/getting float/string values. With the use of these wrappers, it also makes it possible to wire up together opcodes that use data types like F-sigs. Let me know if anyone sees any issues with this API change.

I will be looking at updating WebAudio Csound to add this to the API and to generate some examples that use opcodes within a WebAudio graph.

vlazzarini:
That's great. This is what I meant when I said about making WebAudio nodes based on Csound opcodes. Nice about fsigs, opening up possibilities for spectral processors.

kunstmusik:
I've pushed an implementation of UGEN API and a new libcsound API in WebAudio csound to make using libcsound directly easier. There's an example project that shows using UGen API in an audioworklet in wasm/browser/examples/worklet-demo, as well as additional tests added for c, python, and javascript. 

One important optimization for WebAudio came up in string-pointer.js for memory and strings. 

@vlazzarini and @hlolli : I've moved this to ready for review

kunstmusik:
Sidenote: @rorywalsh could be interesting to make a VST using Csound Ugen API

Engine/ugen.c
```diff
@@ -22,225 +22,295 @@
...
+    ugen->insds = context->insds;
+    opds->insdshead = context->insds;
+    return true;
 }
 
+/* ============================================================
+ *  UGEN Creation / Destruction
+ * ============================================================ */
+
+static OENTRY* ugen_resolve_opcode(OENTRIES* entries,
```
vlazzarini: Would it be possible to refactor this to re-use similar code in csound_orc_semantics.c? Although they are different, they look similar enough. The concern is that they may diverge at some point. If we could have a basic resolve_opcode() function that is used here and there, then it would be easier to maintain.

Engine/ugen.c
```diff
@@ -249,124 +319,612 @@ static CONS_CELL* get_assignable_out_types(CSOUND* csound, char* intypes) {
...
+        return (*oentry->perf)(ugen->csound, ugen->opcodeMem);
+    }
+    return CSOUND_SUCCESS;
 }
 
+/* ============================================================
+ *  Opcode Listing API
+ * ============================================================ */
+
+int32_t csoundUgenListOpcodes(UGEN_FACTORY* factory,
```
vlazzarini, line 733: Similarly, there is the opcodelist API functions in Top/opcode.c. Should we refactor so that these all use the same basic code?

Engine/ugen.c
```diff
@@ -249,124 +319,612 @@ static CONS_CELL* get_assignable_out_types(CSOUND* csound, char* intypes) {
...
+}
+
+void csoundUgenFreeOpcodeList(UGEN_FACTORY* factory, UGEN_OPCODE_INFO* list) {
+    if (factory != NULL && list != NULL) {
+        CSOUND* csound = factory->csound;
+        csound->Free(csound, list);
+    }
+}
+
+bool csoundUgenFindOpcode(UGEN_FACTORY* factory, const char* opname,
```
vlazzarini, line 813: Same here, we have FindOpcode()-like functions elsewhere in the code. It would be good to see if they can all share the same basic functionality.

vlazzarini (commented):
From the C side, it looks good, but I have some questions about code reuse, mainly to help maintenance.

wasm/browser/examples/worklet-demo/src/processor.js
```diff
@@ -0,0 +1,132 @@
...
+
+      const csound = cs.csoundCreate();
+      cs.csoundSetOption(csound, "-d");
+      cs.csoundSetOption(csound, "-n");
+      cs.csoundSetOption(csound, "--nchnls=1");
+      cs.csoundSetOption(csound, "--0dbfs=1");
+      cs.csoundSetOption(csound, `--sr=${this.sr}`);
+      cs.csoundSetOption(csound, `--ksmps=${this.ksmps}`);
+      // Minimal orchestra so csoundStart() succeeds
+      cs.csoundCompileOrc(csound, "instr 1\n  out oscili(p4,p5)\nendin");
```
vlazzarini: Interesting: do we need a fully compiled and started Csound instance? Also if we need a running Csound instance, you should be able to start it with empty code (also using `--daemon`).

Engine/ugen.c
kunstmusik: Good points about these three functions overlapping. I have to review why I went this route years ago (pretty sure the Csound/Aura work was done while we were getting to the finish line for CS6, maybe there was a reason back then that doesn't make sense now). I'll review these.

wasm/browser/examples/worklet-demo/src/processor.js
kunstmusik: I think we do to have sample rate and ksmps set before using the factory, but I will double check about both the requirement and modifying example code to --daemon (if the above is necessary)

kunstmusik: AI Assessment:

```
Yes, a started Csound instance is required — find_opcode2 needs the opcode table, and insds_init_from_csound needs sr/ksmps to be set.
No, a full orchestra isn't needed — --daemon + sr/ksmps overrides via csoundSetOption may suffice, but this needs testing. The current csoundCompileOrc with a dummy instr is the safest current approach.
```

I'll see about revising tests to simplify

vlazzarini: interesting. Just thinking here, a program can use Csound code and the ugen API side by side with the same Csound instance. Like, to add external processing before output.

Engine/ugen.c
kunstmusik: I removed the unnecessary code and reused existing functions.

wasm/browser/examples/worklet-demo/src/processor.js
kunstmusik: So I audited with AI and got the following:

```
Assessment
Yes, a started Csound instance is required — find_opcode2 needs the opcode table, and insds_init_from_csound needs sr/ksmps to be set.
No, a full orchestra isn't needed — --daemon + sr/ksmps overrides via csoundSetOption may suffice, but this needs testing. The current csoundCompileOrc with a dummy instr is the safest current approach.
```

However, I found we don't need either --daemon or the dummy instrument code, calling csoundStart() was sufficient.

kunstmusik:
@vlazzarini I think I've addressed your comments, let me know if there's anything else you find.

kunstmusik:
Note: I see a build issue with cachix for the WASM build here, I've messaged @hlolli for some assistance in diagnosing that. Otherwise, prior to merging this PR, we should do a release of both csound/wasm-bin and csound/browser NPM packages.

vlazzarini:
Ready to approve this, but do you need to fix the wasm build first?

vlazzarini (dismissed):

kunstmusik:
I've released new NPM packages and committed the change. Could I get another approval @vlazzarini or @hlolli?

vlazzarini (approved):

PR #2503, opened 2026-03-03
Array channels

vlazzarini:
Extending array bus channel support to any array subtype, adding tests.


rorywalsh (approved):
Thanks for this Victor.

Issue #2340, opened 2025-10-31
Csound Module System


kunstmusik:
(Summary below generated after brainstorming iteration together with AI)

# Csound Module System Proposal (Revised)

## Overview

I'm proposing a **file-based module system** for Csound, inspired by Python's module system. This would enable better code organization, reusable libraries, and namespace management while maintaining full backward compatibility.

### 1. File-Based Modules
Each `.orc` file is implicitly a module - no explicit declarations needed.

### 2. Python-Style Import Syntax
Clean, familiar import patterns without object destructuring:

```csound
import lib.oscillators                    ; module import
import lib.oscillators as osc             ; module import with alias
from lib.oscillators import MyOsc         ; specific import
from lib.oscillators import MyOsc as Osc  ; aliased import
from lib.oscillators import *             ; wildcard import
```

**Why Python-style?** JavaScript's `import { x }` syntax relies on object destructuring, which doesn't exist in Csound. Python's syntax is simpler and more appropriate.

### 3. Public-by-Default Access
All top-level variables and UDOs are accessible when a module is imported:

```csound
; lib/synth.orc

; All variables are accessible via import
iSampleRate = 44100
kModDepth init 0.5
iDefaultFreq = 440
kGain init 0.5

opcode MySynth, a, k
  kfreq xin
  ; Can access all module variables
  a1 oscil kGain, kfreq
  xout a1
endop

; All UDOs are accessible when module is imported
```

### 4. Singleton Behavior
Files are compiled once and cached - subsequent imports reuse the compiled module (like Python/Node.js).

### 5. Backward Compatible
Existing code works unchanged. Files without imports operate in the global namespace as they do today.

### Example 1: Reusable Oscillator Library

**lib/oscillators.orc:**
```csound
; Module configuration
iDefaultAmp = 0.5
kAmpScale init 1.0

opcode SineOsc, a, k
  kfreq xin
  a1 oscil iDefaultAmp * kAmpScale, kfreq
  xout a1
endop

opcode SawOsc, a, k
  kfreq xin
  a1 vco2 iDefaultAmp * kAmpScale, kfreq, 0
  xout a1
endop
```

**main.orc:**
```csound
from lib.oscillators import SineOsc, SawOsc
import lib.oscillators as osc

instr 1
  ; Modify any module variable
  osc.kAmpScale = 0.8
  osc.iDefaultAmp = 0.7

  ; Use imported opcodes
  a1 SineOsc 440
  a2 SawOsc 220
[...]

PR #2504, opened 2026-03-03
Add #includelib preprocessor directive for UDO libraries

PasqualeMainolfi:
I have always thought it would be interesting to allow sharing of UDO libraries even with users who may not be able to write a full opcode from scratch. For this reason, I propose introducing a new preprocessor directive, `#includelib`, which enables including installable UDO libraries.
The principle of operation is straightforward: libraries can consist of folders with subfolders and a metadata file. 

The `#includelib` directive supports several levels of inclusion: 

- Include the entire library:
```code 
#includelib "test_lib"
```

- Include a submodule or subfolder:
```code 
#includelib "test_lib.mod1"
```

- Include a single UDO from the installed library:
```code 
#includelib "test_lib.mod1.osc"
```
(The extension can be omitted; the system will automatically check for osc.udo)

As with plugins, libraries have a system directory `CS_DEFAULT_UDODIR`, which can also be overridden by the user, following the same principles as plugin directories.
For the problem of sharing, I believe it makes perfect sense to leverage `risset` and also provide the possibility to install libraries.

**NOTE**: This PR contains a draft implementation of the feature. If the idea is well-received, I am happy to finalize the implementation and refine any details based on feedback.


rorywalsh:
Hi @PasqualeMainolfi have you seen the proposal from @kunstmusik regarding a module system? There seems to be some crossover between it and your proposal:

https://github.com/csound/csound/issues/2340

PasqualeMainolfi:
@rorywalsh No, I hadn’t seen that... had just started working on it.

kunstmusik:
I suggest to abandon this as the module system is much further along.

PasqualeMainolfi:
@kunstmusik @rorywalsh I agree with you. I hadn’t seen your proposal. My idea was to create something extremely simple, so that even less experienced users could easily share their libraries without having to package or bundle anything.
That said, I don’t think we need to choose one approach over the other. We could work on both: a simple solution focused on ease of sharing, and, in parallel, a more advanced and structured architecture for module management. 
Let me know!

kunstmusik:
Disagree here, modules aren't simple and we don't need multiple ways to import modules.

rorywalsh:
Personally, I’d rather see a single system here instead of two. If the module system Steven is proposing doesn’t tick the "ease of use" box, I think it would be better to focus on way of improving its accessibility to end-users rather than implementing an entirely new system.

vlazzarini:
I also think the module system is the way to go. We have to be careful with multiple overlapping implementations of similar ideas, these are a maintenance nightmare for us managing system development

PasqualeMainolfi:
Well. A package manager for Csound is not particularly complex, certainly less so than one for Python. An intelligent and advanced solution could be developed in Rust, inspired by [uv](https://github.com/astral-sh/uv.git), with a deterministic resolver, a lockfile that fixes versions and checksums, and a centralized global cache that downloads each package only once, copying it or linking it via symlink into the project’s `cs_modules/` directory.
The system should include an abstract registry compatible with both Git and JSON, and a complete CLI with commands such as `init` to initialize projects, `add` to install modules, and `sync` to apply the lockfile versions.
The connection between Csound and the modules would be handled through an entry point, acting as a bridge between the user’s directive and the module. I believe this can be implemented without major difficulties. 
@vlazzarini you’re right. If you want, we can wrap this up and move on to Steven’s pr.

vlazzarini:
I think this PR has been superseded by the upcoming module work. I am closing it now, since we've had no further activity for a while.

PR #2507, opened 2026-03-07
improvements to prints and fix to android release build

vlazzarini:
`prints` had a number of shortcomings

- it did not add a newline, which was always an annoyance and out of line with other printing opcodes.
- if a string var or parameter was given as the single parameter, it printed nothing (pretty awful behaviour when we're debugging!)

This PR fixes both issues.


vlazzarini:
Also fixes an issue with NaNs in the android release build.

.github/workflows/csound_builds.yml
```diff
@@ -875,7 +875,7 @@ jobs:
           sh release.sh
 
       - uses: actions/upload-artifact@v4
-        if: github.ref == 'refs/heads/develop'
+        #if: github.ref == 'refs/heads/develop'
```
hlolli, line 878: this would then run on any pull request targeting develop branch, going to be a lot of (unnecessary?) uploads?

vlazzarini: you can reinstate it, but it's useful when we need to check the binary packages as was the case here.

rorywalsh: So are we leaving this in for now?

vlazzarini: up to you

rorywalsh (approved):
I'm easy, I'll merge as it stands.

PR #2510, opened 2026-03-12
cmake fixes

vlazzarini:
There was a problem with the cmake script to build with plugins (-DBUILD_PLUGINS=ON) instead of builtin opcodes. Firstly, the static library build ignored this option and built opcodes into the library (which defeated the purpose). Secondly, the token BUILD_PLUGINS was not being added to the library build and so the code in csmodule.c was looking for the the static module functions, which do not exist in this case.

With this now we can build static libs that are much smaller, and then we may be able to link specific opcodes when using it instead of the complete set. On the arm baremetal build, the reduction in size is substantial (from 3.8 to 1.8MB).

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -102,7 +102,7 @@ jobs:
           sudo apt-get install cmake libjack-dev libpulse-dev ladspa-sdk dssi-dev autoconf libtool
 
       - name: Configure Build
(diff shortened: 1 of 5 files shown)


vlazzarini:
Also added csoundAppendOpcodes() to the API and extended the LINKAGE_BUILTIN(name) macro to declare a pointer to the localops array and a length constant. This allows static-link opcode modules in place of plugins for platforms with no dynamic linking.

amanjagwani (approved):

PR #2513, opened 2026-03-14
User-defined types support for software bus channels

### LINKED ISSUE #691: Channels and User-Defined Types
This issue is to track work in addition to #444 for Csound 7 in the exposing of User-Defined data types through channels.

vlazzarini:
This PR adds support for struct variables to the software bus:

- chnset, chnget set and get channel data at init and perf-time
- chnseti, chngeti  work at i-time only
- chnsetk, chngetk work at perf-time only

Example:
```
struct Test val1:k, val2:i

instr 1
var:Test init 0, 2
var.val1 = 3
chnset var, "test"
var2:Test chnget "test"
print var2.val2
printk2 var2.val1
endin
```

closes #691


kunstmusik:
Just a note that I've been looking at this together with AI and there were some edge cases and things. I'm working on some changes and added tests.

kunstmusik:
I've added some fixes for initialization and extended the tests. I've also added and exitnowk so that we can exit at perf-time, and also added a fix for struct initializer init opcode entries to clear the deprecated flag. 

I think these changes could use a review from @vlazzarini @hlolli

vlazzarini:
@kunstmusik thanks, it's looking good, ~just wondering if the test for arg type and channel type should always happen. Left a comment there.~ I misread the code in my phone! All good.

This code should work for any new type (not only udts) should it not? If so we could test that.

EDIT: it does work for any generic type. I've added a couple of tests for that. I also added a mismatch test for UDTs, which seems good.

vlazzarini:
Question for another PR: how can we better support these generic channels for external use via the API? Currently we have dedicated support for ksigs, asigs, fsigs, strings and arrays. Accessing data UDTs and other types is possible via the generic get channel pointer. @kunstmusik do you see a way in which this can be improved?

kunstmusik (approved):
@vlazzarini I added some comments to the new tests to clarify what's being tested for posterity. I definitely think having structs be usable from the API would be a good thing and also think it should be done in a separate PR. (Will add an issue in a moment.) Otherwise this looks good to me, will approve now.

Issue #2092, opened 2025-02-05
Rework data type creation: use "new" opcode or "new" keyword or generate constructor opcodes


kunstmusik:
From reviews of #2047, it came up to look at using either "new" as an opcode in place of "create", or introduce "new" as a keyword. Also, for consideration is the direct use of user-defined struct name or instr names as constructor functions. 

Instead of:

```
ival:MyStruct init 1, 2, 3
```

we could have:

```
ival:MyStruct = new MyStruct(1,2,3,)
;; or
ival:MyStruct = new(MyStruct, 1,2,3)
```

and that could extend to Instrument instances such as:

```
instance:Instr = new MyInstr1()
;; or
instance:Instr = new(MyInstr1)
```

alternatively, we could remove new/create altogether and have:

```
ival:MyStruct = MyStruct(1,2,3)
instance:Instr = MyInstr1()
```

Considering UDOs that work with structs or Instr instances as args, we could imagine inline anonymous values such as:

```
perform(new MyInstr1(), 1,2,3)
;; or
perform(MyInstr1(), 1,2,3)

my_udo(1,2, new MyStruct(1,2,3))
;; or
my_udo(1,2, MyStruct(1,2,3))
```

Note: the original impelmentation of subinstruments in Csound 4 *did* allow for calling instruments by name as if it was an opcode. That implementation got lost over time but there's some prior work to reference.

vlazzarini:
As decided in #2514, we won't go ahead with this.

PR #2514, opened 2026-03-15
Operator new

### LINKED ISSUE #2505: initialiasing empty struct
@kunstmusik @hlolli, quick request for you both. Could we get a way to initialise an empty struct?

When working with large structs, it’s a bit of a headache to remember the exact member order. It would be much cleaner if we could initialise the container first and then set members individually, like this:

```
var@global:Test init 0
var.val1 init "Hello"
var.val2 init "Goodbye"
var.val3 init 4
var.val6 init 7
var.val5 init "something else"
```

This would definitely save some mental load on those long structs!

vlazzarini:
This PR introduces an operator `new` to the language. The syntax is

```
var new type
```

or

```
var new type(init-list)
```

The first version creates an empty variable of a given `type` - not initialised. The second is an alias for `init()`.

For example,

```
struct Test var1:i, var2:i

instr 1
val new Test
val.var1 = 1
val2 new Test(1,2)
print val.var1
print val2.var1
endin
```

Supports any type. For example, we can have

```
ins new InstrDef({{
print 4
}})
schedule(ins, 0, 1)
```

and

```
opc new Opcode(oscil)
opcodeinfo(opc)
```

Closes #2505 and #2092

EDIT: support for `new` in expressions has not been added, but may be implemented later. At the moment, only statements are supported.


kunstmusik:
Something feels a little off about this, as it seems to mix up some concepts in my head. Reminds me a little bit of the way objective-c has its alloc/init system for object creation, but here new is doing multiple things. I know this is in regards to Rory's request about allocating objects in an uninitialized state, but I'm wondering if there isn't a different path to take that wont' introduce 'new' here. Need to have a think about this, but brainstorming out loud:

1. Struct constructors: having the type name of the struct be registered as a constructor in addition to init
2. from here, add default values for arguments, so something like:

```
;; multiline-struct
struct MyStruct 
  var1:i ;; required arg
  var2:i, var3:k ;; required args again, but permit n-number of args separated by comma
  var4:i = 0 ;; optional arg
endstruct

;;usage
struct1 = MyStruct(0,2,3)
;; keyword args, compile-time checked that all required args are filled in
struct1 = MyStruct(var1 = 0, var2 = 0, var3 = 0, var4 = 3) 

struct AllOptional
  var1:i = 0
  var2:k = 0
endstruct

struct2 = AllOptional()
struct3 = AllOptional(1)
struct4 = AllOptional(0, var2 = 3)
```

vlazzarini:
This is partly in answer to Rory and in answer to the other open issue. I thought we wanted a new operator in the language. I am following more or less what was proposed in that issue #2092 

New does two things here: 1) allocates, and optionally 2) initialises. It is a bit like in C++. 

Your proposal is ok but it is just for struct initialisation. That's fine but then we have decide that we do not want an operator new in the language and close that issue.

vlazzarini:
@kunstmusik @rorywalsh @hlolli 
I guess we need to make a decision on this. The main question is whether we want a `new` operator that takes a type and creates a variable. This is what was proposed in #2092.

If we don't, I will close this PR and that issue and call it quits. #2505 can be resolved with an overload of `init` taking no parameters. It would be good to move on as we have a few other PRs and issues to consider.

hlolli:
I would vote for exploring init taking no parameters and default to either 0 or that which was declared as default. And fully explore that path before adding 'new'. I don't know if that's the best path. If possible to avoid passing around uninitialised variables would be user facing simplest without nulls existing. But I know I made an exception to that already by allowing recursive declared structs where the input struct needs to be uninitialised for evaluation but hopefully not in compilation.

kunstmusik:
I'm of the mindset that the implementation of new here is not the best path forward for the problem of allocating/initializing data types in convenient ways. I don't think having unitialized data should be allowed. We might be better off investigating optional params and multiple constructors for data. Suggest we close this as good exploratory work that we can reference later, and leave the branch in repo so that we can review it again if necessary.

vlazzarini:
Ok closing this and the issue, since we're not doing a new operator.

PR #2516, opened 2026-03-18
Add some optimizations to help generate SIMD code

gesellkammer:
Tested on x86-64 with avx2 and macos arm with -mcpu=apple-m1


vlazzarini:
do we need changes to CMake or its options for these to compile to fast code?

gesellkammer:
In x86-64 (linux and windows) avx2 can be safely added. When building for macos arm64, -mcpu=apple-m1 is a good lowest target which enables all simd optimizations. For external plugins I have this in the cmake file but I could make a PR to adapt that to csound's cmake:

```cmake

# set optimization flags
if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_COMPILER_IS_CLANG)
    add_definitions(-fvisibility=hidden)
    include(CheckCCompilerFlag)
    include(CheckCXXCompilerFlag)
    
    if(LINUX AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|i[3-6]86|AMD64")
        add_compile_options(-mavx2)   
   endif()

    set(CMAKE_C_FLAGS_RELEASE "-O3")
endif()

# AVX2 for Windows (MSVC)
if(MSVC AND CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64|AMD64|x86")
    check_cxx_compiler_flag(/arch:AVX2 MSVC_HAS_AVX2)
    if(MSVC_HAS_AVX2)
        add_compile_options(/arch:AVX2)
        message(STATUS "AVX2 enabled for Windows (MSVC, x86)")
    endif()
endif()

if(APPLE AND CMAKE_SYSTEM_PROCESSOR MATCHES "arm64")
    add_compile_options(-O3 -mcpu=apple-m1)
endif()
```

vlazzarini:
can you add the CMake options here to this PR?

NB: probably we don't need the`-O3` as CMake adds it to the Release config (I think)

gesellkammer:
I updated the PR to add the changes to cmake.

vlazzarini (approved):

PR #2518, opened 2026-03-19
added memory alignment option

### LINKED ISSUE #2517: Add AuxAlloc and Calloc versions with alignment
Some SIMD autovectorization optimizations can only be ensured if memory is aligned at allocation. At the moment there is no way to ask csound to allocate memory with a specific alignment.

vlazzarini:
Closes #2517 - a CMake option is added to align all memory to MYFLT size.


gesellkammer:
Notice that the current solution aligns to the size of MYFLT (64 **bits** = 8 **bytes**) but for some algorithms to autovectorize to avx2 and avx512 we need an alignment of 32 **bytes**.

gesellkammer (dismissed):

vlazzarini:
@gesellkammer I've implemented CallocAligned() with aligned_alloc(). Not sure if it is what you need, note that it requires that the size is a multiple of the alignment, otherwise it will crash. Alternatively we could use a macro to set the size to make the memory be a multiple of the alignment  - perhaps that's better?

EDIT: I've set it now that memory size is always a multiple of alignment - stops crashing on undefined behaviour.

vlazzarini:
@gesellkammer need another approval

gesellkammer (approved):

gesellkammer:
This looks great, thanks.

PR #2519, opened 2026-03-21
added per-k-cycle callback support for Csound debugger in WASM

ervoLB:
Adds a non-stopping per-k-cycle callback hook to Csound's performance loop, exposed both in the C API and in the WASM/browser API.

C API (csdebug.h, Top):

- New kcycle_cb_t typedef
- csoundSetKcycleCallback(csound, cb, userdata) and csoundRemoveKcycleCallback(csound)
- Hooked into both kperf (normal) and kperf_debug (debug mode) loops
- Does not require csoundDebuggerInit() — works in any performance mode
- Inside the callback, csoundDebugGetInstrInstances() / csoundDebugGetVariables() can be used for real-time variable inspection

WASM/browser API (wasm):

- csoundSetKcycleCallbackWasi exposed as a WASM export
- "kcycle" event added to the JS event system
- enableKcycleEvents() convenience method added to all main-thread variants (SAB, vanilla, single-thread worklet)
- k-cycle events propagated back to the main thread via messagePort

### DIFF
--- a/wasm/browser/src/events.js
+++ b/wasm/browser/src/events.js
@@ -65,6 +65,7 @@ export class PublicEventAPI {
       "renderEnded",
       "onAudioNodeCreated",
       "message",
+      "kcycle",
     ]);
   }
 
@@ -122,6 +123,10 @@ export class PublicEventAPI {
 
   triggerMessage({ log }) {
     this.eventEmitter.emit("message", log);
+  }
+
+  triggerKcycle() {
+    this.eventEmitter.emit("kcycle");
   }
 
   decorateAPI(exportApi) {
--- a/wasm/browser/src/events.js
+++ b/wasm/browser/src/events.js
@@ -65,7 +65,7 @@ export class PublicEventAPI {
       "renderEnded",
       "onAudioNodeCreated",
       "message",
-      "kcycle",
+      "debugCallback",
     ]);
   }
 
@@ -125,8 +125,8 @@ export class PublicEventAPI {
     this.eventEmitter.emit("message", log);
   }
 
-  triggerKcycle() {
-    this.eventEmitter.emit("kcycle");
+  triggerDebugCallback() {
+    this.eventEmitter.emit("debugCallback");
   }
 
   decorateAPI(exportApi) {
--- a/wasm/browser/src/mains/sab.main.js
+++ b/wasm/browser/src/mains/sab.main.js
@@ -600,6 +600,12 @@ class SharedArrayBufferMainThread {
         }
       }
     }
+
+    this.exportApi["enableDebugCallback"] = async () => {
+      const fn = this.exportApi["setDebugCallbackWasi"];
+      return typeof fn === "function" ? await fn() : -1;
+    };
+
     log(`PUBLIC API Generated and stored`)();
   }
 }
--- a/wasm/src/csound_wasm.c
+++ b/wasm/src/csound_wasm.c
@@ -3,6 +3,7 @@
 #include "ugen.h"
 #include <ctype.h>
 #include <limits.h>
+#include "csdebug.h"
 #include <stdarg.h>
 #include <stdint.h>
 #include <stdio.h>
@@ -109,6 +110,21 @@ void csoundWasiCMessageCallback(CSOUND *csound, int attr, const char *format, va
 __attribute__((used))
 void __wasi_js_csoundSetMessageStringCallback() {
   return csoundSetDefaultMessageCallback(&csoundWasiCMessageCallback);
+}
+
+/* ---- per-k-cycle debug callback: C → JS ---- */
+
+/* JS import: signals JS that a k-cycle just completed (debug mode) */
+void csoundWasiJsDebugCallback(void) __attribute__((
+     used,
+    __import_module__("env"),
+    __import_name__("csoundWasiJsDebugCallback")
+));
+
+static void csoundWasiCDebugCallback(CSOUND *csound, void *userdata) {
+  (void)csound;
+  (void)userdata;
+  (* csoundWasiJsDebugCallback)();
 }
 
 // copy/paste from upstream csound-emscripten
@@ -224,6 +240,12 @@ int csoundResetWasi(CSOUND *csound) {
   csoundReset(csound);
   csoundSetMidiCallbacks(csound);
   return CSOUND_SUCCESS;
+}
+
+__attribute__((used))
+void csoundSetDebugCallbackWasi(CSOUND *csound) {
+  csoundDebuggerInit(csound);
+  csoundSetDebugCallback(csound, &csoundWasiCDebugCallback, NULL);
 }
 
 // Keep a stable non-null pointer for JS string reads.
--- a/wasm/src/csound_wasm.c
+++ b/wasm/src/csound_wasm.c
@@ -244,6 +244,8 @@ int csoundResetWasi(CSOUND *csound) {
 
 __attribute__((used))
 void csoundSetDebugCallbackWasi(CSOUND *csound) {
+  /* csoundDebuggerInit is idempotent — safe to call even if already
+     initialized or if called multiple times. */
   csoundDebuggerInit(csound);
   csoundSetDebugCallback(csound, &csoundWasiCDebugCallback, NULL);
 }
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -1064,6 +1064,8 @@ static const CSOUND cenviron_ = {
   0,              /* modules loaded */
   -1,             /* audio system sr */
   0,              /* csdebug_data */
+  NULL,           /* debug_cb */
+  NULL,           /* debug_cb_data */
   0,              /* which score parser */
   0,              /* print_version */
   1,              /* inZero */
--- a/Top/csound_debug.c
+++ b/Top/csound_debug.c
@@ -660,8 +660,28 @@ int32_t kperf_debug(CSOUND *csound) {
     }
   }
 
+  /* fire per-k-cycle debug callback if set (only when not stopped at a breakpoint) */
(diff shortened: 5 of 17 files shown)


Top/csound_perf.c
```diff
@@ -386,6 +386,10 @@ int32_t kperf(CSOUND *csound) {
       }
     }
   }
+  /* fire per-k-cycle callback if set */
```
vlazzarini: I would prefer if we did not add callbacks to kperf(). We have deliberately removed them in 7.0 to streamline the code. I think it's ok for kperf_debug() but not here.

vlazzarini (changes requested):

ervoL:
@vlazzarini  Done — removed the callback from kperf(), keeping it only in kperf_debug() as suggested.  This is my first PR so tell me if I should do something different

vlazzarini:
> @vlazzarini Done — removed the callback from kperf(), keeping it only in kperf_debug() as suggested. This is my first PR so tell me if I should do something different

Thanks. It looks good from the C csound library code perspective. I'll let @hlolli and @kunstmusik have a look at the wasm side.

vlazzarini (approved):

vlazzarini:
I am approving this from my end, but still need @hlolli to look over the wasm part.

vlazzarini:
@hlolli could you review the wasm part so we can move this onwards?

hlolli:
I really need to start using github notifications :D thanks for pinging me. Yes taking a look now!

hlolli:
Ok, I have question: why we need to implement this on C level.
Because the audio thread is a worker model. And to trigger a callback will (sadly depends how you look at it) be bottle necked by the worker to main thread signaling. Even if we get better performance by running a callback in wasm, the callback trigger of a foreign function, which must exist on the main thread, will always be as accurate as the value of ksmps. For SharedArrayBuffers the story is a little better, where events can be sent and received a little bit faster.

My suggestion here is, and this is something I was stunned I hadn't implemented yet. Maybe I did and I removed it, I forgot, but it makes no sense we don't have it. Is to implement a method to our EventEmitter. Event.on("performKsmps" or just simply Event.on("perform", that gets triggered every time immedietly before libraryCsound.csoundPerformKsmps is called. This would need to go trough the postMessage or `messagePort.post({ kcycle: true });` (not sure post will succeed?) like you did.

This way we can achieve the same callback accuracy as we see in your PR.

The rule of thumb is to make it work first and optimise second, but I'd be a little demanding on to try as hard to optimise. Because as I write this, I believe I removed this feature after many back and forths at some time in the past about performance bottle-necks. Anyway so. If there isn't any function subscribed to "perform" event, it's not emitted and the message channel doesn't send the messages. This way for minimal csound playbacks, they wont get penalized if they don't use performance callbacks. This may require sending a separate message to the workers to tell them to enable emitting performance events to the message channel (or the SAB Buffer in SharedArrayBuffer's case).

For now, I don't approve except I get convinced that this is better solved in C than with just the js wrapper. I know it's possible for me to get convinced, knowing that there could be more accurate timers and granularity. And this could make sense in more native wasm engines than in browsers, which we intend to support later.

vlazzarini:
I guess it is already implemented in C but just when you run the debugger. I don't know if we need anything extra in JS. @ervoLB can you follow this up with @hlolli and decide the best route forward. Also if we could have a use case and a test for it, it will strengthen the PR.

NB: Also now that I think of it, you should call the C callback something like DebugCallback() to make it explicit as a debugging tool. CsoundSetDebugCallback().

We don't need it otherwise in C/C++ because we always have to call PerformKsmps() (we removed the perform() interface that ran a loop). The only exception is if we use CsoundPerformanceThread() but that has support for callback insertion.

hlolli:
Thanks, that makes sense. I think I was mixing two separate concerns in my earlier comment.

For the browser API side, I’m still not convinced we need a general C-level per-ksmps callback just to emit a JS event. In the worker wrappers we already control the `csoundPerformKsmps()` call sites, so if the goal is something like `csound.on("perform")` or `csound.on("kcycle")`, I’d rather handle that in the JS worker layer. We could also make it lazy and only enable the `postMessage` / SAB signaling when a listener is actually registered.

That said, if the goal is specifically debugger support, then I agree that having something on the C side can make sense, as long as it stays clearly scoped to that use case. In that case, I’d suggest:

* keeping the hook in `kperf_debug()` only, not normal `kperf()`
* naming it explicitly as a debugger callback, for example `csoundSetDebugCallback()` or similar
* making sure the WASM API initializes or uses debugger mode if it depends on `kperf_debug()`
* adding a small use case or test showing that the callback is actually used for debugger inspection

So my preferred middle ground would be to treat the C callback as a debug-only facility, and avoid exposing it as a general browser performance event mechanism.

If we want a browser-level `perform` / `kcycle` event later, I think that should be implemented separately in the JS wrapper around `performKsmps()`.

ervoL:
Thanks for the feedback, I'll make those changes (rename to csoundSetDebugCallback, auto-init the debugger in the WASM wrapper, add a test).

Just to give some context on the motivation: I'm building a web-based Csound debugger as part of my app papad.dev. The per-k-cycle callback is the foundation that makes real-time variable inspection in the browser possible — being able to see live instrument state update in the UI every k-cycle. Without it there was no clean way to hook into the debugger at the right moment from the WASM side.

On top of what's in this PR, I've also been working on further extensions to support UDO debugging, which I'll propose as a follow-up PR once this one is finished.

And of course i agree that a separate JS perform event would be a welcome addition for general browser use. If you want i can put it as part of my next PR, just let me know.

vlazzarini:
It's great to see the debugger being used outside just C/C++ context and see new developments for it. I think what @hlolli suggested/summarised is the right way to go. Having in the C API then wrapping in JS follows the rest of the browser implementation. If there are other bits of the debugger part of the API that are not yet exposed, we should also fix that.

ervoL:
Ok, ive made (i think) all requested changes:

C:

- Renamed csoundSetKcycleCallback → csoundSetDebugCallback and csoundRemoveKcycleCallback → csoundRemoveDebugCallback throughout (csdebug.h, csoundCore.h, csound_debug.c, csound.c)
- Callback remains in kperf_debug() only — not in normal kperf()

WASM/JS:

- Renamed WASM wrapper to csoundSetDebugCallbackWasi and updated needed JS bindings accordingly
- csoundSetDebugCallbackWasi now auto-calls csoundDebuggerInit() so the browser user doesn't have to know about that requirement

Tests:

- Added csound_debug_callback_test.cpp 

I rebuilt the WASM and tested it in my web-based debugger in papad.dev and everything seems to work fine.

And as i mentioned before, ive already implemented some support for debugging UDOs and i plan to make a new PR for that once this one is closed.

vlazzarini:
It's looking good from my perspective; @hlolli I'll leave the last word to you.

wasm/src/csound_wasm.c
```diff
@@ -226,6 +242,12 @@ int csoundResetWasi(CSOUND *csound) {
   return CSOUND_SUCCESS;
 }
 
+__attribute__((used))
+void csoundSetDebugCallbackWasi(CSOUND *csound) {
+  csoundDebuggerInit(csound);
```
hlolli, line 249: This is much closer, thanks. One thing I’m still wondering about: `csoundDebuggerInit()` allocates debugger state unconditionally and the header says it should be called before performance starts. Could this leak/reinitialize state if `enableDebugCallback()` is called more than once, or if it is called after performance has already started? Maybe this should be guarded/idempotent, or the browser API should enforce that it is enabled before start.

hlolli (commented):
Thanks for the updates, this is much closer to what I had in mind. I left a few small questions mostly around the browser-side API/lifecycle details.

wasm/browser/src/events.js
```diff
@@ -124,6 +125,10 @@ export class PublicEventAPI {
     this.eventEmitter.emit("message", log);
   }
 
+  triggerKcycle() {
+    this.eventEmitter.emit("kcycle");
```
hlolli: Small API naming question: since this now depends on the debugger path, should this event name be debug-scoped too, e.g. `debugKcycle` or `debugCallback`? `kcycle` reads a bit like a general perform/k-cycle event, and I’d like to avoid confusing it with a possible future general JS perform event.

wasm/browser/src/mains/sab.main.js
```diff
@@ -600,6 +600,12 @@ class SharedArrayBufferMainThread {
         }
       }
     }
+
+    this.exportApi["enableDebugCallback"] = async () => {
```
hlolli, line 604: If this is becoming public browser API, could we also add it to `wasm/browser/index.d.ts`? Same for the new event name, whichever one we settle on, so TypeScript users see it.

ervoL:
Thanks @hlolli for the heads-up. I updated:

1. Idempotent csoundDebuggerInit — made csoundDebuggerInit() itself idempotent in csound_debug.c (early return if already initialized), rather than guarding it in the WASM wrapper. This is the correct place since csound_wasm.c only has access to the opaque CSOUND* type and can't inspect internal fields. csoundSetDebugCallbackWasi now just calls it unconditionally.

2. Event renamed to "debugCallback" — renamed the JS event from "kcycle" to "debugCallback" across events.js, messages.main.js, and module.js. Subscription now looks like: csound.on("debugCallback", () => { ... })

3. TypeScript declarations added — added enableDebugCallback(): Promise<void> to CsoundObj and "debugCallback" to the PublicEvents type in index.d.ts, with documentation comments.

I've rebuilt the WASM and re-tested in papad.dev — still working correctly.

hlolli (approved):
thanks for your patience and thank you for contributing :)

ervoL:
great! thanks! im more then happy to contribute :beers:

PR #2526, opened 2026-03-29
Software bus support for CS_VARIABLE 

### LINKED ISSUE #2524: Add struct support to Csound API
In #2513, support was added for structs in channels. Following up from that, I've created this issue so that we do additional work to add API methods for working with structs. I think one path we might try is to actually make channels a little more generic and return CS_VAR's, and have API work around working with CS_VAR's and types. That should allow us to expose any variable type and, if there are custom data types, at least expose it as a CS_VAR that the calling code might know what to do with. For standard signal types and structs, the Csound API should have methods for interrogating var's and retrieving/setting values.

vlazzarini:
This PR refactors the channel data structure to support the use of CS_VARIABLE as a means of passing data to/from hosts.

Main Changes:
  - CHENTRY now includes a `CS_VARIABLE *` member
  - API functions `csoundGetChannel()` and `csoundSetChannel()` to access channel data as `CS_VAR_MEM *`.

Also added unit test for the new API functions.

Closes #2524


vlazzarini:
@kunstmusik @hlolli - this is ready for review

vlazzarini:
I found that the generic channel opcodes can replace all of the specific opcodes, including the new array channel opcodes. In fact, they were above the array ones in the entry, so all tests were using these already. For the moment, I moved them down the list and made sure the array-specific code is working correctly (this covers only 1-dim though).

I am wondering whether we should remove the array-specific functions and use only the generic code. It may be easier to maintain.

kunstmusik:
### Code review

Found 4 issues:

1. **NULL pointer dereference before NULL check in `alloc_channel`** (CLAUDE.md says "Memory management and thread safety")

`pp->var->memBlock->varType` is written on line 447 before the NULL check on line 448. If `Calloc` returns NULL, this is a crash.

https://github.com/csound/csound/blob/f4195358143fdef54dd2b3669d096b23c85d8630/OOps/bus.c#L445-L453

2. **Memory leak in `delete_channel_db` -- `pp->var` and `pp->var->memBlock` are never freed** (CLAUDE.md says "Memory management and thread safety")

`alloc_channel` now allocates `pp->var` (CS_VARIABLE) and `pp->var->memBlock` (CS_VAR_MEM), but `delete_channel_db` only frees `hints.attributes` and delegates to `cs_hash_table_mfree_complete` which frees the CHNENTRY structs but not the nested allocations. This leaks on every `csoundReset()`.

https://github.com/csound/csound/blob/f4195358143fdef54dd2b3669d096b23c85d8630/OOps/bus.c#L357-L382

3. **`csoundSetChannel` missing spin lock -- data race on public API** (CLAUDE.md says "Memory management and thread safety")

`csoundSetChannel` calls `copyValue` without acquiring `pp->lock`, while all other channel write paths (`chnset_opcode_generic_init` at line 795, `chnset_opcode_generic_perf` at line 848, etc.) consistently use `csoundSpinLock`/`csoundSpinUnLock`. A host calling this from a non-audio thread while Csound is running will cause a data race.

https://github.com/csound/csound/blob/f4195358143fdef54dd2b3669d096b23c85d8630/OOps/bus.c#L2956-L2967

4. **`csoundGetChannelVarType` dereferences NULL `pp->var` for uninitialized generic channels** (CLAUDE.md says "API compatibility is important - changes to csound.h should be carefully considered")

For `CSOUND_VAR_CHANNEL` channels, `pp->var` is NULL until `chn_generic_initialise` runs. `csoundGetChannelVarType` only checks `if(pp)` but not `if(pp->var)`, while the sibling `csoundGetChannelVarTypeName` correctly checks `if(pp && pp->var)`. Calling this public API function on a generic channel before initialization will crash.

https://github.com/csound/csound/blob/f4195358143fdef54dd2b3669d096b23c85d8630/OOps/bus.c#L511-L516

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kunstmusik:
Was trying out claude code with glm-5.1 to do a code review and it auto-posted the above. @vlazzarini could you check those?  I'm also concerned about introducing varType to CS_VARIABLE, when we've been using CS_VAR_MEM throughout the app which has a ref to both CS_VAR_TYPE and CS_VARIABLE. I'm doing a separate review of the code for that and will reply again once there's a coherent picture on how to proceed around this.

kunstmusik:
I asked for a summary from gpt-5.4 and requested an analysis of the changes and to compare to CS_VAR_MEM system currently in use. In general, I'm concerned about overlap and think CS_VAR_MEM should probably be the abstraction that is used given it's history in the codebase. Definitely concerned about the possible alignment issues too.  AI Response below:


## Summary of changes

- **Primary feature**: Adds generic software-bus channel support for non-primitive types, especially UDTs and `Complex`. New generic `chnget`/`chnset` paths were added, and channel internals were refactored so a channel now carries a `CS_VARIABLE` wrapper instead of only raw `data` plus `varType`.

- **Type-system changes**: `CS_TYPE.createVariable` and `copyValue` were made more const-correct. Standard type constructors now set `var->varType` explicitly, and variable creation now passes a typed `const CS_TYPE *` argument through the creation path.

- **Composite type/runtime support**: Struct runtime handling was tightened around member ownership and copying. Array copy/free behavior was also adjusted.

- **Public API**: Adds csoundGetChannel() and csoundSetChannel(), exposing channel values as `CS_VARIABLE`s.

- **Tests**: Good coverage was added for UDT channels, generic `Complex` channels, mismatch/failure cases, and host-side channel API behavior.

Net: this is a substantial channel/type-system PR, not a small localized change.

## Issues

- **[Blocking] Channel allocation bypasses `CS_VAR_MEM`’s alignment abstraction**: In the new channel wrapper paths, storage is allocated with `sizeof(CS_TYPE *) + memBlockSize`. Core Csound code uses `CS_VAR_TYPE_OFFSET + memBlockSize` for this exact reason. This is exactly the kind of overlap with `CS_VAR_MEM` that is concerning: the PR re-implements the header-size rule instead of reusing the existing abstraction. On 32-bit or padded layouts, this can under-allocate or misalign `CS_VAR_MEM.value`.

- **[Blocking] Channel state now has two sources of truth**: The channel entry now stores both `data` and `var->memBlock->value`. In the normal path they alias each other, but the `chnexport` initialization path breaks that by freeing and replacing `chn->data` with `p->arg` while leaving `chn->var->memBlock` untouched. After that, pointer-based channel access sees one storage location while csoundGetChannel() exposes another. That is a real architectural incoherency and likely behavioral bug.

- **[High] Teardown no longer follows type-aware cleanup**: Channel teardown still mainly frees hash entries and attributes. With the new design, a channel may also own a `CS_VARIABLE`, a `CS_VAR_MEM`, and type-specific nested allocations. The teardown path does not call `freeVariableMemory` through the type system. Even if the global allocator eventually reclaims raw blocks, this skips the canonical destructor path for composite and custom types.

- **[Medium] String channel creation now double-allocates**: String variable initialization allocates `STRINGDAT.data`, and then the channel allocation path special-cases string channels and assigns a new 128-byte buffer without freeing the first one. That looks like a new leak introduced by the `CS_VARIABLE`-backed path.

- **[Low] Null-check order bug in allocation**: The code writes `pp->var->memBlock->varType = ...` before checking whether `pp->var->memBlock` is `NULL`. Rare in practice, but still incorrect.

## Architectural view on `CS_VAR_MEM`

This is the main design point.

`CS_VAR_MEM` already exists as the runtime value cell:

- **Type-aware storage**: It stores the type.
- **Payload container**: It stores the value payload.
- **Layout abstraction**: It hides layout and alignment details from callers.

That makes it the natural canonical representation for channel storage.

By contrast, `CS_VARIABLE` is more of a declaration and allocation descriptor:

- **Metadata**: Name, type, subtype, dimensions.
- **Lifecycle hooks**: Init and update callbacks.
- **Compilation/runtime context**: Context and pool/list metadata.

That is useful when creating storage, but not necessarily as the long-lived storage representation of a channel.

### What this PR currently does

It effectively keeps all three of these around:

- **Standalone pointer**: `CHNENTRY.data`
- **Variable wrapper**: `CHNENTRY.var`
- **Wrapped storage**: `CHNENTRY.var->memBlock`

That overlap is why the implementation is fragile:

- **Split ownership**: One path updates `data`.
- **Split reads**: Another path reads `memBlock`.
- **Reimplemented allocation**: Allocation logic manually reconstructs `CS_VAR_MEM`.
- **Unclear destruction**: Cleanup logic no longer clearly owns the right layer.

## Recommendation

- **Best direction**: Make channel runtime storage canonical on `CS_VAR_MEM`, not `CS_VARIABLE`. Use `CS_VARIABLE` only transiently to compute `memBlockSize`, run type-specific initialization, and capture subtype or dimension info during creation.

- **If `CS_VARIABLE` stays in `CHNENTRY`**: Remove standalone `data`, make every consumer go through `var->memBlock`, forbid pointer redirection that bypasses the wrapper, and make `chnexport` explicitly wrap or alias storage in a way the wrapper understands.

- **For the public API**: I would question exposing `const CS_VARIABLE *` directly. That leaks compiler/runtime internals into the host API. A `const CS_VAR_MEM *`-style runtime-facing API, or explicit copy-in/copy-out helpers, would be a cleaner boundary.

## Merge readiness

- **Recommendation**: Not ready to merge as-is.

- **Why**: There are at least two blocking issues:
  - **Alignment/layout contract bypass**: The implementation sidesteps the existing `CS_VAR_MEM` allocation contract.
  - **Divergent channel representations**: `data` and `var->memBlock` can diverge.

## Short assessment

- **What’s good**: The feature direction is valuable. The tests are stronger than average for this kind of runtime/type work. The const-correctness cleanup around `createVariable` is sensible.

- **What’s concerning**: The implementation is fighting an existing abstraction instead of using it. `CS_VAR_MEM` was already the compatibility layer for typed variable memory. This PR partially rebuilds that mechanism beside it, which is why coherence issues show up.

vlazzarini:
CS_VAR_MEM doesn't carry its size, so that's a disadvantage, it's not self-contained. I think it's better to pass CS_VARIABLE, as we've agreed and leave the  CS_VAR_MEM for internal use. We have already exposed it in a API header so that argument does not count. I think what we've discussed is still good. In any case, we can hold the data in CHNENTRY as a CS_VARIABLE.

There is no advantage of exposing CS_VAR_MEM over the current status: we can already access channel data and type with separate API calls. So no changes are needed. However if we want a fully self contained data unit, then CS_VARIABLE is more appropriate.

Aliasing the memory to data simplified the changes so that old opcodes did not need to be modified, but it can be done.

The changes from Claude are ok, but there is no leak because reset deallocates all memory, so there is no bug there.

vlazzarini:
I've removed the data and datasize members of CHNENTRY. The channel data is now solely encapsulated in the `CS_VARIABLE *var` member. I changed the code to access the memory stored in the memBlock, and the data size is given by memBlockSize. 

This also allowed me to improve chnexport to export any global variable via a channel without any hacks. 

So now we can decide whether to expose CS_VARIABLE or CS_VAR_MEM via the API functions, it's possible to do either, while maintaining the internal data as a CS_VARIABLE. 

Separate from this, and strangely, the MSVC build started failing in udo.c because of `alloca`, which doesn't exist in that build environment. That part of the code was never changed at any time and we never had a problem; there was never any macro redefining `alloca` as `_alloca` on MSVC, so it should have always failed to build. I have now added that to fix the problem.

vlazzarini:
@kunstmusik what should we do about this one? I think the channel data structure is in a better shape than what we have on develop, which is an enhancement. Plus we have improved chnexport as well and made it work for all types.

The API functions can be taken off the public headers for now and we can decide on this later. In any case, it is possible to give access via CS_VARIABLE or to CS_VAR_MEM or both.

vlazzarini:
@kunstmusik Just wondering whether we could make a decision on this PR?

kunstmusik:
Currently away on a trip, will review this when I return on Tuesday.

OOps/bus.c
```diff
@@ -516,7 +510,7 @@ int32_t csoundGetChannelPtr(CSOUND *csound, void **p,
     if ((pp->type ^ type) & CSOUND_CHANNEL_TYPE_MASK)
       return pp->type;
     pp->type |= (type & (CSOUND_INPUT_CHANNEL | CSOUND_OUTPUT_CHANNEL));
-    *p = pp->data;
+    *p = &(pp->var->memBlock->value);
```
hlolli, line 535: can this dereference NULL, or rather can pp->var->memBlock->value ever be NULL?

OOps/bus.c
```diff
@@ -1510,23 +1541,17 @@ int32_t chnexport_opcode_init(CSOUND *csound, CHNEXPORT_OPCODE *p)
...
-  /* csoundSpinLock(lock); */
   if (err)
     return print_chn_err(p, err);
 
   /* Now we need to find the channel entry */
   chn = find_channel(csound, (char*) p->iname->data);
-  /* free the allocated memory (we will not use it) */
-  csound->Free(csound, chn->data);
+  /* Free any existing chn var memBlock */
+  if(chn->var->memBlock)
```
hlolli: Can chn->var be NULL for CSOUND_VAR_CHANNEL? 

In alloc_channel(), the generic channel case leaves varType as NULL, so it looks like no CS_VARIABLE gets allocated for the channel entry.

vlazzarini: yes, alloc_channel() does that because it does not have enough information about the channel to complete initialisation. It's done later, when a new generic channel is initialised by an opcode.
EDIT: actually for generic vars, this opcode also needs to intialise the channel; I've added code for that. We did not have a chnexport for generic vars but now we do.

OOps/bus.c
vlazzarini: There is possibly an edge case. If a CSOUND_VAR_CHANNEL is requested by a host through the API, then we might have a problem if the channel has not been created by Csound code. I think I'll need to protect for that. In that case pp->var will be NULL.

vlazzarini: done

vlazzarini:
Added a generic chnexport and a test for it.

hlolli (dismissed):

kunstmusik (dismissed):
Did a review with AI (gpt5.5) and it found some issues which look valid to me. Report posted below. (Note: will be traveling for the next week to visit family; I'll try to look into more detail regarding the CS_VAR_MEM/CS_VARIABLE issue when I have free time.)


**Findings**

- **Blocking:** channel `CS_VAR_MEM` allocation still bypasses the existing layout abstraction. [OOps/bus.c](/Users/stevenyi/work/csound/cs7-prs/OOps/bus.c:414) and [OOps/bus.c](/Users/stevenyi/work/csound/cs7-prs/OOps/bus.c:755) allocate `sizeof(CS_TYPE *) + memBlockSize`; this should use `CS_VAR_TYPE_OFFSET + memBlockSize` from [include/csound_type_system.h](/Users/stevenyi/work/csound/cs7-prs/include/csound_type_system.h:61). This was one of your core CS_VAR_MEM/alignment concerns and is still unresolved.

- **Blocking:** `chnexport` aliases the channel `memBlock` to the global variable’s `memBlock`, but `delete_channel_db` later frees channel `var->memBlock` as if the channel owns it. See aliasing at [OOps/bus.c](/Users/stevenyi/work/csound/cs7-prs/OOps/bus.c:1572) and freeing at [OOps/bus.c](/Users/stevenyi/work/csound/cs7-prs/OOps/bus.c:336). The duplicate `data`/`memBlock` representation was removed, which is good, but ownership is still not explicit or safe.

- **High:** `csoundGetChannelVarType` still dereferences `pp->var` without checking it. [OOps/bus.c](/Users/stevenyi/work/csound/cs7-prs/OOps/bus.c:487). The sibling `csoundGetChannelVarTypeName` checks `pp && pp->var`; this one should too. This was one of the original reviewed issues and is not fixed.

- **Medium:** string channel allocation still double-allocates. `var_init_memory_string` allocates `STRINGDAT.data` at [Engine/csound_standard_types.c](/Users/stevenyi/work/csound/cs7-prs/Engine/csound_standard_types.c:362), then `alloc_channel` overwrites it with a second buffer at [OOps/bus.c](/Users/stevenyi/work/csound/cs7-prs/OOps/bus.c:435). Cleanup also still uses raw frees rather than the type’s `freeVariableMemory` hook.

- **Medium:** the public API question is still open. `csoundGetChannel`/`csoundSetChannel` remain in [include/csound.h](/Users/stevenyi/work/csound/cs7-prs/include/csound.h:976), exposing `CS_VARIABLE *` as public API. `csoundSetChannel` now locks, so that original race concern is fixed, but `csoundGetChannel` returns an internal pointer and is not inherently thread-safe despite nearby docs saying channel get/set helpers are thread-safe.

- **Low:** the PR comments out an unrelated browser test in [wasm/browser/tests/tests.js](/Users/stevenyi/work/csound/cs7-prs/wasm/browser/tests/tests.js:1265), and `externs.js` adds duplicate `isRequestingRtMidiInput` entries. That should be reverted or split out.

**State vs Your Earlier Concerns**

Fixed or improved: the `CHNENTRY.data`/`datasize` duplicate storage was removed; `csoundSetChannel` now locks; the alloc null-check order is improved; hlolli approved; GitHub checks are green.

Still concerning: CS_VAR_MEM layout is still bypassed, `CS_VARIABLE` is still the public-facing API, ownership around exported channels is unclear, type-aware cleanup is incomplete, and one original null-deref remains.

I would not merge this as-is. The feature direction is good, and the branch is closer than it was, but I’d ask for the allocation/layout fix, `chnexport` ownership fix, `csoundGetChannelVarType` null guard, and removal or explicit deferral of the public `CS_VARIABLE` API before merging.

Verification run locally: rebuilt `unittests`, then `ctest -R ChannelTests --output-on-failure` passed 12/12.

vlazzarini:
Seems reasonable. I had actually thought of the channel freeing issue after I fixed the var channel type yesterday. Strangely enough asan did not detect a double free in the tests.

vlazzarini:
I believe all points are addressed in the latest commit:

1) CS_VAR_TYPE_OFFSET is now used for allocation.
2) A flag for external ownership is introduced and used to prevent a double free.
3) Dereferencing check was added.
4) String channels now bypass variable memory initialisation, and allocate memory at alloc_channel(). ~All channel types still use csound->Free() to deallocate as I do not see a big enough reason to do otherwise.~  Actually, this is a good idea, done.
5) The API has not been changed but properly documented. The fact that csoundGetChannel() is not threadsafe by default is not a good reason to remove it; csoundGetChannelPtr() is both not threadsafe by default AND allows overwriting of channel memory. This has not been questioned since it was introduced in Csound 5.0. I would think if we remove csoundGetChannel() for this reason, then we need to remove csoundGetChannelPtr() as well.
6) The final issue is spurious; those changes were to fix a CI build failure that was preventing me to continue work on this. I have merged from develop to get the latest JS tests.

kunstmusik:
FYI: Still traveling, guessing I won't be able to do deeper review until Thursday. Suppose might be good to merge latest from develop into this to make sure things are in sync. 

Did do another review pass with a couple AI agents (GPT5.5, Gemini 3.5 flash) and go the following joint report for some issues:


Re-reviewed the latest `feature/chnapi` state. The branch is much closer: `CS_VAR_TYPE_OFFSET` is now used, `CHNENTRY.data`/`datasize` are removed, generic `chnexport` has explicit external ownership, and the focused channel tests pass.

## Findings

- **High:** `delete_channel_db()` still skips cleanup for control channels. In `OOps/bus.c:345`, nested cleanup only runs when the channel is *not* `CSOUND_CONTROL_CHANNEL`. Since control channels now allocate `entry->var` and `entry->var->memBlock`, normal `k` channels leak those allocations on reset. This also preserves the older issue where control-channel `hints.attributes` are not freed. The hash table cleanup only frees the `CHNENTRY`, not nested allocations.

  Suggested direction: run `delete_channel_varmem()` and `Free(entry->var)` for all channel entries, and free `hints.attributes` when present, or at least for control channels.

- **Medium:** host-side `CSOUND_VAR_CHANNEL` creation remains incomplete. `csoundGetChannelPtr()` can create a generic channel with `var == NULL`, then immediately return `CSOUND_ERROR`, leaving a partially initialized channel entry in the DB. Later APIs such as `csoundGetChannelDatasize()` can still dereference `pp->var`.

- **Medium/Low:** `csoundSetChannel()` should validate public API inputs. It dereferences `var` and `var->memBlock` without checking them first.

- **Low:** there is a redundant `pp->var->memBlock == NULL` check after the successful allocation path in `alloc_channel()`.

- **Low / unrelated noise:** wasm browser changes are still mixed into this PR: one browser test is commented out, `isRequestingRtMidiInput` is duplicated in `wasm/browser/externs.js`, and the fixture has whitespace-only churn. These should be reverted or split out.

vlazzarini:
Thanks for the review.

1) Fixed it so all channels are deallocated, and hints freed if they exist. NB: although it is good form to have memory deletion in this function, this seems to be *only* called on reset, where all memory is freed in any case. So while the review called this issue *high*, it actually had no practical impact on operation.

2) Checked for de-referencing in CsoundGetChannelDataSize().

3) Validated var etc in CsoundGetChannel()

4) Removed redundant code.

vlazzarini:
Having thought about this for a while, I think using CS_VAR_MEM in the API makes more sense than CS_VARIABLE. So I have adjusted the functions to that now.

OOps/bus.c
```diff
@@ -527,14 +527,21 @@ int32_t csoundGetChannelDatasize(CSOUND *csound, const char *name){
 
   CHNENTRY  *pp;
   pp = find_channel(csound, name);
-  if (pp == NULL) return 0;
+  
+  if (pp == NULL // no channel
+      && pp->var == NULL) // no variable data created
```
hlolli: if (pp == NULL || pp->var == NULL || pp->var->memBlock == NULL)

kunstmusik:
Dealing with travel delays home, will see if I can get a review in at airports, or will take a look when I get home.

kunstmusik:
Used multiple models (glm, mimo, gpt 5.3 codex, gpt 5.6 sol) and consolidated to get a meta review, posted below. Seems like things are almost there with just a few things to tidy up. 

# PR #2526 consolidated review

Reviewed the current PR head (`c431013d4`) against `origin/develop` (`c0b016882`), together with the three independent reviews, PR discussion, commit history, CI, and focused local tests.

## Overall assessment

I would request changes before merging.

The move to `CS_VAR_MEM` resolves most of the original memory-layout and ownership concerns. However, there are still problems involving custom type construction, missing Wasm exports, partially initialized generic channels, API validation, and lost coverage. The PR also contains unrelated web changes and a noisy commit history.

## Findings

### 1. High: custom type construction can produce invalid channel variables

`createVariable` changed its callback parameter from `void *` to `const void *` in `include/csound_type_system.h`. This is source-incompatible for external type implementations. The repository’s own `examples/types/tuple.c` still uses the old signature; compiling with incompatible function-pointer warnings enabled exposes the mismatch. The internal `CREATEF` casts hide these errors rather than updating or preserving the public callback contract.

There is a related runtime invariant problem. `chnexport_generic_initialise()` in `OOps/bus.c:1516-1525` accepts whatever `createVariable()` returns but does not centrally assign:

```c
pp->var->varType = argtype;
```

The standard constructors were updated to initialize this field, but existing or third-party constructors may not. The tuple example is one such constructor. This can leave `varType == NULL`, after which:

- `csoundSetChannel()` rejects or mishandles the variable.
- `csoundGetChannelVarTypeName()` can dereference `pp->var->varType`.
- Generic type comparisons no longer have a reliable invariant.

Recommendation:

- Set `varType` centrally after every successful construction, rather than relying on every constructor.
- Either restore the original callback signature or explicitly update the public contract, examples, and compatibility documentation.
- Add a test using a custom type constructor that follows the previous public contract.

### 2. High: the new public APIs are not exported by the Wasm build

`csoundGetChannel()` and `csoundSetChannel()` were added to `include/csound.h`, but they are absent from `wasm/src/exports.json`. That file drives the Emscripten linker exports through `wasm/src/csound.nix`.

Consequently, the new APIs are available in native builds but not through the Wasm module. Existing CI remains green because it does not assert that these symbols are exported.

Recommendation:

- Add both functions to `wasm/src/exports.json`.
- Update the Wasm export fixture.
- Add a browser/Wasm test that calls the new functions or at least verifies their exported symbols.

### 3. Medium: generic `csoundGetChannelPtr()` leaves a half-created channel behind

`alloc_channel()` can create and register a `CSOUND_VAR_CHANNEL` entry with `var == NULL`. `csoundGetChannelPtr()` then discovers that no concrete type was supplied and returns an error, but the invalid entry remains in the channel database (`OOps/bus.c:500-524`).

That entry can affect later lookups and exposes a state the surrounding code is not consistently prepared to handle.

Recommendation:

- Reject an untyped generic channel before inserting it, or remove the entry on failure.
- Clarify whether `csoundGetChannelPtr()` supports creating generic channels. If it does, it needs a way to receive the type descriptor.

### 4. Medium: new API validation and lifetime contracts are incomplete

The new functions have several unchecked assumptions:

- A null channel name can reach `find_channel()`, which reads `name[0]` once a channel database exists.
- `csoundSetChannel()` assumes the destination `memBlock` and type `copyValue` hook are valid.
- The `if (var)` check is redundant after the preceding null rejection.
- `csoundGetChannel()` returns a pointer into live engine-owned storage, but its validity period, reset behavior, and synchronization requirements are not documented.
- Hosts need `csound_type_system.h` to construct or inspect `CS_VAR_MEM`, despite the API being declared in `csound.h`.

These may be valid internal assumptions, but a public C API should either validate them or document them explicitly.

### 5. Medium: existing `chncleararray` coverage was removed

`tests/commandline/test_array_channels.csd` still declares `clearchan`, but the instrument that exercised `chncleararray` was replaced. The replacement adds useful audio-array coverage, but it should not remove the prior regression test.

Restore the `chncleararray` assertion and add the new test separately.

## `CS_VAR_MEM` concerns

The principal `CS_VAR_MEM` concerns have mostly been handled:

- Channel data is now represented by `CS_VAR_MEM`, eliminating the former duplicate `CHNENTRY.data`/`datasize` storage.
- Allocation uses `CS_VAR_TYPE_OFFSET + memBlockSize`, addressing alignment and wrapper-layout concerns.
- Owned and externally aliased memory are distinguished with `varmem_is_external`.
- Channel cleanup invokes the type-specific free hook and frees owned wrapper storage.
- The public API now passes `CS_VAR_MEM` rather than exposing `CS_VARIABLE`.
- `csoundSetChannel()` takes the API lock.

The remaining architectural concern is that correctness now depends heavily on every wrapper having a valid, canonical `varType`. The custom-constructor path currently violates that invariant. `CS_VAR_MEM` also contains no independent payload size, so copying is only safe while the descriptor and `memBlockSize` remain trustworthy.

For `chnexport`, the ownership flag prevents double-free, but the channel aliases global-variable storage. A reset/destruction test should confirm that cleanup ordering cannot leave the channel database holding an invalid external pointer.

Finally, `csoundGetChannel()` returns live storage rather than a synchronized snapshot. The documentation should state whether callers may read it concurrently with performance or mutation.

## Comments and documentation

Several comments are no longer synchronized with the implementation:

- The PR description still describes `csoundGetChannel()` and `csoundSetChannel()` as accepting `CS_VARIABLE *`; the code now uses `CS_VAR_MEM *`.
- `OOps/bus.c` still contains `// NEED TO CHECK FOR ARRAYS` immediately before an implemented array branch.
- Public documentation does not explain the required type-system header, pointer lifetime, reset invalidation, or thread-safety distinction between get and set.

The earlier review concerns about duplicate storage, alignment, cleanup, and exposing `CS_VARIABLE` were substantively addressed by code changes. The claim that all review points are handled is still too broad because the custom-constructor invariant and integration gaps remain.

## Missing coverage

The added tests are useful, particularly the scalar host API, stack-allocated `CS_VAR_MEM`, host array, audio channel, and Complex `chnexport` cases. Additional tests are needed for:

- A custom type constructor that does not initialize `varType`.
- Wasm symbol exports.
- Null names and null variables.
- Type mismatch errors.
- Untyped generic `csoundGetChannelPtr()` failure without database mutation.
- Non-scalar `csoundSetChannel()`.
- String and array `chnexport`.
- Reset/free behavior for owned and externally aliased storage.
- Concurrent get/set expectations.
- Restoration of `chncleararray` coverage.

## Scope and code quality

The following web changes appear unrelated to the channel/type-system work:

- `wasm/browser/externs.js` adds a duplicate `isRequestingRtMidiInput` declaration.
- `wasm/browser/tests/tests.js` comments out an unrelated failing test.
- `wasm/browser/tests/types/exports.fixture.txt` contains a whitespace-only change.

These should be reverted or moved to a separate PR. A failing unrelated test should not be silently disabled as part of this feature.

The small MSVC `alloca` change in `Engine/udo.c`, the lexer `ECHO` change, and broad `include/sysdep.h` restructuring should also be split or their dependency on this feature documented. The Win32 alignment/atomic changes appear more plausibly connected because they were needed for the PR’s 32-bit builds.

`git diff --check` currently reports 52 whitespace/EOF issues; these should be cleaned up.

## Commit history

The branch contains 56 commits, including 12 merge commits. It also contains temporary/debugging messages such as:

- “trying to find where the crash occurs”
- “fixed conflicts and merged from develop”
- “typo”
- “commenting out browser test that is failing”

Before merge, the branch should be rebased and consolidated into a small set of coherent commits: core channel representation, public API, generic `chnexport`, tests, and any required platform fixes.

## Verification

Current automated results are otherwise healthy:

- Native `ctest`: 159/159 passed.
- Focused channel tests: 13/13 passed.
- Relevant command-line CSD tests passed.
- GitHub Linux, macOS, Windows 32-bit, Wasm, and browser checks are green.

Those results are encouraging, but they do not exercise the custom-constructor failure, Wasm export omission, generic half-created entry, or the missing error/lifecycle cases above.

## Recommended merge conditions

1. Restore the custom type-construction invariant and resolve the callback source-compatibility break.
2. Export and test the new APIs in Wasm.
3. Prevent insertion of untyped/invalid generic channel entries.
4. Add public API validation and lifetime/threading documentation.
5. Restore `chncleararray` coverage and add the missing failure/lifecycle tests.
6. Remove or split unrelated web/platform changes.
7. Fix whitespace and consolidate the commit history.

vlazzarini:
Handled recommendations

1. Kept const, modified example, added var type setting to code.
2. Exported API functions. Wasm tests are beyond the scope of this PR.
3. Modified csoundGetChannelPtr() so it does not create incomplete channels 
5. Improved threading code - handling control channels with atomics
5. Added test for CSOUND_VAR_CHANNEL.
6. Restored test for `chncleararray`

Not touched unrelated wasm code which is beyond the scope of this PR

hlolli (dismissed):
I can follow up with more wasm tests, testing the updated channels

vlazzarini:
thanks, let's see what @kunstmusik says

kunstmusik:
@vlazzarini I ran some additional reviews and found a couple of new issues from the new code. I'll work with gpt to fix those up and commit those later today and provide a report.

vlazzarini:
@kunstmusik thanks. I'll be away from a computer until 1st Aug, so I guess I won't be able to contribute until then. Hopefully this work can be completed without my input

kunstmusik:
@vlazzarini leave this with me, I'll get it through merge. I'm back from travels now and don't plan to travel for a while so I'll take up reviewing and processing PR's while you're away.

kunstmusik:
FYI: I pushed some commits to fix up issues from the last report. There's one issue left but it's going to redo some things from this PR. For reference, the issue is regarding createVariable changes. I read through the following feedback to see about proceeding with it or not; I'm leaning towards yes and will likely proceed to with this path. Putting the feedback here for reference:

## What existed before this PR

`CS_VAR_MEM` is not new. It has existed in the type system since 2014, and `CS_VARIABLE` already contained:

- A `const CS_TYPE *varType`.
- Size, subtype, dimensions, and initialization metadata.
- A `CS_VAR_MEM *memBlock`.
- The `CS_VAR_MEM` header containing a type pointer followed by variable-sized value storage.

The channel system did not fully use that representation. `CHNENTRY` independently stored:

- A raw `MYFLT *data` pointer.
- A byte count in `datasize`.
- A separate `const CS_TYPE *varType`.

Channel allocation manually selected sizes for control, audio, string, PVS, and array channels. Generic channels sometimes created a temporary `CS_VARIABLE` to learn the required size and initialization function, but retained only the raw allocation and type pointer.

The constructor contract was:

```c
CS_VARIABLE *createVariable(void *cs, void *p, INSDS *ctx);
```

The meaning of `p` was already overloaded:

- Most built-in constructors ignored it.
- Array construction interpreted it as `ARRAY_VAR_INIT *`.
- User-defined struct construction interpreted it as `CS_TYPE *`.
- `csoundCreateVariable()` supplied the explicit `typeArg` when present, otherwise the `CS_TYPE *`.
- After construction, `csoundCreateVariable()` assigned `var->varType` centrally.

Direct constructor calls with `NULL` were—and still are—common when the caller only needs `memBlockSize` or `initializeVariableMemory`.

## What the PR changed

The PR unified channel storage around the existing type-system representation:

- `CHNENTRY` now owns a `CS_VARIABLE *` instead of separately tracking data, size, and type.
- Channel storage is allocated as `CS_VAR_MEM`, with the type descriptor in its header.
- Existing channel pointer APIs continue returning the address of the value portion.
- Generic channels can use each type’s constructor, initializer, copier, and destructor.
- Exported channels can alias a global variable’s existing `CS_VAR_MEM`.
- An ownership flag prevents exported memory from being freed by the channel database.
- New `csoundGetChannel()` and `csoundSetChannel()` APIs expose and consume `CS_VAR_MEM`.

To make channel-created variables immediately know their type, the PR then changed constructor behavior:

1. Constructors began assigning `var->varType` themselves.
2. The callback parameter first became `const CS_TYPE *`.
3. Array construction still needed `ARRAY_VAR_INIT *`, so the public callback was widened again to `const void *`.
4. Built-in callbacks retained narrower `const CS_TYPE *` signatures.
5. A `CREATEF` macro and explicit casts were added to fit those narrower callbacks into the `const void *` function-pointer field.

That evolution explains the current inconsistency: the API says “opaque const configuration,” while most implementations say “type descriptor,” and arrays say “array initialization descriptor.”

## Issues in the current change

### 1. Incompatible function-pointer calls

This is the clearest correctness issue.

The `CS_TYPE` member expects:

```c
CS_VARIABLE *(*)(void *, const void *, INSDS *);
```

Most implementations actually declare:

```c
CS_VARIABLE *function(void *, const CS_TYPE *, INSDS *);
```

They are stored using `(CREATEF)` casts. Those are incompatible C function types. Converting the pointer is permitted, but calling the function through the incompatible type is formally undefined behavior.

It works on normal platforms because both arguments are represented as pointers, but the casts suppress useful compiler checking and may conflict with stricter optimization, sanitizers, control-flow integrity, or less conventional toolchains.

This should be fixed before merging. It does not require rolling back `const`.

### 2. The constructor argument still has two meanings

The final `const void *` type is honest about being opaque, but the implementation is not consistent about what it contains:

- Standard scalar constructors expect `CS_TYPE *`.
- Struct constructors expect `CS_TYPE *`.
- Array constructors expect `ARRAY_VAR_INIT *`.
- Some older direct callers deliberately pass `NULL`.

`csoundCreateVariable()` currently makes this worse by declaring its local argument as `const CS_TYPE *` even when the object is really `ARRAY_VAR_INIT *`. It happens to pass the pointer through without dereferencing it, but the local type is false and invites future errors.

Its public `typeArg` parameter also remains mutable `void *`, despite the callback becoming `const void *`.

### 3. The “constructor always sets varType” invariant is not true

The motivating commit says that the type is set by `createVariable`, but older direct call sites still pass `NULL`, including array helpers, `diskin2`, `framebuffer`, and array opcodes.

For built-in types, that can produce a temporary `CS_VARIABLE` whose `varType` is `NULL`. Those callers currently use only its size or initializer and then free it, so this is not presently producing a test failure. It nevertheless means the new invariant cannot safely be relied upon.

Type assignment is also duplicated in several layers:

- Individual constructors.
- `csoundCreateVariable()`.
- Channel initialization.
- Generic channel initialization.
- Export-channel setup.

That duplication caused the exported-channel metadata defect we already fixed. It remains an architectural source of drift.

### 4. `CS_VAR_MEM` is a low-level host API without lifecycle helpers

The channel refactor’s internal use of `CS_VAR_MEM` is reasonable. The concern is exposing it as the general host-facing value-transfer API.

For a scalar, a host can construct a small `CS_VAR_MEM` relatively easily. For strings, arrays, structs, audio blocks, and managed values, the host must know how to:

- Determine `memBlockSize`.
- Allocate `CS_VAR_TYPE_OFFSET + memBlockSize`.
- Run the type’s initializer.
- Construct nested storage correctly.
- Eventually call the correct destructor.
- Use the exact canonical `CS_TYPE *`.

No public helper currently performs that complete allocate/initialize/destroy lifecycle. `csoundCreateVariable()` creates the variable descriptor but does not allocate and initialize its `CS_VAR_MEM`.

Consequently, `csoundSetChannel()` is currently most practical for copying an already-existing engine variable or channel block, rather than constructing arbitrary host values.

Exact type-pointer equality is conservative and safe, but it also means independently defined or cross-engine user types cannot be copied merely because their names and layouts agree.

### 5. Thread-safety needs narrower wording

`csoundSetChannel()` serializes writes to the destination channel. It cannot automatically protect a source `CS_VAR_MEM` returned from another channel because it does not know which channel owns the source.

Therefore:

- Host-owned immutable source memory is safe.
- A live source channel must be locked separately by the caller.
- Arbitrary `copyValue` implementations may allocate or otherwise be unsuitable for a real-time thread.
- `copyValue` receives a `NULL` instrument context through this host API, which custom types must tolerate.

The API should promise destination serialization, not unconditional end-to-end thread safety or real-time safety.

## Recommended direction

I would preserve the pre-7.0 API change and repair it forward.

The cleanest constructor interface would be:

```c
CS_VARIABLE *createVariable(
    void *cs,
    const CS_TYPE *type,
    const void *typeArg,
    INSDS *ctx);
```

This separates the two concepts currently overloaded in `p`:

- `type` is always the type being constructed.
- `typeArg` is optional type-specific initialization data such as `ARRAY_VAR_INIT`.

A properly named function-pointer typedef should replace `CREATEF`, all casts should be removed, and a shared constructor helper should assign `var->varType` consistently after successful creation. This would make `NULL` initialization arguments safe while preserving type identity.

For the channel API, I would either:

- Add public helpers to allocate, initialize, and destroy a complete `CS_VAR_MEM`, or
- Explicitly describe `csoundSetChannel()` as a low-level API intended for copying existing compatible variable blocks.

Given that 7.0 is unreleased, I favor fixing the constructor signature properly now rather than carrying the overloaded argument and incompatible casts into the released API.

kunstmusik:
Final summarization of work in this PR (generated by GPT 5.6 Sol):

## Final PR summary

This PR refactors Csound channels to use the existing type-system storage model instead of maintaining a separate raw-data representation.

### Channel architecture

- Replaces `CHNENTRY`’s separate data pointer, size, and type fields with a `CS_VARIABLE`.
- Stores channel values in `CS_VAR_MEM`, including the canonical `CS_TYPE`.
- Uses each type’s constructor, initializer, copy function, and destructor.
- Tracks externally owned memory so exported variables are not freed by the channel database.
- Preserves existing raw-value access through `csoundGetChannelPtr()`.

`CS_VAR_MEM` itself is not new; this PR integrates the channel system with that existing representation.

### Generic and exported channels

- Extends generic channel handling to arrays, structs, and other registered types.
- Allows `chnexport` to expose additional variable types.
- Makes exported channels alias the global variable’s existing memory safely.
- Keeps wrapper and memory-block type metadata consistent.
- Rejects incompatible or incompletely initialized generic channels without leaving invalid channel registrations behind.

### Public channel API

Adds:

```c
const CS_VAR_MEM *csoundGetChannel(
    CSOUND *csound,
    const char *name);

int32_t csoundSetChannel(
    CSOUND *csound,
    const char *name,
    const CS_VAR_MEM *var);
```

The setter now validates:

- Channel name and source pointers.
- Source and destination type information.
- Channel wrapper and memory-block consistency.
- Exact source/destination type compatibility.
- Availability of the type’s copy function.

Control-channel writes use the platform atomic implementation where available and the channel lock otherwise.

This is deliberately a low-level type-system API. The caller must supply initialized source storage, synchronize live source memory separately, and account for copy functions that may allocate or require non-real-time work.

### Variable-constructor API

The previous constructor argument had become overloaded between a `CS_TYPE` and type-specific initialization data. This PR now separates those concepts:

```c
typedef CS_VARIABLE *(*CS_CREATE_VARIABLE_FUNC)(
    void *cs,
    const CS_TYPE *type,
    const void *typeArg,
    INSDS *ctx);
```

It also adds:

```c
CS_VARIABLE *csoundCreateVariableForType(
    CSOUND *csound,
    const CS_TYPE *type,
    const void *typeArg,
    INSDS *ctx);
```

The helper:

- Invokes the constructor with the canonical type.
- Passes optional initialization data independently.
- Establishes `var->varType` centrally.
- Removes the need for incompatible function-pointer casts.

All built-in, array, struct, opcode, example, test, and channel call sites were migrated. This is an intentional pre-7.0 API and calling-contract change.

### Correctness fixes

The review follow-up also addressed:

- Incorrect control-channel type-mask testing in `csoundSetChannel()`.
- Broken syntax in the non-atomic fallback path.
- Constructor results being dereferenced before null checks.
- Exported init-rate channels retaining control-rate wrapper metadata.
- Invalid generic setter state returning success.
- Missing validation of copy functions and internal type metadata.
- Trailing whitespace and stale comments.
- Incomplete ownership, lifetime, synchronization, and error documentation.

### Platform handling

The PR retains the associated 32-bit Windows atomic adjustments required for float and double `MYFLT` builds.

The separately vetted browser-test changes were intentionally retained.

### Coverage and verification

Coverage now includes:

- Generic `CS_VAR_MEM` channel get/set behavior.
- Invalid setter inputs and inconsistent channel state.
- Failed generic channel creation not registering a channel.
- Exported init-rate type metadata.
- Separation of constructor type, initialization argument, and context.
- Central `varType` assignment.
- WebAssembly exports for the new channel APIs.
- Array, struct-array, bus, and command-line channel behavior.

Verification completed:

- Native library, unit tests, and `csound-bin` build successfully.
- Full CTest suite: **163/163 passed**.
- Focused channel/type/new-type suite: **25/25 passed**.
- Array-channel and bus-channel command-line tests passed.
- Exact C callback initialization passed with incompatible-function-pointer warnings treated as errors.
- The new constructor helper is exported by the shared library.
- Cumulative diff and whitespace checks pass.

kunstmusik:
@hlolli Could you do a quick check on the updates? Would like a second pair of eyes on this before merging.

kunstmusik (dismissed):

kunstmusik:
Fixed a linking issue I missed, and also fixed the failing CSD test with AI-assistance. Report for why the old test was removed and updated below:

### Change report

The previous test was stale after variable redefinition semantics changed in 2025. Explicitly typed local variables are now allowed to shadow globals of the same name.

The test nevertheless continued “passing” because of an unrelated crash:

1. The local `k[]` declaration shadowed a global `i[]`.
2. The redeclaration path recognized an array but did not construct `ARRAY_VAR_INIT` containing its element type and dimensions.
3. The old overloaded constructor substituted `CS_TYPE` when `typeArg` was missing.
4. `create_array()` interpreted that `CS_TYPE` pointer as `ARRAY_VAR_INIT`, resulting in invalid metadata and a segmentation fault.
5. Because the test expected any nonzero return code, the crash was incorrectly counted as its expected syntax failure.

The constructor refactor stopped that invalid substitution, so the CSD correctly completed with return code zero—revealing the stale expectation.

### Implementation

- Centralized resolution of explicit scalar and array annotations.
- Array declarations now consistently receive:
  - element type;
  - dimension count;
  - a valid `ARRAY_VAR_INIT`.
- The same resolution path is used for initial declarations and redeclarations.
- Scalar constructors no longer receive a redundant type descriptor as `typeArg`.
- Removed a redundant variable-name allocation.

### New regression test

The old negative test was replaced with `test_mismatched_type_shadowing.csd`. It verifies that:

- A local `k` scalar can shadow a global `i` scalar.
- A local `k[]` array can shadow a global `i[]` array.
- The local array has the correct length and supports element access.
- Changes to the local variables do not affect the corresponding globals.
- The original global values and array metadata remain intact after the local instrument finishes.

vlazzarini:
just noting that if 1. is right then all the casts in entry.c have exactly the same issue. Do they need to be fixed?

tests/c/channel_tests.cpp
```diff
@@ -23,7 +23,6 @@ class ChannelTests : public ::testing::Test {
     virtual void SetUp ()
     {
       csound = csoundCreate (NULL, NULL);
-      //csoundCreateMessageBuffer (csound, 0);
```
vlazzarini: This should be put back instead of removed. I commented it out so I could see msgs in stderr, but forgot to comment it back in. It should be restored.

kunstmusik: Restored in https://github.com/csound/csound/pull/2526/commits/09aa214bde6c6dacb11a99d10b4c2a1bb055a34d

kunstmusik:
> just noting that if 1. is right then all the casts in entry.c have exactly the same issue. Do they need to be fixed?

@vlazzarini: I'm not sure I understand the comment, could you elaborate further?

kunstmusik:
> > just noting that if 1. is right then all the casts in entry.c have exactly the same issue. Do they need to be fixed?
> 
> @vlazzarini: I'm not sure I understand the comment, could you elaborate further?

FYI: I think the PR is in good enough shape to merge. I'm going to proceed with this so we can move forward on processing other PR's. If there's an issue with entry.c, we can open a separate followup PR.

kunstmusik (approved):

vlazzarini:
> > just noting that if 1. is right then all the casts in entry.c have exactly the same issue. Do they need to be fixed?
> 
> @vlazzarini: I'm not sure I understand the comment, could you elaborate further?

In entry.c and elsewhere we use a similar cast method to CREATEF - SUBR. Does it need to be fixed?

kunstmusik:
I think we're good, CREATEF macro was removed and replaced with:

  typedef struct csvariable* (*CS_CREATE_VARIABLE_FUNC)(
      void *cs, const struct cstype *type, const void *typeArg,
      struct insds *ctx);

in
https://github.com/csound/csound/pull/2526/changes/0cc77a829458064fc25e69774f12e6e6c09c9ccc#diff-aa348080b27636d22878c4e85f27522a3237c85feea6a47e0cd6a41d812cc8e6R42

It should be a bit more type-safe this way.

vlazzarini:
Do we need to do the same for SUBR?

kunstmusik:
If I understand correctly, I don't think there's anything we can do, many
of the opcode functions don't have void* anymore as second arg and we'll
need the cast to make them work. Only way to get rid of the casts is to get
rid of specific types and go back to void* in the functions.

vlazzarini:
yes, and my question is whether what the review said under 1 is strictly true, in which case we need to make that change across the board.

NB: I guess we could use OPDS * instead of void * as it would be a safe cast to all opcode dataspace structs. Not sure if it makes any difference.

kunstmusik:
I'm unsure what #1 you are talking about, as this was in regards to the type for the createVariable function for a CS_TYPE, and not the opcode init/perf funcs and SUBR.

vlazzarini:
I mean this:

"They are stored using (CREATEF) casts. Those are incompatible C function types. Converting the pointer is permitted, but calling the function through the incompatible type is formally undefined behavior."

We similarly use SUBR to store function pointers in oentry. If CREATEF can't be used here, then SUBR can't be used there either.

PR #2528, opened 2026-04-06
Bugfix/arraytype

vlazzarini:
A bug was found where explicit-typed arrays with multi-character type names were failing semantic tests, for example,
at instrument 0

```
test:Complex[] init 2
```
gave

```
syntax error, Unable to find opcode entry for 'init'
  with matching argument types, line 15 columns 16-19
Found:
  Complex[] init c
```

The problem is that `Complex[]` is not an internally-recognised type,  but `:Complex;[]`.  The bug is that when an argument with token type  `T_TYPED_IDENT`, the `tree->value->optype` is passed as `Complex[]`, but the semantic analysis wants `[Complex]`.  If the token is `T_IDENT`, however, the type string is generated by `create_array_arg_type()` correctly.

When the array declaration is inside an instrument, the lexer creates a `T_IDENT`, otherwise it creates a `T_TYPED_IDENT`. So any declaration outside instruments, including those in UDOs, fail. 

This PR fixes the bug by adding a conversion function for `T_TYPED_IDENT` cases. It is a non-op for non-array type strings.

Tests added.

NB: the lexer creating a different token for declarations if they are inside an instrument is a bit dodgy - may need to be reviewed later.

### DIFF
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -53,7 +53,6 @@ int32_t add_udo_definition(CSOUND *csound, bool newStyle, char *opname,
 const char* SYNTHESIZED_ARG = "_synthesized";
 const char* UNARY_PLUS = "_unary_plus";
 
-
 char* csoundStrdup(CSOUND* csound, const char* str) {
   size_t len;
   char* retVal;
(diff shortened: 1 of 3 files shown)


hlolli (approved):
I remember dealing with this or very similar issues. At some point I'd like to take a step back and see how often we are moving square brackets around for intermediary internal representation of array and multi dimensional arrays. I think there's future refine work we can do to make it more consistent. But I see how it fixes the issue and let's get it fixed.

PR #2530, opened 2026-04-06
Fixed timing of complex ops

vlazzarini:
Ops with real numbers of i-type and Complex numbers were only taking place at i-time leading to non-intuitive results. This PR fixes it so that when i and C types are mixed operators run at init and perf time.


kunstmusik:
I haven't looked at this deeply, but should there maybe be finer control to have I op C be init-time, and k op C be perf-time?

vlazzarini:
K op C is perf time only, I op C is now both. The reason being that in some expressions involving arrays with k-rate indices the I op C needs to occur at both times and we get unexpected results.

kunstmusik:
Is there any way to do init time only op then?

vlazzarini:
Not arithmetic, but there are overloads of some functions for i-time only. We should consider Complex types as k-rate mostly but  with i-time results when mixed with i-time operators.

amanjagwani (approved):

kunstmusik:
I wish this wasn't merged when we haven't finished the conversation. I really don't like how this is operating differently than how i op X works elsewhere for other data, and would have preferred more discussion.

vlazzarini:
This was per the original design, and there was a bug. I merged it because it needed fixing. It could not have stayed how it was. As I explained, the Complex type is a perf-time type, with some limited init-time access. We can't have it behaving in such a way that an expression such as 0.5*C[k] always produces zero as a result regardless of the value of the operand, which was happening. 

NB: I spent about 4 hours yesterday debugging a real-to-complex FFT only to realise there was a bug in the Complex type implementation.

kunstmusik:
This is an area of Csound's language where rates fail as a concept with
newer types. At least with i op C being init-only, you can tell users to
use `k(0.5) *C[k]` to ensure perf-time calc. But that doesn't even
accommodate "C * C" which you might want at init-time too. Seems like the
closest thing to make this csound-ish is to have a Complex and ComplexK
type, to mimic i- and k-vars, which are just numbers with different rates.

Language-wise, C has const vs non-const. Csound might one day benefit from
a rate modifier like 'init Complex' or 'perf Complex', so that we can have
init String vs perf String, with a default to perf. Maybe a syntax like:

```
init W:Complex = x, y
```

Regardless, this problem is a general one and is not limited to Complex
data types.

vlazzarini:
I contemplated a Complexi type, but thought it was an over complication, since most of the DSP it's intended for is perf time. I thought I could resolve it on an expression basis, but there are issues like this one. Using k() as you mentioned would work with but it's odd looking, and an exception to general cases, since 0.5*k does not need it. 

For now, the adopted solution should work on the surface as assignment from C to i is supposed to be i-time only, so we do not get i-time vars changed at perf time, even if the expression is actually computed.

A rate converter may be an interesting idea to consider.

Issue #2381, opened 2025-12-08
Csound Array acess (init- vs perf-time)


kunstmusik:
Continuing on from https://github.com/csound/csound/pull/2378, we need to figure out how to deal with arrays to denote init- vs perf-time read/write. The concerns so far are:

1. String array access (array_get) currently accesses data at both init- and pef-time. This has caused issues as seen in #2378 with index variables getting incremented at init-time past the length of the array and then causing perf-time issues with array index out of bounds.
2. We can not use array index type to determine which rate to access array members, as users may want to use an i-var but have it accessed at perf-time. 
3. We need a solution that will work generically with any data type (e.g., UDTs).

IMO, the language syntax should be clear in showing when the array access will happen so that there is no ambiguity. Becuase of historical usage, we may need new syntax to achieve this.

For example, we might have something like:

```
// init-time array acccess using current value of global k-index
ival = iarray[i gkIndex] 
// perf-time array access, always retrieve element 0 each perf pass
kval = karray[k 0]
```

using [i indexVal] and [k indexVal] to force access at one rate or another. 

For the current syntax, we may have to continue using multiple oentries for array_get and array_set to maintain the current semantics for standard (traditional) types. We can document what the array access rate is for the standard types when using `someArray[indx]`, but use a more consistent set of rules for other types, such as:

1. If index is i-rate, use init-time access
2. If index is k-rate, use perf-time access
3. Use whatever explicit syntax we come up with to force using rates.

In terms of rules, the above might also work without explicit syntax if we tell users to use karray[k(ival)] to force  the index to be k-rate. 

Note: the above is brainstorming and any suggestions welcome. I will update the above as we field ideas on how to address the issue.

vlazzarini:
I want to give a bit more context on the issue. Using the old code, before
the fix was applied, we encounter the problem described above in a for-in loop.

Here is a simple test taken from our csdtests

```
instr 1
for j in ["hello\n","world\n","!!!\n"] do
 prints j
od
endin
```

Running this with -v64 we can see the issue with array access quite clearly.
The code compiles to

```
instantiated instr 1
 i__synthetic_300:i = 0:c
 x__synthetic_301[]:S[] fillarray "hello\n":S,"world\n":S,"!!!\n":S
 i__synthetic_302:i lenarray x__synthetic_301[]:S[],1:c
 __synthetic_303 
 j:S ##array_get x__synthetic_301[]:S[],i__synthetic_300:i
 prints j:S
 __synthetic_304 
[...]

PR #2534, opened 2026-04-07
Type 2 for - in loop with explicitly typed index variable

vlazzarini:
As discussed in #2381 this PR implements a proposal put there to use an explicitly typed loop index variable to define the loop type (init vs perf time).

- Currently the index variable type and the loop type is defined by the array variable type ('k' for k, a, and complex, 'i' for the rest) 
- This PR adds the means of using an explicitly type index variable to define the loop type.

The cases are as follows

(a) non-typed array and index variables -> loop type defined by the array type

```
karr[] = [1,2,3]
for val, index in karr do   // perf-time loop; val and index are k-vars
```

(b) typed array variable, non-typed index variable

```
for val:k, index in [1,2,3]  // perf-time loop, val and index are k-vars
```

(c) non-typed array variable, typed index variable

```
for val, index:k in [1,2,3]  // perf-time loop, val and index are k-vars 
```

(d) typed array and typed index variable

```
for val:S, index:k in ["1","2","3"]   // perf-time loop val is a string and index is a k-var 
```

Trying to run an i-time loop with a k-var array variable is user error and gives
a syntax error

```
for val:k, ndx:i in [1,2,3] do
```

```
syntax error, cannot run a perf-time loop with an i-time index, line 11
```

### DIFF
--- a/Engine/csound_orc.y
+++ b/Engine/csound_orc.y
@@ -640,6 +640,24 @@ for_in : FOR_TOKEN identifier in expr DO_TOKEN statement_list OD_TOKEN
           $$->next = $4;
           $$ = make_node(csound,LINE,LOCN, FOR_TOKEN, $$, $5);
           }
+         | FOR_TOKEN typed_identifier ',' typed_identifier in expr DO_TOKEN statement_list OD_TOKEN
+        {
+          $5->left = $6;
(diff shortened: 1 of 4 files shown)


hlolli (approved):
lgtm

PR #2535, opened 2026-04-08
Ensuring T_TYPED_IDENT

### LINKED ISSUE #2532: Lexer issue: T_IDENT and T_TYPED_IDENT
While fixing a bug (#2528) I noticed that the lexer produces different token types for typed identifiers depending whether they are found in an instrument or not. Here's an outline, let's consider a variable initialisation such as `var:MyType init 0`

-   If this is placed in an instrument, we get a T_IDENT token, because of the flex code

```
"instr"       {
                  namedInstrFlag = 1;
                  return INSTR_TOKEN;
                }
```
followed by

```
{TYPED_IDENTIFIER} { *lvalp = lookup_token(csound, yytext, yyscanner);
                  return (*lvalp)->type; }
```

and

```
ORCTOKEN *lookup_token(CSOUND *csound, char *s, void *yyscanner)
{
    int32_t type = T_IDENT;
    ORCTOKEN *ans;

    if(UNLIKELY(csoundGetDebug(csound) & DEBUG_SEMANTICS))
      csound->Message(csound, "Looking up token for: %s\n", s);
    ans = new_token(csound, T_IDENT, yyscanner);
    if (strchr(s, ':') != NULL) {
        char* th;
        char* baseName = strtok_r(s, ":", &th);
        char* annotation = strtok_r(NULL, ":", &th);
        ans->lexeme = csoundStrdup(csound, baseName);
        ans->optype = csoundStrdup(csound, annotation);
        type = T_TYPED_IDENT;
    } else {
        ans->lexeme = csoundStrdup(csound, s);
    }
    if (csound->parserNamedInstrFlag == 1) {
        return ans;
    }
    ans->type = type;
    return ans;
}
```

where it returns before `ans->type = type;` could set it to `T_TYPED_IDENT`(`namedInstrFlag` is a macro for `csound->parserNamedInstrFlag`).

- if this is placed anywhere else (instr0, UDO),  `namedInstrFlag` is not set to 1, therefore `lookup_token()` sets the type to `T_TYPED_IDENT`.

So the same code generates two different tokens. The bison code seems to be able to handle this without any apparent issues, but if we start relying on the difference between the two, we may have a problem. In fact this was what introduced a bug in semantic analysis, fixed in #2528.

I checked and forcing all code in the form checked by `lookup_token()` to make a T_TYPED_IDENT (as I would have expected the idea was) works for all tests except for one that has a line with the form

```
var1:i = 0
var2:k = 1
if var1:i == var2:k then
```

where it fails because a T_TYPED_IDENT is not legal in that expression. Changing it to  `if var1 == var2 then` fixes it. This also means that in the current state that line is only legal inside an instrument. So it's not such a benign issue but needs attention.

From what I can see we have two choices:

- Force any `x:y` to be a T_TYPED_IDENT and document that we can't have such a construct everywhere, but only at variable declaration. That's fairly simple, we can move the `ans->type` assignment and it's done.
- Abandon T_TYPED_IDENT and just have T_IDENT, as the code uses `optype` to carry a type for T_IDENT. This one I have not fully tested but from my recollection gives more trouble.

Aside from that, I wonder what the `namedInstrFlag` was supposed to be used for. It's not specific to named instruments, I wonder if it is a left over from the previous parser.

@kunstmusik @holli thoughts?

vlazzarini:
In #2532 we have noted that the use of T_IDENT and T_TYPED_IDENT tokens is not consistent across the Csound code. This is due to the use of a legacy parser flag `namedInstrFlag`, which is not relevant anymore. 

This PR removes that flag altogether from the code, enabling all explicitly typed identifiers to be made into T_TYPED_IDENT regardless of where they appear in the code. This means that using explicit typing for variables generally is only legal on the left-hand side of an expression, or in special cases such as UDO parameter lists or for-in variable lists.

A test had to be adjusted for this, since it used that discrepancy and it's not anymore allowed.

Closes #2532

### DIFF
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -47,7 +47,7 @@ YYLTYPE *yylloc_param;
  ORCTOKEN *make_num(CSOUND *, char *,  void*);
  ORCTOKEN *make_token(CSOUND *, char *s,  void*);
  ORCTOKEN *make_label(CSOUND *, char *s,  void*);
-#define namedInstrFlag csound->parserNamedInstrFlag
+
 #include "parse_param.h"
 
 #define YY_EXTRA_TYPE  PARSE_PARM *
@@ -235,7 +235,6 @@ ERSTR           "}R"
                   return STRUCT_TOKEN;
                 }
 "instr"         {
-                  namedInstrFlag = 1;
                   return INSTR_TOKEN;
                 }
 "endin"         { *lvalp = make_token(csound, yytext, yyscanner);
@@ -633,9 +632,6 @@ ORCTOKEN *lookup_token(CSOUND *csound, char *s, void *yyscanner)
         type = T_TYPED_IDENT;
     } else {
         ans->lexeme = csoundStrdup(csound, s);
-    }
-    if (csound->parserNamedInstrFlag == 1) {
-        return ans;
     }
     ans->type = type;
     return ans;
--- a/Engine/csound_orc.y
+++ b/Engine/csound_orc.y
@@ -245,7 +245,6 @@ instr_definition : INSTR_TOKEN instr_id_list NEWLINE
                                   csound_orcget_ilocn(scanner), INSTR_TOKEN,
                                   $2, $5);
                     csp_orc_sa_instr_finalize(csound);
-                    namedInstrFlag = 0;
                  }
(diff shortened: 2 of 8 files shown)


kunstmusik (approved):
Added one test for additional test coverage for this change. Otherwise looks good to me.

PR #2536, opened 2026-04-08
Enabling parallel unit testing

vlazzarini:
This PR adds ctest o the test build so that we can run unit tests (as we do for csd tests). 
See https://geoff.space/2025/03/running-unit-tests-in-parallel-with-googletest-and-cmake/

```
ctest -j 8
```

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -68,10 +68,13 @@ jobs:
           cd build
           make
 
-      - name: Run tests
+      - name: Run unit tests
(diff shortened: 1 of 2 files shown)


kunstmusik:
Code change is straightforward, but I thought it'd be good to update our github actions to run the tests in parallel. Pushed a commit and am awaiting to see if that runs alright.

kunstmusik (approved):
Tested parallel tests locally and updated github actions to use parallel also worked. Approving and merging now.

PR #2537, opened 2026-04-10
Fixed spurious warnings on addin, subin, divin and mulin

vlazzarini:
Since the introduction of proper handling of +=, -=, *= and /= (I think) some spurious compiler warnings started to appear on the console messages. For example `icnt += 1` in instr 0 gives

```
warning: Unknown TREE node of type 265 (??) found in root.
```

which should not be there. This PR fixes so that these operations are recognised by the compiler.


kunstmusik (approved):
Looks good to me

PR #2539, opened 2026-04-15
Fix for rfft; additional intermixed complex-real array ops

vlazzarini:
There was a bug in the complex overload of rfft/irfft creating a double-sized output array. Operations for intermixing real and complex arrays are added.


hlolli:
quadosc we don't want to add to tests?

vlazzarini:
good idea. I have a test I can add.

vlazzarini:
ok done @hlolli

hlolli (approved):

PR #2540, opened 2026-04-16
rtauhal fix

### LINKED ISSUE #2538: auhal not recognizing input 9
on a mac with https://github.com/CsoundQt/CsoundQt/releases/download/v7.0.0-beta1/CsoundQt-7.0.0-beta1.3-Csound-7.0.0-beta10-MacOS.dmg i got no input on channel 9.
i changed to portaudio, and that worked.
(hardware is RME fireface 802.)
could someone have a look in the code whether something is wrong?
if nothing shows up, i can test again the next days for clear diagnostics.

vlazzarini:
This PR fixes and closes #2538 where input channels are rotated around.

### DIFF
--- a/InOut/circularbuffer.c
+++ b/InOut/circularbuffer.c
@@ -37,14 +37,14 @@ void *csoundCreateCircularBuffer(CSOUND *csound, int32_t numelem, int32_t elemsi
          csound->Malloc(csound, sizeof(circular_buffer))) == NULL) {
       return NULL;
     }
-    p->numelem = numelem;
+    p->numelem = numelem+1;
     p->wp = p->rp = 0;
     p->elemsize = elemsize;
 
-    if ((p->buffer = (char *) csound->Malloc(csound, numelem*elemsize)) == NULL) {
+    if ((p->buffer = (char *) csound->Malloc(csound, p->numelem*elemsize)) == NULL) {
       return NULL;
     }
-    memset(p->buffer, 0, numelem*elemsize);
+    memset(p->buffer, 0, p->numelem*elemsize);
     csoundSpinLockInit(&p->lock);
     return (void *)p;
 }
--- a/InOut/rtauhal.c
+++ b/InOut/rtauhal.c
@@ -405,7 +405,7 @@ int32_t AuHAL_open(CSOUND *csound, const csRtAudioParams * parm,
        AudioObjectSetPropertyData(dev, &prop, 0, NULL, psize, &srate);
        AudioObjectGetPropertyData(dev, &prop, 0, NULL, &psize, &sr);
        // try another 5 times max (2.5 sec wait)
-       if(++attempts > 5) { 
+       if(++attempts > 5) {
          csound->Warning(csound, "could not set sr to %.1f after %d attempts",
                          srate, attempts);
          break;
@@ -662,13 +662,15 @@ static int32_t recopen_(CSOUND *csound, const csRtAudioParams * parm)
     cdata->inunit = NULL;
     *recordata = (void *) cdata;
     cdata->inParm =  (csRtAudioParams *) parm;
-    cdata->csound = cdata->csound;
+    cdata->csound = csound;
     cdata->inputBuffer =
       (MYFLT *) csound->Calloc(csound,
                                csound->GetInputBufferSize(csound)* sizeof(MYFLT));
+    /* CircularBuffer keeps one slot empty internally. */
     cdata->incb =
       csound->CreateCircularBuffer(csound,
-                                   parm->bufSamp_HW*parm->nChannels, sizeof(MYFLT));
+                                   parm->bufSamp_HW*parm->nChannels + 1,
+                                   sizeof(MYFLT));
 
 
     int32_t ret = AuHAL_open(csound, parm, cdata, 1);
@@ -696,9 +698,11 @@ static int32_t playopen_(CSOUND *csound, const csRtAudioParams * parm)
                                csound->GetOutputBufferSize(csound)*sizeof(MYFLT));
     memset(cdata->outputBuffer, 0,
            csound->GetOutputBufferSize(csound)*sizeof(MYFLT));
+    /* CircularBuffer keeps one slot empty internally. */
     cdata->outcb =
       csound->CreateCircularBuffer(csound,
-                                   parm->bufSamp_HW*parm->nChannels, sizeof(MYFLT));
+                                   parm->bufSamp_HW*parm->nChannels + 1,
+                                   sizeof(MYFLT));
 
     return AuHAL_open(csound, parm,cdata,0);
 }
@@ -717,7 +721,7 @@ OSStatus  Csound_Input(void *inRefCon,
     int32_t j,k,i,chns;
     Float32 *buffer;
     int32_t n = inNumberFrames*inchnls;
-   
+
     AudioUnitRender(cdata->inunit, ioActionFlags, inTimeStamp, inBusNumber,
                     inNumberFrames, cdata->inputdata);
     ioData = cdata->inputdata;
@@ -725,7 +729,7 @@ OSStatus  Csound_Input(void *inRefCon,
     if(chns == 1) { // non-interleaved
       for (i = 0; i < ioData->mNumberBuffers; i++) {
         buffer = (Float32 *) ioData->mBuffers[i].mData;
-        for(j = 0, k = 0; (uint32_t) k < inNumberFrames; j+=inchnls, k++) 
+        for(j = 0, k = 0; (uint32_t) k < inNumberFrames; j+=inchnls, k++)
           inputBuffer[j+i] = (MYFLT) buffer[k];
       }
     } else { // interleaved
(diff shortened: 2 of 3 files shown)


kunstmusik:
@vlazzarini: I had gpt-5.4 do a review of the issue and code and it reported below: 

**Findings**
1. High: the second commit breaks blocking input reads in rtauhal.c. The loop now exits on `while(l)` instead of `while(n)`, but circularbuffer.c returns `0` whenever the circular buffer is temporarily empty, not when the stream is finished. Other blocking backends keep looping until the requested sample count is satisfied, for example rtpa.c and rtopensl.c. With this PR, any transient capture underrun will return early with only part of `inbuff_` refreshed while still claiming a full `nbytes` read.

2. Medium: the rewritten output callback no longer handles underruns safely in rtauhal.c. The non-interleaved branch still walks the full callback frame count in rtauhal.c, but the post-copy clear in rtauhal.c only zeros the `n` samples that were actually read. If `ReadCircularBuffer()` returns short, the tail of `outputBuffer` is not zero-filled, so stale samples can be replayed across callbacks; in the interleaved branch, part of `ioData` can also be left unwritten. The expected pattern is to zero-fill the unread remainder first, as in rtwasapi.c.

** Recommendations **

1. Restore `rtrecord_()` to block until the remaining count reaches zero.
2. Zero-fill the unread portion of the output block before copying to `ioData`.
3. Add a regression test or at least a documented manual check for both interleaved and non-interleaved AuHAL channel layouts.

vlazzarini:
The first recommendation is against the actual fix. Checking for n in this case was causing a circular shift in the data, swapping the channels. If changed back the problem returns. 

The zero-fill is ok, I forgot about that.

kunstmusik:
I went through with gpt and opus and both converged on changes regarding the circular buffer sizing. I pushed changes regarding this. 

Regarding the changes, response was:

```
1. Circular buffer allocation: +1 element

The root cause of the original one-sample rotation is that CreateCircularBuffer internally keeps one slot empty to distinguish full from empty (in circularbuffer.c, checkspace returns rp - wp + numelem - 1). When we allocate exactly bufSamp_HW * nChannels, the effective usable capacity is one sample short of a full hardware block. The input callback writes inNumberFrames * inchnls samples, but only inNumberFrames * inchnls - 1 fit — the last sample wraps around and appears at the front of the next read, rotating channels by one position.

Fix: allocate bufSamp_HW * nChannels + 1 for both incb and outcb.

2. rtrecord_: restore blocking while(n) loop

The while(n && l) change exits the loop when ReadCircularBuffer temporarily returns 0 (buffer momentarily empty), but returns nbytes to the caller as if a full read completed. This delivers partially-filled buffers to the engine. The other backends (PortAudio, OpenSL, WASAPI) all use a blocking while(n) pattern — the buffer being temporarily empty is normal and just means the hardware callback hasn't fired yet.

3. Csound_Render: zero-fill on underrun

If the ring buffer has fewer samples than the hardware block requests, the tail of outputBuffer would contain stale data from a previous cycle. Added memset to zero-fill the gap between what was actually read and what the hardware expects. This matches what rtwasapi.c does explicitly.

4. Pre-existing bug fix: cdata->csound self-assignment in recopen_

Line 665 had cdata->csound = cdata->csound; (self-assignment, should be cdata->csound = csound;). This was already on develop. It happens to work when both input and output are open because playopen_ sets it correctly and they share the same cdata struct — but input-only mode would leave cdata->csound as NULL (from Calloc) and crash in the Csound_Input callback.
```

And requesting a further explanation of the +1 change, I got:

```
The short version is: the circular buffer implementation can only hold $N - 1$ items when you allocate $N$ slots, because it deliberately leaves one slot empty to tell full from empty. The AuHAL code was allocating exactly one hardware block, so the last sample of each block did not fit. Adding 1 gives the buffer one extra bookkeeping slot, which restores an effective capacity of exactly one full hardware block.

You can see the allocation change in rtauhal.c and rtauhal.c. The reason it is needed is in circularbuffer.c: for writes, the free-space calculation subtracts 1. In the empty-buffer case it explicitly returns circularbuffer.c as $numelem - 1$, not $numelem$.

Concretely, before the fix the code did this:

- Requested buffer size: $bufSamp\_HW \times nChannels$
- Allocated circular buffer slots: that exact number
- Actual writable capacity: one less than that

So if the hardware callback produced, say, a 16-channel block with 512 frames, it wanted to push $512 \times 16 = 8192$ samples. The ring buffer could only accept 8191. Because rtauhal.c writes into the circular buffer and ignores the returned count, that last sample was silently dropped.

That one dropped sample is enough to create the channel rotation bug. The reader still asks for a full block, so it gets:

- all but the final sample from block $N$
- then the first sample from block $N + 1$

That is a one-sample shift in the interleaved stream. With multichannel audio, a one-sample shift means channel boundaries move, so what should have been channel 1 of a frame shows up in the wrong channel position on the next read. That matches the symptom from the PR: hardware channel 1 appearing as the last channel and everything else shifted.

Why +1 is the right fix:

- We still want the ring buffer’s full/empty logic unchanged.
- We just need the usable capacity to be one full hardware block.
- Since usable capacity is $allocated - 1$, allocating $blocksize + 1$ makes usable capacity exactly $blocksize$.

So this change:

- old: allocate $bufSamp\_HW \times nChannels$
- new: allocate $bufSamp\_HW \times nChannels + 1$

does not increase the intended audio buffering by a whole extra sample of latency in any meaningful way. It only pays the ring buffer’s internal sentinel-slot cost explicitly, instead of accidentally making the usable capacity too small.
```

vlazzarini:
I dug a bit more and found the root cause of the issue:

(1) An offset in the circular buffer is caused at the start when the coreaudio callback drops a single sample as it fills the buffer when it does its initial runs. Since Csound is not reading the data, following this the callback will actually run a few times dropping complete buffers before the circular buffer is first read. When the data starts circulating, an offset inserted because of that single sample dropped, and so each output frame is 1 sample ahead. This causes the channel mapping to be offset by 1 (1 to 0, ..., 0 to N-1). 

(2) My earlier fix work because in `rt_record()`, checking for zero written samples guarantees that we only go ahead if no samples are to be read from the buffer so the offset is resolved. In the *unlikely* scenario that we cannot retrieve all requested samples (remaining samples > 0) when we get zero written samples, then we could try again. For this we can re-introduce the check for samples written in addition to (2). This would happen if the coreaudio callback got delayed somehow, but it runs at an interrupt, so it's unlikely. That was tested and works.

(3) However, this does not address the root cause, which is the circular buffer needing an extra sample. My approach is to internally add one extra sample whenever it is allocated.

EDIT: I tried to push and realised you have made similar changes. I wonder if we should change the circular buffer instead. I think that's a more general solution - other applications may step on this.

kunstmusik:
Did a quick check via gpt about scope of circular buffer changes and it seems like there's enough where it'd be good to do but in a separate PR. 

I'm not sure how to test this issue, but assume since you were able to reproduce there and the latest fixes things, I'm going to go ahead and approve the PR.

kunstmusik (approved):

kunstmusik (changes requested):
sorry, didn't see you pushed changes while I was writing my last reply, will do a quick re-review

vlazzarini:
So my proposal is to fix the circular buffer and restore the auhal code. A couple of things about the supposed bugs reported above:

(1) the `csound->cdata = csound;` is corrrect but we never ran into trouble because we always used full-duplex with input (hence the bug lasting so long).
(2) the memset() I had in the code was enough to clear the buffer, it did it after use, and so it would be reset for the next call of the function. It would be zero to start with,  that was a misdiagnosis. Since it did not matter whether it was done before or after, I moved it to before and restored my code.

I think it's better to change the circular buffer, because the expectation is that its capacity is what we set, not one less than that. That's what I would expect - if we don't change this we need to document it explicitly.

vlazzarini:
I am actually testing this using blackhole for input. I can play audio from another application and reproduce the problem. It's handy.

kunstmusik:
Added a test for circular buffer capacity. AI review had some minor issues that aren't blockers (pasted below) I'm going to re-approve now.

Low: the interleaved output fallback in rtauhal.c:784-790 still copies only the n samples actually read after an underrun. Because AuHAL is configured non-interleaved in rtauhal.c:464-468, that path should be dead today, so I would not treat it as a blocker or as a regression from the circular-buffer change. It is still a latent correctness issue if the stream format ever changes.

Low: diskin2.c:758-763l directly calls the internal checkspace() helper instead of using the public circular-buffer API. I do not see a regression there from the new semantics; it now just sees one more frame of writable space. But it is the main caller that is tightly coupled to the buffer internals, so it is the first place I would revisit if any follow-up issue appears.

kunstmusik (approved):

vlazzarini:
Thanks. For the interleaved path - I don't think it needs to copy the full buffer, I would expect that coreaudio would furnish a clean buffer to the callback (we don't manage it, coreaudio does). I think in the long run we should just removed the interleaved part, it was never used actually. The canonical approach is non-interleaved and for us there is no performance loss as we can't memcpy the buffers because they can be different float formats.

vlazzarini:
For diskin we could subtract one from checkspace() to conform to previous behaviour. Should we do it?

kunstmusik:
I see it's merged, I'm okay with things as they were in the PR. Cleaning up that dead code for the interleaved path would be good at some point.

vlazzarini:
I have a simple idea to improve the circular buffer API for diskin2. I'll make a PR.

PR #2541, opened 2026-04-18
Circular buffer API improvements

vlazzarini:
This PR improves the circular API by exposing the check space function, while keeping the new behaviour (write space increased by 1 sample to match reading space), adding unit and csd tests for this function and diskin async performance.

The new behaviour introduced in an earlier PR is such that write_space = size - read_space, whereas in the previous scheme we had write_space = size - read_space - 1, which turned out to be confusing for users.

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -533,7 +533,9 @@ jobs:
         run: cmake --build build --config Release
 
       - name: Run tests
-        run: cmake --build build --target csdtests
+        run: |
+          cd build
+          make csdtests
 
       - name: Run unit test framework
         run: ctest --test-dir build -C Release -j --output-on-failure
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -693,7 +693,7 @@ jobs:
       # - name: Run unit tests
       #   run: ctest --test-dir build -j$(nproc) --output-on-failure
       - name: Run CSD tests
-          run: cmake --build build --target csdtests
+        run: cmake --build build --target csdtests
 
   windows_build_32bit:
     name: Windows 32-bit build (mingw/vcpkg)
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -533,9 +533,7 @@ jobs:
         run: cmake --build build --config Release
 
       - name: Run tests
-        run: |
-          cd build
-          make csdtests
+        run: cmake --build build --target csdtests
 
       - name: Run unit test framework
         run: ctest --test-dir build -C Release -j --output-on-failure
@@ -694,8 +692,8 @@ jobs:
 
       # - name: Run unit tests
       #   run: ctest --test-dir build -j$(nproc) --output-on-failure
-      # - name: Run CSD tests
-      #   run: cmake --build build --target csdtests
+      - name: Run CSD tests
+          run: cmake --build build --target csdtests
 
   windows_build_32bit:
     name: Windows 32-bit build (mingw/vcpkg)
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -692,8 +692,8 @@ jobs:
 
       # - name: Run unit tests
       #   run: ctest --test-dir build -j$(nproc) --output-on-failure
-      - name: Run CSD tests
-        run: cmake --build build --target csdtests
+      #- name: Run CSD tests
+      #  run: cmake --build build --target csdtests
 
   windows_build_32bit:
     name: Windows 32-bit build (mingw/vcpkg)
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -533,7 +533,7 @@ jobs:
         run: cmake --build build --config Release
 
       - name: Run tests
-        run: cmake --build build --target csdtests
+        run: cmake --build build --config Release --target csdtests
 
       - name: Run unit test framework
         run: ctest --test-dir build -C Release -j --output-on-failure
@@ -610,7 +610,7 @@ jobs:
         run: cmake --build build --config Release
 
       - name: Run tests
-        run: cmake --build build --target csdtests
+        run: cmake --build build --config Release --target csdtests
 
       - name: Run unit test framework
         run: ctest --test-dir build -C Release -j --output-on-failure
--- a/InOut/circularbuffer.c
+++ b/InOut/circularbuffer.c
@@ -49,7 +49,7 @@ void *csoundCreateCircularBuffer(CSOUND *csound, int32_t numelem, int32_t elemsi
     return (void *)p;
 }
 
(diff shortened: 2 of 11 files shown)


vlazzarini:
I'm trying to figure out why Windows vcpkg does not run the csdtests even though the command is there, nothing runs. I can't see why.

.github/workflows/csound_builds.yml
```diff
@@ -692,8 +692,8 @@ jobs:
 
       # - name: Run unit tests
       #   run: ctest --test-dir build -j$(nproc) --output-on-failure
-      # - name: Run CSD tests
-      #   run: cmake --build build --target csdtests
+      - name: Run CSD tests
+          run: cmake --build build --target csdtests
```
hlolli: the strict yaml indentation is off I think

vlazzarini: That's fine, thanks. It turns out the mingw tests can't be run anyway, missing wine32 dependencies apparently.
But my question above is related to the Windows vcpkg csdtests target - it never runs the tests and I don't know why. See in the actions, it doesn't do anything.

kunstmusik:
I asked AI for a review and asked about Windows CI. For the review, it mentioned csoundCheckCircularBuffer wasn't exposed like the rest of the functions, so it did a change to expose that and update to use it in diskin2.c. Seems legit to me but I think also breaks ABI (though we haven't officially released csound7). I'll leave the git diff here and ask @vlazzarini to review and apply if he approves. (should be able to save as a patch use `git patch apply` with it).

I pushed the change it made for Windows CI. The git patch follows below:

```
diff --git a/OOps/diskin2.c b/OOps/diskin2.c
index 063f3d7b6..b9bbf1edf 100644
--- a/OOps/diskin2.c
+++ b/OOps/diskin2.c
@@ -760,7 +760,7 @@ int32_t diskin2_perf_synchronous(CSOUND *csound, DISKIN2 *p)
 int32_t diskin_file_read(CSOUND *csound, DISKIN2 *p)
 {
     /* nsmps is bufsize in frames */
-    int32_t nsmps = csoundCheckCircularBuffer(csound, p->cb, 1)/p->nChannels;
+    int32_t nsmps = csound->CheckCircularBuffer(csound, p->cb, 1)/p->nChannels;
     int32_t i, nn;
     double  d, frac_d, x, c, v, pidwarp_d;
     MYFLT   frac, a0, a1, a2, a3, onedwarp, winFact;
@@ -1224,7 +1224,7 @@ uintptr_t diskin_io_thread_array(void *p){
 int32_t diskin_file_read_array(CSOUND *csound, DISKIN2_ARRAY *p)
 {
     /* nsmps is bufsize in frames */
-    int32_t nsmps = csoundCheckCircularBuffer(csound, p->cb, 1)/p->nChannels;
+    int32_t nsmps = csound->CheckCircularBuffer(csound, p->cb, 1)/p->nChannels;
     int32_t i, nn;
     double  d, frac_d, x, c, v, pidwarp_d;
     MYFLT   frac, a0, a1, a2, a3, onedwarp, winFact;
diff --git a/Top/csound.c b/Top/csound.c
index 6becb3587..1bd721f9b 100644
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -372,7 +372,7 @@ static const CSOUND cenviron_ = {
     csoundSystemSr,
     csoundGetSr,
     csoundGetKr,
-    csoundGetKcounter,    
+    csoundGetKcounter,
     /* channels */
     csoundGetChannelPtr,
     csoundListChannels,
@@ -495,6 +495,7 @@ static const CSOUND cenviron_ = {
     csoundCreateCircularBuffer,
     csoundReadCircularBuffer,
     csoundWriteCircularBuffer,
+    csoundCheckCircularBuffer,
     csoundPeekCircularBuffer,
     csoundFlushCircularBuffer,
     csoundDestroyCircularBuffer,
diff --git a/include/csoundCore.h b/include/csoundCore.h
index c61ced17b..d599a43a0 100644
--- a/include/csoundCore.h
+++ b/include/csoundCore.h
@@ -1101,7 +1101,7 @@ struct CSOUND_ {
   /* Get engine control rate */
   MYFLT   (*GetEngineKr) (CSOUND *csound);
   /* Get engine kcounter value */
-  uint64_t (*GetEngineKcounter) (CSOUND *csound);  
+  uint64_t (*GetEngineKcounter) (CSOUND *csound);
   /**@}*/
 
   /** @name Software bus */
@@ -1281,6 +1281,7 @@ struct CSOUND_ {
   void *(*CreateCircularBuffer)(CSOUND *, int32_t, int32_t);
   int32_t (*ReadCircularBuffer)(CSOUND *, void *, void *, int32_t);
   int32_t (*WriteCircularBuffer)(CSOUND *, void *, const void *, int32_t);
+  int32_t (*CheckCircularBuffer)(CSOUND *, void *, int32_t);
   int32_t (*PeekCircularBuffer)(CSOUND *, void *, void *, int32_t);
   void (*FlushCircularBuffer)(CSOUND *, void *);
   void (*DestroyCircularBuffer)(CSOUND *, void *);
@@ -1462,7 +1463,7 @@ struct CSOUND_ {
   int32_t (*Sscanf)(char *str, const char *format, ...);
   /* Set opcode as deprecated */
   int32_t (*Deprecate)(CSOUND *csound, char *name,
-                       char *o, char *i, int32_t deprec);  
+                       char *o, char *i, int32_t deprec);
   /**@}*/
   /** @name Placeholders
       To allow the API to grow while maintining backward binary compatibility.

```

and the analysis from AI is below:

Findings

High: the native Windows CSD test target is miswired for a multi-config generator, which is why the GitHub Actions Windows vcpkg job is effectively not running the intended tests. In CMakeLists.txt:47 the MSVC branch hardcodes both OPCODE7DIR64 and the Csound executable path from CMAKE_BUILD_TYPE, but on Visual Studio generators that variable is empty. The workflow then builds Release and invokes the target without specifying a config in csound_builds.yml:529. I verified locally that multi-config builds leave CMAKE_BUILD_TYPE empty and default a plain build invocation to Debug. That matches the Actions summary, where the Windows “Run tests” step completes in about 1 second. The fix is to generate the executable path from the target itself and pass the Release config explicitly in the workflow.

Medium: the new circular-buffer query API is not fully exposed through the existing CSOUND callback surface. The new symbol is declared in csound_circular_buffer.h:94, but the callback table in csoundCore.h:1281 and its initializer in csound.c:495 still stop at Create, Read, Write, Peek, Flush, Destroy. That means host/plugin code using the callback table still cannot access the new function, so the API exposure is incomplete.

vlazzarini:
patch applied; it's a good idea to expose this for plugin opcodes.

kunstmusik (approved):
Think this is in good shape. One note is the ABI incompatibility introduced; given CS7 is not released as 7.0 yet, this should be fine, but downstream consumers of Csound may have to rebuild against latest csound release. (We may want to do another CS7 beta release at this point.)

vlazzarini:
Let's trigger one now. Thanks.

Issue #2515, opened 2026-03-16
Pass by reference xout access to members of struct arg


vlazzarini:
Accessing struct argument members directly by xout is not working correctly.

```
struct Test var1:i, var2:i

opcode Testop(var:Test):i
  xout var.var1
endop

instr 1
val:Test init 1, 1
i1 = Testop(val)
print i1
endin
```

This should print 1 but prints 0. It works if we use an intermediary variable, but direct access is broken. It only affects pass-by-reference.

hlolli:
I can take a look, this for sure should be printing 1

vlazzarini:
this is fixed now

PR #2544, opened 2026-04-22
Optimize pass-by-ref: Precompute UDO pass-by-ref rewiring at compile time

kunstmusik:
Worked with AI to do an optimization of pass-by-ref to remove memory allocations at init-time and move that to compile-time. The initial pass-by-ref implementation was done with the intent to get the general system working and to later optimize to remove the init-time allocations, as they are now in this PR. 

AI summary of work and rationale below: 

# Summary

This changes UDO pass-by-reference setup from a per-instance runtime lookup path to a compile-time rewire plan. Instead of building a hash map and scanning opcode arguments during initialization, the compiler now records the needed rewires once and the runtime applies a flat init and perf plan.

## What changed

- Add pass-by-ref rewire metadata to opcode info and build it when UDO compilation finishes.
- Replace the runtime hash-table rewiring path in UDO setup with direct application of the precomputed rewire entries.
- Preserve new-style aliasing behavior, including struct-member pass-by-ref cases by reconstructing member paths when parsed argument metadata does not carry them through.
- Free opcode-info chain and rewire-plan allocations during reset so the optimization does not leak across resets.

## Why

UDO pass-by-ref was doing deterministic setup work on every initialization. That work only depends on the compiled UDO body, so moving it to compile time removes repeated allocation and lookup overhead while keeping existing behavior for arrays, structs, recursion, and xout-driven aliasing.

## Validation

- Built successfully with the existing macOS build configuration.
- Ran the full command-line suite with make csdtests.
- Result: 276 passed, 0 failed.

### DIFF
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -2068,6 +2068,8 @@ int32_t csound_compile_tree(CSOUND *csound, TREE *root, int32_t async)
         opinfo->oentry->perf = NULL;
       }
 
+      csoundBuildUserOpcodeRewirePlan(csound, opinfo);
+
       break;
     case T_OPCALL:
     case LABEL_TOKEN:
@@ -2602,7 +2604,7 @@ static ARG *create_arg(CSOUND *csound, INSTRTXT *ip, char *s,
     else if(csound->engineState.varPool != NULL && (uintptr_t)csound->engineState.varPool >= 0x1000 &&
             setup_arg_for_var_name(csound, arg, csound->engineState.varPool, s) != NULL) {
     arg->type = ARG_GLOBAL;
-    
+
   }
   /* otherwise we have a local arg */
   else {
@@ -2610,7 +2612,7 @@ static ARG *create_arg(CSOUND *csound, INSTRTXT *ip, char *s,
       if (s[0] == '#') {
         csoundDie(csound, Str("compile error:"
                               " missing temporary variable '%s'"), s);
-      }   
+      }
       else {
         csoundDie(csound, Str("compile error:"
                               " missing variable '%s'"), s);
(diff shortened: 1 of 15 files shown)


kunstmusik (commented):
Note: I have created this as a draft as I still need to do some further testing, particularly with WASM build.

vlazzarini:
This is excellent. I think moving things to compile time whenever possible it is a great improvement. 
By the way, maybe you can also take care of this outstanding issue #2515 while you are at it.

vlazzarini:
It also reminded me of another possible compile-time action, which is struct access (x.y), currently done at insert/instantiation (insert.c).

gogins:
Nice!


Michael Gogins
Irreducible Productions

kunstmusik:
@vlazzarini I think https://github.com/csound/csound/pull/2544/commits/d1461b1799caaede4a71f83af9f65adbd1488b2a fixes #2515. Going to do some more testing here with Blue, livecoding, and WASM, but I think this'll be done today.

kunstmusik:
Requesting review from @vlazzarini and @hlolli

kunstmusik:
FYI: I found an issue that was introduced in beta29 of csound/browser where introducing UGEN API caused a problem with types like CsoundObj no longer being exported from the package. Saw this problem in beta30 too, but is now fixed in https://github.com/csound/csound/pull/2544/commits/41ed9c17c84912262b66e537b2a3769f1fdd3a71 and published as beta31

vlazzarini (approved):
Seems good to me. Tested a few things here and all seems to be running

vlazzarini:
~~Actually I found a new issue.~~ No, I didn't, it was all correct behaviour.

kunstmusik:
Thanks for the review @vlazzarini, merging now

PR #2546, opened 2026-04-25
fix for brew arm build

vlazzarini:
fixing failed workflow


kunstmusik (approved):
Don't really have experience with daisy build and this arm-toolchain path but I will assume since it's building it's good to go.

PR #2550, opened 2026-05-10
Add linux portable build (no dependencies)

gesellkammer:
This PR adds a minimal csound 7 build for linux without any external dependencies. It builds a static libsndfile and liblo and generates an artifact with the csound executable, libcsound and all plugins which run without the need of being installed. The user is required to set OPCODE7DIR64 to point to the distributed plugins directory. The resulting binary and libcsound.so is able to load external plugins and correctly loads the system libraries needed if available (alsa, jack). In particular it does not include portaudio since csound itself has native support for linux backends.

### DIFF
--- a/.github/workflows/linux_minimal.yml
+++ b/.github/workflows/linux_minimal.yml
@@ -61,6 +61,7 @@ jobs:
       LIBOPUS_VERSION:     "1.4"
       LIBFLAC_VERSION:     "1.4.3"
       LIBSNDFILE_VERSION:  "1.2.2"
+      LIBLO_VERSION:       "0.31"
 
     steps:
       # ------------------------------------------------------------------ #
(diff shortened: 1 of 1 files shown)


vlazzarini:
Can you integrate this in the csound_builds.yml so that the artefact can come out automatically with the release?

gesellkammer:
Ok, let me withdraw this PR so that I create a new branch with the build integrated.

gesellkammer:
The problem I see is that the csound_builds workflow is only run on releases. It would be more convenient to always run after each commit but upload the artifacts as a pre-release. This allows testing any commit without the need to build locally.  
A release would in this case just imply editing the release in github, giving it a version and unsetting its "pre-release" status.

vlazzarini:
Some artefacts are always built - not all, on every commit. You could add the linux artefact to these. They are all built when merged to the develop.

The upload to release, yes, that's only on actual release.

PR #2551, opened 2026-05-10
Testbuild2

gesellkammer:
This PR adds a minimal csound 7 build for linux without any external dependencies. It builds a static libsndfile and liblo and generates an artifact with the csound executable, libcsound and all plugins which run without the need of being installed. The user is required to set OPCODE7DIR64 to point to the distributed plugins directory. The resulting binary and libcsound.so is able to load external plugins and correctly loads the system libraries needed if available (alsa, jack). In particular it does not include portaudio since csound itself has native support for linux backends.


.github/workflows/csound_builds.yml
```diff
@@ -115,6 +115,367 @@ jobs:
...
+  # Builds inside ubuntu:20.04 (glibc 2.31). The full codec chain
+  # (libogg / libvorbis / libopus / libFLAC / libsndfile / liblo) is compiled
+  # from source as static PIC libraries and merged into libcsound.so or linked
+  # into the relevant plugin .so. libgcc_s and libstdc++ are baked in via
+  # -static-libgcc / -static-libstdc++.
+  #
+  # Runs on Ubuntu 20+, Debian 11+, Fedora 33+, RHEL 9+ and any other distro
+  # with glibc >= 2.31.
+  # ---------------------------------------------------------------------------
+  linux_minimal_build:
```
vlazzarini, line 130: I think it is good idea to add this to the worklow

```
  needs: update_version
```

to keep the builds destined to be uploaded to the release all under the same dependency structure (see the other related builds)

vlazzarini (approved):

PR #2552, opened 2026-05-12
fixed midi tempo scaling

vlazzarini:
A bug in midi file tempo scaling existed because the code assumed default tempo as the initial bpm for all midi files. This PR fixes the issue.


kunstmusik:
@vlazzarini I reviewed with AI and found issues:

> The negative default_tempo was only intended as an internal sentinel, but it leaked into real tempo math. That caused two regressions: MIDI files with no tempo event at tick 0 were converted with a negative tempo, which broke event timing, and midifiletempo/miditempo could report or scale from the sentinel state when tempo was overridden before playback. This fix keeps the real default at 120 BPM and initializes currentTempo from any tick-0 tempo event, so the sentinel no longer affects time conversion or runtime tempo reporting.

Ended up working with it to put together a change that restored (positive) 120 as the default_tempo but address the scaling issue. 

Things worked here with the one file I was testing but I just found issues with another file http://piano-midi.de/midis/debussy/deb_clai_format0.mid. Playback was strange but I opened in Logic and found the tempo was changing frequently within the file which was also odd. I'm guessing this is another issue altogether?

Could you take a look at the commit I pushed and let me know what you think of all this?

vlazzarini:
That debussy midifile plays correctly on csound 6, so I suspect the tempo scaling is not quite right yet.
NB: after some thought I think the temposcal adjustment code is not useful and it causes more problems (as in this case). I am removing it now ~and calling it done~.

vlazzarini:
I've redesigned the tempo scaling code - now it's correct. The main idea is to replace the global kcounter used in the reading by a local midifile counter, and apply the tempo control to it. With this, everything works now (midi playback tempo etc) as it should. It is also a much simpler code without need for extra adjustments. Tested here with the debussy file as well.

kunstmusik:
@vlazzarini I did another audit with AI and it found some issues:

1. High: midifile.c:890 no longer rewinds the new per-file playback clock. The rework makes the local counter authoritative for scheduling at midifile.c:867, but the rewind path only resets currentTempo and the list indices at midifile.c:903, midifile.c:904, and midifile.c:905. After rewindscore() on an active -F MIDI file, the next read still sees the old counter value, so everything up to the pre-rewind position is treated as already elapsed instead of replaying from time 0.

2. High: midifile.c:1046 seeks the indices and counter, but it never restores the tempo state for the target position. It updates midifile.c:1063, midifile.c:1067, and midifile.c:1069, while currentTempo is only changed later when the reader encounters a future tempo event at midifile.c821. If the seek target is already inside a later tempo region, miditempo is stale immediately after midifilepos(), and if the target is after the last tempo event it stays wrong for the rest of playback. Any later positive midifiletempo call then computes its scaling from the wrong base tempo.

3. Medium: Positive midifiletempo overrides no longer remain absolute across later tempo-map changes. The setter stores a one-time ratio at midifile.c:1038, later tempo events update currentTempo at midifile.c:821, and the reported effective tempo is derived from that unchanged ratio at midifile.c:1017. On a file whose internal tempo later halves, an earlier midifiletempo 90 call becomes 45 BPM rather than staying at 90. That is the multi-tempo failure mode discussed in the PR thread.

Looks like some hidden complexity arising from this issue as a whole. Thoughts on these?

vlazzarini:
@kunstmusik genuine issues, but no showstoppers

1. I forgot to reset the counter - fixed and test added.

2. Made sure tempo is reset to last tempoEvent found.

3. This one is a question of semantics - what do we mean by setting `miditempo 90` against a backdrop of changes in tempo in the MIDI file? It is not possible to have an effective tempo control with `miditempo` if we take it to mean an absolute BPM - it will be soon overridden and the control will not do anything. I was working with the Debussy file that is a pathological example. It just does nothing if we try to keep an absolute tempo measure - alternatively it just overrides all the tempo changes - do we want that?

Instead I think we should mean it to be always a relative tempo change in this case (against whatever tempo the MIDI file had when the command was issued). For MIDI files that do not change - it'll mean an absolute tempo. For the ones that do, it will be relative.

vlazzarini:
Found another issue in the new code. It assumed `csoundMIDIFileRead` was called exactly *once* every kcycle, and that is wrong. So the local counter could race ahead depending on how many times it was called. That is now fixed, we make sure the increment only happens once per kcycle.

kunstmusik:
@vlazzarini: Did another review with AI, looks like Issue 1 is good now. Issue 2 still had an issue so I went through with the AI to add some tests and introduce a fix. AI summarized as:

> Issue 2 was still real. midifilepos could restore the wrong tempo state when seeking, specifically by picking a future tempo event instead of the tempo actually active at the target position. I fixed that by restoring tempo from the last tempo event at or before the seek position, and I also aligned the related loop and midifilerewind paths to reset back to the file’s initial tempo.

I think Issue 3 is just awkward design since opcodes were assuming a single tempo rather than possibility of series of tempos. I think we're okay for now as it's a long-standing issue and probably for most use cases it's fine, and scaling seems like well-enough behavior. I'll approve the PR now and leave it to you to review the last changes I pushed and merge if you're okay with them.

kunstmusik (approved):

vlazzarini:
Thanks. I think this is a big improvement, not only the bugfix but also we gained a better tempo control and the capacity to loop automatically, which wasn't there before.

PR #2560, opened 2026-06-01
Suppressing misleading indentation warning

vlazzarini:
With clang 21, the -Wall flag adds -Wmisleading-indentation, which causes ~~bison~~flex-generated code to fail to compile on Debug configurations. This PR suppresses this warning so we can continue to build for Debug. Because of the order the flags are added, this cannot be suppressed in Custom.cmake since -Wall is added after any C flags added there - no effect. Since this is a low-risk warning, we can suppress it without much difficulty.


kunstmusik:
Quick note as I'm travelling and replying from the airport, but quick review here makes me think the problem is happening due to the way our actions are written in the .y file and that we could avoid the problem by reformatting the actions so generated code passes the misleading indentation rule. Alternatively, rather than apply it to all compiler flags, we could apply it to just the generated bison file using set_source_files_properties in CMake, so that we keep the benefit of check misleading indentation elsewhere. My initial thought is that we should fix the bison file. WIll look into this further when I get home.

vlazzarini:
You're right, a better solution is to add the flag to the file itself. In fact there was already a warning suppression there for switch fallthrough, so I just added this one to it. I also fixed the few instances where there were misleading indentations in the code base. Surprisingly, since we never checked for it, there were only a few.

vlazzarini:
Also, I checked the offending code and I do not think our lexer script has any influence on it,

```
        /* Create the reject buffer large enough to save one state per allowed character. */
        if ( ! yyg->yy_state_buf )
            yyg->yy_state_buf = (yy_state_type *)yyalloc(YY_STATE_BUF_SIZE  , yyscanner);
            if ( ! yyg->yy_state_buf )
                YY_FATAL_ERROR( "out of dynamic memory in yylex()" );

		if ( ! yyg->yy_start )
			yyg->yy_start = 1;	/* first start state */
```

it seems just an issue with flex formatting the output.

kunstmusik (approved):
This works for me. In general I'd rather like to see braces added wherever we have if-statements even if there's just a single statement in the branch, but that's my own preference. I'm sure we could come up with a rule and revisit using clang formatter at some point, but we can deal with it in another PR.

PR #2564, opened 2026-06-11
Fix: udo passbyref xin xout collision

kunstmusik:
Did work with AI Assistance (GPT5.5 and Opus 4.6) to implement a fix to issues related to xin/xout collisions. Has different mechanism than #2548 and should match UDO processing more closely with native C opcodes. Generated summary below:

## Summary

Fixes new-style UDO pass-by-reference handling when a `xin`-backed variable is returned directly through `xout`.

The implementation now treats direct `xin -> xout` pass-through values like native opcode output storage: the UDO output slot is seeded from the input before init/perf execution, internal UDO opcodes operate on that output-backed working value, and mutable inputs are written back after the UDO chain runs. Read-only sources such as constants and p-fields are not written back, so literal constants are no longer corrupted.

Additional regression coverage was added for constants, p-fields, k-rate and audio pass-through, arrays, struct members, mixed pass-through/expression outputs, and duplicate pass-through outputs.

## Difference from #2548

[PR #2548](https://github.com/csound/csound/pull/2548) fixed the immediate constant corruption by detecting constant-backed `xin` arguments and special-casing the existing pass-by-ref rewiring path.

This PR uses a broader model instead: it removes the caller-chain alias rewrite for direct `xout` pass-through and routes UDO-local pass-through work through the declared output slot, matching native C opcode input/output pointer behavior more closely. Constants and p-fields are protected as a consequence of the general seed/writeback model rather than by a constant-only exception.

## Testing

- `build/csound --nosound tests/commandline/udo/pass_by_ref.csd`
- `ctest --output-on-failure -j4`
- `cmake --build build --target csdtests`

### DIFF
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -900,7 +900,7 @@ jobs:
         run: .\vcpkg\bootstrap-vcpkg.bat
 
       - name: Configure build
-        run: cmake -B build -S . -G x64 -DBUILD_TESTS=1 -DUSE_VCPKG=1 -DCUSTOM_CMAKE="./platform/windows/Custom-vs.cmake"
+        run: cmake -B build -S . -A x64 -DBUILD_TESTS=1 -DUSE_VCPKG=1 -DCUSTOM_CMAKE="./platform/windows/Custom-vs.cmake"
 
       - name: Build Csound
         run: cmake --build build --config Release
--- a/.github/workflows/csound_builds.yml
+++ b/.github/workflows/csound_builds.yml
@@ -88,10 +88,10 @@ jobs:
           fetch-depth: 1
           submodules: true
 
-      - uses: lukka/get-cmake@v3.29.0
+      - uses: lukka/get-cmake@v4.3.3
 
       - name: Export GitHub Actions cache environment variables
-        uses: actions/github-script@v6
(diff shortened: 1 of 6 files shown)


vlazzarini:
@kunstmusik I wonder if this fixes #2061, which at the time I deemed user error.

kunstmusik:
@vlazzarini Will take a look at #2061 when I get home later today. 

Looks like the latest changes have gotten the windows build to complete. Would be good if @rorywalsh or @stekyne or someone else could test the Windows build still works as expected, as there was a change to use "Visual Studio 18 2026".

vlazzarini:
@kunstmusik yeh, I started trying to fix these earlier but then other work got in the way.

Good to see it fixed.

kunstmusik:
@vlazzarini did another session with Opus to review #2061 and it looks like it was not fixed. Went ahead and added a test to cover the issue in 2061 and a fix to udo.c in https://github.com/csound/csound/pull/2564/commits/e5b4cefc2d548113dd577900aba5749825b894c6.

vlazzarini (approved):
all good, thanks.

rorywalsh:
I’m going to be Windows-free for the next few months. We had a long chat,
realised we both want different things, and decided it’s best to give each
other some space :rofl

PR #2569, opened 2026-06-18
Moving CsoundGetOutputBufferSize to local in CsoundObj

vlazzarini:
Since CsoundGetOutputBufferSize is now out of the API, we had to change CsoundGetOutputBufferSize to a local function in CsoundObj.m accessing the parameter in OPARMS.


kunstmusik:
Could you re-create this PR but do:

1. add xcuserdata to .gitignore
2. commit just the CsoundObj.m change

PR #2572, opened 2026-06-21
Various updates/fixes to iOS Examples

vlazzarini:
The iOS examples needed some updates:

1) Obj-C UI needed to be updated to scene lifecyle
2) Obj-C and Swift graphic UI controls needed updating for new iphone screens
3) CsoundObj run to disk needed a call to csoundStart()


vlazzarini:
@hlolli this is an easy one we should try to clear now

hlolli (approved):

kunstmusik:
I see this one was approved already, going to merge it in.

PR #2573, opened 2026-06-21
fix: handle struct array member copy without segfault

hlolli:
Fixes a crash when copying an array of structs out of a struct member.

  Example:

  holder:Holder init boxes
  copied:Box[] = holder.items

  What was wrong:

  - Csound tried to create struct array elements without passing the struct type, so it could crash.
  - The copy also happened too late for i-time checks, so copied values could still be 0.
  - A later cleanup path could free memory it did not own.

  What changed:

  - Struct array elements now get created with the right type.
  - Struct arrays are copied during init when needed.
  - Struct array aliases now avoid double-free cleanup.
  - Added the failing CSD to the commandline tests as a normal passing test.

### DIFF
--- a/Engine/csound_orc_expressions.c
+++ b/Engine/csound_orc_expressions.c
@@ -101,8 +101,12 @@ char *create_out_arg(CSOUND *csound, char* outype, int32_t argCount,
         snprintf(s, 16, "#%c%d[]", type[1], argCount);
         add_array_arg(csound, s,  NULL, 1, typeTable);
       } else { // type[]
+        size_t typeLen = strlen(type);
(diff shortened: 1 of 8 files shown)


vlazzarini:
Thanks. I'll be looking at this in detail soon. Do you want to ask CodeRabbit for a review too?

hlolli:
@coderabbitai review

vlazzarini (commented):
All ok from my perspective

hlolli:
@vlazzarini thanks, if you smack that approve button then I can merge it :)

vlazzarini (approved):

PR #2576, opened 2026-06-22
Fix for diskin2 async race condition

### LINKED ISSUE #2575: diskin2 async reader can use freed instrument memory during cleanup
See failing workflow: https://github.com/csound/csound/actions/runs/27932238185/job/82646282986

ASan reports a heap-use-after-free in diskin2 async mode. The async IO thread walks the global DISKIN_INST list and dereferences current->diskin, but that pointer refers to a DISKIN2 opcode struct owned by an instrument instance. During cleanup, free_inactive_instances() can free the owning INSDS while the async thread is still reading from it.

Relevant path:
- OOps/diskin2.c:994: async thread walks DISKIN_INST
- OOps/diskin2.c:998: calls diskin_file_read(current->csound, current->diskin)
- OOps/diskin2.c:763: dereferences freed p->cb
- OOps/diskin2.c:517: deinit removes/frees list entries without synchronization

Likely cause:
DISKIN_INST.diskin is a borrowed pointer to instrument-owned memory, but async list traversal/removal and instrument cleanup are not synchronized.

Potential fix:
Synchronize DISKIN_INST list access and lifetime. Either lock traversal/mutation, stop/join the async thread before freeing DISKIN2 instances, or defer unlink/free ownership to the IO thread and flush it before INSDS memory is released.

vlazzarini:
A race condition leading to after-free access was detected by asan. This PR fixes the issue by simplifying the code and removing any premature deallocations, keeping it lock-free as required.

Closes #2575 

PS: a more stringent test was added.

### DIFF
--- a/OOps/diskin2.c
+++ b/OOps/diskin2.c
@@ -291,6 +291,7 @@ int32_t soundin(CSOUND *csound, DISKIN2 *p){
 
 static uintptr_t diskin_io_thread(void *p);
 static int32_t diskin2_cleanup(CSOUND *csound, void *p);
+int32_t csoundCircularBufferGetSize(CSOUND *csound, void *p); 
 
 static int32_t diskin2_init_(CSOUND *csound, DISKIN2 *p, int32_t stringname)
 {
@@ -412,13 +413,16 @@ static int32_t diskin2_init_(CSOUND *csound, DISKIN2 *p, int32_t stringname)
   if (csound->oparms->realtime==1 && p->fforceSync==0) {
     DISKIN2  **top, *current;
     p->csound = csound;
+    int32_t numelem =  p->bufSize*p->nChannels;
+    
      /* circular buffer is allocated once per opcode
         instance and will be freed by csoundReset
+        we also make sure size is compatible
       */
-    if(p->cb == NULL) {
+    if(p->cb == NULL ||
+       csoundCircularBufferGetSize(csound,p->cb)) {
        p->cb = csound->CreateCircularBuffer(csound,
-                                    p->bufSize*p->nChannels,
-                                             sizeof(MYFLT));
+                                    numelem, sizeof(MYFLT));
     }
 
    if (UNLIKELY(p->cb == NULL)) {
@@ -1534,7 +1538,17 @@ static int32_t diskin2_init_array(CSOUND *csound, DISKIN2_ARRAY *p,
     if (csound->oparms->realtime==1 && p->fforceSync==0){
       DISKIN2_ARRAY **top, *current;
       p->csound = csound;
-
+    int32_t numelem =  p->bufSize*p->nChannels;
+    
+     /* circular buffer is allocated once per opcode
+        instance and will be freed by csoundReset
+        we also make sure size is compatible
+      */
+    if(p->cb == NULL ||
+       csoundCircularBufferGetSize(csound,p->cb)) {
+       p->cb = csound->CreateCircularBuffer(csound,
+                                    numelem, sizeof(MYFLT));
+    }
       // create circular buffer if not already created
       // freed on CsoundReset()
       if(p->cb == NULL) {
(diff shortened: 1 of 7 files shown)


OOps/diskin2.c
```diff
@@ -412,12 +409,27 @@ static int32_t diskin2_init_(CSOUND *csound, DISKIN2 *p, int32_t stringname)
...
+    DISKIN2  **top, *current;
+    p->csound = csound;
+    int32_t numelem =  p->bufSize*p->nChannels;
+    
+     /* circular buffer is allocated once per opcode
+        instance and will be freed by csoundReset
+        we also make sure size is compatible
+      */
+    if(p->cb == NULL ||
+       csoundCircularBufferGetSize(csound,p->cb)) {
```
hlolli: Should this compare the buffer size to numelem instead of treating every existing buffer as incompatible?

OOps/diskin2.c
```diff
@@ -542,13 +536,28 @@ int32_t diskin2_async_deinit(CSOUND *csound, DISKIN2 *p){
...
+    if ((top = (DISKIN2 **)
+         csound->QueryGlobalVariable(csound, "DISKIN_INST")) == NULL)
+      return NOTOK;
+    int i = 0;
+    current = *top;
+    while(current != (DISKIN2 *)p) {
+      prv = current;
+      current = current->nxt;      
+    }
+    if (prv == NULL) *top = current->nxt;
```
hlolli, line 554: Could this list removal still race with the reader thread walking the same list while another async instance keeps the thread alive?

OOps/diskin2.c
```diff
@@ -1187,42 +1188,18 @@ int32_t diskin2_async_deinit_array(CSOUND *csound,  DISKIN2_ARRAY *p){
...
-        diskin_file_read_array(current->csound, (DISKIN2_ARRAY *)current->diskin);
-        current = current->nxt;
-      }
-    }
-    return 0;
-}
 
-int32_t diskin_file_read_array(CSOUND *csound, DISKIN2_ARRAY *p)
-{
+void diskin_file_read_array(CSOUND *csound, DISKIN2_ARRAY *p) {
```
hlolli, line 1200: Inside this function, should the write loop use DISKIN_THREAD_START_ARRAY instead of DISKIN_THREAD_START?

OOps/diskin2.c
```diff
@@ -454,19 +466,18 @@ static int32_t diskin2_init_(CSOUND *csound, DISKIN2 *p, int32_t stringname)
...
+    }
     current->nxt = NULL;
 
 #ifndef __EMSCRIPTEN__
-    if ( *(start = csound->QueryGlobalVariable(csound,
+    if(*(start = csound->QueryGlobalVariable(csound,
                                                "DISKIN_THREAD_START")) == 0) {
       void **thread = csound->QueryGlobalVariable(csound, "DISKIN_PTHREAD");
       *thread = csound->CreateThread(diskin_io_thread, *top);
+      printf("thread start\n");
```
hlolli: Can we drop this printf so normal runs do not get unexpected debug output?

hlolli (changes requested):
Left a few inline questions.

OOps/diskin2.c
vlazzarini: I don't think so. The file reader thread loop starts by checking that *top is not null. So if it is in the middle of walking the list, it will stop at the next iteration. The key thing is to not free any memory prematurely while the other thread is still running the loop.
I have tested it here quite extensively and the test I put in would also catch any access after free.

OOps/diskin2.c
vlazzarini: that's right. I'll fix that.

OOps/diskin2.c
vlazzarini: Forgot to remove tracing. Will do.

OOps/diskin2.c
vlazzarini: it can be bigger but not smaller so we don't need to allocate in that case. Oh I see, I forgot to add the logic expression.

hlolli (approved):

PR #2577, opened 2026-06-30
feat(debugger extensions):support for UDOs, globals, f-signals, and array serialisation

ervoL:
Changelog compared to PR #2519:
- Builds on merged #2519 (per-k-cycle callback)
- csoundDebugGetUdoFrames — nested + recursive UDO inspection
- csoundDebugGetGlobalVariables, csoundDebugSerializeFsig, csoundDebugSerializeArray
- Globals return full pool; clients filter to g* + engine whitelist
- WASM exports-only (same pattern as existing csoundDebugGet*)

Explanation: ive added debugger support for UDOs, globals, f-signals and array serialization. I built new .wasm file and I tested these changes in my web app (papad.dev) and they are working correctly.

Im fully open to any feedback and needed changes.


vlazzarini:
Thanks. Given that this is UDO-related, I will defer the full review to @kunstmusik, but I'll have a look as well.

kunstmusik:
@ervoL  I ran a code review with AI (GPT5.5) and it found some issues worth investigating:


I would not merge this as-is. The existing API/ABI surface is not broken by removal/reordering, but the new public debugger API has a couple of stability/correctness issues that should be fixed before it becomes contract.

### Findings

**[P1] Sliding fsig serialization can read past the frame buffer**

`Top/csound_debug.c` uses `csound->ksmps` in `csoundDebugSerializeFsig()` to choose the last sliding sub-frame. Sliding PVS frames are allocated using the producer opcode’s local `CS_KSMPS`, not necessarily the global engine ksmps. For local-ksmps instruments/UDOs, this can point beyond `frame.auxp`.

Suggested fix: derive the available sub-frame count from `fsig->frame.size / (NB * sizeof(CMPLX))`, or pass enough context into the API to use the producer’s local ksmps.

**[P2] UDO frame traversal can drop active sibling UDOs when nesting is present**

`csoundDebugUdoChainNext()` only follows `p->ip->opcod_deact` if the next `UOPCODE` has the same parent. But UDO setup links older siblings through the child instance first, and nested calls can overwrite the child’s `opcod_deact` head. In a shape like “two top-level UDO calls, where the later call invokes a nested UDO”, the earlier top-level sibling can be skipped.

Suggested fix: add a regression test with two top-level UDO calls where the later call invokes a nested UDO, then adjust traversal to preserve both sibling and nested chains.

**[P3] `frameIndex` docs do not match implementation**

`debug_udo_frame_t.frameIndex` is documented as ordering siblings on the same parent, but the implementation increments one shared counter across the full preorder traversal. Since this is a new public struct field, either the implementation should reset/index per parent, or the docs should describe it as a global traversal index.

ervoL:
Hi Steven! 

Thanks for the review and sorry for my delayed response. I think i fixed all 3 points you mentioned:
**P1 (sliding fsig OOB):** `csoundDebugSerializeFsig` now derives the sliding sub-frame count from `fsig->frame.size / (NB * sizeof(CMPLX))` instead of `csound->ksmps`, so local-ksmps producers (`setksmps` UDOs) stay within the allocated buffer. Added `testSerializeFsigSlidingLocalKsmps`.
**P2 (UDO sibling dropped when nesting):** `csoundDebugUdoChainNext` now follows direct siblings when `next->parent_ip == parent_ip`, and otherwise walks saved `opcod_deact` tails through nested call heads via `csoundDebugUdoFindSavedSibling`. Collection also skips `UOPCODE`s where `p->parent_ip != ip` (saved sibling tails on child instances) and uses a visited set to avoid cycles on recursive UDOs. Added `testUdoFramesSiblingAfterNestedCall`.
**P3 (frameIndex):** `frameIndex` is now a per-parent sibling index (resets at each `INSDS` level); docs updated to say 0 = head/most recent on the parent's chain. `testUdoFramesDualCallSites` asserts indices 0 and 1.

I rebased my changes to the newest csound version.
I verified it by compiling and running the unit tests and i built new WASM file and tested it in my web app (papad.dev) and everything seems to be working correctly.

Could you review the changes and tell me if you think it's good to go?

kunstmusik:
@ervoL I did another review of the latest, this time with Luna Max, and got:

**Findings**

- **[P1] Continue through nested sibling chains**

  When an outer UDO instance contains multiple nested UDO calls, `csoundDebugUdoFindSavedSibling()` can stop at a sibling belonging to the nested instance instead of continuing to the older sibling of the requested parent. For example, two outer calls with two nested calls each produce only three frames instead of four. The traversal should continue until it finds a candidate belonging to the requested parent.

- **[P2] Avoid silently truncating UDO frame enumeration**

  The hard limits of 64 saved-chain levels and 256 visited UDO opcodes can omit valid active frames without notifying the caller. This conflicts with the API contract of returning one entry per active UDO. Either remove these limits or expose an explicit truncation/error status.

kunstmusik:
@ervoL I redid the review using Sol 5.6 xHigh and recieved the following:

## Review summary

I recommend **requesting changes**. The linked review comment is substantively valid, although its expected frame count should be corrected.

### Findings

1. **[P1] Nested UDO traversal still drops older outer siblings**

   In `Top/csound_debug.c`, `csoundDebugUdoFindSavedSibling()` terminates when a second nested sibling belongs to the nested parent rather than the newest child.

   I reproduced this with two outer UDO calls, each containing two inner UDO calls:

   - Actual: 3 frames — the newest outer call and its two inner calls
   - Expected: 6 frames — two outer calls and four inner calls

   Therefore, [[the linked comment](https://github.com/csound/csound/pull/2577#issuecomment-5159007541)](https://github.com/csound/csound/pull/2577#issuecomment-5159007541) is correct about the traversal bug. However, “three frames instead of four” should be “three frames instead of six.”

2. **[P1] Sliding fsig serialization treats allocation capacity as the current local `ksmps`**

   `csoundDebugSerializeFsig()` derives the last sliding subframe from `fsig->frame.size`. However, `AUXCH.size` represents allocation capacity and can be retained when a UDO instance is reused.

   For example, a reused UDO can currently have local `ksmps=1` while the retained buffer capacity still represents 32 subframes. `pvssanal()` writes only the current local-ksmps range, but the serializer selects the last subframe implied by the retained capacity. It can therefore return an unwritten or stale frame.

   The serializer needs the producer’s current local `ksmps`, explicit logical subframe metadata, or sufficient UDO-frame context. Allocation size alone is not reliable.

3. **[P2] The traversal limits silently violate the API contract**

   The public documentation says `csoundDebugGetUdoFrames()` returns one entry per active UDO, but the implementation:

   - Stops saved-chain searching after 64 levels.
   - Silently skips frames after 256 visited UDO opcodes.

   The caller cannot distinguish a complete result from a truncated result. These limits should either be removed through dynamically sized cycle tracking or exposed through a detectable truncation/error result.

### ABI compatibility

Backward ABI compatibility looks good.

I built the PR base and head as shared libraries and compared their exported symbol sets:

- No existing exported symbols were removed or changed.
- The only additions were the five new debugger functions.
- Existing public struct layouts and function signatures are unchanged.
- WASM exports are additive.
- Windows 32/64-bit, macOS, Linux, iOS, Android, and WASM builds complete successfully in CI.

The new public structs become an ABI commitment once merged, so the correctness and contract issues above should be resolved before release.

### Tests and coverage

- All 29 debugger and serialization tests passed locally.
- The complete local unit suite passed: 188/188.
- The sole failing GitHub job reports an ASan use-after-free in `diskin2`, outside this PR’s changed files and apparently unrelated to this change.

Important coverage gaps remain:

- `testUdoFramesSiblingAfterNestedCall` contains one inner call per outer call and asserts only `frameCount >= 3`; the correct count for that test is exactly 4.
- There is no test with two inner calls per outer call, which is the confirmed failing traversal shape.
- There are no tests for the 64- or 256-entry truncation paths.
- The local-ksmps fsig test covers only a fresh allocation, not UDO-instance reuse after changing local `ksmps`.
- The tests compile as `__BUILDING_LIBCSOUND`, so they do not test the installed public header as an external consumer.
- CI does not publish line or branch coverage, so a coverage percentage cannot be established.

As a cleanup item, `tests/c/csound_debugger_test.cpp` was converted from LF to CRLF. This creates hundreds of unnecessary changed lines and causes `git diff --check` warnings; it should be normalized before merge.

ervoL:
Hi @kunstmusik ! Thanks for the review and feedback. I pushed an update addressing those points you mentioned:

P1 — UDO saved-sibling walk
csoundDebugUdoFindSavedSibling() no longer aborts when a candidate’s parent_ip doesn’t match the target parent; it continues along nested saved-sibling chains. Added testUdoFramesDoubleInnerPerOuter (2 outer × 2 inner → 6 frames) and tightened testUdoFramesSiblingAfterNestedCall to assert exactly 4 frames.

P1 — Sliding fsig sub-frame selection
csoundDebugSerializeFsig() now takes a localKsmps argument (0 = use csound->ksmps). For sliding analysis the last active sub-frame uses min(frame.size / stride, localKsmps) instead of capacity alone. Added testSerializeFsigSlidingReuseAfterKsmpsChange covering UDO instance reuse after setksmps change (sequential calls with setksmps 32 then setksmps 1 on a recycled instance).

P2 — Silent traversal limits
Replaced the fixed 256-entry visited stack with a dynamically grown set. Added optional truncatedOut to csoundDebugGetUdoFrames(); set when allocation fails or the depth safety limit (4096) is hit. 


I normalized csound_debugger_test.cpp line endings to LF. All 31 debugger/fsig unit tests pass locally.

Regarding the localKsmps as input param, i think it makes sense to have it like this because it then covers cases where we have different ksmps value (here i mean, different then global) in UDO or some instrument.

Again, i run test locally and generated new WASM file and tested it and everything seems to be working correctly.


If I missed something just let me know 🍻

kunstmusik:
@ervoL Looking good, I asked GPT 5.6 Sol for a re-review and it recommended a couple of small cleanup tasks:

Remaining issues

[P2] Correct the localKsmps=0 documentation
[include/csdebug.h](https://github.com/csound/csound/blob/db6f1df7e3f51331c5adc23115857fcc118b3047/include/csdebug.h#L404-L406) says that top-level instruments can pass zero to use csound->ksmps. However, setksmps is supported in top-level instruments as well as UDOs. If such an instrument instance is recycled after changing its local ksmps, passing zero can again select a stale retained subframe.
The documentation should say something like:
Pass 0 only when the producer is actually using the engine-global ksmps; otherwise pass the producer’s current local ksmps, whether it is a UDO or a top-level instrument.

[P3] The committed reuse test does not actually test sequential instance reuse
[testSerializeFsigSlidingReuseAfterKsmpsChange](https://github.com/csound/csound/blob/db6f1df7e3f51331c5adc23115857fcc118b3047/tests/c/csound_debug_fsig_globals_arrays_test.cpp#L244-L304) creates two simultaneous slidingAnal call sites, one with 32 and one with 1. It does not run sequential notes through one call site and inspect a recycled instance, as described in the PR comment.
The production fix passes a true sequential-reuse reproducer, but the regression test should be changed to cover that exact lifecycle.
Additional minor coverage gaps:
No test grows the visited set beyond its initial 64 entries.
No test exercises truncatedOut == 1.
The documentation says allocation failure sets truncatedOut, but Csound’s standard Malloc performs a LongJmp on failure instead of returning NULL, making that path unreachable under the normal allocator.

ervoL:
Hi @kunstmusik — thanks for the cleanup notes. I pushed update:

- Clarified localKsmps=0: pass 0 only when the producer actually uses engine-global ksmps; otherwise pass the producer’s current local ksmps (UDO or top-level instrument).
- Clarified truncatedOut: it signals the depth safety limit (dropped the unreachable default-Malloc NULL claim).
- Rewrote testSerializeFsigSlidingReuseAfterKsmpsChange to run sequential notes (i … 32, then i … 1) through one call site so a recycled UDO instance is inspected.

Again, i run tests locally and generated new WASM file and tested it and everything seems to be working correctly.
If you think that i should add some more tests, just let me know.

kunstmusik (approved):
Thanks @ervoL for working through this one and for the contribution. I think it's in a good state to merge.

ervoL:
cool! thanks. happy to contribute 🍻

PR #2579, opened 2026-07-04
Added code to exit cleanly on stack overflow

### LINKED ISSUE #2578: Recursive UDO after xout can crash Csound with SIGSEGV
I found a case where a recursive user-defined opcode crashes the process with signal 11.

I realize the example below is bad user code because it recurses without a real exit path. Still, it would be much nicer if Csound reported a recursion or stack exhaustion error instead of segfaulting.

Version:

Csound version 7.0, double samples
Build date shown by my binary: Jul 2 2026
Commit shown by my binary: 0967355ce3800beeeddbdba4efdbb5a0d5d38345

Reproduction:

Save this as repro-xout-recursion-segfault.orc:

```csound
declare Crashy():i

opcode Crashy():i
  ikeepGoing = 1

  while (ikeepGoing == 1) do
    if (ikeepGoing == 1) then
      xout -1
    endif

    ivalue = Crashy()
  od

  xout 0
endop

instr TEST
  ivalue = Crashy()
  prints "value=%d\n", ivalue
endin

schedule("TEST", 0, 0)
event_i("e", 0, 0)
```

Run:

```sh
csound -n -d -m128 --orc repro-xout-recursion-segfault.orc
```

Actual result:

Csound compiles the orchestra, starts SECTION 1, and then exits with signal 11.

In my local runner this appears as:

```text
Process exited with code 139
```

Expected result:

Csound should not segfault. Ideally it would stop with a controlled diagnostic such as a recursion depth, stack exhaustion, or runtime error message.

vlazzarini:
As per the title, some code has been added to handle stack overflow in UDO recursion more gracefully.
Closes #2578

### DIFF
--- a/Engine/insert.c
+++ b/Engine/insert.c
@@ -1951,6 +1951,11 @@ static INSDS *instantiate(CSOUND *csound, int32_t insno, int32_t link) {
 }
 
 INSDS *instance(CSOUND *csound, int32_t insno) {
+  // no stack checks when running in multiple threads
+  if(csound->multiThreadedThreadInfo != NULL ||
+     csound->oparms->realtime)
+    return instantiate(csound, insno, 1);
+  // otherwise do stack overflow check
   if(!csound_stack_exhaustion_check(csound))
     return instantiate(csound, insno, 1);
   else return NULL;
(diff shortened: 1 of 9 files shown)


vlazzarini:
OK, hit a dead end here. It's not possible to protect against stack overflow after all.

PR #2580, opened 2026-07-05
setting UDO recursion depth limits

### LINKED ISSUE #2578: Recursive UDO after xout can crash Csound with SIGSEGV

vlazzarini:
This PR is a second attempt at handling stack exhaustion in infinite recursion UDO due to user error. Earlier attempt tried to check for stack exhaustion and exit gracefully but that proved to be difficult to catch as it could happen in different places.

This attempt is a second-best approach, it sets a hard limit for recursion depth in UDOs, which is controlled by the an engine parameter. Set to 1000 by default, this can be changed using the option `--recursion-depth=n`. If `n` is set to zero, the check is bypassed.

A test was added to check that an infinite recursion is caught while finite recursion passes.

Closes #2578 (hopefully).

### DIFF
--- a/Top/argdecode.c
+++ b/Top/argdecode.c
@@ -348,7 +348,7 @@ static const char *longUsageList[] = {
     Str_noop("--run-unit-tests        enable assertion opcodes and report test failures"),
     Str_noop("                          (assertions are ignored by default)"),
     Str_noop("--error-deprecated      trigger compilation error on deprecated opcodes"),
-    Str_noop("--recursion-dept=n    set max UDO recursion depth to n"),
+    Str_noop("--recursion-depth=n    set max UDO recursion depth to n"),
     Str_noop("--help                  long help"),
     NULL};
 
--- a/Top/argdecode.c
+++ b/Top/argdecode.c
@@ -348,6 +348,7 @@ static const char *longUsageList[] = {
     Str_noop("--run-unit-tests        enable assertion opcodes and report test failures"),
     Str_noop("                          (assertions are ignored by default)"),
     Str_noop("--error-deprecated      trigger compilation error on deprecated opcodes"),
+    Str_noop("--recursion-dept=n    set max UDO recursion depth to n"),
     Str_noop("--help                  long help"),
     NULL};
 
@@ -1235,9 +1236,7 @@ static int32_t decode_long(CSOUND *csound, char *s, int32_t argc, char **argv) {
     s += 14;
     O->kr_default = O->sr_default / atof(s);
     return 1;
-  }
-
-  else if (!(strcmp(s, "aft-zero"))) {
+  } else if (!(strcmp(s, "aft-zero"))) {
     csound->aftouch = 0;
     return 1;
   } else if (!(strncmp(s, "limiter=", 8))) {
@@ -1248,10 +1247,14 @@ static int32_t decode_long(CSOUND *csound, char *s, int32_t argc, char **argv) {
                        Str("Ignoring invalid limiter\n"));
       O->limiter = 0;
     }
-    return 1;
-
+   return 1;
   } else if (!(strcmp(s, "limiter"))) {
     O->limiter = 0.5;
+    return 1;
+  } else if (!(strncmp(s, "recursion-depth=", 16))) {
+    s += 16;
+    int32_t depth = atof(s);
+    O->recursion_depth = depth > 0 ? depth : 0;
     return 1;
   } else if (!(strcmp(s, "vbr"))) {
 #ifdef SNDFILE_MP3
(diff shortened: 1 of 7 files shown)


hlolli:
I will review it today. I didn't look before into how other languages handle this and what's a standard practice. I wonder if it's a thing that the recursion limit doesn't count except if the arguments and output don't change. I don't know. But I'll do some digging then review. I'd like it to be specific enough that it becomes an edge case for either actual inf-recursion or something out of the beaten path. I feel 1000 which is smaller than the sampling rate, could be on the low side, that could also depend on the stack size this creates in memory. And we could also measure it by stack size and not count. Anyway will follow up.

vlazzarini:
I think stack sizes may be difficult to assess for this. Looking at the backtraces, the max frame count before the segfault seemed to vary from time to time.

I just checked and Python has also a default max recursion depth of 1000.

vlazzarini:
Re: checking for stack sizes, if you look at the branch from the closed PR #2579  you will see an attempt to check for stack sizes and exit from there. We can definitely check for it and I had cross-platform code to do it but the problem is that you will need to put a check almost everywhere because the stack exhaustion could happen at any call frame traversed. I spent the whole evening (while Paraguay was holding on) trying to get this to work and in the end, like Paraguay, I had to give up. Setting recursion depth limits might not be pretty but seems to be the only solution.

NB: interestingly enough an instrument infinite recursion using `schedule(this_instr, 0,1)` does not cause stack exhaustion, only an i-time hang.

Top/argdecode.c
```diff
@@ -348,6 +348,7 @@ static const char *longUsageList[] = {
     Str_noop("--run-unit-tests        enable assertion opcodes and report test failures"),
     Str_noop("                          (assertions are ignored by default)"),
     Str_noop("--error-deprecated      trigger compilation error on deprecated opcodes"),
+    Str_noop("--recursion-dept=n    set max UDO recursion depth to n"),
```
hlolli: dept missing 'h'?

vlazzarini: yes it is!

hlolli:
I am getting on board with this limit actually. Stack counting is a massive undertaking it seems. And it's fine, Python as you say, it flies there it can fly here in csound. Then to the review.

This test is set to max 2, but there is 3 depth recursion.


```csound
  <CsoundSynthesizer>
  <CsOptions>
  -n -d -m0 --recursion-depth=2
  </CsOptions>

  <CsInstruments>
  0dbfs = 1

  opcode Branch(cnt:i):i
    cnt += 1

    if (cnt < 3) then
      i1 = Branch(cnt)
      i2 = Branch(cnt)
    endif

    xout cnt
  endop

  instr 1
    iret = Branch(0)
    print iret
  endin
  </CsInstruments>

  <CsScore>
  i1 0 0.1
  e
  </CsScore>
  </CsoundSynthesizer>
```

we could consider book-keeping the jumps, not just set to 0, but decrement

- increment on UDO entry
- decrement on normal exit
- reset/clean up on the longjmp error path

Currently this increments inm->recurse_depth, but on successful return it resets it to 0 instead of decrementing it. That a recursive UDO with branching calls can undercount the real call depth after one inner recursive branch returns.

The counter is incremented on UDO entry at `Engine/udo.c:1232`, so on normal unwind this should behave like a stack depth counter. Resetting to zero here can undercount branching recursion: one inner recursive call returns, resets the shared depth to zero, then a sibling recursive call starts from zero even though the outer UDO frames are still active.

Something like `inm->recurse_depth--` on normal exit would preserve the active call depth. The error path from `setjmp` around `Engine/udo.c:1225` should probably reset/clean up the counter before returning `InitError`, since `longjmp` skips the normal unwind path.

vlazzarini:
looks like the it's right way to go.

vlazzarini:
@hlolli implemented your idea.

hlolli (approved):
I've tested with different recursion depths and they match the depth passed as option.

PR #2583, opened 2026-07-05
Improved array_dat method so that it returns a numpy array instead of…

fggp:
Improved array_dat method so that it returns a numpy array instead of a type void pointer


vlazzarini (approved):
Thanks

PR #2584, opened 2026-07-07
Allow typed struct declarations to use default initialization

hlolli:
This PR allows a typed struct variable to be declared without calling its `init`
opcode explicitly. This is already needed and partially implemented when doing recursive declarations of structs. But there's a bug in it.

Before this change, this did not compile:

```csound
struct Point x:i, y:i

instr 1
  point:Point
endin
```

That was awkward because Csound already knows how to create initialized storage
for struct values. Numeric members start at `0`, string members start as empty
strings, and nested member storage is initialized by the type system.

The fix adds parser and semantic support for bare typed declarations. The
declaration creates the variable and then does not emit an opcode call. Runtime
storage initialization still happens through the existing variable-pool
initialization path.

This does not make struct constructor arguments optional. Required members are
still required when using `init`:

```csound
struct Point x:i, y:i

instr 1
  point:Point init 7
endin
```

That still fails, because `y:i` is not optional. This keeps constructor behavior
strict while allowing declared typed storage to have a canonical default value.

### DIFF
--- a/Engine/csound_orc.y
+++ b/Engine/csound_orc.y
@@ -383,6 +383,18 @@ opcall  : identifier NEWLINE
             $2->left = $1;
             $2->right = $3;
           }
+        | out_arg_list identifier NEWLINE
+          { $$ = make_leaf(csound, LINE,LOCN, T_OPCALL, NULL);
+            if ($2->value != NULL && $2->value->lexeme != NULL &&
+                strcmp($2->value->lexeme, "init") == 0) {
+              $$->left = $2;
+              $2->type = T_OPCALL;
+              $2->left = $1;
+            } else {
+              $$->left = $1;
+              $$->right = $2;
+            }
+          }
         | out_arg_list_array expr_list NEWLINE
           { $$ = make_leaf(csound, LINE,LOCN, T_OPCALL, NULL);
             $$->left = $1;
@@ -399,6 +411,18 @@ opcall  : identifier NEWLINE
             $2->type = T_OPCALL;
             $2->left = $1;
             $2->right = $3;
+          }
+        | out_arg_list_array identifier NEWLINE
+          { $$ = make_leaf(csound, LINE,LOCN, T_OPCALL, NULL);
+            if ($2->value != NULL && $2->value->lexeme != NULL &&
+                strcmp($2->value->lexeme, "init") == 0) {
+              $$->left = $2;
+              $2->type = T_OPCALL;
+              $2->left = $1;
+            } else {
+              $$->left = $1;
+              $$->right = $2;
+            }
           }
         | function_call NEWLINE
           { $$ = $1; }
--- a/Engine/csound_orc_semantics.c
+++ b/Engine/csound_orc_semantics.c
@@ -3104,10 +3104,14 @@ int32_t initStructVar(CSOUND* csound, void* p) {
   CS_STRUCT_VAR* structVar = (CS_STRUCT_VAR*)init->out;
   CS_TYPE* type = csoundGetTypeForArg(init->out);
   int32_t len = cs_cons_length(type->members);
+  int32_t incnt = (int32_t)init->h.optext->t.inArgCount;
   int32_t i;
   if(csoundGetDebug(csound) & DEBUG_SEMANTICS) {
      csound->Message(csound, "Initializing Struct...\n");
      csound->Message(csound, "Struct Type: %s\n", type->varTypeName);
+  }
+  if (incnt == 0) {
+    return CSOUND_SUCCESS;
   }
   for (i = 0; i < len; i++) {
     CS_VAR_MEM* mem = structVar->members[i];
@@ -3407,6 +3411,10 @@ int32_t add_struct_definition(CSOUND* csound, TREE* structDefTree) {
     member = member->next;
   }
   temp[index] = 0;
+  oentry.intypes = csoundStrdup(csound, "");
+  csoundAppendOpcodes(csound, &oentry, 1);
+  oentry.opname = csoundStrdup(csound, oentry.opname);
+  oentry.outypes = csoundStrdup(csound, oentry.outypes);
   oentry.intypes = csoundStrdup(csound, temp);
   csoundAppendOpcodes(csound, &oentry, 1);
   return 1;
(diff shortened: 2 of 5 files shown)


vlazzarini:
What about global vars? Can we do

```
point@global:Point
```

vlazzarini:
Also it's a bit odd that we could do it for structs and not for builtin types. It would be good to allow for default init to all builtin types:

instr 1
y:a 
y = in() + delay(y)*0.5
endin 

Another question - what happens when an instance is re-used? Does it keep previous contents or is initialised?

Looking at this example and the one for structs, I'd say it is a bit awkward. I would think init with no arguments would be better (but I seem to remember some problem with that last time I tried).

point:Point = init() or point:Point init
y:a = init() or y:a init

kunstmusik:
@hlolli didn't you create a declare keyword for something similar to this?

hlolli:
I found it ackward too initially. I find defaulting to 0 bit arbitrary but it does avoid introducing null, and this is already the case for arrays. Maybe a future idea would be `struct Point x:i = 3, y:i = 4` or smth for giving default arguments.

As for not calling init. Yeh we could decide that we need it. But I also wasn't planning on making arguments just fully optional always, finding balance between flexibility (allowing all to be optional) and type strictness. I feel this could allow for it. But I tagged @kunstmusik because maybe you had some similar ideas in mind.

All I need it for is recursiveness and making placeholders.

hlolli:
@vlazzarini I force pushed example using global struct `point@global:Point`

hlolli:
Notebene we also have optional type declarations in the classical UDO, like 'o', 'j', 'p', 'V' etc. I could go the direction to just make sure those work. It's a workaround that's acceptable to me.

hlolli:
@kunstmusik 

> @hlolli didn't you create a declare keyword for something similar to this?

You're likely referring to the keyword `declare` that is more for declaring UDO ahead of time. I forgot if it works for struct too.

vlazzarini:
They work in new-style UDOs.

I think if we could have something like 'declare' for default initialisation, it would be better than just var:type. And we could roll it out for all types.

hlolli:
(future pie in the sky whishfulness ahead, but could affect how we move forward from here)

My preference would be that UDT should also support it, but I'd argue it would be more readable if it was similar to python

```python
@dataclass
class Point:
    x: int = 3
    y: int = 4
```

or js

```js
function point({ x = 3, y = 4 } = {}) {
  return { x, y };
}
```

For opdoces it would imo fit fell like this

```csound
struct InstrumentNote pitch:i

opcode MakeNote(pitch:i = 69):InstrumentNote
  note:InstrumentNote init pitch
  xout note
endop

instr 1
  n:MakeNote init // or not init
  print n // ;; => 440
endin
```

hlolli:
I should ask before I get ahead of myself, UDT, you mean the type returned from a struct declaration? Can you give example, I may be thinking about the wrong thing here.

vlazzarini:
My idea was:

instead of 

point:Point

have

declare point:Point

to use the default initialisation.

Similarly,

declare sig:a

declare violin:InstrDef

etc

hlolli:
I see the appeal, but I’m hesitant to put default construction behind `declare`.
To me `declare` reads as a compile-time declaration, especially because that is
already how it is used for UDO forward declarations. Default construction feels
like a value-level operation, so I think `point:Point init` is a better fit.

That also keeps the door open for field defaults later:

```csound
struct Point x:i = 3, y:i = 4
point:Point init
```

For my current use case, I mainly need recursive structs and placeholders without introducing null. A zero-argument `init` for structs seems like the narrowest way to support that while staying close to existing Csound syntax.

vlazzarini:
I am happy with using init with no arguments. I think that was Rory's proposal as well. Let me know when you have this in and I can review.

hlolli:
@vlazzarini 

Behavior now:

 ```csound
 
 struct Point x:i, y:i
...
point:Point init       ; allowed
point:Point init 1,2 ; allowed when all members are provided
point:Point init 1    ; rejected
point:Point            ; rejected
```

I could be convinced that

```
point:Point init 1    ; rejected
```

should be allowed too.

I feel csound doesn't need to be strictly typed here. Could also just be a warning. Open for ideas.

Then again you can't make struct without argument. So, init without arguments could be just the exception to the rule, and it would be consistent. This is also ok as the behaviour is now as I implented it in my last commit.

**Comment by vlazzarini** [comment]
The last option (partial initialisation) may be ok but could also be confusing. Let's hear from @kunstmusik and @rorywalsh

**Comment by PasqualeMainolfi** [comment]
If I may suggest an idea, the bare `point:Point` syntax solves the runtime initialization issue, but it is implicit: it looks like an incomplete statement or even a typo. I think an explicit marker would make the intent much clearer. One possibility would be introducing `new()` as the default constructor.

The idea would be to distinguish two initialization modes that are currently conflated:`new()` → default-aware,  unspecified members are initialized using their canonical default values.

```csound
struct Point x:i, y:i

instr 1
  a:Point = new() ; x=0, y=0        full default initialization
  c:Point init 5, 6 ; x=5, y=6        strict
endin
```

This makes the programmer’s intent explicit while preserving the existing strict semantics of `init`.
As for partial initialization (b:Point = new(5)), the positional form seems sufficient for a first implementation. Its limitations could be addressed later by introducing named arguments, for example:

```csound
b:Point = new(y=4)
```

This would allow arbitrary subsets of members to be initialized without depending on their declaration order, while keeping the initial implementation simple.

vlazzarini:
We already had a PR that I put through, introducing new() for this and it was rejected, so let's not reopen that discussion.

https://github.com/csound/csound/pull/2514

PasqualeMainolfi:
@vlazzarini Thanks, I had completely missed #2514. I don’t want to reopen that discussion. I was only referring to readability.

That said, I still believe default initialization is absolutely necessary. I also think partial initialization has its place, provided it’s presented in a way that doesn’t introduce confusion, just as you pointed out.

rorywalsh:
This looks good, but I am not in favour of the partial initialization. I
don't think we need it.

hlolli:
@rorywalsh looks good the PR in current state you mean?

rorywalsh:
Yeah looks good, shall I merge?

hlolli:
@rorywalsh yes please

vlazzarini (approved):

PR #2585, opened 2026-07-07
Fix direct indexing of recursive struct array members

hlolli:
While testing recursive structs, I ran into another compiler edge case: indexing an array field directly from a recursive struct could fail even though the same code worked if the array field was first copied into a local variable.

For example, a struct can own an array of its own type, and code like this should be valid:

```csound
note = LookupInstrument(shelf.parent[0], name)
```

Before this change, that path could lose part of the element type while lowering the expression. The compiler then created an invalid temporary variable with no real type, which surfaced as an internal error like:

```text
cannot create variable #EnvS3: NULL type
```

The fix keeps the full array element type when creating temporary array values, including user-defined struct types. It also asks for the type of the single array expression being lowered, instead of accidentally including sibling function arguments in the type string.

This preserves existing behavior for normal arrays and struct arrays, but makes direct indexing behave the same as the local-variable workaround. A new regression test uses a small orchestra instrument lookup example with recursive shelves to cover direct reads, writes, and parent lookup through a recursive struct array field.

### DIFF
--- a/Engine/csound_orc_expressions.c
+++ b/Engine/csound_orc_expressions.c
@@ -98,13 +98,17 @@ char *create_out_arg(CSOUND *csound, char* outype, int32_t argCount,
     char *type = remove_type_quoting(csound, outype);
     if (find_brace(type)) {
       if(*type == '[') { // [type]
-        snprintf(s, 16, "#%c%d[]", type[1], argCount);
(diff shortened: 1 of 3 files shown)


hlolli:
@coderabbitai review

vlazzarini:
@coderabbitai full review

vlazzarini (approved):
looking good

Issue #2568, opened 2026-06-15
Feature Proposal: Semantic Retrieval and Embedding Opcodes


PasqualeMainolfi:
Over the last few years I have been working extensively on latent representations in the context of algorithmic composition and computational creativity. I am currently preparing a few papers on these topics, and I also expect to release a project on agentic composition in Csound in the near future.

This work has led me to think more deeply about the potential of latent spaces, especially in relation to sound synthesis and musical control. Since I primarily work with Csound and try to contribute to its ecosystem whenever possible, I started wondering whether it would make sense to introduce a small family of opcodes dedicated to semantic representations.

The idea is not to embed a large language model into Csound, nor to perform expensive inference in the audio thread. Instead, I envision a lightweight asynchronous infrastructure that allows users to load an embedding model of their choice and interact with it safely from Csound.

A possible first step could be an opcode family such as:

- **sem_init** – initialize a user-provided embedding backend.
- **sem_embed** – asynchronously compute an embedding from text (or potentially other modalities).
- **sem_ready** – poll the status of a pending request.
- **sem_get** – retrieve the resulting embedding vector.

On top of this low-level interface, additional opcodes could provide semantic operations, for example:

- storing and indexing vectors,
- nearest-neighbor retrieval,
- similarity computation,
- interpolation between semantic points,
- mapping latent representations to synthesis presets or algorithmic structures.

The important point is that the embedding space itself would remain model-agnostic. Csound would simply expose a generic interface, while users would be free to build their own semantic associations between latent vectors and synthesis parameters, presets, orchestration strategies, or compositional processes.

I believe this could open interesting possibilities for semantic control of synthesis, retrieval of sound materials, adaptive algorithmic composition, and human-AI interaction, while remaining fully compatible with Csound’s philosophy of modularity and explicit control.

If the community finds this direction worthwhile, I would be happy to prototype such an opcode family and discuss possible designs and use cases.

kunstmusik:
This sounds interesting but I wonder if it wouldn't be better to have something like this developed and maintained in a separate opcode library?

vlazzarini:
@PasqualeMainolfi I also find it interesting and agree with @kunstmusik - could we have this in the csound plugins repository? Or else you can maintain it in your 3rd-party repo as other people do.

PasqualeMainolfi:
I’m glad to hear that the idea resonates with you.

Either solution would work for me. My initial thought is that starting as a 3rd-party plugin might be the most practical approach.

[...]

PR #2587, opened 2026-07-08
Add stm: state-machine graph opcodes for Csound

PasqualeMainolfi:
# Add `stm`: state-machine graph opcodes for Csound

## Summary

This PR introduces **`stm`**, a small opcode family for declaring and running
stateful computational graphs directly inside a Csound orchestra.

The model is inspired by [LangGraph](https://github.com/langchain-ai/langgraph):
the graph owns the control-flow structure, while each node's actual computation
remains ordinary Csound code. In practice:

- the **graph** declares named nodes, legal edges, the entry node, and the
  current state;
- the **orchestra** implements what each node does, using UDOs, instruments,
  shared state, DSP, analysis, MIDI/OSC input, or any other Csound logic.

This keeps transitions explicit and validated without forcing a new execution
model onto existing Csound code.

## Motivation

Csound patches that model evolving processes often encode state transitions with
`if`/`goto`, flags, counters, and scattered conditionals. That works for small
patches, but it becomes hard to answer basic questions:

- What are the valid states?
- Which transitions are legal?
- Where does the process start?
- Can this branch ever jump to an invalid state?

`stm` gives that control-flow structure a first-class home. A transition is only
applied if it matches a declared edge, so illegal moves are rejected by
construction. Once compiled, the graph structure is immutable, keeping the
runtime state space stable and inspectable.

Typical targets include adaptive music systems, gesture-driven patches,
analysis-to-synthesis pipelines, intro/loop/outro form control, and agent-like
musical behaviors.

## Core Model

- **Graph**: registry-managed `i`-time handle containing nodes, edges, entry,
  and current state.
- **Node**: named state. The node's behavior is user code, commonly a UDO or a
  bound instrument.
- **Shared state struct**: user-owned data passed to node code. `stm` tracks
  which node is current; fields such as RMS, density, gesture values, counters,
  or synthesis parameters live in a normal Csound struct.
- **Edge**: legal transition from one node to another.
- **Conditional edge**: legal set of possible targets; the running node chooses
  one target at k-time.
- **Entry node**: initial node, defaulting to the first added node unless
  overridden with `stmentry`.
- **Request / advance**: node code calls `stmnext`; `stmadvance` validates and
  applies the pending request.

The canonical runtime loop is:

```text
stmcurrent -> dispatch current node -> node calls stmnext -> stmadvance
```

`stmadvance` is the gatekeeper. A node may request any target, but the graph only
changes state when the requested transition is legal.

`stm` deliberately does not own the patch's musical or analytical state. The
graph stores control state: current node, pending request, entry, and compiled
topology. The shared working state stays in user code, usually as a typed struct
that is passed to every node implementation. This makes the state machine
explicit without hiding DSP, analysis, or musical data inside the graph.

### Build API

Build opcodes run at i-time before `stmcompile`.

| Opcode | Signature | Purpose |
| --- | --- | --- |
| `stmcreate` | `ig = stmcreate()` | Create a graph and return its handle. |
| `stmaddnode` | `stmaddnode(g:i, name:S)` | Add a named node. |
| `stmaddedge` | `stmaddedge(g:i, from:S, to:S)` | Add one legal transition. |
| `stmaddcondedge` | `stmaddcondedge(g:i, from:S, targets:S[])` | Add a set of legal targets from one source node. |
| `stmentry` | `stmentry(g:i, node:S)` | Set the entry node. Defaults to the first node. |
| `stmcompile` | `stmcompile(g:i)` | Validate and freeze the graph structure. |

### Runtime API

Runtime opcodes operate at k-time.

| Opcode | Signature | Purpose |
| --- | --- | --- |
| `stmcurrent` | `name:S = stmcurrent(g:i)` | Return the current node name. |
| `stmcurrentid` | `id:k = stmcurrentid(g:i)` | Return the current node id. |
| `stmnext` | `stmnext(g:i, name:S)` / `stmnext(g:i, id:k)` | Request a transition by name or id. |
| `stmadvance` | `kchanged = stmadvance(g:i)` | Apply the pending request if it is legal. Returns `1` when the node changed. |
| `stmreset` | `stmreset(g:i)` | Return the graph to its entry node. |
| `stmonenter` | `trig:k = stmonenter(g:i, node:S)` | Return `1` on the k-cycle when `node` becomes current. |
| `stmonexit` | `trig:k = stmonexit(g:i, node:S)` | Return `1` on the k-cycle when `node` stops being current. |

### Introspection API

| Opcode | Signature | Purpose |
| --- | --- | --- |
| `stmnodecount` | `n:k = stmnodecount(g:i)` | Return the number of nodes. |
| `stmedgecount` | `e:k = stmedgecount(g:i)` | Return the total number of edges. |
| `stmnodeid` | `id:k = stmnodeid(g:i, name:S)` | Resolve a node name to an id. |
| `stmnodename` | `name:S = stmnodename(g:i, id:k)` | Resolve a node id to a name. |

### 1. Declare a graph

```csound
graph:i = stmcreate()

stmaddnode(graph, "Analyze")
stmaddnode(graph, "Texture")
stmaddnode(graph, "Freeze")

stmaddedge(graph, "Analyze", "Texture")
stmaddedge(graph, "Texture", "Freeze")

; Freeze may branch to either Texture or Analyze.
stmaddcondedge(graph, "Freeze", ["Texture", "Analyze"])

stmcompile(graph)
```

### 2. Define shared state

The graph decides which node may run next. The data that nodes read and mutate
belongs to the patch. A typical pattern is to keep that data in a struct and pass
the same instance to every node:

```csound
struct AudioState rms:k density:k freeze:k
state@global:AudioState = init(0, 0, 0, 0.05, 0)
```

In this example, `AudioState` is the shared state object for the process:

- analysis code updates `state.rms`;
- node code reads `state.rms` to choose transitions;
- synthesis code reads fields such as `state.density` and `state.freeze`;
- each node can update the same struct before requesting the next state.

This keeps two kinds of state separate: `stm` owns the current graph node, while
`AudioState` owns the patch-specific data that survives across node calls.

### 3. Dispatch nodes to UDOs

A node can be an ordinary UDO that reads or writes shared state and requests a
transition:

```csound
opcode FreezeNode(g:i, st:AudioState):void
    st.density = 1

    if st.rms < 0.05 then
        stmnext(g, "Analyze")
    else
        stmnext(g, "Texture")
    endif
endop
```

The supervisor asks the graph which node is current, runs the matching node code,
then advances the graph:

```csound
opcode dispatch(g:i, st:AudioState, cur:S):void
    if strcmpk(cur, "Analyze") == 0 then
        AnalyzeNode(g, st)
    elseif strcmpk(cur, "Texture") == 0 then
        TextureNode(g, st)
    elseif strcmpk(cur, "Freeze") == 0 then
        FreezeNode(g, st)
    endif
endop

instr 1
    cur:S = stmcurrent(graph)
    dispatch(graph, state, cur)
    changed:k = stmadvance(graph)
endin
```

If `FreezeNode` requests `"Texture"`, the transition is accepted because
`Freeze -> Texture` was declared. If it requests an undeclared target, the
request is ignored and `stmadvance` returns `0`.

### 4. Bind nodes to instrument lifecycles

When a node should own a long-running process instead of per-cycle UDO code, use
the enter/exit triggers to spawn and stop instruments:

```csound
instr supervisor
    if stmonenter(graph, "A") == 1 then
        schedulek(30, 0, -1)
    endif

    if stmonexit(graph, "A") == 1 then
        turnoff2(30, 0, 0)
    endif

    changed:k = stmadvance(graph)
endin
```

This pattern is useful when the node represents a sustained layer, voice,
texture, or process with its own lifecycle.

### 5. Dispatch by id and inspect the graph

For larger graphs, id-based dispatch avoids repeated string comparisons:

```csound
id:k   = stmcurrentid(graph)
name:S = stmnodename(graph, id)

next:k = stmnodeid(graph, "Freeze")
stmnext(graph, next)

n:k = stmnodecount(graph)
e:k = stmedgecount(graph)
```

## Binding Styles

`stm` does not prescribe how node behavior is implemented. It only answers
"which node is current?" and validates requested transitions.

| Style | Node behavior | Use when |
| --- | --- | --- |
| Name dispatch | UDO selected with `stmcurrent` and `strcmpk` | The graph is small and readability matters. |
| Id dispatch | UDO selected with `stmcurrentid` | The graph is larger or string comparisons should be avoided. |
| Instrument binding | Instrument spawned on enter and stopped on exit | The node is a sustained process, layer, or voice. |

## Testing

`tests/commandline/test_stm.csd` covers:

1. deterministic transitions, including conditional edges and
   `stmonenter`/`stmonexit` trigger timing;
2. seeded random-drive invariants and coverage, verifying that all nodes are
   visited and transitions remain legal;
3. instrument lifecycle binding, asserting that only the current node's process
   is active;
4. introspection, entry/reset behavior, id-based transition requests, and
   rejection of illegal id-based moves.

All scenarios pass with `0 errors in performance`.

## Notes and Limitations

- Graph structure is immutable after `stmcompile`.
- Nodes, edges, and entry must be declared at i-time before compilation.
- `stmadvance` applies at most one pending transition per call.
- Illegal transition requests are dropped and reported as `changed = 0`.
- Exact-cycle tests involving `schedulek` or `turnoff2` should use their own time
  slot. When a lower-numbered instrument mutates the active instrument list in a
  k-cycle, a higher-numbered instrument can skip that cycle's body. This is
  general Csound scheduling behavior, not specific to `stm`.

If you like the direction of this proposal and there are other capabilities you
would want to see in `stm`, please let me know.


PasqualeMainolfi:
## Add internal graph clock

This commit introduces an internal runtime clock for STM graphs.
Each graph now maintains its own notion of time, independent of Csound's global performance clock. The runtime exposes three new k-rate opcodes:

- `stmtick`: number of execution steps since `stmcompile` or `stmreset`
- `stmtime`: elapsed graph time in seconds
- `stmnodetime`: elapsed time since the current node became active

Unlike `times()` or `timeinstk()`, the STM clock advances only when `stmadvance` is executed. Time therefore becomes a property of the graph runtime rather than the orchestra scheduler, allowing deterministic time-driven behaviours to be expressed directly within the state machine.
Both graph time and node time are derived from the internal tick counter rather than accumulated incrementally, preventing long-term floating-point drift while remaining consistent with instruments using local `setksmps`.
This feature enables state transitions and orchestration logic based on the graph's own execution timeline, without requiring user-defined timers in the shared state.

CMakeLists.txt
```diff
@@ -1136,6 +1136,7 @@ set(stdopcod_SRCS
     Opcodes/wterrain2.c
     Opcodes/stdopcod.c
     Opcodes/dbap.c
+    Opcodes/stm.c
```
hlolli: Would it make sense to incubate this as a standalone plugin before adding it to `stdopcod_SRCS`? My concern is that this introduces 21 public opcode names and a long-term compatibility surface while the lifetime, multicore, clock, and trigger contracts are still evolving. A plugin could let the API mature through real Csound use, with core inclusion reconsidered once those contracts and the manual pages are settled.

hlolli (commented):
Thanks for the detailed rationale and examples. I built and exercised the branch, including targeted reproducers around graph lifetime, local `setksmps`, numeric IDs, and late observers. I am leaving these as non-blocking questions and suggestions for discussion.

Opcodes/stm.c
```diff
@@ -0,0 +1,676 @@
...
+static uint32_t stm_register_graph(CSOUND *csound, GRAPH *g) {
+    STM_REGISTRY *reg = stm_registry_query(csound);
+    if (reg == NULL) {
+        if (csound->CreateGlobalVariable(csound, STM_REGISTRY_NAME, sizeof(STM_REGISTRY)) != 0)
+            return 0;
+        reg = stm_registry_query(csound);
+        if (reg == NULL) return 0;
+        reg->capacity = INITIAL_GRAPH_CAPACITY;
+        reg->count = 0;
+        reg->graphs = csound->Calloc(csound, sizeof(GRAPH *) * reg->capacity);
```
hlolli: Could the registry have an explicit lifetime and slot-reuse strategy? Each graph receives a monotonically increasing slot, deinit only nulls that entry, and this nested `graphs` allocation has no reset cleanup. My concern is unbounded registry growth across repeated graph creation, plus the backing allocation surviving the named-global cleanup. A registry reset callback together with a safe slot/generation scheme, or an engine-owned graph lifetime, may make this ownership clearer.

Opcodes/stm.c
```diff
@@ -0,0 +1,676 @@
...
+
+    return OK;
+}
+
+/* ------------------------------------------------------------------ */
+/* opcodes                                                            */
+/* ------------------------------------------------------------------ */
+
+int32_t graph_create_deinit(CSOUND *csound, GRAPH_CREATE *p) {
+    GRAPH *g = stm_handle_to_graph(csound, *p->handle);
```
hlolli: Could the owning handle be stored privately in `GRAPH_CREATE` instead of being read back from `*p->handle` during deinit? This output is user-visible and can be overwritten by assignment or `reinit`; in a targeted reinit case I saw handles 1 through 8 created, with handle 1 still usable after the owner ended because only the final output value was freed. My concern is both leaked graphs and freeing a graph owned by another opcode instance. Keeping an `owned_handle` privately, and releasing any prior owned graph before reinitialization, would decouple ownership from the output variable.

Opcodes/stm.c
```diff
@@ -0,0 +1,676 @@
...
+        g->requested_node = NO_NODE;
+
+        if (target < g->node_count && graph_has_edge(node, target)) {
+            g->previous_node = g->current_node;
+            g->current_node = target;
+            changed = 1;
+        }
+    }
+
+    g->kperiod = (MYFLT) CS_KSMPS / CS_ESR;
```
hlolli: How should graph time behave when successive `stmadvance` calls use different local `setksmps` values? Because this replaces `kperiod`, `stmtime` and `stmnodetime` later multiply every historical tick by the latest period, retroactively rescaling earlier steps. In a targeted case, 33 ticks representing 64 samples reported `33 / 48000` seconds rather than `64 / 48000`. Would tracking a `uint64_t` total sample-frame count and node-entry sample count preserve drift-free timing while supporting varying control periods?

Opcodes/stm.c
```diff
@@ -0,0 +1,676 @@
...
+
+static OENTRY stm[] = {
+    { "stmcreate",      S(GRAPH_CREATE),        0, "i", "",      (SUBR) graph_create,        NULL,                     (SUBR) graph_create_deinit },
+    { "stmaddnode",     S(GRAPH_ADD_NODE),      0, "",  "iS",    (SUBR) graph_add_node,      NULL,                     NULL },
+    { "stmaddedge",     S(GRAPH_ADD_EDGE),      0, "",  "iSS",   (SUBR) graph_add_edge,      NULL,                     NULL },
+    { "stmaddcondedge", S(GRAPH_ADD_COND_EDGE), 0, "",  "iSS[]", (SUBR) graph_add_cond_edge, NULL,                     NULL },
+    { "stmcompile",     S(GRAPH_COMPILE),       0, "",  "i",     (SUBR) graph_compile,       NULL,                     NULL },
+    { "stmcurrent",     S(GRAPH_CURRENT),       0, "S", "i",     NULL,                       (SUBR) graph_current,     NULL },
+    { "stmcurrentid",   S(GRAPH_CURRENT_ID),    0, "k", "i",     NULL,                       (SUBR) graph_current_id,  NULL },
+    { "stmadvance",     S(GRAPH_ADVANCE),       0, "k", "i",     NULL,                       (SUBR) graph_advance,     NULL },
```
hlolli: Is sharing one graph handle across instruments intended to be supported under multicore performance (`-j`)? The graph is mutated behind an i-rate handle, all of these OENTRY interlock flags are `0`, and no mutex protects `requested_node`, `current_node`, the registry, or graph lifetime. My concern is that the scheduler cannot see these hidden writes, so a driver, observers, and deinit may race, including a possible resolve/free race. Could ownership be restricted and enforced to one supervisor, or could graph access be synchronized and represented in the scheduler interlocks?

Opcodes/stm.c
```diff
@@ -0,0 +1,676 @@
...
+    g->requested_node = (uint32_t) requested_node;
+    return OK;
+}
+
+int32_t graph_next_id(CSOUND *csound, GRAPH_NEXT_ID *p) {
+    GRAPH *g = stm_handle_to_graph(csound, *p->handle);
+    if (g == NULL)
+        return csound->PerfError(csound, &(p->h), "[stm] stmnextid: invalid graph");
+
+    int32_t requested_node = (int32_t) *p->next_node;
```
hlolli: Could numeric IDs be validated as finite, exact integers before conversion? This cast currently truncates values such as `1.75` to node `1` before the range check; the handle resolver and `stmnodename` have the same fractional-value behavior, while NaN or out-of-range conversion is undefined. My concern is that malformed k-rate input silently selects a different node or graph. A shared checked-conversion helper used by handles and node IDs would keep these paths consistent.

Opcodes/stm.c
```diff
@@ -0,0 +1,676 @@
...
+    GRAPH *g = stm_handle_to_graph(csound, *p->handle);
+    if (g == NULL) {
+        return csound->PerfError(csound, &(p->h), "[stm] stmonenter: invalid graph (freed?)");
+    }
+    if (!g->compiled) {
+        return csound->PerfError(csound, &(p->h), "[stm] stmonenter: graph not compiled");
+    }
+
+    int32_t is_curr = g->current_node == (uint32_t) p->node_id;
+    *p->trig = is_curr && !p->was_current ? FL(1.0) : FL(0.0);
```
hlolli: Should `stmonenter` describe an actual graph transition, or an observer-local rising edge? Since `was_current` starts at zero, an observer scheduled long after node B became current still receives `stmonenter(B) == 1` on its first pass; an observer can also miss transitions that occur between its passes. My concern is duplicate or omitted process lifecycle actions in the binding example. A graph-level transition sequence/timestamp could let observers identify real enter/exit events, or the observer-local semantics could be made explicit in the API and documentation.

PasqualeMainolfi:
@hlolli thank you for the thoughtful feedback.
My own view is that this opcode family is a good fit for Csound’s standard opcode set. I see STM as a general control-flow and orchestration facility rather than a domain-specific extension, and I think that kind of functionality integrates naturally with the Csound ecosystem.
A few days ago I published another opcode family as a standalone plugin [semsys](https://github.com/PasqualeMainolfi/csound-semsys) (see #2568), and I completely agree that it is well suited to live outside the core. STM feels different to me in that it aims to provide a general execution model for structuring stateful musical processes, rather than adding a specific processing capability.
That said, I completely understand the concern about committing to a public API before the semantics have fully settled. If the consensus is that STM should first mature as a standalone plugin, I’m perfectly happy to take that route. It would provide the opportunity to validate the API and refine the runtime contracts before considering inclusion in the core.

I’ll also start addressing your comments on the runtime semantics and ownership model as soon as possible. They raise important points, and I’d like to make those aspects more robust before moving forward.

PasqualeMainolfi:
Based on @hlolli review.... this report summarizes the changes made to `Opcodes/stm.c`, `Opcodes/stm.h`, and the related STM tests. The goal is to make ownership, lifetime, slot reuse, handle validity, the internal clock, multicore synchronization, and the semantics of `stmonenter` / `stmonexit` explicit.

The STM registry does not semantically own graphs. It is a global table associated with a `CSOUND` instance through a global variable, and it is only used to resolve handles to live graphs.

Each `GRAPH` is owned by the `stmcreate` opcode instance that created it.

The handle returned to the orchestra is no longer just the registry slot index. It now encodes both the slot and a generation counter.

```text
create graph A in slot 0, generation 1 -> handle 1
deinit graph A -> slot 0 becomes empty, generation becomes 2
create graph B in slot 0, generation 2 -> handle 4097
use the old handle 1 -> lookup fails
```

This fixes the stale-handle problem that could happen when an old handle remained usable after the same slot was reused for a new graph.

I decided not to enforce a strict single-owner model, because cross-instrument use can be intentional: for example, one instrument can drive the graph while others observe its state, or multiple instruments can coordinate transitions on a shared graph. All main graph access now goes through a mutex.

Previously, time was computed by multiplying historical ticks by the latest known control period. This was risky when successive calls to `stmadvance` used different `setksmps` values, because changing `kperiod` retroactively rescaled previous ticks.

The old problem was:

```text
33 real ticks represent 64 sample frames
but using the latest period produced 33 / sr instead of 64 / sr
```

Now `stmadvance` does two separate things:

1. increments `graph_tick`, which remains a logical advance counter;
2. adds the current `CS_KSMPS` value to `total_sample_frames`.

Times are computed as:

```text
stmtime     = total_sample_frames / sr
stmnodetime = (total_sample_frames - node_sample_on_enter) / sr
```

Risky direct casts from `MYFLT` to integer types were removed for inputs that represent ids or indexes.

The final model is:

1. `stmcreate` creates and owns a `GRAPH`.
2. The graph is registered in a global STM slot associated with `CSOUND`.
3. The returned handle encodes both slot and generation.
4. Every handle lookup checks that:
   - the slot exists;
   - the slot still contains a graph;
   - the handle generation matches the slot generation.
5. When the owner is deinitialized:
   - the graph is freed;
   - the slot is invalidated;
   - the generation is incremented;
   - old handles can no longer resolve to the new slot contents.
6. All registry and graph access is synchronized through a mutex.
7. Graph time is computed from real sample frames, not from rescaled ticks.
8. `stmonenter` and `stmonexit` describe real graph transitions observed after the observer is initialized, not the observer's local initial state.

## Remaining limitations

- The mutex is registry-wide, so it also serializes access to different graphs. This is a conservative choice for correctness and lifetime safety.
- Synchronization is per opcode call, not across multi-opcode orchestra blocks.
- `stmonenter` / `stmonexit` are not an event queue: a slow observer can miss intermediate transitions.
- The practical limit is 4096 live graphs and 4096 generations per slot, chosen so handles remain exact even when `MYFLT` is single precision.
- The registry and its mutex have a lifetime tied to the `CSOUND` global variable; individual `GRAPH` objects are freed by the deinit callback of their owning `stmcreate`.

Opcodes/stm.c
```diff
@@ -0,0 +1,1027 @@
...
+        csound->Free(csound, g);
+        p->graph = NULL;
+        p->slot = 0;
+        p->generation = 0;
+        *p->handle = 0;
+    }
+    return OK;
+}
+
+int32_t graph_create(CSOUND *csound, GRAPH_CREATE *p) {
```
hlolli: Could `graph_create()` release an existing `p->graph` before installing the new one when this opcode instance is reached through `reinit`? The private owner state is a good improvement, but Csound reruns init callbacks without calling deinit first. In a small reproducer, handles 1 and 2 were created, and handle 1 still resolved after the owner ended. Perhaps the cleanup could be factored into a helper shared by deinit and repeated initialization.

Opcodes/stm.c
```diff
@@ -0,0 +1,1027 @@
...
+            g->previous_node = g->current_node;
+            g->current_node = target;
+            g->event_seq++;
+            g->event_exited_node = g->previous_node;
+            g->event_entered_node = target;
+            changed = 1;
+        }
+    }
+
+    if (g->reset_pending) {
```
hlolli: Would it make sense to associate `reset_pending` with the actual performance cycle in which the reset occurred? The sample-frame clock change looks good, but when I reset at 0.05 and did not advance again until 0.10, that later advance was still suppressed, leaving `stmtick` at zero for two driver passes. Tracking the reset cycle could preserve the intended same-cycle behavior without discarding a genuinely later advance.

tests/commandline/test.py
```diff
@@ -695,6 +695,9 @@ def runTest():
         ["test_dbap.csd", "test dbap and dbapgains opcodes"],
         ["test_generic_chan.csd", "testing generic bus channel"],
         ["test_generic_chan_no_match.csd", "testing type mismatch for bus channel", 1],
+        ["test_stm.csd", "testing stm opcdes"],
+        ["test_stm_lifetime.csd", "expected failure: stm graph observed after being freed", 1],
```
hlolli: Could this regression distinguish the intended controlled performance error from a crash? This is useful coverage, but the expected-failure handling accepts any nonzero return, so a `PerfError`, segfault, or ASan abort would all pass. Checking the expected diagnostic or exercising this through an API-level test would make it guard the original crash reliably.

hlolli:
On high level. Knowing that we want to utilise plugins more. I see things playing in favour of making it internal opcode.

1. written in 'c' with no additional dependencies
2. this are finite state operators, while users can build their own, this offers nice and readable abstraction.
3. Useful for learning, teaching and composing.

There are things that could be added later

* checkpointing;
* save and restore;
* interruption and resumption;
* retries and failure states;
* fan-out and fan-in;
* subgraphs;
* workflow cancellation.

Opcodes/stm.c
```diff
@@ -0,0 +1,1027 @@
...
+     become valid again just because the same slot was reused. */
+
+/* Query-only: never creates the registry (safe in deinit). */
+static STM_REGISTRY *stm_registry_query(CSOUND *csound) {
+    return (STM_REGISTRY *) csound->QueryGlobalVariable(csound, STM_REGISTRY_NAME);
+}
+
+static void stm_registry_lock(CSOUND *csound, STM_REGISTRY *reg) {
+    if (reg != NULL && reg->mutex != NULL)
+        csound->LockMutex(reg->mutex);
```
hlolli: Could the k-rate path avoid taking a blocking registry-wide mutex before this becomes part of the core opcode set? I realize that may require a substantial redesign, but this mutex gives the performance thread no upper bound on how long it can wait if another worker is descheduled while holding it. The lock also covers string lookup, edge traversal, and sometimes `ReAlloc`, so unrelated graphs can contend behind work whose duration is not strictly bounded. That creates a genuine priority-inversion and audio-dropout risk even though the memory accesses themselves are now race-free.

Would an immutable numeric runtime representation with one scheduler-ordered writer and atomically published state for observers be feasible, with reclamation deferred outside performance? The string APIs could remain as convenience wrappers. Addressing this later may be considerably harder because the current public API explicitly permits concurrent cross-instrument mutation, and changing that contract after users depend on it could be disruptive.

Opcodes/stm.c
```diff
@@ -0,0 +1,1027 @@
...
+        g->reset_pending = 0;
+    } else {
+        g->graph_tick++;
+        g->total_sample_frames += (uint64_t) CS_KSMPS;
+        if (changed) {
+            g->node_sample_on_enter = g->total_sample_frames;
+        }
+    }
+
+    *p->changed = (MYFLT) changed;
```
hlolli: Could `stmadvance` report why a transition did or did not occur rather than returning only `changed`? Centralizing edge validation here makes sense, but `0` currently represents both "nothing was requested" and "the requested edge was illegal," while a legal self-edge returns `1` even though the current node ID did not change. Because requests occupy a single shared slot, one writer replacing a request from another writer is also invisible. In a live interactive system, these cases can make configuration and coordination errors look like ordinary inactivity.

Would a status result such as `STM_CHANGED`, `STM_NO_REQUEST`, `STM_ILLEGAL_EDGE`, `STM_ALREADY_CURRENT`, `STM_CONFLICT`, or `STM_RESET`, potentially accompanied by the source and target IDs, make the contract clearer? Invalid handles and unknown node names currently become performance errors earlier in the path, so those would need a deliberate and consistent error policy. If concurrent writers remain supported, reporting conflicts would also require request sequencing or writer identity rather than only the current `requested_node` slot. This seems worth settling before the opcode API becomes public, since extending the return signature later could be disruptive.

Opcodes/stm.c
```diff
@@ -0,0 +1,1027 @@
...
+    if (stm_get_graph_perf_locked(csound, &(p->h), *p->handle,
+            "[stm] stmnext: invalid graph", &reg, &g) != OK) return NOTOK;
+
+    int32_t requested_node = graph_find_node(g, p->next_node->data);
+    if (requested_node < 0) {
+        stm_registry_unlock(csound, reg);
+        return csound->PerfError(csound, &(p->h), "[stm] stmnext: node '%s' not found", p->next_node->data);
+    }
+
+    g->requested_node = (uint32_t) requested_node;
```
hlolli: Could the request-conflict semantics be defined explicitly rather than inherited from assignment order here? At present, the last `stmnext` call to acquire the mutex replaces any earlier request, making the winner potentially nondeterministic under multicore execution. `stmadvance` clears the request before validating it, so rejected requests are dropped; calling `stmadvance` twice advances the clock twice; and a self-request only succeeds with an explicit self-edge, after which it emits an exit and entry for the same node. The result of combining `stmreset`, `stmnext`, observers, and `stmadvance` within one cycle also depends on their call order.

Since the API intentionally permits multiple instruments to mutate a graph, would it be worth specifying whether requests use first-writer, last-writer, conflict-error, or queued semantics, along with when observer-triggered requests take effect? Whichever model is chosen could then be covered by focused tests for self-transitions, reset/request ordering, repeated advances, rejected requests, and competing writers. These are observable public state-machine semantics, and changing them after users depend on the current incidental behavior could be difficult.

Opcodes/stm.c
```diff
@@ -0,0 +1,1027 @@
...
+    }
+
+    if (!g->compiled) {
+        stm_registry_unlock(csound, reg);
+        return csound->PerfError(csound, &(p->h), "[stm] stmonenter: graph not compiled");
+    }
+
+    int32_t entered = g->event_seq > p->last_seen_event_seq && g->event_entered_node == (uint32_t) p->node_id;
+    *p->trig = entered ? FL(1.0) : FL(0.0);
+    p->last_seen_event_seq = g->event_seq;
```
hlolli: Could observers detect that transition events were skipped before advancing `last_seen_event_seq` directly to the latest sequence here? The graph-level sequence is a good fix for the original false-enter behavior, and I see that retaining only the latest event is documented. However, an `A -> B -> C` sequence between observer passes means an observer waiting for `stmonexit(A)` sees only the latest `B -> C` event and may leave the process associated with A running indefinitely.

Would a single atomic snapshot opcode returning the latest sequence, source, target, and status be useful? An observer could then detect `sequence > previous_sequence + 1` and reconcile with the current graph state. A single multi-output opcode would avoid separate getters observing different transitions. If replay is needed later, a small preallocated ring buffer with an explicit overflow result could provide bounded history without allocation or an unbounded event queue.

Opcodes/stm.h
```diff
@@ -0,0 +1,379 @@
...
+
+typedef struct {
+    char *name;
+    uint32_t id;
+    uint32_t *edges; // node id in graph
+    uint32_t edge_count;
+    uint32_t edge_capacity;
+} GRAPH_NODE;
+
+typedef struct {
```
hlolli: Could the immutable graph definition be separated from each mutable execution instance before the public API is settled? The compile-and-freeze boundary is a good start, but `GRAPH` currently combines nodes, edges, names, and entry configuration with the current node, pending request, clock, and event sequence. Consequently, reusing one topology requires rebuilding it, while sharing its handle also shares all runtime state and synchronization.

Would a compiled definition plus runner model be feasible, for example:

```csound
idefinition = stmcompile(ibuilder)
irunner1 = stminstance(idefinition)
irunner2 = stminstance(idefinition)
```

The compiled definition could be immutable and safely shared, while each runner would own independent numeric state, timing, requests, and transition events. With explicit reference-counted or registry-managed definition lifetime, this would also decouple topology lifetime from one supervisor instrument and make lock-free topology access and single-writer runner semantics easier to establish. This is admittedly a larger change, but introducing the distinction later may be difficult once the current handle is publicly understood to represent both a definition and its execution state.

hlolli:
@PasqualeMainolfi btw these first 3 reviews before my comment are more blocking that 3 comments after my last comment there, a they are more speculative about the design. Feel free to argue against them, and that would be fine too from my side.

vlazzarini:
> On high level. Knowing that we want to utilise plugins more. I see things playing in favour of making it internal opcode.
> 
> 1. written in 'c' with no additional dependencies
> 2. this are finite state operators, while users can build their own, this offers nice and readable abstraction.
> 3. Useful for learning, teaching and composing.
> 
> There are things that could be added later
> 
> * checkpointing;
> * save and restore;
> * interruption and resumption;
> * retries and failure states;
> * fan-out and fan-in;
> * subgraphs;
> * workflow cancellation.

Since this is added to Opcodes, it is not internal and may be built as a plugin lib or included in the library at compile time. In fact we need to check that it does not break the plugin build (not sure if one of the CI builds tests that, we may need to add one specifically).

PasqualeMainolfi:
@hlolli I think your review has opened a really interesting direction.
My original idea was to model STM as a classic finite state machine, where the graph itself represented both the topology and the current execution state. With your suggestion of separating a builder/definition from mutable instances, the design starts moving toward a proper workflow engine. I think that’s an extremely powerful and versatile direction.
You’re reasoning in terms of a runtime, and I really like that perspective. The instance remains the finite state machine—it owns the current state and execution. The definition, on the other hand, is no longer a state machine in itself; it’s simply the states and transitions, the “program” that describes the workflow.
That separation between describing the machine and executing it is really elegant, and I think it provides a much stronger foundation for future extensions and reuse.

I actually preferred to start with this proposal first, and then work my way toward the smaller ones.

# STM follow-up changes

Six areas:

1. safe ownership across repeated initialization;
2. an explicit builder/definition/runner object model with typed handles;
3. reset behavior tied to the actual performance cycle;
4. explicit `stmadvance` results with deterministic conflict handling;
5. bounded transition-event replay through `stmevent`;
6. a lock-free single-writer/multi-reader performance path.

The public construction sequence is:

```csound
ibuilder = stmcreate()
; stmaddnode, stmaddedge, stmentry, ...
idefinition = stmcompile(ibuilder)
irunner = stminstance(idefinition)
```

Runtime opcodes (`stmcurrent`, `stmnext`, `stmadvance`, `stmreset`, the time queries, observers, and `stmevent`) accept only runner handles; builder and definition handles are rejected.

## 1. Safe ownership across deinit and reinit

Each object-producing opcode (`stmcreate` → builder, `stmcompile` → definition, `stminstance` → runner) now shares one cleanup helper between the start of its init callback and its deinit callback. The helper clears the public output handle first, if the replacement allocation then fails, the output holds 0 rather than a stale handle, and validates a private owner token (slot, generation, object type) before invalidating the registry slot and advancing its generation. A delayed deinit therefore can never destroy a different object that has reused the same slot.

Runtime opcodes follow the same repeated-init rule: their init first releases any runner reference and writer claim already held by that opcode instance, then resolves and retains the new runner.

## 2. Builder, immutable definition, mutable runner

The earlier monolithic `GRAPH` mixed topology with execution state, so reusing a topology meant rebuilding it, and two independent executions could not share compiled nodes and edges. There are now three object roles:

- **`GRAPH_BUILDER`** (mutable): nodes, edges, entry configuration. Topology opcodes operate only on builders. `stmcompile` validates, deep-copies, and freezes it; adding nodes or edges, or changing the entry, after compilation is rejected.
- **`GRAPH_DEFINITION`** (immutable): copied nodes, names, edge arrays, entry ID, and a reference count. One definition may back many runners; each runner retains it, so it outlives the instrument owning the public definition handle.
- **`GRAPH_RUNNER`** (mutable): current/previous node, pending request and conflict marker, clocks, reset-cycle tracking, latest enter/exit info, event ring, writer ownership, and snapshot version. Two runners of one definition share topology but no execution state: resetting or advancing one cannot affect the other.

Handles remain numeric `MYFLT` values, but every registry slot records its object kind, generation, and a tagged union. Lookup validates the numeric encoding, slot bounds, generation, expected type, and non-null object, so passing a builder where a definition is required produces a controlled type-specific init error instead of an invalid cast. Handles are encoded within `2^24`, exact even in single-precision builds, giving 4096 slots with 4096 generations each.

## 3. Reset suppression follows the performance cycle

`stmreset` restarts the runner clocks. A previous boolean flag made the *next* `stmadvance` skip its clock increment regardless of when it ran, call order, not time, so an advance genuinely later than the reset was still suppressed.
Reset now records the engine performance-cycle counter. Advance suppresses its clock increment only when it runs in that same engine cycle; the first advance in any later cycle counts normally, even if no advance occurred during the reset cycle.

The clock remains advance-driven: every non-suppressed `stmadvance` adds one tick plus the writer's current `CS_KSMPS` sample frames, so multiple advances in one cycle tick the clock multiple times.

## 4. Explicit advance status and request conflicts

```csound
status:k, from:k, to:k = stmadvance(irunner)
```

| Value | Status | Meaning |
| ---: | --- | --- |
| 0 | `STM_NO_REQUEST` | No transition target was pending. |
| 1 | `STM_CHANGED` | A legal edge was applied and changed the current node ID. |
| 2 | `STM_ILLEGAL_EDGE` | A target was requested but no corresponding edge exists. |
| 3 | `STM_SELF_TRANSITION` | A legal explicit self-edge was applied. |
| 4 | `STM_CONFLICT` | Different targets were requested before the advance; none was applied. |

`from:k` reports the evaluated source node. `to:k` reports the unique requested target, or `-1` for `STM_NO_REQUEST` and `STM_CONFLICT`. Invalid handles, unknown node names, and fractional or out-of-range IDs remain controlled errors rather than transition statuses.
Requests share one pending slot: the first target is retained, repeating it is idempotent, and requesting a different target marks a conflict that further requests cannot undo. A conflicted advance applies neither target and clears the pending state. The temporarily retained first target deliberately does not win, hat would expose call or scheduling order as a hidden priority policy; priority or queueing belongs in supervisor logic. The pending request survives sequential writer handoff, so requests from successive writer instruments conflict the same way; *overlapping* writer instruments are instead rejected at init.

Only `STM_CHANGED` and `STM_SELF_TRANSITION` update previous/current state, restart node time, and record an event (a self-edge is a real exit and entry of the same node). The other statuses still consume the pending request and still advance the runner clock unless suppressed by a same-cycle reset. Reset clears any pending request: request-then-reset discards the request, reset-then-request evaluates it.

## 5. Bounded transition history and `stmevent`

`stmonenter` and `stmonexit` are lightweight latest-event observers: they avoid false initial triggers, but transitions occurring between two passes of a slow observer are missed without indication. `stmevent` adds loss detection and ordered consumption.

Each runner preallocates a ten-record transition ring at creation, no performance-time allocation. Records hold a sequence number, source, target, and an event status separate from the advance statuses:

| Value | Event status | Meaning |
| ---: | --- | --- |
| 1 | `STM_EVENT_CHANGED` | An accepted transition changed node ID. |
| 2 | `STM_EVENT_SELF_TRANSITION` | An accepted self-edge exited and entered the same node. |
| 3 | `STM_EVENT_RESET` | Reset moved the runner from a non-entry node to its entry node. |

No-request, illegal-edge, and conflict outcomes occupy no ring slot; neither does a reset that is already on the entry node.

```csound
status:k, sequence:k, overflow:k, available:k, from:k, to:k = stmevent(irunner)
```

There is no shared read index: every `stmevent` instance owns a private cursor, initialized past the latest event, so observation is future-only. Per pass: nothing new returns `available = 0`; a cursor older than the oldest retained record reports `overflow = 1` and the same call returns that oldest event; otherwise one coherent event is returned with `available = 1` and the cursor advances past it. `stmonenter`/`stmonexit` remain appropriate for simple same-supervisor scheduling; consumers that must detect loss should use `stmevent`.

## 6. Lock-free runner performance path

The concurrency contract is: one active top-level writer instrument per runner, any number of concurrent observers, multiple mutating opcode instances inside that writer, and sequential writer handoff after the previous writer deinitializes. `stmnext` (string and numeric), `stmadvance`, and `stmreset` claim the writer role at init; claims made inside UDOs are normalized to the top-level instrument. An overlapping writer instrument fails init with

```text
[stm] runner already has another active writer
```

which removes nondeterministic multicore write ordering from the public contract.

**Runner pinning.** Every runtime opcode resolves its typed handle once at init, retains the runner, and releases it at deinit; performance callbacks use the retained pointer and never acquire the registry mutex. Removing the public registry handle does not invalidate already initialized opcodes, the runner and its definition are freed only after every reference is released, and a new opcode initializing with the stale handle gets a controlled init error.

The resulting lifetime distinction is: `runner still alive != public handle still resolvable`. Retained references protect existing users from use-after-free; they are not transferable handles for creating new users. If new observers or a replacement writer must initialize later, the instrument owning `stminstance` must still be active, in practice, `stminstance` belongs in a long-lived supervisor whose lifetime covers every writer handoff and dynamically started observer.

**Snapshots.** The runner carries a `state_version` counter. The writer brackets compound updates (advance, reset, clocks, latest event, ring publication) between two increments, odd while writing, even when published, and publishes fields with atomic stores. Observers copy the fields they need with atomic loads and retry if the version was odd or changed: a seqlock. Readers never take a lock and never block the writer on atomic builds. The protocol is confined to snapshot helpers (`stm_latest_event_snapshot`, `stm_node_time_snapshot`, `stm_transition_snapshot`) and the writer's `stm_runner_update_begin/end`; opcode implementations never touch versions, fences, or retry loops. Single-field queries use plain atomic loads. Compound values are therefore always coherent — an event's sequence can never be paired with another event's source or target. Reference counts are atomic as well.

**Remaining performance-time allocation.** `stmcurrent` and `stmnodename` may grow their opcode-local `STRINGDAT` output the first time a node name does not fit. That reallocation is per opcode instance, lock-free, and bounded by the longest name in the immutable definition; the numeric queries (`stmcurrentid`, `stmnodeid`) avoid it entirely.

**NOTE on portability.** Without `HAVE_ATOMIC_BUILTIN`, each runner owns a fallback state mutex preserving the same snapshot contract. The registry mutex remains limited to registry mutation, typed lookup at init, writer claims, and deinit bookkeeping.

## 7. Regression coverage

Expected-failure tests require return code 1 plus the exact STM diagnostic, and forbid crash and sanitizer markers.

- `test_stm.csd` — dispatch, accepted/rejected transitions, enter/exit, introspection, graph clocks, variable local `ksmps`, all advance statuses, reset/request ordering, repeated advances, idempotent requests, and conflicts.
- `test_stm_sequential_trace.csd` — a complete workflow example: a `WorkflowState` struct passed by reference to node UDOs, a dispatch UDO, one supervisor pass per cycle. Eight printed, fully asserted cycles cover start/process/retry/done, a self-edge, no-request, reset, and restart — demonstrating the intended boundary: STM owns control flow, node UDOs own application state.
- `test_stm_reinit_reset.csd` — reinit releases and replaces the owned object (slot reuse with a new generation); same-cycle reset suppression versus a genuinely later advance.
- `test_stm_instances.csd` — two runners share one definition with independent state; a runner retains its definition after the owner ends. Ends with an overload burst: 300 runners created in one control cycle (growing the registry), each driven for 200 cycles by its own writer with deterministic assertions, plus a wall-clock report making the real-time margin visible.
- `test_stm_type_mismatch.csd`, `test_stm_invalid_id.csd`, `test_stm_unknown_node.csd` — controlled errors for a builder passed where a definition is required, fractional node IDs, and unknown node names.
- `test_stm_long_names.csd` — a 621-character node name exercises the `STRINGDAT` growth path of `stmcurrent`/`stmnodename` and the long-name lookups of `stmnext`/`stmnodeid`.
- `test_stm_event.csd` — event ring: empty reads, queued events, self-transition and reset events, wraparound, overflow recovery, two independent readers.
- `test_stm_lifetime.csd` — a retained observer stays safe after the public handle is removed; a new observer on the stale handle gets the controlled init error.
- `test_stm_lockfree.csd` — `-j2`: one writer with two concurrent observers; snapshot validity and monotonicity asserted without depending on scheduler order.
- `test_stm_parallel_runners.csd` — `-j2`: two writers concurrently drive two independent runners of one shared definition with fully deterministic per-runner assertions; unrelated graphs do not contend.
- `test_stm_writer_conflict.csd` — `-j2`: an overlapping writer instrument is rejected at init with the active-writer diagnostic.


...for now, I’ve focused on this larger architectural change. My next step will be to start implementing them on top of this new design.

PasqualeMainolfi:
One more thing, while working on Opcodes/stm.c, I needed to include H/udo.h to use OPCOD_IOBUFS. However, udo.h uses the private OPCODINFO typedef, whose definition is only available through cs_internal.h. Since stm.c is also built as part of the opcode plugin, this causes the plugin build to fail. Would it make sense to replace the pointer occurrences in udo.h with the opaque struct tag `(struct opcodinfo *)`? This would keep the implementation private while allowing udo.h to be included without depending on `cs_internal.h`.  Is `udo.h` intentionally meant to remain strictly core-only, or would it be appropriate for me to open a small PR for this change?

vlazzarini:
OPCODINFO is not part of the API, so you either use an opaque pointer if it works for you or you have to move your opcodes to become internal (in `OOPS/*.c` and `entry.c`). Opcodes in `Opcodes/*.c` can *only* use the API via the CSOUND object public functions (csound->*()). They should not be linked to the Csound library in any form.

PasqualeMainolfi:
@vlazzarini, thanks. 
While implementing Opcodes/stm.c, I needed to identify the top-level instrument instance so that STM opcodes called from nested UDOs are recognized as belonging to the same writer. The current implementation forward-declares OPCODINFO as an opaque type and includes udo.h only to follow OPCOD_IOBUFS::parent_ip. It does not call or link against any internal Csound function.  Does this match the opaque-pointer approach you suggested, or is accessing OPCOD_IOBUFS::parent_ip itself considered an internal API dependency and therefore not allowed from Opcodes/stm.c?

vlazzarini:
anything you can access safely through the API is fair game. So if you can make it work for your purposes it's fine.

PasqualeMainolfi:
@hlolli I have implemented some of the features you proposed. For the time being, I believe the remaining ones are either redundant or too specific to justify inclusion. Below is a description of the features that have been implemented.

# STM feature 1: stmcall and stmrecall

## Overview

This feature adds named checkpoints to an STM runner. A checkpoint freezes the
runner's logical execution state at k-rate and can later restore that state by
name.

Two opcodes form the public orchestra interface:

    kcaptured stmcall   irunner, Sname, ktrigger
    krecalled stmrecall irunner, Sname, ktrigger

The call/recall terminology distinguishes checkpoint capture and restoration
from runner control: stmpause and stmresume are reserved for pausing and
continuing the live runner.

The checkpoint belongs to the runner identified by irunner. It is not attached
to the graph builder or to the immutable compiled graph definition.

The interface intentionally exposes names rather than checkpoint buffer
positions. A ring position would not be a stable identifier because the slot
can be reused when the bounded buffer wraps.

## Opcode arguments

| Argument | Rate/type | Meaning |
| --- | --- | --- |
| irunner | i-rate handle | Runner returned by stminstance. |
| Sname | string | Checkpoint identifier. It must contain between 1 and 63 bytes, excluding the terminating NUL. |
| ktrigger | k-rate value | The operation is requested only when the value is exactly 1. Other values, including other non-zero values, do not trigger it. |
| kcaptured | k-rate output | 1 when the capture request succeeds, or when that opcode instance has already handled the same last name; otherwise 0. |
| krecalled | k-rate output | 1 when the checkpoint recall succeeds, or while the same already-handled recall trigger remains high; otherwise 0. |

Both outputs are reset to 0 at the beginning of every performance call. If an
operation reports a performance error, its output remains 0.

## State stored by stmcall

Each successful capture stores the following mutable runner fields:

| Stored field | Observable effect |
| --- | --- |
| current_node | Restores the active graph node. |
| previous_node | Restores the previous-node state used by the runner. |
| requested_node | Restores any pending node request. |
| request_conflict | Restores the conflict state associated with the pending request. |
| graph_tick | Restores the value observed through stmtick. |
| total_sample_frames | Restores the runner's elapsed graph time. |
| node_sample_on_enter | Restores the reference point used to calculate time spent in the current node. |

Capture occurs at the point where stmcall is executed in the instrument.
Therefore opcode order within the k-cycle is significant. For example, placing
stmcall after stmadvance captures the state produced by that advance.

The following state is deliberately not checkpointed:

- the compiled graph definition;
- checkpoint storage itself;
- the transition ring and its write position;
- the transition event sequence;
- graph-reset cycle bookkeeping;
- runner references, lifetime data, or writer ownership.

The graph definition is shared, immutable runner input rather than execution
state. Event history remains live so that observers never see an event sequence
move backwards after a checkpoint recall.

## Naming and bounded storage

Every runner allocates a zero-initialized ring of 100 STM_CHECKPOINT entries.
The ring is released with the runner.

Checkpoint names are unique among the currently valid entries in that runner:

- an empty name is rejected;
- a name of 64 bytes or more is rejected;
- capturing a name that is already present in a valid slot is rejected;
- no two valid ring entries can therefore have the same name.

After 100 distinct captures, the next successful capture overwrites the slot at
the ring's current write position. Consequently, the checkpoint previously in
that slot is no longer available by name. Once a name has been overwritten, it
may be captured again because it is no longer present among the valid entries.

There is no positional checkpoint handle. stmrecall performs an exact string
lookup over the valid entries. This avoids returning an index that could later
refer to a different checkpoint after ring wraparound.

## Capture trigger behavior

stmcall performs at most one capture for its last successfully captured name
within the lifetime of that opcode instance:

1. When ktrigger is not exactly 1, kcaptured is 0 and no runner state changes.
2. On the first successful request, the checkpoint is added to the runner and
   kcaptured becomes 1.
3. If the same opcode instance is called again with the same last name and a
   trigger of 1, it returns 1 without adding another ring entry.
4. A low trigger does not clear stmcall's remembered last name.

The remembered last name is an opcode-local idempotency check. Returning 1 from
this shortcut means that this opcode instance already captured the name; it is
not a new global lookup proving that the entry is still present after other
captures have wrapped the runner's ring.

A duplicate name submitted by another opcode instance reaches the runner-level
uniqueness check and produces a performance error.

## Recall behavior

On a successful stmrecall call, the named entry is found and its stored logical
and timing state is written back to the runner. A missing, expired, empty, or
overlong name produces a performance error.

Recall is edge-like for a stable name:

1. When ktrigger is not exactly 1, krecalled is 0 and the opcode is re-armed.
2. The first call with ktrigger equal to 1 restores the checkpoint and returns
   1.
3. While the trigger remains 1 and the name remains unchanged, subsequent
   calls return 1 without restoring the state again and without publishing
   additional checkpoint-recall events.
4. At least one call with a trigger other than 1 re-arms the same name, allowing
   a later pulse to restore the checkpoint again.

Restoring requested_node and request_conflict is intentional: a checkpoint
represents the complete logical runner state at capture time, including pending
transition intent.

## Recall events and observers

Recalling a checkpoint does not rewind the event subsystem. Instead, it
publishes a new transition event through the normal transition ring:

| Event field | Checkpoint-recall value |
| --- | --- |
| status | STM_EVENT_RECALL, numeric value 4 |
| from | Node active immediately before stmrecall |
| to | current_node stored in the checkpoint |
| sequence | Previous live event sequence plus one |

This event is emitted even if the pre-recall and restored nodes are the same.
The live sequence remains monotonic, and existing event-reader cursors remain
valid. The recall is consequently observable through stmevent and through the
runner's enter/exit event state rather than being confused with an initial reset
or an ordinary graph transition.

The status values currently exposed by the event API are:

| Value | Meaning |
| --- | --- |
| 1 | STM_EVENT_CHANGED |
| 2 | STM_EVENT_SELF_TRANSITION |
| 3 | STM_EVENT_RESET |
| 4 | STM_EVENT_RECALL |
| 5 | STM_EVENT_PAUSE |
| 6 | STM_EVENT_RESUME |
| 7 | STM_EVENT_DELETE |

## Time and reset semantics

graph_tick, total_sample_frames, and node_sample_on_enter are restored together.
This makes stmtick, graph elapsed time, and current-node elapsed time continue
from the frozen state.

reset_pcycle and has_reset_cycle are not restored. reset_pcycle refers to an
absolute engine k-counter, so replaying an old value at a later performance
cycle would be invalid. Both fields are cleared during checkpoint recall, and
the STM_EVENT_RECALL event provides the correct notification for the new live
cycle.

## Concurrency and lifetime

stmcall and stmrecall retain the runner at initialization and release their
reference at deinitialization. Both request writer access and therefore obey
the runner's existing single-writer ownership rules.

Capture and restore are enclosed by stm_runner_update_begin and
stm_runner_update_end. Individual shared runner fields are accessed through
STM_LOAD and STM_STORE. Readers therefore observe the update through the same
coherent-snapshot mechanism used by the rest of the STM runner: an atomic
sequence counter where compiler atomics are available, or the runner fallback
mutex otherwise.

Publishing the checkpoint-recall event is part of the same compound update as
restoring the checkpoint. Readers cannot observe a completed recall without its
corresponding event state.

## Example

The following fragment captures the state after advancing to B and later
restores that exact node and tick:

    if kcycle == 1 then
        stmnext(irunner, "B")
    endif

    kadvanceStatus, kfrom, kto = stmadvance(irunner)

    kcallTrigger = (kcycle == 1 ? 1 : 0)
    kcaptured = stmcall(irunner, "at-B-tick-1", kcallTrigger)

    krecallTrigger = (kcycle == 4 ? 1 : 0)
    krecalled = stmrecall(irunner, "at-B-tick-1", krecallTrigger)

After the successful recall, stmcurrentid returns B's node ID, stmtick returns
the captured tick, and stmevent reports status 4 with from equal to the node
active before recall and to equal to B.

## Regression coverage

tests/commandline/stm/test_stm_checkpoint.csd exercises a non-initial checkpoint so
that an accidental reset cannot satisfy the assertions. Its expected flow is:

| Cycle | Operation and expected state |
| --- | --- |
| 1 | Transition A to B, reach tick 1, and capture at-B-tick-1. |
| 2 | Transition B to C and reach tick 2. |
| 4 | Recall while on C; restore B at tick 1 and publish event sequence 4, status 4, from C to B. |
| 5 | Continue normally from restored B to C and reach tick 2. |
| 6 | Recall the same checkpoint after a low trigger cycle; restore B at tick 1 and publish a second checkpoint-recall event. |

The test prints explicit FREEZE, STATE, RESUME, and EVENT records, verifies both
the restored state and the event metadata, and is registered in
tests/commandline/test.py. The scenario has been run in both float-sample and
double-sample builds, together with the existing STM event, sequential, and
parallel-runner regression tests.

# STM feature 2: stmpause and stmresume

## Overview

This feature adds an explicit execution state to each STM runner:

    STM_RUNNER_RUNNING = 0
    STM_RUNNER_PAUSED  = 1

Two one-shot k-rate control opcodes change that state:

    stmpause  irunner, ktrigger
    stmresume irunner, ktrigger

Both opcodes have no output. A positive trigger edge requests an idempotent
state-setting operation: stmpause changes only a running runner, while
stmresume changes only a paused runner.

Pausing stops progression of the STM runner. It does not suspend the Csound
instrument or the engine k-cycle. Orchestra code continues to run so that it
can inspect the runner and eventually call stmresume.

## Opcode arguments

| Argument | Rate/type | Meaning |
| --- | --- | --- |
| irunner | i-rate handle | Runner returned by stminstance. |
| ktrigger | k-rate value | A transition from a non-positive value to a positive value requests the operation. |

The runner is resolved and retained when each opcode instance initializes.
Both controls require writer access and obey the runner's existing
single-writer ownership rules.

## Trigger and idempotency semantics

Each opcode instance stores whether its trigger was positive on its preceding
performance call. The operation fires only on a positive edge:

- a transition from ktrigger less than or equal to 0 to ktrigger greater than
  0 requests the operation;
- holding any positive value for several k-cycles does not request it again;
- at least one performance call with ktrigger less than or equal to 0 re-arms
  the opcode instance;
- stmpause on an already paused runner is a successful no-op;
- stmresume on an already running runner is a successful no-op.

The trigger latch belongs to the opcode instance, not to the runner or event
ring. The opcode should normally be called every k-cycle with a trigger
expression, including inactive cycles, so that it can observe the non-positive
value that re-arms it.

The already-paused and already-running cases deliberately do not produce
Message, Warning, or performance-error output. They are expected idempotent
operations rather than exceptional conditions. Actual changes are observable
through stmevent, and a paused stmadvance reports STM_PAUSED.

Orchestra order remains significant when pause and resume edges occur in the
same k-cycle. Each opcode instance has an independent latch, so both operations
can fire in statement order.

## Entering pause

When stmpause changes a running runner to PAUSED, it performs one compound
update:

1. read the current node;
2. clear requested_node to STM_NO_NODE;
3. clear request_conflict to STM_REQUEST_OK;
4. store STM_RUNNER_PAUSED in run_state;
5. publish one STM_EVENT_PAUSE event.

Clearing the request state discards a transition requested before the pause.
The runner cannot unexpectedly apply that old request after stmresume.

Holding the stmpause trigger positive does not clear state again and does not
publish additional events. A later positive edge while the runner is already
paused is also a successful no-op.

## Behavior while paused

The following runner operations change behavior while run_state is PAUSED:

| Operation | Paused behavior |
| --- | --- |
| stmadvance | Returns STM_PAUSED without consuming requests, changing node, or advancing the runner clocks. |
| stmnext | Returns successfully without recording a target request. |
| stmnext.id | Returns successfully without recording a target request. |
| stmrecall | Does not restore a checkpoint and leaves its k-rate result at 0. |

For a paused stmadvance, the three outputs are:

| Output | Value |
| --- | --- |
| status | STM_PAUSED, numeric value 5 in STM_ADVANCE_STATUS |
| from | Current node ID |
| to | -1, because no transition is evaluated or applied |

The early return from stmadvance occurs before its clock update. Therefore all
of the following remain unchanged for the duration of the pause:

- current_node;
- previous_node;
- graph_tick, observed through stmtick;
- total_sample_frames, observed through stmtime;
- node_sample_on_enter and the derived stmnodetime.

Queries such as stmcurrent, stmcurrentid, stmtick, stmtime, and stmnodetime
remain available. They expose the frozen runner state.

The node implementation dispatched by the orchestra is not automatically
suspended. In the normal execution loop, node code runs before stmadvance.
While paused, its calls to stmnext are ignored, but any other orchestra-side
effects remain the responsibility of that code. A caller that needs to suppress
all node processing must gate its dispatch separately.

stmreset remains an explicit state reset even while paused, but it does not
change run_state: the reset runner remains paused until stmresume. Checkpoint
capture through stmcall remains possible; pause state itself is not part of
STM_CHECKPOINT.

## Resuming

When stmresume changes a paused runner to RUNNING, it performs one compound
update:

1. read the current node;
2. clear requested_node to STM_NO_NODE;
3. clear request_conflict to STM_REQUEST_OK;
4. store STM_RUNNER_RUNNING in run_state;
5. publish one STM_EVENT_RESUME event.

stmresume does not itself call stmadvance and does not increment any runner
clock. The next effective stmadvance continues from the frozen node and time.

Opcode order determines whether that advance can occur in the same k-cycle:

- stmresume before stmadvance allows that stmadvance to run immediately;
- stmresume after a paused stmadvance restarts progression on the next
  stmadvance call.

## Pause and resume events

Pause and resume are runner-control events, not graph-node transitions. Both
are written to the normal event ring with the current node as both endpoints:

| Operation | Event status | Numeric value | from | to |
| --- | --- | --- | --- | --- |
| stmpause | STM_EVENT_PAUSE | 5 | Current node | Current node |
| stmresume | STM_EVENT_RESUME | 6 | Current node | Current node |

Both events increment the global event_seq and are observable through
stmevent. They are recorded with is_transition equal to 0, so they do not:

- update event_entered_node or event_exited_node;
- increment node_event_seq;
- trigger stmonenter or stmonexit.

Node-changing events continue to update both event_seq and node_event_seq.
This separation prevents a pause or resume from creating false node-entry or
node-exit notifications, and also prevents control events from hiding the most
recent node event from those observers.

## Concurrency and publication

Pause and resume mutations are enclosed by stm_runner_update_begin and
stm_runner_update_end. Shared runner fields are written through STM_STORE, and
the event is published within the same compound update as run_state.

Readers therefore observe a coherent result: a completed PAUSED state includes
its pause event, and a completed RUNNING state after stmresume includes its
resume event. The implementation uses the runner's atomic sequence protocol or
its fallback mutex, according to platform support.

## Example

The following sequence pauses after reaching tick 1, keeps calling stmadvance
without moving the runner, and resumes later:

    kadvanceStatus, kfrom, kto = stmadvance(irunner)

    kpauseTrigger = (kcycle == 1 ? 1 : 0)
    stmpause(irunner, kpauseTrigger)

    kresumeTrigger = (kcycle == 3 ? 1 : 0)
    stmresume(irunner, kresumeTrigger)

If stmresume appears after stmadvance as above, the advance on cycle 3 still
reports STM_PAUSED. The runner becomes RUNNING later in that cycle, and its
tick advances on cycle 4.

## Regression coverage

tests/commandline/stm/test_stm_pause_resume.csd is registered in
tests/commandline/test.py. It uses an A-to-B graph and deliberately holds each
control trigger high for two cycles to verify one-shot edge behavior. The
repeated PAUSE and RESUME trace labels indicate that the corresponding trigger
remains high; only the EVENT records indicate actual state changes.

Its expected flow is:

| Cycle | Operation and expected state |
| --- | --- |
| 1 | Advance to tick 1, pause on A, and publish event sequence 2 with status 5. |
| 2 | Remain paused at tick 1; ignore stmnext; emit no repeated pause event. |
| 3 | A paused advance leaves tick 1 unchanged; resume afterward and publish event sequence 3 with status 6. |
| 4 | Advance normally to tick 2; the held resume trigger emits no second event; no paused request survives. |
| 5 | Accept a fresh stmnext request, transition A to B at tick 3, and publish the normal node event sequence 4. |
| 6 | Remain on B, reach tick 4, and finish without another event. |

The test also verifies that pause and resume never trigger stmonenter for B or
stmonexit for A, while the later real A-to-B transition triggers both. It has
been run successfully with the local double-sample build together with the
checkpoint, event-ring, and sequential-trace STM regressions.

# STM feature 3: one-shot stmreset

## Overview

stmreset restores the logical runner to the entry state selected by stmentry,
or to the first node when no explicit entry was configured. Its k-rate
interface is:

    stmreset irunner, ktrigger

The trigger is required because stmreset has a performance callback. Without a
one-shot trigger, an unconditional opcode statement would reset the runner and
its clocks on every k-cycle.

## Opcode arguments

| Argument | Rate/type | Meaning |
| --- | --- | --- |
| irunner | i-rate handle | Runner returned by stminstance. |
| ktrigger | k-rate value | A transition from a non-positive value to a positive value requests one reset. |

stmreset retains the runner at opcode initialization, releases that reference
at deinitialization, and requires writer access under the normal single-writer
rules.

## Trigger behavior

stmreset uses the same per-opcode one-shot trigger state as stmpause and
stmresume:

1. The first performance call with ktrigger greater than 0 after an inactive
   value performs the reset.
2. Further calls while ktrigger remains positive are no-ops.
3. A call with ktrigger less than or equal to 0 re-arms the opcode instance.
4. A later positive edge can then perform another reset.

The trigger state is local to each stmreset statement. It is not inferred from
the most recent transition event. In particular, a reset executed while the
runner is already on its entry node may produce no reset event, but it still
resets the runner clocks.

Call stmreset every k-cycle with an explicit trigger expression when the same
opcode instance must support multiple reset pulses. Placing
`stmreset(irunner, 1)` inside a conditional block does not let the instance
observe the intervening low cycles and therefore does not re-arm it.

## State reset by stmreset

On a positive trigger edge, stmreset performs one compound runner update:

1. set current_node to the compiled definition's entry node;
2. set previous_node to STM_NO_NODE;
3. clear requested_node and request_conflict;
4. set graph_tick, total_sample_frames, and node_sample_on_enter to zero;
5. record the current engine k-counter as the reset performance cycle;
6. publish STM_EVENT_RESET when the active node actually changes.

The reset does not clear checkpoints, rewind the transition ring, or change
run_state. A paused runner therefore remains paused after stmreset and must be
continued explicitly with stmresume.

If stmadvance is called later in the same engine k-cycle, it may evaluate a
request made after stmreset, but it does not increment the freshly reset graph
clock. An stmadvance in a genuinely later cycle increments the clock normally.

## Events and observers

When reset moves the runner from a non-entry node to its entry node, it
publishes an event with status STM_EVENT_RESET, numeric value 3, and the old and
entry node IDs as from and to. This is a node-changing event, so it also updates
the runner-level enter/exit event state used by stmonenter and stmonexit.

Resetting a runner already on its entry node publishes no duplicate event. The
state and clocks are still reset, and any pending transition request is still
discarded.

## Example

The opcode should normally remain in the performance path while its trigger
expression supplies both active and re-arm cycles:

    kresetTrigger = (kcycle == 7 ? 1 : 0)
    stmreset(irunner, kresetTrigger)

Holding kresetTrigger at any positive value for several cycles performs only
one reset. After at least one non-positive cycle, a new positive edge performs
another reset.

## Regression coverage

The command-line STM tests use the two-argument stmreset interface throughout.
In particular:

- tests/commandline/stm/test_stm_reinit_reset.csd verifies same-cycle and delayed
  stmadvance clock behavior around reset;
- tests/commandline/stm/test_stm_event.csd verifies STM_EVENT_RESET publication;
- tests/commandline/stm/test_stm_instances.csd verifies that resetting one runner
  does not affect another runner sharing the same definition;
- tests/commandline/stm/test_stm.csd covers entry restoration, pending-request
  discard, request ordering after reset, and clock restart semantics.

# STM feature 4: stmdelete

## Overview

`stmdelete` places one runner in a permanent terminal state:

    stmdelete irunner, ktrigger

Deletion is logical rather than immediate memory deallocation. The runner
handle, immutable definition, checkpoints, transition history, and retained
opcode references remain alive for observation and normal lifetime cleanup.
What becomes terminal is the runner's mutable execution state: it can no longer
advance, accept transition requests, reset, pause, resume, capture, or recall.

The terminal state and its public advance result are:

    STM_RUNNER_DELETED = 2
    STM_DELETED        = 6

The values belong to different enums: `STM_RUNNER_DELETED` is the internal
runner state, while `STM_DELETED` is returned by `stmadvance`.

## Opcode arguments

| Argument | Rate/type | Meaning |
| --- | --- | --- |
| irunner | i-rate handle | Runner returned by `stminstance`. |
| ktrigger | k-rate value | A transition from a non-positive value to a positive value requests deletion. |

`stmdelete` retains the runner at opcode initialization, releases that
reference at deinitialization, and claims writer access under the runner's
single-writer rules.

## Trigger and idempotency

`stmdelete` uses the same per-opcode positive-edge latch as `stmreset`,
`stmpause`, and `stmresume`:

1. a transition from `ktrigger <= 0` to `ktrigger > 0` requests deletion;
2. holding a positive trigger does not request deletion again;
3. a non-positive performance call re-arms that opcode instance;
4. deleting an already deleted runner is a successful no-op, including after
   the trigger has been re-armed.

Only the first effective deletion publishes an event. Repeated calls never
move the event sequence or alter runner state.

## Entering the terminal state

On a positive trigger edge, deletion performs one compound update:

1. read the current node;
2. clear `requested_node` to `STM_NO_NODE`;
3. clear `request_conflict` to `STM_REQUEST_OK`;
4. set `run_state` to `STM_RUNNER_DELETED`;
5. publish one `STM_EVENT_DELETE` event.

Deletion is valid from both `STM_RUNNER_RUNNING` and
`STM_RUNNER_PAUSED`. A pending transition requested earlier in the same
orchestra cycle is discarded and cannot be applied after deletion.

Deletion does not modify:

- `current_node` or `previous_node`;
- `graph_tick`, `total_sample_frames`, or `node_sample_on_enter`;
- the immutable graph definition;
- checkpoint storage;
- transition history or existing reader cursors;
- runner ownership, references, or registry lifetime.

There is no operation that changes a deleted runner back to running or paused.
A new execution requires a new runner from `stminstance`.

## Behavior after deletion

Read-only queries remain available and expose the final frozen runner state.
Mutation opcodes become successful no-ops, except that `stmadvance` reports
the terminal status explicitly.

| Operation | Deleted-runner behavior |
| --- | --- |
| `stmadvance` | Returns `STM_DELETED` without consuming state or advancing clocks. |
| `stmnext` / `stmnext.id` | Ignore the request before validating its target. |
| `stmreset` | Does nothing; clocks and node remain frozen. |
| `stmpause` / `stmresume` | Do nothing and publish no event. |
| `stmcall` | Returns 0 and does not validate or capture a checkpoint. |
| `stmrecall` | Returns 0 and does not validate or restore a checkpoint. A non-active trigger still re-arms the opcode-local recall latch before the terminal-state check. |
| Queries | Continue to return the final node, topology, and frozen clocks. |
| `stmevent` | Continues consuming already published events, including the delete event. |
| `stmonenter` / `stmonexit` | Remain available; deletion itself does not trigger either observer. |

For a deleted `stmadvance`, the outputs are:

| Output | Value |
| --- | --- |
| status | `STM_DELETED`, numeric value 6 in `STM_ADVANCE_STATUS` |
| from | Final current node ID |
| to | -1 |

Terminal no-op checks intentionally precede target and checkpoint-name
validation. This makes all mutation paths consistently inert after deletion;
for example, an unknown `stmnext` target or an empty checkpoint name does not
turn a terminal call into a performance error.

## Delete event and observers

Deletion publishes a control event through the normal transition ring:

| Event field | Delete value |
| --- | --- |
| status | `STM_EVENT_DELETE`, numeric value 7 |
| from | Current node at deletion |
| to | Same current node |
| sequence | Previous live event sequence plus one |

The event is recorded with `is_transition = 0`. It increments the global
`event_seq` and is observable through `stmevent`, but it does not update
`event_entered_node`, `event_exited_node`, or `node_event_seq`.
Consequently, deletion does not create false node entry or exit notifications.

If pause and delete occur in the same cycle, statement order remains visible in
the event ring. A pause followed by delete publishes both control events in
order; a delete followed by pause publishes only delete because the later pause
is terminally ignored.

## Concurrency and lifetime

The state change, pending-request cleanup, and delete-event publication are
enclosed by `stm_runner_update_begin` and `stm_runner_update_end`.
Readers therefore observe deletion through the same coherent snapshot protocol
used for transition, reset, recall, pause, and resume updates.

`stmdelete` does not invalidate the public runner handle or release other
opcode references. Existing queries and observers stay safe, and normal owner
and reference-count cleanup eventually frees the runner and its immutable
definition. A deleted runner may be claimed by a later sequential writer after
the previous writer deinitializes, but every mutation remains terminally inert.

## Example

The opcode should stay in the performance path so its trigger latch can observe
both active and re-arm cycles:

    kdeleteTrigger = (kcycle == 8 ? 1 : 0)
    stmdelete(irunner, kdeleteTrigger)

    kstatus, kfrom, kto = stmadvance(irunner)

After the deletion edge, `kstatus` is 6, `kfrom` is the final current node,
`kto` is -1, and the runner clocks no longer advance.

## Regression coverage

`tests/commandline/stm/test_stm_delete.csd` is registered in
`tests/commandline/test.py`. It verifies:

- transition to a non-entry node before deletion;
- a held delete trigger and a later re-armed delete request;
- exactly one `STM_EVENT_DELETE` event with status 7 and equal endpoints;
- `STM_DELETED` status 6 from every later `stmadvance`;
- frozen node and clock values;
- terminal no-op behavior for next, reset, pause, resume, call, and recall;
- continued read-only queries and event consumption.

The checkpoint, pause/resume, event-ring, and delete regressions should be run
together because their state and event precedence rules intentionally overlap.

PasqualeMainolfi:
The WASM CI check is currently failing because of the issue reported in (#2625) The STM build causes the C++ plugin test to reach kperf, exposing a pre-existing callback-table relocation bug in the WASM plugin loader. The plugin binaries are identical between the previously successful and failing builds, and I found no opcode-count or function-table limit related to STM.
I do not see an STM-specific change that could fix this failure without merely hiding the underlying loader problem.

Could you please confirm whether this diagnosis is correct? If STM being placed under Opcodes, and therefore being expected to support plugin builds, is considered a problem, I can move it to `OOps` so that it is always linked directly into the Csound core.

vlazzarini:
Should not the opcodes in ../Opcodes be built as part of the library as we do by default on the desktop platforms?

PasqualeMainolfi:
@vlazzarini you are right. I phrased the last paragraph poorly. 
In the failing WASM configuration, BUILD_PLUGINS is disabled, so stdopcod_SRCS, including stm.c, are already built directly into libcsound, as they are by default on desktop platforms. I mentioned OOps only because of the earlier concern that placing STM under Opcodes might require it to remain compatible with standalone plugin builds. Moving it to OOps is not necessary to include it in the core library and would not fix #2625. The failure described in #2625 concerns the separately loaded plugin_example_cpp.wasm and its callback-table indices. STM is statically linked into the host in this build and only causes the existing loader problem to become observable.

hlolli:
@PasqualeMainolfi there's something going on with the test reporter action, I try making a PR which you're need to rebase over once/if it's merged

PasqualeMainolfi:
@hlolli Ok!

hlolli:
@PasqualeMainolfi it's been merged, can you rebase with develop and resolve the merge conflict too?

hlolli:
are you opening a new PR @PasqualeMainolfi ?

PasqualeMainolfi:
It’s really strange. I tried to do a force-push after the rebase, but somehow the PR got closed. I honestly don’t know what happened. I’ve been checking things on my end, but I can’t see anything unusual.

PasqualeMainolfi:
@hlolli  I can’t reopen the PR. Could you do it?

hlolli:
I cant because it was opened from your fork

PR #2588, opened 2026-07-08
Fix inverted yield callback throttle

hlolli:
## Summary

This fixes the event poll throttle calculation so it matches the intended rate.

The old formula used:

```c
250.0 / ekr
```

At a normal control rate, for example `ekr = 750`, that becomes `0` after integer truncation. That means the counter reload value is zero, so the throttle does not actually space out polling work.

The new formula uses:

```c
ekr / 250.0
```

So with `ekr = 750`, the counter reload value becomes `3`, which matches the goal of polling at about 250 times per second instead of every k-cycle.

## Why this matters

This code sits in the performance setup path and controls how often event polling work is allowed to run.

A throttle counter should answer: “how many k-cycles should pass before the next poll?” Since `ekr` is the number of k-cycles per second, we divide `ekr` by `250` to get the spacing between polls.

For `ekr = 750`:

```text
750 k-cycles/sec / 250 polls/sec = 3 k-cycles per poll
```

The previous calculation asked the opposite question, so normal rates rounded down to zero.

## Testing

Added a GoogleTest regression on `develop` that:

- creates a Csound instance
- sets `sr = 48000` and `ksmps = 64`
- starts the engine
- confirms `ekr = 750`
- confirms `evt_poll_maxcnt = 3`

Verified locally with:

```sh
ninja -C build unittests
./build/tests/c/unittests --gtest_filter=EngineTests.testEventPollCounterUsesControlRate
```

### DIFF
--- a/Engine/musmon.c
+++ b/Engine/musmon.c
@@ -301,7 +301,7 @@ int32_t start_engine(CSOUND *csound)
 
     csound->evt_poll_cnt    = 0;
     csound->evt_poll_maxcnt =
-      (int)(250.0 /(double) csound->ekr); /* VL this was wrong: kr/250 originally */
+      (int)((double) csound->ekr / 250.0);
 
     /* open MIDI output (moved here from argdecode) */
     if (O->Midioutname != NULL && O->Midioutname[0] == (char) '\0')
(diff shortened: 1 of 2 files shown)


vlazzarini:
this is obsolete, we don't have a yield callback anymore...

PR #2589, opened 2026-07-08
Remove unused event poll counters

hlolli:
This removes the old `evt_poll_cnt` and `evt_poll_maxcnt` fields from the engine state.

These counters used to control how often `csoundYield()` was called from the old k-cycle performance loop. On `develop`, that yield path is gone, and the current `kperf()` implementation no longer reads these counters.

So instead of fixing the old throttle formula, this PR removes the dead state entirely.


kunstmusik:
FWIW: This looks fine to me, it'll break ABI compatibility but better to do it now before we release 7.0. Will leave it to @vlazzarini for the final review.

vlazzarini (approved):

PR #2594, opened 2026-07-10
fix(fgens): zero the FUNC header and release the buffers it owns

PasqualeMainolfi:
Starting from a review of the stm opcode clocks #2587, this ended up uncovering what looks like a
long-standing memory bug in Engine/fgens.c.

`csoundFTAlloc()` allocated FUNC headers with Malloc, leaving fields such as `ftp->args`
uninitialized. Later, `csoundFTCreate()` could do:
```c
if (ftp->args != NULL) csound->Free(csound, ftp->args);
```
and attempt to free a pointer that was never written by Csound.
This showed up in Windows CI in the deferred `GEN49` path:

```text
Return Code: 3221226356  ->  0xC0000374  =  STATUS_HEAP_CORRUPTION
```
or, with a different heap layout:
```text
csound command: received signal 11
```
The output always stopped right after: 
`ftable 101`:
which places the crash inside the deferred-length ftgen.

Under AddressSanitizer, the pointer passed to `free()` had the malloc fill pattern:

```text
#0 csoundFree  memalloc.c
#1 ftgen_      ftgen.c

x[1] = 0xbebebebebebebebe
```
So the pointer was not valid-but-corrupted table memory; it was an uninitialized field in
the FUNC header.

## Fix

This patch:

- allocates FUNC headers with Calloc
- fixes the relocation branch of csoundFTAlloc(), which allocated len * sizeof(MYFLT) instead of sizeof(FUNC)
- adds ftfree() to release the FUNC header together with its owned ftable and args buffers
- releases args before ftalloc() clears a reused same-size FUNC header

## Verification

The original failing case now passes with Guard Malloc and filled allocations:

DYLD_INSERT_LIBRARIES=/usr/lib/libgmalloc.dylib MALLOC_FILL_SPACE=1 \
  ./csound tests/commandline/test_gen49_defer.csd

I also used targeted `.csd` reproducers for paths that are not covered by the normal suite:

- `csoundFTFree()` releasing table storage
- `ftalloc()` same-size replacement/reuse
- `ftalloc()` different-size relocation
- `csoundFTAlloc()` relocation via ftload

The leak was measurable. Creating and releasing thirty 8 MB tables changed from:

before: max RSS = 262 MB
after : max RSS =  31 MB

The targeted tests are included below for review/reproduction if useful.
<details>
<summary>Test 1: csoundFTFree leak check</summary>

<CsoundSynthesizer>
<CsOptions>
-n -m0
</CsOptions>
<CsInstruments>
sr=44100
ksmps=32
nchnls=1
0dbfs=1

instr 1
    fn:i = ftgen(1, 0, 1048576, 10, 1)
    ftfree(1, 0)
endin
</CsInstruments>
<CsScore>
r30
i 1 0 0.001
s
e
</CsScore>
</CsoundSynthesizer>
</details>
<details>
<summary>Test 2: ftalloc reuse/free/relocation paths</summary>

<CsoundSynthesizer>
<CsOptions>
-n
</CsOptions>
<CsInstruments>
sr     = 44100
ksmps  = 32
nchnls = 1
0dbfs  = 1

opcode sumtab(fn:i):i
    s:i = 0
    i:i = 0
    while i < ftlen(fn) do
        s += table:i(i, fn)
        i += 1
    od
    xout s
endop

instr 1
    a:i = ftgen(3, 0, 16, 10, 1)
    b:i = ftgen(3, 0, 16, 10, 1, 0.5)
    if ftlen(3) != 16 then
        prints("[FAIL] reuse: ftlen(3)=%d expected 16\n", ftlen(3))
        exitnow(-1)
    endif
    prints("[reuse] ftlen(3)=%d sum=%f\n", ftlen(3), sumtab(3))
endin

instr 2
    c:i = ftgen(4, 0, 32, 10, 1)
    prints("[free] before: ftlen(4)=%d\n", ftlen(4))
    ftfree(4, 0)
    prints("[free] table 4 released\n")
endin

instr 3
    d:i = ftgen(5, 0, 16, 10, 1)
    e:i = ftgen(5, 0, 64, 10, 1)
    if ftlen(5) != 64 then
        prints("[FAIL] relocate: ftlen(5)=%d expected 64\n", ftlen(5))
        exitnow(-1)
    endif
    prints("[relocate] ftlen(5)=%d sum=%f\n", ftlen(5), sumtab(5))
    prints("[ftlifetime] OK\n")
endin
</CsInstruments>
<CsScore>
i 1 0 0.05
i 2 0.05 0.05
i 3 0.10 0.05
e
</CsScore>
</CsoundSynthesizer>
</details>
<details>
<summary>Test 3: csoundFTAlloc relocation via ftload</summary>

<CsoundSynthesizer>
<CsOptions>
-n
</CsOptions>
<CsInstruments>
sr     = 44100
ksmps  = 32
nchnls = 1
0dbfs  = 1

instr 1
    src:i = ftgen(1, 0, 16, 10, 1)
    ftsave("ftrelo.bin", 0, 1)

    dst:i = ftgen(2, 0, 32, 10, 1)
    prints("[relo] before: ftlen(2)=%d\n", ftlen(2))

    ftload("ftrelo.bin", 0, 2)
    prints("[relo] after : ftlen(2)=%d\n", ftlen(2))

    if ftlen(2) != 16 then
        prints("[FAIL] relocation did not resize table 2\n")
        exitnow(-1)
    endif

    sum:i = 0
    idx:i = 0
    while idx < ftlen(2) do
        sum += table:i(idx, 2)
        idx += 1
    od

    prints("[relo] read back %d points, sum=%f\n", ftlen(2), sum)
    prints("[relo] OK\n")
endin
</CsInstruments>
<CsScore>
i 1 0 0.1
e
</CsScore>
</CsoundSynthesizer>
</details>

I may still be missing something in the FUNC ownership model, so I would really appreciate
a careful review to confirm whether this is indeed a memory bug and whether this is the
right fix.

### DIFF
--- a/Engine/fgens.c
+++ b/Engine/fgens.c
@@ -93,6 +93,18 @@ static int32_t GENUL(FGDATA *ff, FUNC *ftp)
     return csoundFtError(ff, Str("unknown GEN number"));
 }
 
+/* Releases a function table and the two buffers its header owns. csoundFree
+   ignores NULL, and args stays NULL until csoundFTCreate fills it in.
+   The caller must have already cleared csound->flist[fno]. */
+static void ftfree(CSOUND *csound, FUNC *ftp)
+{
+    if (ftp == NULL)
+      return;
+    csound->Free(csound, ftp->ftable);
+    csound->Free(csound, ftp->args);
+    csound->Free(csound, (void*) ftp);
+}
+
 /**
  * Create ftable using evtblk data, and store pointer to new table in *ftpp.
  * If mode is zero, a zero table number is ignored, otherwise a new table
@@ -139,7 +151,7 @@ int32_t csoundFTCreate(CSOUND *csound, FUNC **ftpp, const EVTBLK *evtblkp,
         return csoundFtError(&ff, Str("ftable does not exist"));
       }
       csound->flist[ff.fno] = NULL;
-      csound->Free(csound, (void*) ftp);
+      ftfree(csound, ftp);
       if (UNLIKELY(msg_enabled))
         csoundMessage(csound, Str("ftable %d now deleted\n"), ff.fno);
       return 0;
@@ -198,9 +210,8 @@ int32_t csoundFTCreate(CSOUND *csound, FUNC **ftpp, const EVTBLK *evtblkp,
       i = (*csound->gensub[genum])(&ff, NULL);
       ftp = csound->flist[ff.fno];
       if (i != 0) {
-        if(ftp != NULL)
-          csound->Free(csound, ftp);
         csound->flist[ff.fno] = NULL;
+        ftfree(csound, ftp);
         return -1;
       }
       ftp->sr = csound->esr;
@@ -251,7 +262,7 @@ int32_t csoundFTCreate(CSOUND *csound, FUNC **ftpp, const EVTBLK *evtblkp,
       csoundMessage(csound, Str("ftable %d:\n"), ff.fno);
     if ((*csound->gensub[genum])(&ff, ftp) != 0) {
       csound->flist[ff.fno] = NULL;
-      csound->Free(csound, ftp);
+      ftfree(csound, ftp);
       return -1;
     }
 
@@ -295,11 +306,11 @@ int32_t csoundFTAlloc(CSOUND *csound, int32_t tableNum,
         csound->flist[i] = NULL;            /* Clear new section            */
       csound->maxfnum = size;
     }
-    /* allocate space for table */
-    size = (int32_t) (len * (int32_t) sizeof(MYFLT));
+    /* allocate space for table: the header must be zeroed, csoundFTCreate
+       frees ftp->args before replacing it and would otherwise free garbage */
     ftp = csound->flist[tableNum];
     if (ftp == NULL) {
-      csound->flist[tableNum] = (FUNC*) csound->Malloc(csound, sizeof(FUNC));
+      csound->flist[tableNum] = (FUNC*) csound->Calloc(csound, sizeof(FUNC));
       csound->flist[tableNum]->ftable =
         (MYFLT*)csound->Malloc(csound, sizeof(MYFLT)*(len+1));
     }
@@ -311,14 +322,13 @@ int32_t csoundFTAlloc(CSOUND *csound, int32_t tableNum,
                                         "may find this disturbing"), tableNum);
       }
       csound->flist[tableNum] = NULL;
-      csound->Free(csound, ftp);
-      csound->flist[tableNum] = (FUNC*) csound->Malloc(csound, (size_t) size);
+      ftfree(csound, ftp);
+      csound->flist[tableNum] = (FUNC*) csound->Calloc(csound, sizeof(FUNC));
       csound->flist[tableNum]->ftable =
         (MYFLT*)csound->Malloc(csound, sizeof(MYFLT)*(len+1));
     }
     /* initialise table header */
     ftp = csound->flist[tableNum];
-    //memset((void*) ftp, 0, (size_t) ((char*) &(ftp->ftable) - (char*) ftp));
     ftp->flen = (int32) len;
     if (!(len & (len - 1))) {
       /* for power of two length: */
@@ -351,7 +361,7 @@ int32_t csoundFTFree(CSOUND *csound, int32_t tableNum)
     if (UNLIKELY(ftp == NULL))
       return -1;
     csound->flist[tableNum] = NULL;
-    csound->Free(csound, ftp);
+    ftfree(csound, ftp);
 
     return 0;
 }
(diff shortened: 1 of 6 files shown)


PasqualeMainolfi:
last commit: CI also exposed a pre-existing realtime cleanup race in test_async_diskin.csd: `eventi("e")` sets `event_insert_loop` to 0, so `csound_cleanup()` could skip joining `event_insert_thread` and close diskin2 files while the RT insert thread was still initializing an instance. This branch now joins the thread based on the thread handle before deactivating notes.

Engine/fgens.c
```diff
@@ -93,6 +93,18 @@ static int32_t GENUL(FGDATA *ff, FUNC *ftp)
...
 
+/* Releases a function table and the two buffers its header owns. csoundFree
+   ignores NULL, and args stays NULL until csoundFTCreate fills it in.
+   The caller must have already cleared csound->flist[fno]. */
+static void ftfree(CSOUND *csound, FUNC *ftp)
+{
+    if (ftp == NULL)
+      return;
+    csound->Free(csound, ftp->ftable);
+    csound->Free(csound, ftp->args);
```
hlolli, line 104: Could there be one more ownership issue hiding behind this fix? The uninitialized `args` finding is correct, and having `ftfree()` release it is the right invariant for normally created tables. Binary `ftsave`, however, writes the raw `FUNC` prefix including `args`, and `ftload` copies that pointer into the destination table. With this line, a loaded table can therefore free the source table's `args` (or a stale address from another process). I reproduced this by saving table 1, loading it as table 2, then freeing 2 and 1; the second free segfaulted.

Would a pointer-free on-disk header, or explicitly resetting/reconstructing `args` during load, make the new ownership rule safe? I can put together a follow-up PR for the serialization side if you'd prefer; happy to leave that choice to you.

hlolli (commented):
Thanks for tracking this down. The uninitialized `FUNC` header is a real bug, and the switch to `Calloc(sizeof(FUNC))` plus centralized cleanup is a solid improvement. The event-thread join change also matches the lifetime issue described in the PR.

I left one inline question because I think there is still another ownership problem lurking in the binary `ftsave`/`ftload` path. This is a non-blocking comment, and I’m happy to make a follow-up PR for that part if useful; either approach is fine.

PasqualeMainolfi:
@hlolli... you were right

Thanks, that is a good catch, and everything you pointed out here is correct. This is
exactly the ownership edge I missed. I reproduced the case you described: saving table 1,
loading it as table 2, then freeing 2 and 1 can leave the loaded table holding the source
table's raw `args` pointer from the binary header, which makes the second free unsafe.
I think it is better to handle this in the same PR rather than leave it for a follow-up,
because this PR changes the `FUNC` cleanup invariant by making `ftfree()` release `args`.
With that new invariant in place, the existing binary serialization behavior becomes a
known double-free path, so it feels safer to make the ownership rule and the serialization
rule consistent before merging.

I updated the patch on both sides of the binary path. On save, binary `ftsave` now writes a
local copy of the `FUNC` header with `args`, `argcnt`, and `ftable` cleared, so new files
do not contain process-local pointers. On load, the header copy is centralized through a
helper that preserves the destination `ftable`, frees any previous destination `args`,
copies the serialized header, then resets `args` to `NULL` and `argcnt` to `0`. This also
protects older binary `ftsave` files that may already contain a stale raw `args` pointer.

I added a small commandline regression test for the save table 1, load as table 2, free 2,
then free 1 case. It also checks that the loaded table data matches the saved table before
freeing both tables:

```shell
ftload ownership: saving table 1
ftload ownership: loading table 1 as table 2
ftload ownership: checking loaded table data
ftload ownership: freeing table 2
ftload ownership: freeing table 1
ftload ownership: passed

Result: 0 errors in performance
```

I also verified the stale-header case explicitly by injecting a non-NULL fake pointer into
the on-disk args field and confirming that ftload ignores it safely. args is indeed inside
the binary prefix written by ftsave on this build:

```c
offsetof(FUNC, args) = 88
binary header size = 152
sizeof(FUNC) = 1184
```

After injecting 0x1111111111111111 at that offset, ftload, ftfree 2, and ftfree 1 completed
without a segfault.

Thanks again for spotting this; it was the missing ownership edge in the original fix.

Opcodes/ftgen.c
```diff
@@ -245,6 +245,17 @@ static int32_t myInitError(CSOUND *csound, OPDS *p, const char *str, ...)
...
 
+static void ftload_copy_header(CSOUND *csound, FUNC *ftp,
+                               const FUNC *header, size_t size)
+{
+  MYFLT *ftable = ftp->ftable;
+  csound->Free(csound, ftp->args);
+  memcpy(ftp, header, size);
+  ftp->ftable = ftable;
+  ftp->args = NULL;
+  ftp->argcnt = 0;
```
hlolli, line 256: `getftargs` can also return an empty string for this valid no-args state, since setting `argcnt` to 0 with `args == NULL` makes its loop start at 1 and dereference NULL, I reproduced the resulting segfault with `ftload(..., 2)` followed by `Sargs getftargs 2, 1`.

hlolli (commented):
One additional inline note from the follow-up review.

Opcodes/ftgen.c
hlolli:
```csound
<CsoundSynthesizer>
  <CsOptions>
  -n
  </CsOptions>
  <CsInstruments>
  sr = 44100
  ksmps = 32
  nchnls = 1
  0dbfs = 1

  giSrc = ftgen(1, 0, 16, -2, 1)
  giDst = ftgen(2, 0, 16, 2, 0)

  instr 1
    ftsave("args.ftsave", 0, 1)
    ftload("args.ftsave", 0, 2)
    Sargs getftargs 2, 1
  endin
  </CsInstruments>
  <CsScore>
  i 1 0 0.01
  </CsScore>
  </CsoundSynthesizer>
  ```

PasqualeMainolfi:
Fixed. The valid no-args state is now handled explicitly, and the loops are safe: they use `i < argcnt`, so they can no longer start at 1 and continue when `argcnt == 0`.

hlolli (approved):

hlolli:
@PasqualeMainolfi I did rebase and merge, causing my name to appear as co-author, that was unintentional. I sometimes forget what the green merge button is set to, I tend to jump between merge and rebase+merge. I hope it's ok.

PasqualeMainolfi:
@hlolli No worries at all! I’m happy to collaborate, and I really don’t mind. Thanks for letting me know!

PR #2596, opened 2026-07-10
Preserve whitespace inside preprocessor string arguments

hlolli:
The preprocessor removed whitespace while expanding string arguments, even when that whitespace was part of the string itself.

For example, a string containing spaces could be joined together or otherwise changed after macro expansion.

This change preserves whitespace inside quoted strings while keeping the existing whitespace handling outside strings. A regression test covers spaces within a preprocessor string argument.

### DIFF
--- a/tests/commandline/test_string_preprocessor_whitespace.csd
+++ b/tests/commandline/test_string_preprocessor_whitespace.csd
@@ -9,14 +9,13 @@ opcode AssertString(actual:S, prefix:S, suffix:S):i
   expected:S = prefix
   expected strcat expected, space
   expected strcat expected, suffix
+  result:i = 1
 
   if (strcmp(actual, expected) != 0) then
     prints "String whitespace failed: expected '%s', got '%s'\n", \
       expected, actual
+    result = 0
     exitnow(1)
-    result:i = 0
-  else
-    result:i = 1
   endif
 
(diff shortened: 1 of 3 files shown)


tests/commandline/test_string_preprocessor_whitespace.csd
```diff
@@ -0,0 +1,46 @@
+<CsoundSynthesizer>
```
vlazzarini, line 1: I can't see any preprocessor code in this test. What is it supposed to do?

hlolli: I think we may be using “preprocessor” in two slightly different senses here. You are probably looking for an explicit preprocessor construct such as #define, while this test covers the preprocessing pass that every orchestra goes through.

The test checks function-like text such as int ( and sin (. The do_function() helper removed whitespace from those matches, which is fine except when they appear inside quoted strings, "int (3.5)" became "int(3.5)".

The test does go trough the preprocessing path, thel int (3.5) call confirms that normal code retains its existing behavior.

hlolli:
@copilot I addressed both your comments, please re-review

hlolli:
@coderabbitai review

vlazzarini (approved):

PR #2597, opened 2026-07-10
Fix input and output order in declared UDO signatures

hlolli:
Forward-declared UDO signatures stored their input and output types in the opposite order.

This caused calls to a declared UDO to resolve against the wrong signature, particularly when the input and output types differed.

This change keeps the declared signature in the same input/output order as the UDO definition. Regression coverage verifies that forward declarations resolve and compile with distinct argument and return types.


vlazzarini:
@coderabbitai review

vlazzarini (approved):

PR #2607, opened 2026-07-15
fix(wasm): coordinate SAB performance completion

hlolli:
## Summary

This extracts the SAB completion work from #2603 so it can be reviewed independently of UDO recycling and structured-array ownership.

- Wait for both worker release and terminal play-state completion.
- Scope completion signals to the active performance run.
- Prevent delayed signals from an earlier run from completing a newer run.


hlolli:
@coderabbitai review

kunstmusik (approved):
Change looks good to me. I noticed the following (silent) error on our wasm GH actions:

Image

This looks not new but thought it worth mentioning. Approving this one.

hlolli:
@kunstmusik that's the action checkouts still at v4, we can update it to v5

PR #2609, opened 2026-07-16
Resolve race condition and deadlock in --realtime 

vlazzarini:
In recent changes to the `--realtime` mode, when running diskin2 asynchronously `realtime_init_pass()` drops `alloc_spinlock` during `init_pass()` to avoid lock inversion with opcodes like `prealloc` that acquire the spinlock.  This creates a window where the perf thread's `beat_expire() -> deact() -> deinit_pass() -> diskin2_async_deinit()` can race with `diskin2_init_()` running on the `event_insert_thread`.

Two bugs result:

1. NULL dereference (EXC_BAD_ACCESS at 0x430): When the last diskin2 instance is deinited by the perf thread, the `DISKIN_INST` chain becomes empty (`*top == NULL`).  The event_insert_thread then dereferences *top in the else branch of diskin2_init_(), crashing.

2. Deadlock: `diskin2_async_deinit()` called `diskin2_cleanup()` which called `csoundJoinThread()` while holding `alloc_spinlock`.  The `event_ insert_thread` blocked trying to re-acquire `alloc_spinlock` in r`ealtime_init_pass()`, deadlocking with the perf thread.

Fix:

- Add NULL guards in the else branches of `diskin2_init_() `and `diskin2_init_array_()`: when `*top` is `NULL` (chain emptied by a concurrent deinit), re-seed with the new instance.

- Split cleanup into two phases:
  * `diskin2_async_deinit()` now only signals the I/O thread to stop (sets `*start = 0`) — no join, no global destruction, no deadlock.
  * `diskin2_reset_callback()` (registered via `RegisterResetCallback`) performs the actual join and global cleanup at csoundReset time, when alloc_spinlock is not held and it is safe to block.

- Add NULL guards on `QueryGlobalVariable` results for the thread start code to handle the case where globals have not yet been created.

- Add `diskin2_reset_callback_array()` with the same pattern for the array variant (`diskin2_init_array_/diskin2_async_deinit_array`).

Tested with `tests/commandline/test_async_diskin.csd` under --realtime mode: 20/20 consecutive passes, 149/149 unit tests pass.

Fix by OpenCode using Big Pickle.

### DIFF
--- a/OOps/diskin2.c
+++ b/OOps/diskin2.c
@@ -290,7 +290,8 @@ int32_t soundin(CSOUND *csound, DISKIN2 *p){
 }
 
 static uintptr_t diskin_io_thread(void *p);
-static int32_t diskin2_cleanup(CSOUND *csound, void *p); 
+static int32_t diskin2_reset_callback(CSOUND *csound, void *p);
+static int32_t diskin2_reset_callback_array(CSOUND *csound, void *p);
 
 static int32_t diskin2_init_(CSOUND *csound, DISKIN2 *p, int32_t stringname)
 {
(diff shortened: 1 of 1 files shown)


hlolli (approved):
let's get this small fix in, and I'll rebase on top

PR #2619, opened 2026-07-19
fix(parser): preserve newline suppression in nested parentheses

hlolli:
Csound already suppresses newlines inside parenthesized expressions, but the lexer previously tracked this with a single `ignorenewline` state.

That breaks for nested parentheses because encountering any `)` restores normal newline handling, even when an outer parenthesis is still open. Function and opcode-call  tokens also consume their opening `(` directly, bypassing the existing lexer state.

This change:

- Replaces the lexer state with a per-parser parenthesis depth counter.
- Tracks opening parentheses consumed by function and opcode-call tokens.
- Resumes emitting `NEWLINE` only after the outermost parenthesis closes.
- Extends the existing newline parser test with grouped, nested, function-call, and nested-function expressions.

### DIFF
--- a/Engine/csound_orc.lex
+++ b/Engine/csound_orc.lex
@@ -93,7 +93,6 @@ RSTR            "R{"
 ERSTR           "}R"
 
 
-%s ignorenewline
 %x line
 %x sline
 %x src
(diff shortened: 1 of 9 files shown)


vlazzarini:
In Csound 7, we introduced nesting of multiline and raw strings, e.g.

```
{{ 
 {{ ...
 }}
 }}
```

etc. 

We need to make sure there is no regression involved. I think we have tests but we need to check.

hlolli:
@vlazzarini  Existing tests already cover nested {{…}} and R{…}R strings. I extended tests/commandline/test_newlines_within_function_calls.csd to cover both nested string types inside multiline parenthesized calls, including parentheses within the strings.

vlazzarini (approved):
Looks good.

PR #2630, opened 2026-07-21
fix for diskin2 async failure: join I/O threads before freeing instance memory

vlazzarini:
This may be a regression - diskin2 async test started failing again. But it may not be either. 

In reset(), csound_cleanup() was called before the registered reset callbacks. csound_cleanup() calls free_inactive_instances() which frees the DISKIN2/DISKIN2_ARRAY instance structs, but the async I/O threads (diskin_io_thread / diskin_io_thread_array) were still running and accessing that memory. The reset callbacks that join those threads ran too late.

This caused a heap-use-after-free detected by AddressSanitizer in diskin_file_read_array (diskin2.c:1251) during csound teardown.

Fix by:
- Moving the reset callback invocation before csound_cleanup() so I/O threads are stopped and joined before free_inactive_instances() reclaims their memory.
- Removing DestroyGlobalVariable calls from diskin2_reset_callback and diskin2_reset_callback_array, since csound_cleanup() still needs those globals (DISKIN_INST, DISKIN_THREAD_START, etc.) for diskin2_async_deinit_array to unlink instances from the chain. The globals are cleaned up by csoundDeleteAllGlobalVariables() later in reset().

Fix by opencode (Big Pickle).

### DIFF
--- a/OOps/diskin2.c
+++ b/OOps/diskin2.c
@@ -550,9 +550,6 @@ static int32_t diskin2_reset_callback(CSOUND *csound, void *p) {
       pt = csound->QueryGlobalVariable(csound, "DISKIN_PTHREAD");
       if (pt != NULL && *pt != NULL)
         csound->JoinThread(*pt);
-      csound->DestroyGlobalVariable(csound, "DISKIN_PTHREAD");
-      csound->DestroyGlobalVariable(csound, "DISKIN_THREAD_START");
-      csound->DestroyGlobalVariable(csound, "DISKIN_INST");
     }
 #endif
     return OK;
@@ -569,9 +566,6 @@ static int32_t diskin2_reset_callback_array(CSOUND *csound, void *p) {
       pt = csound->QueryGlobalVariable(csound, "DISKIN_PTHREAD_ARRAY");
       if (pt != NULL && *pt != NULL)
         csound->JoinThread(*pt);
-      csound->DestroyGlobalVariable(csound, "DISKIN_PTHREAD_ARRAY");
-      csound->DestroyGlobalVariable(csound, "DISKIN_THREAD_START_ARRAY");
-      csound->DestroyGlobalVariable(csound, "DISKIN_INST_ARRAY");
     }
 #endif
     return OK;
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -1868,14 +1868,16 @@ static void reset(CSOUND *csound) {
     ATOMIC_SET(csound->parflag,!csound->parflag);
 #endif
   }
-  csound_cleanup(csound);
-  /* call registered reset callbacks */
+  /* call registered reset callbacks before csound_cleanup so that
+     any I/O threads (e.g. diskin2 async) are joined before their
+     instance memory is freed by free_inactive_instances. */
   while (csound->reset_list != NULL) {
     resetCallback_t *p = (resetCallback_t *)csound->reset_list;
     p->func(csound, p->userData);
     csound->reset_list = (void *)p->nxt;
     free(p);
   }
+  csound_cleanup(csound);
   /* call local destructor routines of external modules */
   /* should check return value... */
   csoundDestroyModules(csound);


Top/csound.c
```diff
@@ -1868,14 +1868,16 @@ static void reset(CSOUND *csound) {
...
+  /* call registered reset callbacks before csound_cleanup so that
+     any I/O threads (e.g. diskin2 async) are joined before their
+     instance memory is freed by free_inactive_instances. */
   while (csound->reset_list != NULL) {
     resetCallback_t *p = (resetCallback_t *)csound->reset_list;
     p->func(csound, p->userData);
     csound->reset_list = (void *)p->nxt;
     free(p);
   }
+  csound_cleanup(csound);
```
hlolli, line 1880: This reorder swaps all thirteen reset-callback registrants ahead of cleanup, and OSC breaks under the new order: `OSC_reset` frees each port's `lo_server_thread` and destroys `port->mutex_` (neither is NULLed), and `deactivate_all_notes` inside `csound_cleanup` then runs `OSC_listendeinit` for any still-active `OSClisten` note — `LockMutex` on a destroyed mutex, `lo_server_thread_del_method` on a freed thread. Same UAF class this PR fixes for diskin2, recreated for OSC (any CSD exiting with a listener running). A narrower fix — joining the diskin2 threads at the top of `csound_cleanup`, before `free_inactive_instances` — avoids reordering registrants with opposing needs.

Top/csound.c
```diff
@@ -1868,14 +1868,16 @@ static void reset(CSOUND *csound) {
     ATOMIC_SET(csound->parflag,!csound->parflag);
 #endif
   }
-  csound_cleanup(csound);
-  /* call registered reset callbacks */
+  /* call registered reset callbacks before csound_cleanup so that
```
hlolli, line 1871: `udp_console_stop` (Top/server.c) also moves ahead of cleanup, so a host attached via `--port` loses the end-of-performance output (overall amps, error count) — the UDP mirror is torn down before `csound_cleanup` prints it. Conversely the swap likely *fixes* tabaudio's `on_reset_audio`, which is diskin2-shaped: the registrant list contains both pre-cleanup thread-joiners and post-cleanup state-destroyers, so no single ordering of the whole list is correct — which suggests either per-subsystem joins inside cleanup, or splitting the callback list into two phases.

OOps/diskin2.c
```diff
@@ -550,9 +550,6 @@ static int32_t diskin2_reset_callback(CSOUND *csound, void *p) {
       pt = csound->QueryGlobalVariable(csound, "DISKIN_PTHREAD");
       if (pt != NULL && *pt != NULL)
         csound->JoinThread(*pt);
```
hlolli, line 552: This half checks out: the callbacks set `*start = 0` before joining, keeping the `DISKIN_*` globals alive is needed for cleanup's `diskin2_async_deinit` unlink, and `csoundDeleteAllGlobalVariables` frees them later in `reset()`. Worth noting it only covers the teardown-time UAF though — mid-performance, the `DISKIN_INST` list is still traversed lock-free by the IO thread every wakeup while note deinits unlink from it concurrently, and instance dataspace can be freed under an in-flight `diskin_file_read` — so the async test may keep flaking under `--realtime` stress even with this merged.

hlolli:
I don't mind little hard to reproduce edge-case in the cleanup sequence because PR #2610 solves this fully by removing diskin2's reset callbacks altogether and sequencing the teardown inside csound_cleanup itself, both I/O worker threads are joined (diskin2_async_shutdown) before free_inactive_instances reclaims any instance memory, so the fix doesn't depend on reordering the shared reset-callback list that other modules like OSC and the UDP console rely on.

vlazzarini:
So should we just press ahead and wait until that one is merged? This particular issue seems to be only causing problems within ASAN but not outside of it. I made the PR because we had the issue in CI a couple of times yesterday. It doesn't need to be merged.

hlolli:
Yes, I think waiting for #2610 is the better course. I feel what's happening here is that ASan is exposing a race rather than an ASan specific problem. Your PR fixes the teardown case, #2610 handles the lifetime ordering at the diskin2 boundary. If there's a delay, we could definitely use your PR, it likely does fix the symptom. These llm comments seem to show what would need to be done wrt the cleanup ordering.

vlazzarini:
Sounds good. I'll close this now and we can reopen if needed.

PR #2706, opened 2026-07-25
fix: bound the sockrecv string receiver to the bytes received

### LINKED ISSUE #2684: The sockrecv string receiver runs strlen on raw packet data that need not be null-terminated
The UDP string form of `sockrecv` copies a received packet into a fixed `MTU`-sized buffer and then calls `strlen` on it. Nothing guarantees the packet contains a null byte, so a packet that fills the buffer with non-zero bytes makes `strlen` read past the end of the buffer.

`send_recv_S()` at `Opcodes/sockrecv.c:307` reads the string length with `strlen(&p->buf[p->outsamps])` on data that came straight from `recvfrom` in `udpRecv_S()`.

Tested on macOS with an AddressSanitizer build of Csound 7.0, double samples, commit `a7f70409def0a6955a56e6dae599f5f866080cff`.

## Reproducer

`repro_sockrecv_strlen.csd` receives on a UDP port in real time:

```csound
<CsoundSynthesizer>
<CsOptions>
-d -m0 -odac -+rtaudio=null
</CsOptions>
<CsInstruments>
sr = 48000
ksmps = 128
nchnls = 1
0dbfs = 1

instr 1
  Sres sockrecv 7775, 512
  if strlen(Sres) > 0 then
    printf "recv len=%d\n", 1, strlen(Sres)
  endif
endin
</CsInstruments>
<CsScore>
i 1 0 5
e
</CsScore>
</CsoundSynthesizer>
```

Sender (`send_raw_nonull.py`) sends a full `MTU` (1456-byte) packet with no null byte:

```python
import socket
payload = b"A" * 1456
socket.socket(socket.AF_INET, socket.SOCK_DGRAM).sendto(payload, ("127.0.0.1", 7775))
```

Run the CSD, then run the sender while it plays.

## Expected

`sockrecv` treats the packet as at most `MTU` bytes and returns a bounded string without reading past the receive buffer.

## Actual

AddressSanitizer reports a heap-buffer-overflow read in the `strlen`:

```text
==ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 1457 at 0x... thread T0
    #1 send_recv_S sockrecv.c:316
SUMMARY: AddressSanitizer: heap-buffer-overflow sockrecv.c:316 in send_recv_S
```

The read runs one byte past the 1456-byte buffer because no null terminator was found.

## Likely cause

`Opcodes/sockrecv.c`:

```c
len = strlen(&p->buf[p->outsamps]);          /* buf holds raw packet bytes, may lack '\0' */
...
strncpy(str->data, &p->buf[p->outsamps], len+1);
p->outsamps += len+1;
```

`p->buf` is filled by `udpRecv_S()`, which copies up to `MTU` bytes from `recvfrom` with no terminator added. The receiver should bound the search to the number of bytes actually received, for example by using `strnlen` against the remaining buffer length or by writing a terminating null after the received bytes. The `p->outsamps += len + 1` step can also walk past the valid data once `len` is wrong.

## Acceptance criteria

- The string length is bounded by the received byte count, never scanned past the buffer.
- A packet with no null byte returns a string of at most the bytes received and does not read out of bounds.
- The reproducer completes without a memory error under AddressSanitizer.
- Add a malformed-packet test with a full-`MTU`, unterminated payload run under AddressSanitizer.

hlolli:
The UDP string form of `sockrecv` copied a packet into a fixed buffer and then ran `strlen` on it. A packet with no null byte made `strlen` read past the buffer. The receiver now tracks how many bytes it holds and never scans beyond them: `udpRecv_S()` appends a terminating null to each packet before queuing it, and `send_recv_S()` finds the string end with `memchr` bounded by the received byte count, copies exactly that many bytes, and advances by the same amount. The circular buffer and its backing storage are sized and counted in bytes rather than `MYFLT` units, and the buffer-length argument is range-checked before it is cast so an out-of-range value can't produce an undefined conversion.

Fixing the overflow exposed a second, older defect on the same path. The `sockrecv.k` and `sockrecv.S` opcode entries passed four function pointers where `OENTRY` holds three, so their deinit function landed in `useropinfo` and the real deinit slot was `NULL`. The receiver thread was never joined, so at shutdown it kept reading the instrument instance after it was freed ‚Äî a use-after-free the reproducer now reaches because the process no longer aborts on the `strlen` overflow first. Dropping the stray `NULL` puts `deinit_udpRecv_S` back in the deinit slot; the thread is joined and the socket closed before the instance goes away.

A `ServerTests` case sends a full-MTU, unterminated UDP payload to a `sockrecv` string instrument and checks the opcode returns exactly the bytes sent. On the unfixed code it dies with the issue's heap-buffer-overflow in `send_recv_S`; with the fix it passes under AddressSanitizer.

## Before

Listener (`repro_sockrecv_strlen.csd`):

```csound
<CsoundSynthesizer>
<CsOptions>
-d -m0 -odac -+rtaudio=null
</CsOptions>
<CsInstruments>
sr = 48000
ksmps = 128
nchnls = 1
0dbfs = 1

instr 1
  Sres sockrecv 7775, 512
  if strlen(Sres) > 0 then
    printf "recv len=%d\n", 1, strlen(Sres)
  endif
endin
</CsInstruments>
<CsScore>
i 1 0 5
e
</CsScore>
</CsoundSynthesizer>
```

Sender (`send_raw_nonull.py`):

```python
import socket
payload = b"A" * 1456
socket.socket(socket.AF_INET, socket.SOCK_DGRAM).sendto(payload, ("127.0.0.1", 7775))
```

```text
==ERROR: AddressSanitizer: heap-buffer-overflow
READ of size 1457 at 0x... thread T0
    #1 send_recv_S sockrecv.c:316
SUMMARY: AddressSanitizer: heap-buffer-overflow sockrecv.c:316 in send_recv_S
```

## After

The unterminated packet is received as 1456 bytes, the receiver thread is joined at shutdown, and the run completes cleanly under AddressSanitizer.

Closes #2684

### DIFF
--- a/Opcodes/sockrecv.c
+++ b/Opcodes/sockrecv.c
@@ -141,7 +141,7 @@ static uintptr_t udpRecv(void *pdata)
 
 static int32_t deinit_udpRecv_S(CSOUND *csound, void *pdata)
 {
-    SOCKRECV *p = (SOCKRECV *) pdata;
+    SOCKRECVSTR *p = (SOCKRECVSTR *) pdata;
 
     p->threadon = 0;
     csound->JoinThread(p->thrid);
--- a/Opcodes/sockrecv.c
+++ b/Opcodes/sockrecv.c
@@ -47,6 +47,7 @@
 
 #define MAXBUFS 32
 #define MTU (1456)
+#define STRING_RECV_BUFFER_SIZE (MTU + 1)
 
 #ifndef WIN32
 extern  int32_t     inet_aton(const char *cp, struct in_addr *inp);
@@ -167,7 +168,8 @@ static uintptr_t udpRecv_S(void *pdata)
     while (p->threadon) {
       /* get the data from the socket and store it in a tmp buffer */
       if ((bytes = (int32_t) recvfrom(p->sock, (void *)tmp, MTU, 0, &from, &clilen)) > 0) {
-        bytes = (bytes+sizeof(MYFLT)-1)/sizeof(MYFLT);
+        if (tmp[bytes - 1] != '\0')
+          tmp[bytes++] = '\0';
         csound->WriteCircularBuffer(csound, p->cb, tmp, bytes);
       }
     }
@@ -239,7 +241,8 @@ static int32_t init_recv(CSOUND *csound, SOCKRECV *p)
 /* UDP version for strings */
 static int32_t init_recv_S(CSOUND *csound, SOCKRECVSTR *p)
 {
-    MYFLT   *buf;
+    char    *buf;
+    int64_t circular_buffer_size;
 #if defined(WIN32) && !defined(__CYGWIN__)
     WSADATA wsaData = {0};
     int32_t err;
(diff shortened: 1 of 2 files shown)


kunstmusik (approved):
Did a review with GPT 5.6 Sol xHigh and found a few issues. Pushed fixes and CI tests pass on all platforms.

PR #2707, opened 2026-07-25
fix: preserve C-locale parsing in csoundSscanf

### LINKED ISSUE #2685: csoundSscanf runs the va_list twice, which is undefined behaviour and defeats locale-independent parsing
In the `HAVE_SPRINTF_L` build of `csoundSscanf()` (`Top/getstring.c:265`), the function scans the argument list with the C-locale `vsscanf_l`, then immediately scans it again with the locale-dependent `vsscanf`, reusing the same `va_list`:

```c
va_start(args, format);
retVal = vsscanf_l(str, csound_c_locale, format, args);
retVal = vsscanf(str, format, args);   /* reuses an already-passed va_list */
va_end(args);
```

Passing one `va_list` to a scan function and then reusing it without `va_copy` is undefined behaviour. The second call also parses under the current locale, so it overrides the C-locale result the first call just produced. That defeats the whole point of the wrapper, which exists to parse numbers independently of the user's locale.

Confirmed by inspection on Csound 7.0, commit `a7f70409def0a6955a56e6dae599f5f866080cff`. The `HAVE_SPRINTF_L` branch is compiled on this build (`HAVE_SPRINTF_L:INTERNAL=1`).

## Affected paths

`csoundSscanf` is reached through `CS_SSCANF` and `csound->Sscanf`, including:

- `Top/csound_time.c:90` — `CS_SSCANF(s, "%lf", &timeResolutionSeconds)` reading the CPU frequency.
- `Engine/extract.c:92` and nearby — `%d:%lf` section and beat parsing for score extraction (`-x`).
- `util/srconv.c` and `util/dnoise.c` — numeric command-line options.

## Expected

A float such as `3.5` parses to `3.5` regardless of the user's `LC_NUMERIC`, and the argument list is consumed exactly once.

## Actual

The value is parsed twice. The second parse uses the current locale, so in a locale whose decimal separator is a comma, `3.5` parses as `3` and overwrites the correct C-locale result. Reusing the consumed `va_list` is undefined behaviour; on a libc that does not copy the list internally it reads past the passed arguments and can write a parsed value through a garbage pointer.

The sibling `HAVE__SPRINT_L` branch (`Top/getstring.c:289`) calls the scan function once and is correct, which shows the second line here is a stray copy.

## Likely cause

The `vsscanf` line is a leftover. It should be removed so the function scans once with `vsscanf_l`:

```c
va_start(args, format);
retVal = vsscanf_l(str, csound_c_locale, format, args);
va_end(args);
return retVal;
```

If both a locale-aware and a locale-independent scan were ever wanted, each needs its own `va_start`/`va_end` pair or a `va_copy`.

## Acceptance criteria

- `csoundSscanf` consumes the argument list once; no `va_list` is reused after being passed to a scan function.
- Number parsing stays locale-independent: a `%lf` parse of `3.5` returns `3.5` under a comma-decimal `LC_NUMERIC`.
- Add a test that calls `csound->Sscanf` for `%lf` under a comma-decimal locale and checks the result.

hlolli:
Closes #2685.

`csoundSscanf()` scanned the same input twice: first with `vsscanf_l()` and then with `vsscanf()` using the same `va_list`. The second call reused an exhausted argument list and could replace the C-locale result with one parsed under the current locale.

Remove the second scan so the argument list is consumed once, and add coverage for dot-decimal input under a comma-decimal locale.

## Before / after

With `value.txt` containing:

```text
3.5
```

and this CSD:

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d
</CsOptions>
<CsInstruments>
ksmps = 32

instr 1
  kValue readk "value.txt", 7, 1
  printf "value = %.1f\n", 1, kValue
  turnoff
endin
</CsInstruments>
<CsScore>
i 1 0 1
</CsScore>
</CsoundSynthesizer>
```

Run it from a host that sets `LC_NUMERIC` to `de_DE.UTF-8` after `csoundCreate()`. Text format `7` in `readk` reaches `CS_SSCANF`.

Before:

```text
value = 3,0
```

After:

```text
value = 3,5
```

### DIFF
--- a/Top/getstring.c
+++ b/Top/getstring.c
@@ -269,7 +269,6 @@ double csoundStrtod(char* nptr, char** endptr) {
     int32_t retVal;
     va_start(args, format);
     retVal = vsscanf_l(str,csound_c_locale,format,args);
-    retVal = vsscanf(str,format,args);
     va_end(args);
     return retVal;
 }
--- a/tests/c/engine_test.cpp
+++ b/tests/c/engine_test.cpp
@@ -4,6 +4,7 @@
 #include "fftlib.h"
 #include <algorithm>
 #include <atomic>
+#include <clocale>
 #include <cmath>
 #include <cstring>
 #include <stdio.h>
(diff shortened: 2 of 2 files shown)


hlolli:
@coderabbitai review

kunstmusik (approved):

PR #2721, opened 2026-07-27
Fix qinf/qnan classification under fast math

### LINKED ISSUE #2720: qinf/qnan misclassify non-finite values in optimized builds
## Problem

`qinf` and `qnan` should distinguish NaN, positive infinity, negative infinity, and finite values. In an optimized build where finite-math assumptions are active, the current implementation misclassifies some values:

- `qinf(-inf)` returns `1`, but it should return `-1`.
- `qnan(+inf)` returns `1`, but it should return `0`.

This matters because CSD code can use these opcodes to guard non-finite results. Losing the sign of infinity and treating infinity as NaN makes that guard unreliable.

## Reproducer CSD

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d
</CsOptions>
<CsInstruments>
ksmps = 10
nchnls = 1
0dbfs = 1

instr 1
kOne = 1
kZero = 0
kPosInf = kOne / kZero
kNegInf = -kOne / kZero
kNan = kZero / kZero
kFinite = 123

printks "qinf(+inf): expected 1, actual %d\n", 0, qinf(kPosInf)
printks "qinf(-inf): expected -1, actual %d\n", 0, qinf(kNegInf)
printks "qinf(finite): expected 0, actual %d\n", 0, qinf(kFinite)
printks "qnan(nan): expected 1, actual %d\n", 0, qnan(kNan)
printks "qnan(+inf): expected 0, actual %d\n", 0, qnan(kPosInf)
turnoff
endin
</CsInstruments>
<CsScore>
i1 0 0.01
</CsScore>
</CsoundSynthesizer>
```

## Actual output

Built from pre-fix code with `-ffinite-math-only` last in the release C flags, then ran the CSD above:

```text
qinf(+inf): expected 1, actual 1
qinf(-inf): expected -1, actual 1
qinf(finite): expected 0, actual 0
qnan(nan): expected 1, actual 1
qnan(+inf): expected 0, actual 1
```

## Expected output

```text
qinf(+inf): expected 1, actual 1
qinf(-inf): expected -1, actual -1
qinf(finite): expected 0, actual 0
qnan(nan): expected 1, actual 1
qnan(+inf): expected 0, actual 0
```

## Notes

The implementation should not rely on compiler or libc floating-point classification in builds that may use fast-math style assumptions. A bit-pattern check keeps `qinf` and `qnan` stable for scalar and audio-rate inputs.

hlolli:
## Summary

`qinf` and `qnan` were not reliable in optimized builds that use finite-math assumptions. `qinf` called `isinf()`, and `qnan` used bit masks that could treat infinities as NaNs. Under fast-math style flags, those paths can report the wrong class for values that CSD code uses as guards.

Before:

```text
qinf(+inf): 1
qinf(-inf): 1
qnan(nan): 1
qnan(+inf): 1
```

After:

```text
qinf(+inf): 1
qinf(-inf): -1
qnan(nan): 1
qnan(+inf): 0
```

## Fix

This changes scalar and audio-rate `qinf`/`qnan` to inspect the `MYFLT` bit pattern directly. The exponent and fraction bits identify infinity and NaN without relying on `isinf()` or `isnan()`, so the result stays stable when the compiler assumes finite math.

The audio-rate variants use the same helpers, so they keep the same behavior as the scalar opcodes while counting the active samples.

Closes #2720

### DIFF
--- a/OOps/aops.c
+++ b/OOps/aops.c
@@ -2229,66 +2229,67 @@ int32_t divinak(CSOUND *csound, ASSIGN *p)
 /**
  * Identifies both signaling NaN (sNaN) and quiet NaN (qNaN).
  *
- * According to the IEEE 754 standard, all NaN have the sign bit set to 0 and
+ * IEEE 754 NaNs can have either sign bit, and
  * all exponent bits set to 1. qNaN has the most significant bit of the
- * fractional set to 1, while sNaN has most the significant bit of the
- * fraction set to 0 -- but the NEXT most significant bit of the fraction must
- * be set to 1! This is necessary in order to distinguish sNaN from positive
- * infinity. Hence, there are 2 bit masks to test. Doubles have the most
- * significant bit of the fraction in (0-based) bit 52, floats have the most
+ * fractional set to 1, while sNaN has the most significant bit of the
+ * fraction set to 0. In both cases, the remaining fraction bits contain a
+ * nonzero payload that distinguishes NaN from infinity. Doubles have the most
+ * significant bit of the fraction in (0-based) bit 51, floats have the most
  * significant bit of the fraction in bit 22.
  * double qNaN:
- * 0111111111110000000000000000000000000000000000000000000000000000
- * 0x7FF0000000000000ULL
+ * 0111111111111000000000000000000000000000000000000000000000000000
+ * 0x7FF8000000000000ULL
  * double sNaN:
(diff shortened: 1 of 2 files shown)


vlazzarini:
We can't allow fast math optimisations that break nan encoding (see my comments in #2720).

PR #2723, opened 2026-07-27
CI probe for qinf and qnan with preserved NaN builds

hlolli:
## Summary

This is a throwaway CI probe on top of `fix/enforce-fno-finite-math-only`.

It adds the qinf/qnan unit tests from #2721, but does not include the opcode fixes from that PR. The goal is to check whether preserving NaN/Inf semantics in the build flags is enough for the current opcode code to pass under CI.

Expected result: if #2722 fixes the build flags correctly, these tests should pass without the qinf/qnan implementation changes.

### DIFF
--- a/tests/c/engine_test.cpp
+++ b/tests/c/engine_test.cpp
@@ -2,11 +2,19 @@
 #include "csoundCore.h"
 #include "csound_graph_display.h"
 #include "fftlib.h"
+extern "C" {
+#include "aops.h"
+int32_t is_NaN(CSOUND *, void *);
+int32_t is_NaNa(CSOUND *, void *);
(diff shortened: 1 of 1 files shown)


hlolli:
Closing this CI probe because it served its purpose.

This PR showed that the qinf/qnan regression tests pass when based on `fix/enforce-fno-finite-math-only`. That supports the build-flag fix: preserving NaN/Inf semantics is enough for the current opcode code to pass these tests.

I moved the test coverage into `fix/enforce-fno-finite-math-only` as commit `02a21628b`, with a source comment that points future failures at finite-math-only / fast-math flag order first.

PR #2724, opened 2026-07-27
CI probe for qinf and qnan on develop

hlolli:
## Summary

This is a throwaway CI probe based directly on `develop`.

It adds the same qinf/qnan unit tests as #2723, but does not include the build flag fix from #2722 or the opcode implementation changes from #2721.

The goal is to confirm that the tests fail on the current optimized Linux build when `-ffast-math` enables finite-math-only.

Expected result: at least one optimized Linux CI job should fail these tests on `develop`.

### DIFF
--- a/tests/c/engine_test.cpp
+++ b/tests/c/engine_test.cpp
@@ -2,11 +2,19 @@
 #include "csoundCore.h"
 #include "csound_graph_display.h"
 #include "fftlib.h"
+extern "C" {
+#include "aops.h"
+int32_t is_NaN(CSOUND *, void *);
+int32_t is_NaNa(CSOUND *, void *);
+int32_t is_inf(CSOUND *, void *);
+int32_t is_infa(CSOUND *, void *);
+}
 #include <algorithm>
 #include <atomic>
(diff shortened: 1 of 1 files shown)


hlolli:
this confirms that the test here https://github.com/csound/csound/pull/2723 passes when https://github.com/csound/csound/pull/2722 is included. I'll close this PR now as its usefulness is over. And I think I'd suggest we merge together 2723 and 2722 so that we get a fix, and a test that catches a regression. Wdyt @vlazzarini ?

hlolli:
Closing this CI probe because it served its purpose.

This PR ran the same qinf/qnan regression tests directly on `develop`, without the build-flag fix from #2722 and without the opcode hardening from #2721. CI failed there, while #2723 passed on top of `fix/enforce-fno-finite-math-only`.

That confirms the test is catching the optimized-build finite-math-only problem. The useful test coverage has now been added to `fix/enforce-fno-finite-math-only` in commit `02a21628b`.

vlazzarini:
sounds right to me.

PR #2719, opened 2026-07-26
Add stm: state-machine graph opcodes for Csound

PasqualeMainolfi:
This PR supersedes #2587, which was accidentally closed after a force-push following a rebase.

Issue #2625, opened 2026-07-20
WASM C++ plugin callbacks are not relocated from the isolated plugin table


PasqualeMainolfi:
Running the STM (#2587) build exposed a previously hidden bug in the WASM C++ plugin loader.
I initially suspected that adding the STM opcodes had affected function-table allocation, or that we had reached some opcode or indirect-function-table limit. Since I could not identify the cause directly, I performed a guided analysis with GPT 5.6, tracing the allocation and use of the WASM function-table indices and comparing successful and failing CI artifacts.
The analysis showed that there is no fixed opcode limit involved and that the STM implementation is not directly corrupting the table.
The actual problem is that the browser loader instantiates C++ plugins with an isolated indirect-function table. The plugin then calls `csoundModuleInit()`, which registers its opcode callbacks through `AppendOpcode`. However, the callback indices are copied into Csound’s opcode registry without being relocated from the plugin table to the host table. In the `plugin_example_cpp.wasm` case:

```text
hello440.kperf → plugin-table index 3
```

Index 3 is valid inside the plugin’s isolated table. During performance, however, Csound resolves that same numerical index in the host table. In the host table, index 3 refers to `midiDataRead`, which has a different WASM signature. The first execution of `hello440.kperf` therefore produces:

```text
RuntimeError: function signature mismatch 
at csound.kperf 
at csound.csoundPerformKsmps
```

I verified this experimentally by replacing only host-table slot 3 with the function from plugin-table slot 3. After that single substitution, the failing artifact completed `csoundPerformKsmps()` successfully.

The C-style opcode-plugin path does not have this problem because it already relocates init, perf, and deinit callbacks by appending them to the host table and replacing the plugin-local indices with the newly allocated host indices. The `csoundModuleInit()` C++ path does not perform this relocation.
The plugin binaries from the successful and failing builds are byte-for-byte identical. The invalid index was therefore already present in the previously successful build. The old test passed because it did not reliably execute a performance cycle: it called `compileCSD()`, `start()`, and immediately `stop()` without ensuring that `hello440.kperf` had actually run. Changes present in the STM build caused the performance path to reach the callback, exposing the latent loader bug.
Therefore, STM did not introduce the incorrect index. Its execution made an existing bug observable.
The proper fix should relocate C++ plugin callbacks into newly allocated host-table slots before they are registered with Csound, equivalent to what the C opcode-plugin path already does. The regression test should also explicitly execute at least one kperf cycle and verify that the original host-table entries are not overwritten.

PR #2725, opened 2026-07-29
add remap opcode

PasqualeMainolfi:
# Add the `remap` opcode

This adds `remap`, an opcode that maps a value onto an arbitrary curve defined by
a breakpoint table. Given a pair of arrays holding the breakpoints (`xdata`,
`ydata`), it returns the `y` matching the input `x`, interpolating between the
two breakpoints that bracket it.

It is directly inspired by `scipy.interpolate.interp1d`, and the argument list
mirrors it on purpose: `mode` corresponds to `kind`, and `bounds` plus the
optional `fill` cover what SciPy splits between `bounds_error` and `fill_value`
(including `fill_value="extrapolate"`). The four non-cubic modes are named after
SciPy's and produce the same values.

The reason I think this is worth having in Csound is that arbitrary mappings are
everywhere in practice and there is currently no single opcode for them. `remap` takes the
breakpoints as plain arrays, so they can be irregular, computed in the
orchestra, and rewritten during performance simply by assigning to the array.

It also covers the three shapes of the problem with the same opcode: a single
control value, a whole array in one call, and an audio signal. That last case is
the one I find most interesting: mapping an a-rate signal through a breakpoint
table *is* a waveshaper whose transfer function is an ordinary Csound array.

```csound
xd:i[] = fillarray(-1, -0.5, 0, 0.5, 1)
yd:i[] = fillarray(-1, -0.7, 0, 0.7, 1)

sig:a = oscili(0.8, 220)
out(remap(sig, xd, yd, 4, 1))      ; cubic, clamp
```

Changing the curve means writing into `yd`, so the transfer function can be
drawn, morphed or modulated at k-rate without rebuilding a table. The
shape-preserving cubic matters here: it keeps the transfer curve free of the
overshoot a natural spline would add between the points the user actually chose.

## What it does

```
y:k   = remap(x:k,   xdata:k[], ydata:k[], mode:i, bounds:i [, fill:i])
y:k[] = remap(x:k[], xdata:k[], ydata:k[], mode:i, bounds:i [, fill:i])
y:a   = remap(x:a,   xdata:k[], ydata:k[], mode:i, bounds:i [, fill:i])
```

Nine overloads: scalar, array and audio input, each accepting the breakpoint
tables at i- or k-rate. Five interpolation modes (linear, nearest, previous,
next, shape-preserving cubic) and four out-of-range policies (error, clamp,
fill, extrapolate).

The `x` table must be strictly increasing; both tables must hold at least two
points and have the same length.

### The cubic mode is PCHIP, evaluated locally

This is the one place where the opcode deliberately departs from `interp1d`.
SciPy's `kind='cubic'` is a natural cubic spline; `remap` uses the
shape-preserving Fritsch-Carlson cubic instead, the scheme SciPy exposes
separately as `PchipInterpolator`. On a monotonic table the result stays
monotonic and never leaves the range of the tabulated values, which matters when
the curve drives an amplitude or a frequency and an overshoot would be audible as
a bump the user never asked for.

The difference is not academic. On the table `x = [0, 1, 3, 4]`,
`y = [0, 2, 2, 6]`, whose middle segment is flat, the spline dips to 1.5625
between two breakpoints that both hold 2 — a deviation of 0.44 invented out of
nothing. PCHIP stays at exactly 2 across the whole segment.
PCHIP is a *local* scheme: the slope at a breakpoint depends only on the one or 
two intervals adjacent to it, so the coefficients of a single segment can be derived 
on demand from a four-point stencil. 
Evaluating a segment costs a fixed number of operations regardless of
table size, and the opcode needs no auxiliary allocation at all, which also
removes the whole question of allocating from the audio thread when a k-rate
table is resized mid-note.

The array and audio versions memoize the coefficients of the current segment and
rebuild them only when consecutive inputs cross into a different one, so the work
per call is bounded by `min(number of inputs, number of segments)`, never more
than tabulating the whole curve would cost, and usually far less.

### The interval search is branchless

The search is the branchless form of binary search: the trip count depends only
on the table size, so the loop is perfectly predicted and the single
data-dependent step compiles to a conditional move.

The branchless form is faster than a plain binary search on *every* table size and *both*
access patterns, and its cost barely differs between them:

| n | pattern | branchy | branchless |
|---|---|---|---|
| 16 | coherent ramp | 1.89 ns | 1.67 ns |
| 16 | random | 2.54 ns | 1.75 ns |
| 256 | random | 5.38 ns | 4.81 ns |

Predictable behaviour under any input seemed worth more than a best case that
depends on how the user drives the opcode.

### Validation happens where it is free

When both tables are i-rate the breakpoints cannot change during the note, so
they are validated once at init, including a strict-increase check that fails
loudly before the note starts. With k-rate tables the same check would cost O(n)
per control period, so it is not performed and the precondition is documented
instead. The opcode detects which case applies by inspecting the `intypes` of
the resolved overload.

## Notes and limitations

- With k-rate tables the strict-increase precondition on `xdata` is not checked.
  A non-increasing table yields undefined values rather than an error, in every
  mode. Validating it would cost an O(n) scan per control period.
- `previous` and `next` are step functions, so on a breakpoint they return the
  end of the segment the search lands on rather than the breakpoint itself. The
  tests pin this down explicitly so it is not mistaken for a rounding issue.
- `bounds = error` reports through `PerfError`, so it terminates the instrument
  instance rather than clamping and continuing.

### DIFF
--- a/Android/CsoundAndroid/jni/Android.mk
+++ b/Android/CsoundAndroid/jni/Android.mk
@@ -307,7 +307,8 @@ $(CSOUND_SRC_ROOT)/Opcodes/paulstretch.c \
 $(CSOUND_SRC_ROOT)/Engine/cs_new_dispatch.c \
 $(CSOUND_SRC_ROOT)/Engine/cs_par_base.c \
 $(CSOUND_SRC_ROOT)/Engine/cs_par_orc_semantic_analysis.c \
-$(CSOUND_SRC_ROOT)/Java/cs_glue.cpp
+$(CSOUND_SRC_ROOT)/Java/cs_glue.cpp \
+$(CSOUND_SRC_ROOT)/Opcodes/remap.c
 #CsoundObj.cpp
 
 LOCAL_LDLIBS += -llog -lOpenSLES -laaudio -lamidi -ldl -lm -lc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1191,6 +1191,7 @@ set(stdopcod_SRCS
     Opcodes/wterrain2.c
     Opcodes/stdopcod.c
     Opcodes/dbap.c
+    Opcodes/remap.c
     )
 
 set(sock_ops_SRCS
--- a/Opcodes/README.md
+++ b/Opcodes/README.md
@@ -102,6 +102,7 @@ Opcodes: internal and plugin opcodes
 - pvsgendy.c: streaming phase vocoder GENDY processing
 - pvsops.cpp: phase vocoder opcodes, time-varying convolution
 - quadbezier.c: bezier lines
+- remap.c: remaps values through an interpolated lookup curve
 - repluck.c: plucked-string physical models
 - reverbsc.c: FDN reverb
 - scansyn.c: scanned synthesis
--- a/Opcodes/remap.c
+++ b/Opcodes/remap.c
@@ -1,0 +1,668 @@
+/*
+    remap.c:
+
+    Copyright (C) 2026 Pasquale Mainolfi.
+
+    This file is part of Csound.
+
+    The Csound Library is free software; you can redistribute it
+    and/or modify it under the terms of the GNU Lesser General Public
+    License as published by the Free Software Foundation; either
+    version 2.1 of the License, or (at your option) any later version.
+
+    Csound is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU Lesser General Public License for more details.
+
+    You should have received a copy of the GNU Lesser General Public
+    License along with Csound; if not, write to the Free Software
+    Foundation, Inc., 31 Milk Street, #960789, Boston, MA, 02196, USA
+*/
+
+/*
+    remap - map a value onto an arbitrary curve defined by a breakpoint table.
+
+    Given a table of breakpoints (xdata, ydata), remap returns the y matching
+    the input x, interpolating between the two breakpoints that bracket it.
+    The x table must be strictly increasing; both tables must hold at least
+    two points and have the same length.
+
+    SYNTAX
+
+      y:k   = remap(x:k,   xdata:k[], ydata:k[], mode:i, bounds:i [, fill:i])
+      y:k[] = remap(x:k[], xdata:k[], ydata:k[], mode:i, bounds:i [, fill:i])
+      y:a   = remap(x:a,   xdata:k[], ydata:k[], mode:i, bounds:i [, fill:i])
+
+    The two tables may be i- or k-rate arrays. With i-rate tables the
+    breakpoints cannot change during the note, so they are validated once at
+    init; with k-rate tables they may be rewritten at any control period and
+    are re-checked every time.
+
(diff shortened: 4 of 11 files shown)


PasqualeMainolfi:
**CI: the WASM "can load and run c++ plugins" failure is not from this PR**

Nothing here touches WASM or the plugin loader, the diff is the opcode, its registration in stdopcod, and the tests. Not being able to see where the error actually originated, I ran a second review with Claude Opus 5, tracing the loader path against the sources, and it lines up with what I already described in #2587.

The C opcode-plugin path relocates the callbacks. wasm/browser/src/dlinit.js:117-129:

```ts
const remapToHostTable = (funcIndex) => {
  const funcRef = pluginTable.get(funcIndex);
  const hostTableIndex = table.length;
  table.grow(1);
  table.set(hostTableIndex, funcRef);
  return hostTableIndex;
};
```

It is applied to init, perf and deinit while the OENTRY array is rebuilt (lines 151-153), before appendOpcodes is called. So the indices reaching Csound's registry are host-table indices.

The csoundModuleInit path has no equivalent. The plugin calls cs->AppendOpcode() from inside csnd::plugin<T>() (include/plugin.h:1215) passing (SUBR)init<T> / (SUBR)perf<T> / (SUBR)deinit<T> directly, which are indices into the plugin's own table. They are stored unrelocated and later resolved against the host table, where the same numeric index refers to an unrelated function with a different signature.

Two details worth adding to the description in #2625:

- The isolated table is conditional, not unconditional. module.js:413 is useIsolatedPluginTable = !hasLegacyWasiPluginLoader || (hasOpcodeInit && !hasModuleInit). For a modload.h plugin the second term is false, so it comes down to whether the host exports csoundWasiLoadPlugin (module.js:392-394), which is compiled only under #ifdef __wasi__ (Top/csmodule.c:290). The experiment in #2719, substituting host slot 3 with plugin slot 3, shows the isolation is in effect in the failing artifact, so that export is absent there. If it is supposed to be present, that is a second thing to look at.
- The asymmetry is not C versus C++. It is which entry point the plugin exports: LINKAGE produces csound_opcode_init (relocated path), modload.h produces csoundModuleInit (not relocated). A C plugin using modload.h would hit the same bug, and a C++ plugin exporting csound_opcode_init would not. Worth stating so the fix is looked for in the csoundModuleInit branch of the loader rather than on the C++ side.

Registering new built-in opcodes shifts the host function table, which turns the already-invalid index into a live collision. The plugin binary itself is unchanged.

hlolli:
@PasqualeMainolfi wasm has 3857 function-table slots; the PR increases it to 3862. Which is fine but this is exposing a boundry bug we never notices. I'll make a fix for this.

PasqualeMainolfi:
@hlolli good, this should also fix the failed check in #2719

hlolli:
@PasqualeMainolfi ok this commit fixes it, I'll cherry pick it and open new PR. I'd say for now we can ignore the wasm errors. I'll force push and remove the commit from your PR, we have the proof we need that this fixes the issue.

hlolli:
@PasqualeMainolfi try to rebase, fix the merge conflict, and the tests should pass now

PasqualeMainolfi:
@hlolli thanks. I saw the merge. I’m away at the moment, but as soon as I’m back I’ll rebase and push.

PR #2726, opened 2026-07-29
fix(wasm): enable signal-flow graph opcodes

hlolli:
## What

- Build the existing `signalflowgraph.cpp` in WASI command and browser builds.
- Register the module in the WASI static module list.
- Skip the null mutex check only for WASI.
- Run both signal-flow graph command-line tests on WASM and native builds.

## Why

PR #2717 added a second C implementation because the C++ file was thought not to work with WASI.

The [PR discussion](https://github.com/csound/csound/pull/2717#issuecomment-5107978698) raised the cost of keeping two versions. Testing showed that the current WASI toolchain builds the C++ file.

The only runtime issue was `Create_Mutex()` returning null on WASI. WASI mutex calls are no-ops, but `ftgenonce_()` treated the null value as an init failure. This change skips that check only under `__wasi__`.

This keeps one signal-flow graph implementation.

## Checks

- WASI command build
- WASI browser build
- WASI command-line tests: 317 passed, 0 failed
- Native signal-flow graph opcode test
- Native signal-flow graph lifetime and `ftgenonce` test

### DIFF
--- a/Opcodes/signalflowgraph.cpp
+++ b/Opcodes/signalflowgraph.cpp
@@ -1343,11 +1343,12 @@ static int32_t ftgenonce_(CSOUND *csound, FTGEN *p, bool isNamedGenerator,
   }
   SignalFlowGraphState *sfg_globals = nullptr;
   QueryGlobalPointer(csound, "sfg_globals", sfg_globals);
-#ifndef __wasi__
+#ifdef __wasi__
+  // signal_flow_ftables_lock is expected to be NULL on WASI.
+  if (UNLIKELY(sfg_globals == nullptr)) {
+#else
   if (UNLIKELY(sfg_globals == nullptr ||
                sfg_globals->signal_flow_ftables_lock == nullptr)) {
-#else
-  if (UNLIKELY(sfg_globals == nullptr)) {
 #endif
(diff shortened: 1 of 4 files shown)


gogins:
This looks correct to me.

Opcodes/signalflowgraph.cpp
```diff
@@ -1343,8 +1343,12 @@ static int32_t ftgenonce_(CSOUND *csound, FTGEN *p, bool isNamedGenerator,
   }
   SignalFlowGraphState *sfg_globals = nullptr;
   QueryGlobalPointer(csound, "sfg_globals", sfg_globals);
+#ifndef __wasi__
```
kunstmusik: @hlolli could you flip the branches to use ifdef rather than ifndef?  Could also add a comment to explan sgnal_flow_ftables_lock is expected to be NULL on WASI.

kunstmusik (changes requested):

Opcodes/signalflowgraph.cpp
hlolli: cases are swapped

kunstmusik (approved):

PR #2734, opened 2026-08-04
Honor optional i-rate connect gain in signal-flow graph inlets.

gogins:
Apply Connection.gain when inlets accumulate outlets (default 1), and add a regression test covering omit, unity, scale, and fan-in. At this time, no k-rate or a-rate gain exists; only i-rate. Review shows no prior use of the gain parameter in available Csound orchestras.

### DIFF
--- a/Opcodes/signalflowgraph.cpp
+++ b/Opcodes/signalflowgraph.cpp
@@ -762,6 +762,8 @@ struct Inletf : public OpcodeBase<Inletf> {
     float *source = 0;
     CMPLX *sinkFrame = 0;
     CMPLX *sourceFrame = 0;
+    int32_t newLastframe = lastframe;
+    bool mergedNonSliding = false;
     // Loop over the source connections...
     for (size_t sourceI = 0, sourceN = sourceOutlets->size(); sourceI < sourceN;
          sourceI++) {
@@ -832,6 +834,8 @@ struct Inletf : public OpcodeBase<Inletf> {
           source = (float *)sourceOutlet->fsignal->frame.auxp;
           // Compare against the source frame counter; using the sink's
           // framecount here stopped updates after the first merge.
+          // Defer updating shared lastframe until all sources are checked so
+          // fan-in from multiple outlets with the same framecount still merges.
           if (lastframe < int(sourceOutlet->fsignal->framecount)) {
             for (size_t binI = 0, binN = fsignal->N + 2; binI < binN;
                  binI += 2) {
@@ -841,10 +845,16 @@ struct Inletf : public OpcodeBase<Inletf> {
                 sink[binI + 1] = source[binI + 1];
               }
             }
-            fsignal->framecount = lastframe = sourceOutlet->fsignal->framecount;
+            if (int(sourceOutlet->fsignal->framecount) > newLastframe) {
+              newLastframe = int(sourceOutlet->fsignal->framecount);
+            }
+            mergedNonSliding = true;
           }
         }
       }
+    }
+    if (mergedNonSliding) {
+      fsignal->framecount = lastframe = newLastframe;
     }
     return result;
   }
--- a/Opcodes/signalflowgraph.cpp
+++ b/Opcodes/signalflowgraph.cpp
@@ -91,7 +91,9 @@
  *
  * Optional igain is an i-rate linear amplitude applied to that connection
  * when the sink inlet accumulates sources. It defaults to 1 (unity), so
- * omitted gain and explicit unity match historical behavior.
+ * omitted gain and explicit unity match historical behavior. If the same
+ * source outlet is connected more than once to the same sink inlet, the
+ * gains are summed.
  *
  * alwayson Tinstrument [p4, ..., pn]
  *
@@ -263,6 +265,13 @@ struct SignalFlowGraphState {
   std::vector<std::vector<std::vector<Outletf *> *> *> foutletVectors;
   std::vector<std::vector<std::vector<Outletv *> *> *> voutletVectors;
   std::vector<std::vector<std::vector<Outletkid *> *> *> kidoutletVectors;
+  // Gain vectors are heap-pooled like sourceOutlets: opcode structs are
+  // zeroed raw memory and must not own std::vector by value.
+  std::vector<std::vector<MYFLT> *> againVectors;
+  std::vector<std::vector<MYFLT> *> kgainVectors;
+  std::vector<std::vector<MYFLT> *> fgainVectors;
+  std::vector<std::vector<MYFLT> *> vgainVectors;
+  std::vector<std::vector<MYFLT> *> kidgainVectors;
   SignalFlowGraphState(CSOUND *csound_) {
     csound = csound_;
     signal_flow_ports_lock = csound->Create_Mutex(0);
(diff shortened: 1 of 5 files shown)


tests/commandline/test_connect_gain.csd
```diff
@@ -0,0 +1,94 @@
...
+    if abs(kin - 0.25) > 1e-6 then
+      printks "connect.S gain failed: expected=0.25 got=%f\n", 0, kin
+      exitnowk(1)
+    endif
+  endif
+endin
+</CsInstruments>
+<CsScore>
+e 0.05
+</CsScore>
```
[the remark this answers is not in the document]
gogins: Addressed in the follow-up commit: `timeinstk()` counts **k-cycles**, not seconds. With `sr=48000` / `ksmps=32`, `e 0.05` is many cycles, so the checks do run. Clarified that in a comment and switched the guard to `timeinstk() >= 2`.

tests/commandline/test_connect_gain.csd
```diff
@@ -0,0 +1,94 @@
...
+
+; Non-unity gain scales the connection.
+connect "SourceA", "out", "SinkHalf", "in", 0.5
+
+; Fan-in with per-edge gains: 1 + 0.5 = 1.5
+connect "SourceA", "out", "SinkFan", "in", 1
+connect "SourceB", "out", "SinkFan", "in", 0.5
+
+; String-instrument overload with gain (connect.S).
+connect "SourceA", "out", "SinkString", "in", 0.25
```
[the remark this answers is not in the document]
gogins: Fixed: that connection now uses an explicit `connect.S` call so the string-instrument overload entry is exercised.

Opcodes/signalflowgraph.cpp
```diff
@@ -360,6 +369,7 @@ struct Inleta : public OpcodeBase<Inleta> {
    */
   char sinkInletId[MAX_STRING];
   std::vector<std::vector<Outleta *> *> *sourceOutlets;
+  std::vector<MYFLT> sourceGains;
```
hlolli: **Blocker:** Csound gives opcode structs zeroed raw memory. It does not run C++ constructors or destructors. A direct `std::vector` member is unsafe, and its buffer leaks at reset. This applies to every `sourceGains` member below. Please pool it in `SignalFlowGraphState`, like `sourceOutlets`, or pool `(outlet, gain)` pairs.

Opcodes/signalflowgraph.cpp
```diff
@@ -764,9 +787,10 @@ struct Inletf : public OpcodeBase<Inletf> {
           if (lastframe < int(fsignal->framecount)) {
             for (size_t binI = 0, binN = fsignal->N + 2; binI < binN;
                  binI += 2) {
-              if (source[binI] > sink[binI]) {
-                source[binI] = sink[binI];
-                source[binI + 1] = sink[binI + 1];
+              float amp = source[binI] * gain;
```
hlolli, line 887: **Blocker:** This non-sliding merge is in the `else` of the active-source check, so it runs only for inactive sources. Active non-sliding fsigs never use the gain. This also fixes the old reversed copy with no fsig test or note. Please handle the active fsig case and add an fsig test. Note the copy fix, or split it into another PR.

Opcodes/signalflowgraph.cpp
```diff
@@ -397,18 +408,19 @@ struct Inleta : public OpcodeBase<Inleta> {
...
-      const std::string &sourceOutletId = sourceOutletIds[i];
+    for (size_t i = 0, n = sourceConnections.size(); i < n; i++) {
+      const Connection &connection = sourceConnections[i];
       std::vector<Outleta *> &aoutlets =
-          sfg_globals->aoutletsForSourceOutletIds[sourceOutletId];
+          sfg_globals->aoutletsForSourceOutletIds[connection.sourceOutletId];
       if (std::find(sourceOutlets->begin(), sourceOutlets->end(), &aoutlets) ==
           sourceOutlets->end()) {
         sourceOutlets->push_back(&aoutlets);
+        sourceGains.push_back(connection.gain);
```
hlolli: `std::find` above drops later links to the same outlet, so only the first gain is kept. For example, `0.5` followed by `0.25` still gives `0.5`. Please add duplicate gains together, or say in the header that the first gain wins.

tests/soak/connect.csd
```diff
@@ -16,17 +16,17 @@ nchnls = 2
 
 ; Connect up the instruments to create a signal flow graph.
 
-connect "SimpleSine",   "leftout",     "Reverberator",     	"leftin"
-connect "SimpleSine",   "rightout",    "Reverberator",     	"rightin"
+connect "SimpleSine",   "leftout",     "Reverberator",     	"leftin", 1
```
hlolli: This adds no useful coverage; the new regression test already checks gain `1`. It also makes the example harder to read and copy. Please revert this file.

HUMAN EDIT: this is a nit, don't worry about this, take it as optional non-blocking comment.

hlolli (changes requested):
Requesting changes for two blockers:

- Direct `sourceGains` vectors are unsafe in raw opcode memory and leak their buffers.
- Active, non-sliding fsig sources never apply the gain.

The core per-edge design looks good. Please also define how duplicate edges handle gains and revert the example-only churn.

Opcodes/signalflowgraph.cpp
gogins: Done: `sourceGains` is now a pooled `std::vector<MYFLT>*` owned by `SignalFlowGraphState` (parallel to `sourceOutlets`), for all inlet types. Cleared/deleted in `clear()`.

Opcodes/signalflowgraph.cpp
gogins: Done in this PR: non-sliding merge now runs for **active** sources, copies source→sink (fixing the old reversed copy), and uses the source frame counter so frames keep updating. Added `test_connect_gain_fsig.csd` (`pvsbin` amplitude ratio ≈ 0.5).

Opcodes/signalflowgraph.cpp
gogins: Done: duplicate edges to the same inlet **sum** their gains (documented in the header). Covered by `SinkDup` in `test_connect_gain.csd` (0.5 + 0.25 → 0.75).

tests/soak/connect.csd
gogins: Reverted `tests/soak/connect.csd` to the prior unity-omit form.

gogins:
Addressed the requested changes in b883ff621:

- Pooled gain vectors in `SignalFlowGraphState` (no by-value `std::vector` in opcode structs)
- Fixed active non-sliding `inletf` gain path (+ old reversed/stale frame merge); added fsig regression test
- Duplicate edges sum gains (documented + tested)
- Reverted soak `connect.csd` example churn

gogins:
The 32 bit Windows vcpkg build failed the first time, succeeded when I re-ran the failed job. That's not relevant to this code anyway.

@hlolli after making the change you requested, the workflow runs failed on some platforms due to lack of resources while setting up jobs, etc.; the builds that did run (Mac brew, Linux, etc.) passed.

Opcodes/signalflowgraph.cpp
hlolli: Thanks, the active single-source path is fixed. One non-sliding fsig fan-in case still fails.

`lastframe` belongs to the whole inlet. After the first source sets it at line 844, a second source with the same `framecount` fails the check at line 835 and never reaches the merge. The new fsig test has one source per sink, so it misses this.

<details>
<summary>Reproducer on <code>b883ff621</code></summary>

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d -m0
</CsOptions>
<CsInstruments>
sr = 48000
ksmps = 64
nchnls = 1
0dbfs = 1

connect "SilentFirst", "fout", "Sink", "fin", 1
connect "ToneSecond", "fout", "Sink", "fin", 1

instr SilentFirst
  azero = 0
  fsig pvsanal azero, 256, 64, 256, 1
  outletf "fout", fsig
endin

instr ToneSecond
  atone oscili 0.5, 1875
  fsig pvsanal atone, 256, 64, 256, 1
  outletf "fout", fsig
endin

instr Sink
  fsig inletf "fin"
  kamp, kfreq pvsbin fsig, 10
  if timeinstk() == 80 then
    printf "fanin bin10 amp=%f freq=%f\n", 1, kamp, kfreq
    if kamp < 0.01 then
      exitnowk 1
    endif
  endif
endin
</CsInstruments>
<CsScore>
i "SilentFirst" 0 0.2
i "ToneSecond" 0 0.2
i "Sink" 0 0.2
e
</CsScore>
</CsoundSynthesizer>
```

```sh
csound test_connect_gain_fsig_fanin.csd
```

```text
fanin bin10 amp=0.000000 freq=0.000000
```

Csound exits with status 1.
</details>

Bin 10 should be nonzero.

Please add this two-source case. A fix could defer the shared frame update until every source has been checked. Per-source frame state would also work.

gogins: Fixed in 2783f873a: non-sliding merge now defers updating shared `lastframe` / `framecount` until every active source has been checked for that k-cycle. Your SilentFirst+ToneSecond reproducer now reports `fanin bin10 amp=0.500000`, and the same case is in `test_connect_gain_fsig.csd`.

hlolli: Pulled `2783f873a` and reran this path. Thanks, the equal-frame case now works.

I still hit the staggered-start case because `lastframe` belongs to the inlet, while each outlet instance has its own `framecount`. With `SilentEarly` at 0 and `ToneLate` plus the sink at 0.05, the early source keeps the shared count ahead, bin 10 stays at zero, and Csound exits 1.

The sink also keeps old peaks across new frames. In a single-source decay test, the direct fsig reaches 0 but the inlet remains at 0.5.

Per-outlet-instance last-seen state may fit better here. When any source advances, clear the sink once, merge the current frames from every active source with a matching layout, then publish the new frame count. The sliding path seems to need the same reset before its merge.

hlolli: https://github.com/gogins/mkg-csound/pull/1

hlolli (approved):

PR #2740, opened 2026-08-06
global audio/audio-array vars make setting local ksmps fail

### LINKED ISSUE #2737: Global a-rate variables use the wrong block with local ksmps
A direct write to a global a-rate variable only updates the first local block when `setksmps` is smaller than the global `ksmps`.

## Reproducer

Save this as `global-a-local-ksmps.csd` and run `csound global-a-local-ksmps.csd`.

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d
</CsOptions>
<CsInstruments>
sr = 32
ksmps = 8
nchnls = 1
0dbfs = 1

gaValue init 0

instr 1
  setksmps 4
  aOne = 1
  gaValue = aOne
endin

instr 2
  kMean downsamp gaValue, 8
  printks "global mean = %.1f\n", 0, kMean
endin
</CsInstruments>
<CsScore>
i 1 0 0.25
i 2 0 0.25
</CsScore>
</CsoundSynthesizer>
```

## Expected

```text
global mean = 1.0
```

All eight samples should be `1`.

## Actual

```text
global mean = 0.5
```

Only four of the eight samples are written. Direct reads have the same offset problem: each local pass starts at the first sample of the global block.

## Needed fix

Use the right global-buffer offset for each local pass. Until then, reject direct global a-rate access when local and global `ksmps` differ.

vlazzarini:
This PR introduces a check for the presence of global audio variables in a context where local ksmps or local sr are attempted. If these are detected, an init error is issued.
Closes #2737.

### DIFF
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -2610,6 +2610,48 @@ static CS_VARIABLE *setup_arg_for_var_name(CSOUND* csound, ARG* arg,
   return arg->argPtr;
 }
 
+typedef struct audio_type_path {
+  const CS_TYPE *type;
+  const struct audio_type_path *parent;
+} AUDIO_TYPE_PATH;
+
+static int32_t type_contains_audio(const CS_TYPE *type,
+                                   const AUDIO_TYPE_PATH *path);
+
+static int32_t var_contains_audio(const CS_VARIABLE *var,
+                                  const AUDIO_TYPE_PATH *path) {
+  if (var == NULL || var->varType == NULL)
+    return 0;
+  return type_contains_audio(var->varType == &CS_VAR_TYPE_ARRAY
+                               ? var->subType : var->varType,
+                             path);
+}
+
+static int32_t type_contains_audio(const CS_TYPE *type,
+                                   const AUDIO_TYPE_PATH *path) {
+  const AUDIO_TYPE_PATH *current = path;
+  CONS_CELL *members;
+
+  if (type == &CS_VAR_TYPE_A)
+    return 1;
+  if (type == NULL || !type->userDefinedType)
+    return 0;
+  while (current != NULL) {
+    if (current->type == type)
+      return 0;
+    current = current->parent;
+  }
+
+  AUDIO_TYPE_PATH next = { type, path };
+  members = type->members;
+  while (members != NULL) {
+    if (var_contains_audio((CS_VARIABLE *) members->value, &next))
+      return 1;
+    members = members->next;
+  }
+  return 0;
+}
+
 static int32_t is_audio_var(CSOUND *csound, CS_VARIABLE *var,
                             const char *structPath) {
   if (structPath != NULL) {
@@ -2626,10 +2668,7 @@ static int32_t is_audio_var(CSOUND *csound, CS_VARIABLE *var,
     }
     csound->Free(csound, path);
   }
-  return var != NULL && var->varType != NULL &&
-    (var->varType == &CS_VAR_TYPE_A ||
-     (var->varType == &CS_VAR_TYPE_ARRAY &&
-      var->subType == &CS_VAR_TYPE_A));
+  return var_contains_audio(var, NULL);
 }
 
 
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -2609,6 +2609,16 @@ static CS_VARIABLE *setup_arg_for_var_name(CSOUND* csound, ARG* arg,
   return arg->argPtr;
 }
 
+static int32_t is_audio_var(CSOUND *csound,
+                    CS_VARIABLE *var) {
+  if(var->varType == &CS_VAR_TYPE_A ||
+    (var->varType == &CS_VAR_TYPE_ARRAY &&
+     var->subType ==  &CS_VAR_TYPE_A))
+     return 1;
+   else return 0;
+}
+
+
 /* get storage ndx of const, pnum, lcl or gbl */
 /* argument const/gbl indexes are positiv+1, */
 /* pnum/lcl negativ-1 called only after      */
@@ -2697,11 +2707,14 @@ static ARG *create_arg(CSOUND *csound, INSTRTXT *ip, char *s,
   else if(engineState->varPool != NULL && (uintptr_t)engineState->varPool >= 0x1000 &&
           setup_arg_for_var_name(csound, arg, engineState->varPool, s) != NULL) {
        arg->type = ARG_GLOBAL;
+       if(is_audio_var(csound, arg->argPtr))
+          ip->glbvarcnt++;
     }
     else if(csound->engineState.varPool != NULL && (uintptr_t)csound->engineState.varPool >= 0x1000 &&
             setup_arg_for_var_name(csound, arg, csound->engineState.varPool, s) != NULL) {
     arg->type = ARG_GLOBAL;
-
+    if(is_audio_var(csound, arg->argPtr))
+          ip->glbvarcnt++;  
   }
   /* otherwise we have a local arg */
   else {
--- a/Engine/csound_orc_compile.c
+++ b/Engine/csound_orc_compile.c
@@ -2610,48 +2610,6 @@ static CS_VARIABLE *setup_arg_for_var_name(CSOUND* csound, ARG* arg,
   return arg->argPtr;
 }
 
-typedef struct audio_type_path {
-  const CS_TYPE *type;
-  const struct audio_type_path *parent;
-} AUDIO_TYPE_PATH;
-
-static int32_t type_contains_audio(const CS_TYPE *type,
-                                   const AUDIO_TYPE_PATH *path);
-
-static int32_t var_contains_audio(const CS_VARIABLE *var,
-                                  const AUDIO_TYPE_PATH *path) {
-  if (var == NULL || var->varType == NULL)
-    return 0;
-  return type_contains_audio(var->varType == &CS_VAR_TYPE_ARRAY
-                               ? var->subType : var->varType,
-                             path);
-}
-
-static int32_t type_contains_audio(const CS_TYPE *type,
-                                   const AUDIO_TYPE_PATH *path) {
-  const AUDIO_TYPE_PATH *current = path;
-  CONS_CELL *members;
-
-  if (type == &CS_VAR_TYPE_A)
-    return 1;
-  if (type == NULL || !type->userDefinedType)
-    return 0;
-  while (current != NULL) {
-    if (current->type == type)
-      return 0;
-    current = current->parent;
-  }
-
-  AUDIO_TYPE_PATH next = { type, path };
-  members = type->members;
-  while (members != NULL) {
-    if (var_contains_audio((CS_VARIABLE *) members->value, &next))
-      return 1;
-    members = members->next;
-  }
-  return 0;
-}
-
 static int32_t is_audio_var(CSOUND *csound, CS_VARIABLE *var,
                             const char *structPath) {
   if (structPath != NULL) {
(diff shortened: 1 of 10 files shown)


Engine/udo.c
```diff
@@ -2473,6 +2473,9 @@ int32_t useropcd_pass_by_ref(CSOUND *csound, UOPCODE *p)
 int32_t setksmpsset(CSOUND *csound, SETKSMPS *p)
 {
 
+  if(p->h.insdshead->instr->glbvarcnt > 0)
```
hlolli: Could we also check the callee when it inherits the caller's rate? I tried an outer UDO with `setksmps 4` and an inner UDO that writes a global a-rate variable, with global `ksmps = 8`. It ran successfully and produced `1 1 1 1 0 0 0 0`. `subinstr` inherits the rate too, through `subinstrset_`, so that entry point needs the check as well.

Engine/csound_orc_compile.c
```diff
@@ -2609,6 +2609,16 @@ static CS_VARIABLE *setup_arg_for_var_name(CSOUND* csound, ARG* arg,
   return arg->argPtr;
 }
 
+static int32_t is_audio_var(CSOUND *csound,
+                    CS_VARIABLE *var) {
+  if(var->varType == &CS_VAR_TYPE_A ||
```
hlolli: `is_audio_var()` only sees a scalar `a` or an `a[]` base type. A global struct slips through when the selected member is audio-rate. I tested `bus@global:AudioBus` with `value:a`, global `ksmps = 8`, and local `setksmps 4`. The score ran and printed a mean of `0.5`. Can this follow `arg->structPath` to the selected member type?

Engine/udo.c
```diff
@@ -2473,6 +2473,9 @@ int32_t useropcd_pass_by_ref(CSOUND *csound, UOPCODE *p)
...
+    return csoundInitError(csound, "local ksmps not permitted with global audio vars\n");
```
hlolli, line 2492: These checks run before the existing no-op cases. On this head, `setksmps 8` with global `ksmps = 8`, `setksmps 0`, `oversample 1`, and `undersample 1` all return an init error. The base branch accepts all four. Could the global-a check run after the requested rate is known and fire only when the effective ksmps or sr differs from the global value?

tests/commandline/test.py
```diff
@@ -494,6 +494,7 @@ def runTest():
         ["test46.csd", "if-then with expression in boolean comparison"],
         ["test47.csd", "until loop and k[]"],
         ["test48.csd", "expected failure with variable used before defined", 1],
+        ["test_local_ksmps_global_fail.csd", "test failing use of global var with local ksmps",  1],
```
hlolli, line 497: This one file mixes the allowed global k case with the scalar and array failures, while the harness checks one exit code. Any one init error makes the test pass. Splitting these into one success test and two expected failures would prove all three behaviors, and would leave room for separate no-op checks.

Engine/csound_orc_compile.c
```diff
@@ -2697,11 +2707,14 @@ static ARG *create_arg(CSOUND *csound, INSTRTXT *ip, char *s,
...
        arg->type = ARG_GLOBAL;
+       if(is_audio_var(csound, arg->argPtr))
+          ip->glbvarcnt++;
     }
     else if(csound->engineState.varPool != NULL && (uintptr_t)csound->engineState.varPool >= 0x1000 &&
             setup_arg_for_var_name(csound, arg, csound->engineState.varPool, s) != NULL) {
     arg->type = ARG_GLOBAL;
-
+    if(is_audio_var(csound, arg->argPtr))
+          ip->glbvarcnt++;
```
hlolli, line 2728: `git diff --check` reports trailing spaces here and on the added blank lines in `Engine/udo.c` near `setksmpsset()` and `undersampleset()`. Could you trim those?

tests/commandline/test.py
hlolli: human follow up, I don't agree with this, leave it here 4wiw

Engine/csound_orc_compile.c
hlolli: HUMAN: I don't care :)

Engine/csound_orc_compile.c
vlazzarini: @hlolli not sure how to do that, can you add this extra check?

hlolli: ok, I'll try

hlolli: https://github.com/csound/csound/commit/82e8cce41a4a992614dc023789ccfa4307bb2fea I realise we could make a helper for struct member lookups, later on as cleanup. But for now it should do it.

Engine/udo.c
vlazzarini: fixed in [e50a125](https://github.com/csound/csound/pull/2740/commits/e50a125714e81cd2a5faccf3a504dda235f45383)

Engine/udo.c
vlazzarini: Fixed in [e50a125](https://github.com/csound/csound/pull/2740/commits/e50a125714e81cd2a5faccf3a504dda235f45383)

Engine/csound_orc_compile.c
```diff
@@ -2609,6 +2610,29 @@ static CS_VARIABLE *setup_arg_for_var_name(CSOUND* csound, ARG* arg,
   return arg->argPtr;
 }
 
+static int32_t is_audio_var(CSOUND *csound, CS_VARIABLE *var,
+                            const char *structPath) {
+  if (structPath != NULL) {
```
hlolli, line 2615: I think this still misses whole-struct access. With `struct Bus signal:a`, `local:Bus = bus` inside a `setksmps 4` UDO runs against global `ksmps = 8`. I get `source[4] = 0.5` and `result[4] = 0.0`, so it reads the first half twice. `structPath` is null for the whole `bus` value. Can `is_audio_var()` also walk the UDT members in that case?

Engine/udo.c
```diff
@@ -1550,6 +1550,15 @@ int32_t useropcdset(CSOUND *csound, UOPCODE *p)
   lcurip->onedksmps = CS_ONEDKSMPS;
   lcurip->kicvt = CS_KICVT;
 
+  if(lcurip->instr->glbvarcnt > 0 &&
```
hlolli: This return happens after `inm->recurse_depth++`, and the normal decrement is down at line 1668. I tried `--recursion-depth=1`: one rejected local-rate call leaves the count at 1, then a later call at the global rate fails with `max recursion depth 1 reached`. Could this check run before the increment and allocation?

Engine/udo.c
```diff
@@ -2473,6 +2482,10 @@ int32_t useropcd_pass_by_ref(CSOUND *csound, UOPCODE *p)
 int32_t setksmpsset(CSOUND *csound, SETKSMPS *p)
 {
 
+  if(p->h.insdshead->instr->glbvarcnt > 0 &&
```
hlolli, line 2489: `setksmps 0` still trips this guard, even though the existing code treats zero as a no-op a few lines below. I ran it with a global `gaValue` and got `local ksmps not permitted`. Can we preserve the zero return before checking the global audio count?

Engine/udo.c
```diff
@@ -2529,6 +2542,10 @@ int32_t setksmpsset(CSOUND *csound, SETKSMPS *p)
    with audio/control array arguments.
 */
 int32_t oversampleset(CSOUND *csound, OVSMPLE *p) {
+  if(p->h.insdshead->instr->glbvarcnt > 0 &&
```
hlolli, line 2550: This comparison looks flipped. With a global `gaValue`, `oversample 1` fails and `oversample 2` runs, even though 1 is the no-op and 2 changes the rate. `undersampleset()` has the same `<= 1` check. Can we parse the factor first, keep 1 as the no-op, then reject a rate change?

Engine/udo.c
```diff
@@ -2726,6 +2747,15 @@ int32_t subinstrset_(CSOUND *csound, SUBINST *p, int32_t instno)
   p->ip->onedksmps = CS_ONEDKSMPS;
   p->ip->kicvt = CS_KICVT;
 
+
+  if(p->ip->instr->glbvarcnt > 0 &&
```
hlolli: `subinstrset_()` also handles `subinstrinit`, where `init_op` is true and only the child init chain runs. A local-rate caller now gets this error if that child contains a global a-rate statement. The same CSD passed before this change. Could we gate this check on `!init_op`?

Engine/csound_orc_compile.c
hlolli: this one's on me, fixing

hlolli: Fixed in [`69619c198`](https://github.com/csound/csound/commit/69619c19862b4cbebd455566f452ec7605c66627). Whole-struct access now checks each UDT member, including nested structs and arrays, and stops cleanly on recursive types. I also added a test for this case.

Engine/udo.c
vlazzarini: fixed in a718884a3

Engine/udo.c
vlazzarini: a718884a3

Engine/udo.c
vlazzarini: a718884a3

Engine/udo.c
vlazzarini: fixed in 21ca2f3b2

hlolli (commented):
Inline note below.

Engine/udo.c
```diff
@@ -2473,6 +2483,11 @@ int32_t useropcd_pass_by_ref(CSOUND *csound, UOPCODE *p)
 int32_t setksmpsset(CSOUND *csound, SETKSMPS *p)
 {
 
+  if(p->h.insdshead->instr->glbvarcnt > 0 &&
+     *p->i_ksmps > 0 /* no-op */ &&
+     csound->ksmps != *p->i_ksmps /* no-op */)
+    return csoundInitError(csound, "local ksmps not permitted with global audio vars\n");
```
hlolli, line 2492: Part of the rejected UDO still runs. This error reaches the first UDO init scan as `err`, but `useropcdset()` clears it before the normal init pass. I tried `setksmps 4`, then `prints "INIT AFTER REJECTION"`, with a global a-rate assignment at global `ksmps = 8`. Csound reports the error and prints the message anyway. Could we check `err` after the first scan and jump to `finish_init` before pass-by-ref setup, so later init side effects never run?

vlazzarini: I don't think that is a problem. It needs to fail before the perf pass, it doesn't matter that it runs some init pass before it fails. If it enters the perf pass, then it is not right. Given that the local ksmps is always less than the global ksmps, there is no issue of overflow to contend with in an init pass variable initialisation so the question is moot in my opinion.

hlolli: Codex Sol Ultra speaking here: Fair point on overflow. My concern is the observable init work after the UDO has already been rejected. `prints` makes it easy to see, and the same pass can run later i-time opcodes that schedule events or change shared state. Clearing `err` also means `useropcdset()` returns a generic `NOTOK` after `instance_init_finish()` instead of the original failure. Checking `err` after the first scan would leave the perf path unchanged and stop at the rejection point.

hlolli: human: let me ask it to make a suggestion, not sure how heavly lift it's asing for. I understand what you mean, it's going to fail anyway so it's good it fails before perf pass.

hlolli: https://github.com/csound/csound/commit/d1966b244 ok I just pushed it, the suggestion didn't fit in the diff here, looks ok to me?

vlazzarini:
ok as long as it doesn't cause any other issue inadvertently.

hlolli (approved):

PR #2746, opened 2026-08-07
Fix PARCS completion race between control cycles

### LINKED ISSUE #2742: PARCS can enter the same instrument instance twice at once
I hit this while testing a stateful opcode with Csound's multicore scheduler. No plugin is involved.

`Slow` has one k-rate flag named `kInside`. Each performance pass checks that flag, sets it to 1, does some work, then clears it before returning, which must happen before Csound enters the only `Slow` instance again.

With four threads, Csound enters that same instance again before the first pass has cleared the flag. The CSD records the overlap and checks the result after both worker instruments have stopped.

Tested with:

```text
Csound 7.0, double samples
macOS 26.5.2, arm64
Apple clang 21.0.0
```

Commands:

```sh
csound --num-threads=1 parcs_instance_overlap.csd
csound --num-threads=4 parcs_instance_overlap.csd
```

The first command prints:

```text
PASS: no overlap detected
```

The four-thread command reliably prints something like this and returns status 1:

```text
PARCS_OVERLAP: Slow was entered twice at control cycle 6922
FAIL: PARCS entered one instrument instance twice at once
```

Two threads also reproduce it on this machine, though four has been consistent. The CSD uses `ksmps = 1` to give the scheduler many rounds during a short test. The original program that exposed it used `ksmps = 32`.

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d -m128
</CsOptions>
<CsInstruments>

sr = 48000
ksmps = 1
nchnls = 1
0dbfs = 1

gkOverlap init 0

instr Fast
  kTick = 0
endin

instr Slow
  kInside init 0
  kCycle timeinstk

  if kInside != 0 then
    printf "PARCS_OVERLAP: Slow was entered twice at control cycle %d\n", \
        1, kCycle
    gkOverlap = 1
  endif

  kInside = 1
  kIndex = 0
  kSum = 0
  while kIndex < 5000 do
    kSum += sqrt(kIndex + 1)
    kIndex += 1
  od
  kInside = 0
endin

instr Check
  if i(gkOverlap) != 0 then
    prints "FAIL: PARCS entered one instrument instance twice at once\n"
    exitnow -1
  else
    prints "PASS: no overlap detected\n"
  endif
endin

</CsInstruments>
<CsScore>
i "Fast" 0 0.25
i "Slow" 0 0.25
i "Check" 0.30 0.01
e
</CsScore>
</CsoundSynthesizer>
```

The lock-free completion flags look like the likely source. A worker leaves `taskflag[index]` set to 1 when it finishes, then clears it only after it sees the next `parflag` change. The main thread changes `parflag`, runs its own task, and may read the old 1 before the worker has cleared it. The task graph can reset early.

[PR #2338](https://github.com/csound/csound/pull/2338) added this path. The two places that seem to matter are the [worker completion flag](https://github.com/csound/csound/blob/a16365817092ba1ff10c7f4861c5f517dfadfb04/Top/csound_perf.c#L203-L220), which survives into the next round, and the [main-thread completion check](https://github.com/csound/csound/blob/a16365817092ba1ff10c7f4861c5f517dfadfb04/Top/csound_perf.c#L275-L291), which can read that old value.

I would expect Csound to finish every active instrument task for one control cycle before it resets the task graph for the next one.

hlolli:
PARCS could reuse a worker's completion flag from the last control cycle. Workers now publish the cycle they finished, so the DAG waits for fresh work.

The lock-barrier path now uses the right guard and stops workers once.

Fixes #2742.

### DIFF
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -1878,9 +1878,11 @@ static void reset(CSOUND *csound) {
   int32_t n = 0;
 
   // stop multicore threads if still running
-  if (csound->oparms->numThreads > 1) {
+  if (csound->oparms->numThreads > 1 &&
+      csound->multiThreadedThreadInfo != NULL &&
+      ATOMIC_GET(csound->multiThreadedComplete) == 0) {
     ATOMIC_SET(csound->multiThreadedComplete, 1);
-#ifdef PARCS_USE_LOCK_BARRIER
+#ifdef PARCS_USE_THREAD_BARRIER
     csound->WaitBarrier(csound->barrier1);
 #else
     ATOMIC_SET(csound->parflag,!csound->parflag);
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -1882,7 +1882,7 @@ static void reset(CSOUND *csound) {
       csound->multiThreadedThreadInfo != NULL &&
       ATOMIC_GET(csound->multiThreadedComplete) == 0) {
     ATOMIC_SET(csound->multiThreadedComplete, 1);
-#ifdef PARCS_USE_THREAD_BARRIER
+#ifdef PARCS_USE_LOCK_BARRIER
     csound->WaitBarrier(csound->barrier1);
 #else
     ATOMIC_SET(csound->parflag, !ATOMIC_GET(csound->parflag));
--- a/Top/csound.c
+++ b/Top/csound.c
@@ -1885,7 +1885,7 @@ static void reset(CSOUND *csound) {
 #ifdef PARCS_USE_THREAD_BARRIER
     csound->WaitBarrier(csound->barrier1);
 #else
-    ATOMIC_SET(csound->parflag,!csound->parflag);
+    ATOMIC_SET(csound->parflag, !ATOMIC_GET(csound->parflag));
 #endif
   }
   csound_cleanup(csound);
--- a/Top/csound_perf.c
+++ b/Top/csound_perf.c
@@ -200,14 +200,14 @@ unsigned long kperf_thread(void *cs) {
     return ULONG_MAX;
   }
   index++;
-  int32_t parflag, last_parflag = 0;
+  int32_t parflag, taskflag = 0;
   while (1) {
-#ifdef PARCS_USE_THREAD_BARRIER
+#ifdef PARCS_USE_LOCK_BARRIER
     csound->WaitBarrier(csound->barrier1);
 #else
     do parflag = ATOMIC_GET(csound->parflag);
-    while(parflag == last_parflag);
-    last_parflag = parflag;
+    while(parflag == taskflag);
+    taskflag = parflag;
 #endif
     if (ATOMIC_GET(csound->multiThreadedComplete) == 1) {
       // exit thread on performance end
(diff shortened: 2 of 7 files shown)


Top/csound_perf.c
```diff
@@ -416,7 +418,7 @@ int32_t kperf(CSOUND *csound) {
         csoundUnlockMutex(csound->API_lock);
       if (csound->oparms->numThreads > 1) {
         ATOMIC_SET(csound->multiThreadedComplete, 1);
-#ifdef PARCS_USE_LOCK_BARRIER
+#ifdef PARCS_USE_THREAD_BARRIER
```
vlazzarini: why this change? it seems pretty useless

Top/csound_perf.c
```diff
@@ -200,25 +200,25 @@ unsigned long kperf_thread(void *cs) {
     return ULONG_MAX;
   }
   index++;
-  int32_t parflag, taskflag = 0;
+  int32_t parflag, last_parflag = 0;
```
vlazzarini: again, these changes are just variable renaming. Let's keep the old names.

Top/csound.c
```diff
@@ -1878,9 +1878,11 @@ static void reset(CSOUND *csound) {
...
-  if (csound->oparms->numThreads > 1) {
+  if (csound->oparms->numThreads > 1 &&
+      csound->multiThreadedThreadInfo != NULL &&
+      ATOMIC_GET(csound->multiThreadedComplete) == 0) {
     ATOMIC_SET(csound->multiThreadedComplete, 1);
-#ifdef PARCS_USE_LOCK_BARRIER
+#ifdef PARCS_USE_THREAD_BARRIER
     csound->WaitBarrier(csound->barrier1);
 #else
     ATOMIC_SET(csound->parflag,!csound->parflag);
```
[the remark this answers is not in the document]
hlolli: Fixed in 4661c5ab5. Reset now reads `parflag` with `ATOMIC_GET` before the atomic store.

Top/csound_perf.c
```diff
@@ -416,7 +418,7 @@ int32_t kperf(CSOUND *csound) {
         csoundUnlockMutex(csound->API_lock);
       if (csound->oparms->numThreads > 1) {
         ATOMIC_SET(csound->multiThreadedComplete, 1);
-#ifdef PARCS_USE_LOCK_BARRIER
+#ifdef PARCS_USE_THREAD_BARRIER
         csound->WaitBarrier(csound->barrier1);
 #else
         ATOMIC_SET(csound->parflag,!csound->parflag);
```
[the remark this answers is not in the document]
hlolli: Fixed in 4661c5ab5. Both shutdown toggles in `csound_perf.c`, including line 471, now read `parflag` with `ATOMIC_GET`.

Top/csound_perf.c
hlolli: Good point. This exposed a name mismatch: CMake and main.c used `PARCS_USE_THREAD_BARRIER` while the runtime used `PARCS_USE_LOCK_BARRIER`. I kept the old runtime name and made CMake and main.c match it in 91f60db0b.

Top/csound_perf.c
hlolli: Agreed. I restored `taskflag` in 91f60db0b. The cycle-tag fix does not need the local rename.

vlazzarini:
all looking good, just waiting for CI now. Thanks for the quick action.

vlazzarini (approved):

PR #2747, opened 2026-08-07
Fix/parcserror

### LINKED ISSUE #2744: PARCS keeps running an instrument's opcode chain after a perf error
When an opcode reports a performance error, the serial engine stops that instrument's opcode chain for the cycle. With more than one thread, the engine reports the error but runs the rest of the chain anyway, so later opcodes work on data the failed opcode never produced.

The serial loop in `kperf` (Top/csound_perf.c:325) keeps the perf return value and stops on `error == 0 && ... && ip->actflg`. Both loops in `node_perf` (Top/csound_perf.c:120 and 154) discard the return value and check neither the error nor the active flag.

## Reproducer

Save this as `parcs_perf_error.csd`. The array read is out of range, so `##array_get` raises a perf error on the first cycle. The print after it should never run.

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d -m0
</CsOptions>
<CsInstruments>
sr = 48000
ksmps = 32
nchnls = 1
0dbfs = 1

instr 1
  kvalues[] init 1
  kindex init 2
  kvalue = kvalues[kindex]
  kcycle timeinstk
  printks "after-error cycle=%d value=%f\n", 0, kcycle, kvalue
endin
</CsInstruments>
<CsScore>
i 1 0 0.01
e
</CsScore>
</CsoundSynthesizer>
```

Run it twice:

```sh
csound --num-threads=1 parcs_perf_error.csd
csound --num-threads=2 parcs_perf_error.csd
```

## Expected

Both runs print the error and nothing more from the instrument:

```text
PERF ERROR in instr 1 (opcode ##array_get) line 14: Array index 2 out of range (0,0) for dimension 1
```

## Actual

One thread behaves as expected. Two threads print the error and then keep going:

```text
PERF ERROR in instr 1 (opcode ##array_get) line 14: Array index 2 out of range (0,0) for dimension 1
after-error cycle=1 value=0.000000
```

Tested with:

```text
Csound 7.0, double samples
macOS 26.5, arm64
Apple clang 21.0.0
```

### LINKED ISSUE #2745: Debugger with multicore: worker audio buffers are never cleared and output grows every cycle
When the debugger is attached and Csound runs with more than one thread, audio from instruments scheduled on worker threads piles up cycle after cycle. Output gets louder until it clips.

`kperf` clears the whole set of per-thread output buffers each cycle (Top/csound_perf.c:257, `nspout * numThreads` samples). `kperf_debug` clears only the first thread's block (Top/csound_debug.c:915, `nspout` samples), yet still mixes every worker block into the output (Top/csound_debug.c:976). Output opcodes add into these buffers, so whatever a worker wrote last cycle is still there and gets mixed again.

The debugger path only runs after a host calls `csoundDebuggerInit()`, so the reproducer is a small API host rather than a CSD.

## Reproducer

Save this as `debug_spout_clear.c`. It performs the same embedded CSD twice with `--num-threads=2`, first without the debugger and then with it, and prints the peak of `spout` per cycle. Four identical instruments make sure at least one lands on the worker thread.

```c
#include <csound.h>
#include <csdebug.h>
#include <stdio.h>
#include <math.h>

static const char *csd =
  "<CsoundSynthesizer>\n<CsOptions>\n-n -d -m0\n</CsOptions>\n"
  "<CsInstruments>\n"
  "sr = 48000\nksmps = 32\nnchnls = 1\n0dbfs = 1\n"
  "instr 1\n aSig = 0.125\n out aSig\nendin\n"
  "instr 2\n aSig = 0.125\n out aSig\nendin\n"
  "instr 3\n aSig = 0.125\n out aSig\nendin\n"
  "instr 4\n aSig = 0.125\n out aSig\nendin\n"
  "</CsInstruments>\n<CsScore>\n"
  "i 1 0 1\ni 2 0 1\ni 3 0 1\ni 4 0 1\ne\n</CsScore>\n</CsoundSynthesizer>\n";

static double run(int with_debugger)
{
  CSOUND *cs = csoundCreate(NULL, NULL);
  double peak = 0.0;
  int i, k;
  csoundSetOption(cs, "--num-threads=2");
  if (csoundCompileCSD(cs, csd, 1, 0) != 0 || csoundStart(cs) != 0) {
    fprintf(stderr, "compile/start failed\n");
    csoundDestroy(cs);
    return -1.0;
  }
  if (with_debugger)
    csoundDebuggerInit(cs);
  for (k = 0; k < 8; k++) {
    if (csoundPerformKsmps(cs))
      break;
    const MYFLT *sp = csoundGetSpout(cs);
    uint32_t n = csoundGetKsmps(cs) * csoundGetChannels(cs, 0);
    double m = 0.0;
    for (i = 0; i < (int)n; i++)
      if (fabs(sp[i]) > m) m = fabs(sp[i]);
    printf("  cycle %d: max |spout| = %.4f\n", k + 1, m);
    if (m > peak) peak = m;
  }
  if (with_debugger)
    csoundDebuggerClean(cs);
  csoundDestroy(cs);
  return peak;
}

int main(void)
{
  printf("debugger OFF, 2 threads:\n");
  double a = run(0);
  printf("debugger ON, 2 threads:\n");
  double b = run(1);
  printf("peaks: off=%.4f on=%.4f\n", a, b);
  return 0;
}
```

Build and run against a multicore Csound build:

```sh
cc -o debug_spout_clear debug_spout_clear.c \
   -I/path/to/csound/include -F /path/to/build -framework CsoundLib64 \
   -Wl,-rpath,/path/to/build
./debug_spout_clear
```

## Expected

Four instruments at 0.125 each give a steady peak of 0.5 in both runs.

## Actual

Without the debugger the peak stays at 0.5. With the debugger it grows by 0.125 every cycle, one instrument's worth of stale worker audio:

```text
debugger OFF, 2 threads:
  cycle 1: max |spout| = 0.5000
  ...
  cycle 8: max |spout| = 0.5000
debugger ON, 2 threads:
  cycle 1: max |spout| = 0.5000
  cycle 2: max |spout| = 0.6250
  cycle 3: max |spout| = 0.7500
  cycle 4: max |spout| = 0.8750
  cycle 5: max |spout| = 1.0000
  cycle 6: max |spout| = 1.1250
  cycle 7: max |spout| = 1.2500
  cycle 8: max |spout| = 1.3750
```

Tested with:

```text
Csound 7.0, double samples
macOS 26.5, arm64
Apple clang 21.0.0
```

vlazzarini:
This PR does two things:

(1) Adds perf error detection to parcs (Closes #2744)
(2) Disables parcs in debugger, stopping Csound if parcs run is attempted (Closes #2745)


hlolli:
I agree with the approach, just disable debugger in multi threaded mode for now. Still grabbing the llm in case something's there.

Top/csound_perf.c
```diff
@@ -113,11 +113,11 @@ inline static int32_t node_perf(CSOUND *csound, int32_t index,
         insds->spout = csound->spout_tmp + index * csound->nspout;
         insds->kcounter = csound->kcounter;
         csound->mode = 2;
-        while ((opstart = opstart->nxtp) != NULL) {
+        while (error == 0 && (opstart = opstart->nxtp) != NULL) {
           /* In case of jumping need this repeat of opstart */
           opstart->insdshead->pds = opstart;
           csound->op = opstart->optext->t.opcod;
-          (*opstart->perf)(csound, opstart); /* run each opcode */
+          error = (*opstart->perf)(csound, opstart); /* run each opcode */
```
hlolli, line 121: Could we carry the same return check into the local `ksmps` loop below? That path still drops the opcode result. With global `ksmps = 32` and local `ksmps = 8`, I get four perf errors and `after-error` runs four times. So `setksmps` instruments still show the behavior from #2744.

Top/csound_perf.c
```diff
@@ -214,13 +213,15 @@ unsigned long kperf_thread(void *cs) {
...
     }
-    node_perf(csound, index, numThreads);
+    err = node_perf(csound, index, numThreads);
 #ifdef PARCS_USE_LOCK_BARRIER
     csound->WaitBarrier(csound->barrier2);
 #else
     ATOMIC_SET(csound->taskflag[index], parflag);
 #endif
   }
+  ATOMIC_SET(csound->multiThreadedComplete, 1);
```
hlolli: This worker leaves the pool after a perf error, but the master still expects its task flag on later control cycles. I reproduced a hang under `-j4` by keeping another instrument active after the error. Could the worker stay alive through the final sync and let the master handle shutdown?

Top/csound_perf.c
```diff
@@ -278,7 +279,8 @@ int32_t kperf(CSOUND *csound) {
       int32_t parflag = !ATOMIC_GET(csound->parflag);
       ATOMIC_SET(csound->parflag, parflag);
 #endif
-      node_perf(csound, 0, n);
+      error = node_perf(csound, 0, n);
+      if(error == 0) {
```
hlolli: Could we keep the completion wait outside this error check? Every worker still reaches `barrier2` after `node_perf()`. With `USE_LOCK_BARRIER=ON`, the master skips the matching wait here and the new error test hangs. I confirmed it with a five-second timeout.

Top/csound_debug.c
```diff
@@ -973,11 +970,12 @@ int32_t kperf_debug(CSOUND *csound) {
         } while(i < n);
       }
 #endif
-      /* do the mixing of thread buffers */
       for (k = 1; k < csound->oparms->numThreads; k++)
           mix_out(csound->spout_tmp, csound->spout_tmp +
                   k * csound->nspout, csound->nspout);
-#endif /* PARCS */
+#else /* currently disabled */
+      csoundDie(csound, "csound debugger cannot run in multiple threads\n");
```
hlolli: Could we reject this combination before performance starts? `csoundDie()` jumps out of `csoundPerformKsmps()` while it holds `API_lock`, so the lock stays held. I reproduced it by cleaning the debugger after the error and calling `csoundPerformKsmps()` from another thread. That call blocks. An earlier check would report the same limit without leaving the instance locked.

tests/commandline/test_parcs_perf_error.csd
```diff
@@ -0,0 +1,23 @@
...
+ksmps = 32
+nchnls = 1
+0dbfs = 1
+
+instr 1
+  kvalues[] init 1
+  kindex init 2
+  kvalue = kvalues[kindex]
+  kcycle timeinstk
+  printks "after-error cycle=%d value=%f\n", 0, kcycle, kvalue
```
hlolli, line 16: Can we make reaching this line fail for a reason other than the array error? On current `develop`, this line prints and Csound still exits with code 1, so the expected-failure test passes before and after the patch. An assertion tied to `after-error` would make the regression prove that the rest of the chain stopped.

hlolli (commented):
Thanks for taking these two PARCS issues on. I left a few notes from local runs. This is a comment-only review.

tests/commandline/test_parcs_perf_error.csd
vlazzarini: currently there is no way to differentiate performance errors in the tests, it's either fail or pass. If this is checking for perfErrors then it's a fail and we can't say whether the failure was on line 14 or after line 16. So there is no assertion possible here - if it is set to fail.

Top/csound_perf.c
vlazzarini: fixed in d685eec35

Top/csound_perf.c
vlazzarini: fixed in d685eec35

Top/csound_debug.c
vlazzarini: fixed in d685eec35

Top/csound_perf.c
vlazzarini: fixed in d685eec35

tests/commandline/test_parcs_perf_error.csd
hlolli: we have another PR for asserting specifically the errors, llm agrees, that you're correct, we leave it as is

vlazzarini:
I realised we of course need to keep going after a perfError, so the node doesn't exit immediately on finding an error in an instance, it can move on to the next if there is one. It returns the total error count which we might find a use for.

hlolli (commented):
Pulled the latest changes at `8807590e7` and gave them another pass. The earlier PARCS issues look fixed in my local runs. I found two control-flow cases worth checking and left possible fixes inline. This is a comment-only review.

Top/csound_perf.c
```diff
@@ -77,21 +77,22 @@ inline static int32_t node_perf(CSOUND *csound, int32_t index,
...
     int32_t done;
+    int32_t error = 0;
     which_task = dag_get_task(csound, index, numThreads, next_task);
     if(csoundGetDebug(csound) & DEBUG_PARCS)
       csound->Message(csound, "Select task %d %d\n", which_task, index);
     if (which_task == WAIT)
       continue;
     if (which_task == INVALID)
-      return played_count;
+      return error;
```
hlolli: Could this return `errorcnt` here? `error` is set to zero at the start of each loop, so the `INVALID` path always returns zero and both checks after `node_perf()` stay quiet. The `return errorcnt` below cannot run. Returning `errorcnt` on this path should keep the per-task reset and pass the full count back to the caller.

Top/csound_debug.c
```diff
@@ -973,12 +970,13 @@ int32_t kperf_debug(CSOUND *csound) {
...
 #endif
-      /* do the mixing of thread buffers */
       for (k = 1; k < csound->oparms->numThreads; k++)
           mix_out(csound->spout_tmp, csound->spout_tmp +
                   k * csound->nspout, csound->nspout);
-#endif /* PARCS */
-      csound->multiThreadedDag = NULL;
+      csound->multiThreadedDag = NULL; 
+#else /* currently disabled */
+      return CSOUND_ERROR;
```
hlolli: Could the callers treat a negative `kperf()` result as an error here? `CSOUND_ERROR` is `-1`, while `csoundPerformKsmps()` and `perform_buffer()` repeat on any nonzero result. With an indefinite note, one API call stayed in that loop until my timeout. One possible fix is to save the result, repeat only while it is positive, then unlock and return a negative value. An earlier check before performance starts would also avoid this loop.

vlazzarini: fixed in ab4bd6445

Top/csound_perf.c
vlazzarini: fixed in ab4bd6445

Top/csound_debug.c
```diff
@@ -67,6 +67,10 @@ void csoundDebuggerBreakpointReached(CSOUND *csound)
 
  void csoundDebuggerInit(CSOUND *csound)
 {
+    /* to be removed if/when multicore is supported */
+    if(csound->multiThreadedThreadInfo != NULL)
+      csoundDie(csound, "cannot initialise debugger for multicore\n");
```
hlolli: I pulled the latest changes at `5b4326ad8`. `csoundDie()` is unsafe here. The documented call order runs `csoundDebuggerInit()` after `csoundStart()`, when the matching `setjmp()` frame has already returned, so this can jump through a stale frame. My repro let init return and then `csoundDebuggerClean()` aborted because init never set `csdebug_data`. Calling init before start also skips this check because `multiThreadedThreadInfo` is still null. That path returned success but produced silence.

Could we check `csound->oparms->numThreads` here, report the error without a longjmp, let `csoundDebuggerClean()` return early when `csdebug_data` is null, and add a matching guard in `csoundStart()` for code that initializes the debugger before multicore setup? If the API can change, returning a status from init would also give callers a clean result.

hlolli (commented):
I left one inline note after checking the latest changes.

Top/csound_debug.c
vlazzarini: ok, this is proving a PITA to get completed. Fixed in [f29b561](https://github.com/csound/csound/pull/2747/commits/f29b56172dfe33d8504866889521bbac08375ef2), where I changed the API to return an error code.

vlazzarini:
Left some comments, this is getting messy. It should be a localised disable.

hlolli (dismissed):

hlolli (approved):

PR #2749, opened 2026-08-07
Fix for local ksmps kcounter in parcs

### LINKED ISSUE #2743: PARCS gives instruments with local ksmps the wrong control-cycle count
Instruments that use `setksmps` see a wrong `kcounter` when Csound runs with more than one thread. Anything that derives time from the local control cycle, such as `timeinstk`, reports wrong values.

The serial path in `kperf` (Top/csound_perf.c:341) starts the local counter at `(csound->kcounter - 1) * csound->ksmps / lksmps` and increments it once per local pass. The worker path in `node_perf` (Top/csound_perf.c:133) starts it at `csound->kcounter * csound->ksmps`. It drops the `- 1` and never divides by the local ksmps, so the counter is shifted by one global cycle, scaled wrong, and jumps between global cycles.

## Reproducer

Save this as `parcs_local_kcounter.csd`:

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d -m0
</CsOptions>
<CsInstruments>
sr = 64
ksmps = 8
nchnls = 1
0dbfs = 1

instr 1
  setksmps 4
  kwhen timeinstk
  printk 0, kwhen
endin
</CsInstruments>
<CsScore>
i 1 0 0.25
e
</CsScore>
</CsoundSynthesizer>
```

Run it twice:

```sh
csound --num-threads=1 parcs_local_kcounter.csd
csound --num-threads=2 parcs_local_kcounter.csd
```

## Expected

Both runs count the local cycles 1 through 4:

```text
instr 1:  time     0.06250:     1.00000
instr 1:  time     0.06250:     2.00000
instr 1:  time     0.18750:     3.00000
instr 1:  time     0.18750:     4.00000
```

## Actual

One thread prints the expected values. Two threads print:

```text
instr 1:  time     0.06250:     8.00000
instr 1:  time     0.06250:     9.00000
instr 1:  time     0.18750:    16.00000
instr 1:  time     0.18750:    17.00000
```

Note the jump from 9 to 16 between global cycles. With `setksmps 1` the same run prints 8 through 15 instead of 1 through 8.

Tested with:

```text
Csound 7.0, double samples
macOS 26.5, arm64
Apple clang 21.0.0
```

vlazzarini:
This PR fixes the kcounter update with local ksmps in PARCS.
Closes #2743

### DIFF
--- a/tests/commandline/test.py
+++ b/tests/commandline/test.py
@@ -854,6 +854,11 @@ def runTest():
             ]
         )
 
+
+
+        tests += [[ "test_parcs_local_ksmps_kcounter.csd",
+                   "PARCS kcounter test with local ksmps"]]
+
     arrayTests = [
         ["arrays/arrays_i_local.csd", "local i[]"],
         ["arrays/arrays_i_global.csd", "global i[]"],
--- a/Top/csound_perf.c
+++ b/Top/csound_perf.c
@@ -132,7 +132,7 @@ inline static int32_t node_perf(CSOUND *csound, int32_t index,
         
         insds->spin = csound->spin;
         insds->spout = csound->spout_tmp + index * csound->nspout;
-        insds->kcounter = csound->kcounter * csound->ksmps;
+        insds->kcounter = ((csound->kcounter - 1) * csound->ksmps / lksmps) + 1;
         /* we have to deal with sample-accurate code
            whole CS_KSMPS blocks are offset here, the
            remainder is left to each opcode to deal with.
--- a/tests/commandline/test_parcs_local_ksmps_kcounter.csd
+++ b/tests/commandline/test_parcs_local_ksmps_kcounter.csd
@@ -1,0 +1,27 @@
+<CsoundSynthesizer>
+<CsOptions>
+-n -d  -j 2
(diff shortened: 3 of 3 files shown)


tests/commandline/test.py
```diff
@@ -854,6 +854,11 @@ def runTest():
             ]
         )
 
+
+
+        tests += [[ "test_parcs_local_ksmps_kcounter.csd",
+                   "PARCS kcounter test with local ksmps"]]
```
hlolli, line 860: I think we could combine the append above into this, since this += concat scales but append doesn't

hlolli (approved):

tests/commandline/test.py
hlolli: no stress about it, I am going to have wasm just ignore the -j flag in another PR so I'll simplify these exceptions soon

hlolli (approved):

tests/commandline/test.py
vlazzarini: yes, I was just mindful of the changes in the other PR and did not want to get a merge conflict out of the box.

PR #2763, opened 2026-08-22
fix(debug): report live values for typed pass-by-ref UDO arguments

ervoLB:
## Summary

- `csoundDebugGetUdoFrames()` / `csoundDebugGetVariables()` reported 0 for every named argument of a Csound 7 typed-header UDO (`opcode name(ain:a, kGain:k):a`), while audio was correct. Classic `opcode name, a, ak` + `xin` UDOs were unaffected.
- Cause: when caller `ksmps`/`esr` match, `useropcdset()` enables pass-by-ref and rewires the body onto `UOPCODE.ar[]`. The instance pool slots for those names are allocated but never written; the debugger was reading `lclbas + memBlockIndex`.
- Fix: keep the compile-time name → `ar_index` table already built by `csoundBuildUserOpcodeRewirePlan()`, and resolve through `csoundUserOpcodeRefArgStorage()`. That table already encodes pass-through xout and in-place `aSig = udo(aSig)` collisions, which a hand-rolled xin/xout walk would get wrong.

Engine runtime behaviour is unchanged: the alias table is retained instead of freed, plus a read-only accessor. No hot-path change. `debug_variable_t` layout and WASM exports are unchanged.

Not in this PR: `csoundDebugGetUdoFrames()` still enumerates O(2^depth) frames for recursive UDOs; clients de-duplicate per call site.

### DIFF
--- a/Engine/udo.c
+++ b/Engine/udo.c
@@ -188,11 +188,6 @@ static inline void rewire_argpp(CSOUND *csound, OPDS *chain, int32_t index,
 static MYFLT *pbr_resolve_struct_target(CSOUND *csound, MYFLT *argPtr,
                                         const char *structPath);
 
-typedef struct pbr_alias_entry {
-  const char *name;
-  int32_t ar_index;
-} PBR_ALIAS_ENTRY;
-
 typedef struct pbr_plan_builder {
   PBR_ALIAS_ENTRY *aliases;
   int32_t alias_count;
@@ -451,7 +446,11 @@ static int32_t pbr_add_alias(CSOUND *csound,
     return NOTOK;
   }
 
-  builder->aliases[builder->alias_count].name = name;
+  builder->aliases[builder->alias_count].name =
+    cs_strndup(csound, name, strlen(name));
+  if (builder->aliases[builder->alias_count].name == NULL) {
+    return NOTOK;
+  }
   builder->aliases[builder->alias_count].ar_index = ar_index;
   builder->alias_count++;
 
@@ -834,6 +833,22 @@ static void pbr_free_seed_entries(CSOUND *csound,
   csound->Free(csound, entries);
 }
 
+static void pbr_free_aliases(CSOUND *csound,
+                             PBR_ALIAS_ENTRY *aliases,
+                             int32_t alias_count) {
+  int32_t i;
+
+  if (aliases == NULL) {
+    return;
+  }
+
+  for (i = 0; i < alias_count; i++) {
+    csound->Free(csound, aliases[i].name);
+  }
+
+  csound->Free(csound, aliases);
+}
+
 static void pbr_free_plan(CSOUND *csound, PBR_REWIRE_PLAN *plan) {
   if (plan == NULL) {
     return;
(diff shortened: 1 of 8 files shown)


vlazzarini:
@coderabbitai review

Engine/udo.c
```diff
@@ -1407,6 +1425,70 @@ static int32_t handle_pass_by_ref(CSOUND* csound, UOPCODE* p,
...
+   than read from OPCODINFO.passByRef, which only records the most recent
+   init. */
+static int32_t udo_call_is_pass_by_ref(const UOPCODE *p,
+                                       const OPCODINFO *udoinfo) {
+  return udoinfo != NULL && udoinfo->newStyle && p->ip != NULL &&
+    p->parent_ip != NULL && p->parent_ip->ksmps == p->ip->ksmps &&
+    p->parent_ip->esr == p->ip->esr;
+}
+
+MYFLT *csoundUserOpcodeRefArgStorage(const UOPCODE *p, const char *varName) {
```
vlazzarini: is this function PUBLIC, part of the API or just internal to Csound? If it is the latter then please use snake case: user_opcoderef_arg_storage(), dropping the "csound" prefix.

Top/csound_debug.c
```diff
@@ -344,9 +344,17 @@ void csoundDebugFreeOpcodeList(CSOUND *csound, debug_opcode_t *opcode_list)
...
  *   [       -> ARRAYDAT*, decode with csoundDebugSerializeArray()
+ *
+ * udo is the invocation a UDO instance belongs to, NULL for top-level
+ * instruments and the global pool. Csound 7 typed headers
+ * (opcode name(a:k):a) run pass-by-ref: the xin/xout names are rewired
+ * onto the caller's storage and the instance's own pool slots for them
+ * stay untouched, so those names must be read through
+ * csoundUserOpcodeRefArgStorage() instead of lclbas.
  */
 static debug_variable_t *csoundDebugBuildVarList(
```
vlazzarini: since this is static, please use snake case: debug_build_var_list(), dropping the csound prefix. Please review other function names in this file. The rule is (a) public api functions are in the form csoundXxx() - camel case; (b) static and non-public extern functions are snake case xxx_xxx(), generally without the need for a csound_ prefix.

H/udo.h
```diff
@@ -116,4 +116,10 @@ void recycle_init_only_udo_instances(CSOUND *, INSDS *);
 void csoundBuildUserOpcodeRewirePlan(CSOUND *, OPCODINFO *);
```
vlazzarini: My comment above about snake_case applies here too even though this was not part of the current PR. If this is not supposed to be PUBLIC, then it should be "build_user_opcode_rewire_plan()" and the next one  below too.

vlazzarini (changes requested):
some naming convention points.

ervoLB:
Hi @vlazzarini ! Thanks for the review — these are all internal (not PUBLIC / not in the WASM export list).

Renamed per the rule:
- user_opcode_ref_arg_storage()
- build_user_opcode_rewire_plan() / free_opcode_info_chain()
- static helpers in Top/csound_debug.c to snake_case (debug_build_var_list(), etc.)

The public debugger API (csoundDebugGetUdoFrames, csoundDebugGetVariables, …) is unchanged.

vlazzarini (approved):
All good, many thanks

PR #2766, opened 2026-09-01
fix: prevent realtime audio pops during interrupt shutdown

kunstmusik:
Render a short engine fade when the command-line frontend receives an interrupt, then let realtime backends drain their queued output before closing.

Use one frame-synchronous fade helper for PortAudio callback and AUHAL output. For blocking PortAudio on CoreAudio, feed 450 ms of genuine silence before stop/reset so the device tail clears without replaying audio.

Close audio before teardown, keep a second interrupt as a forced exit, and cover interleaved fade behavior with focused unit tests.

Note: Work done with assistance by GPT 5.6 Sol.

### DIFF
--- a/Frontends/csound/csound_main.c
+++ b/Frontends/csound/csound_main.c
@@ -38,7 +38,6 @@
 extern int set_rt_priority(int argc, const char **argv);
 #endif
 extern int csoundErrCnt(CSOUND*);
-extern int32_t csoundPerformOutputFade(CSOUND *, uint32_t);
 static FILE *logFile = NULL;
 
 static void msg_callback(CSOUND *csound,
--- a/Frontends/csound/csound_main.c
+++ b/Frontends/csound/csound_main.c
@@ -38,6 +38,7 @@
 extern int set_rt_priority(int argc, const char **argv);
 #endif
 extern int csoundErrCnt(CSOUND*);
+extern int32_t csoundPerformOutputFade(CSOUND *, uint32_t);
 static FILE *logFile = NULL;
 
 static void msg_callback(CSOUND *csound,
@@ -87,8 +88,9 @@ static void psignal(int sig, char *str)
 }
 #endif
 
-static CSOUND *_csound = NULL;
-static int perf_flag = 1;
+static volatile sig_atomic_t perf_flag = 1;
+static volatile sig_atomic_t received_signal = 0;
+
 static void signal_handler(int sig) {
 #if defined(SIGPIPE)
     if (sig == (int) SIGPIPE) {
@@ -98,19 +100,21 @@ static void signal_handler(int sig) {
       return;
     }
 #endif
+    if ((sig == (int) SIGINT || sig == (int) SIGTERM)) {
+      if (received_signal) {
+        /* Second Ctrl-C: force-exit immediately */
+        _exit(1);
+      }
+      received_signal = sig;
+      /* Let main() return from csoundPerformKsmps so it can render the
+         interrupt fade and shut down from a safe context. */
+      perf_flag = 0;
+      return;
+    }
 #ifndef __wasm__
     psignal(sig, "\ncsound command");
 #endif
-    if ((sig == (int) SIGINT || sig == (int) SIGTERM)) {
-      if (_csound) {
-        perf_flag = 0;
-        csoundDestroy(_csound);
-      }
-      if (logFile != NULL)
-        fclose(logFile);
-      exit(1);
-    }
-    exit(1);
+    _exit(1);
 }
 
 static const int sigs[] = {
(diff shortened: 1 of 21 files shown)


kunstmusik:
@vlazzarini: This gets rid of pops when interrupting playback (ctrl-c, as well as from apps like Blue when hitting stop) with auhal, pa_cb, and pa_bl. If you could give this a try and it looks/sounds good to you, we can extend the fade out windows and linux drivers before completing this PR.

vlazzarini:
Thanks, I'll have a look.
What about Linux and Windows: alsa, jack, pipewire, pulseaudio and wasapi?

InOut/libsnd.c
```diff
@@ -51,6 +51,25 @@ static inline void alloc_globals(CSOUND *csound)
     csound->libsndStatics.nframes = (uint32)1;
 }
 
+static void reset_output_fade(CSOUND *csound)
+{
+  STA(outputFadeFrames) = 0;
+  STA(outputFadePos) = 0;
+}
+
+PUBLIC int32_t csoundPerformOutputFade(CSOUND *csound, uint32_t frames)
```
vlazzarini: don't declare PUBLIC in the implementation. Leave this to be set at the header prototype.

Frontends/csound/csound_main.c
```diff
@@ -38,6 +38,7 @@
 extern int set_rt_priority(int argc, const char **argv);
 #endif
 extern int csoundErrCnt(CSOUND*);
+extern int32_t csoundPerformOutputFade(CSOUND *, uint32_t);
```
vlazzarini: Should this not be in the host API? csound_rtaudio.h in this case.

InOut/rtauhal.c
```diff
@@ -26,6 +26,7 @@
 #include <unistd.h>
 #include <stdint.h>
 #include "csdl.h"
+#include "rt_audio_fade.h"
```
vlazzarini, line 29: ideally we would not want backends to depend on internal headers as they might be built as plugins. We should depend on the public API.

vlazzarini (changes requested):
I think we need to decide whether this is part of the API. Headers for backends need to be public and changes need to be rolled out to all other backends we support. We need to have a clear module API support for this for anyone implementing audio IO on other platforms as a plugin.

An alternative would be to design a system with no changes to backends whatsoever, all the fading being managed inside the engine. I would be favouring that as the addition of fade support may be a complication we do not need to add to backend development.

kunstmusik:
Thanks for the review; I'm not sure we can generalize completely since there's some quirks to each backend (e.g., portaudio blocking issue that required 450ms silence to push data out). Seems you're okay with this in principle so I'll look at cleaning up and having it applied to other backends.

vlazzarini:
ok, I guess the worse that can happen is that a backend doesn't support fades.

kunstmusik:
@vlazzarini just a heads up, I did a pass with one agent for applying the fade/drain work to all backends. That agent both took a long time and also resulted in an over-engineered solution. I'm redoing that work now but due to 5h credit limits I may have have to wait until the 5h reset later today to complete the work.

hlolli:
I tested this on Linux at commit `5c2f230abac944af4eb6cf31cf4ee0e7b1e76897` and found no blocking issue.

What I checked:

- Read the PR description, discussion, inline review comments, and the relevant shutdown/audio code.
- Built Csound and the ALSA, JACK, PipeWire, PortAudio, and PulseAudio backends with Ninja.
- Ran the new realtime fade unit tests; all passed.
- Sent Ctrl-C while Csound was producing audio through virtual Linux devices:
  - ALSA null device
  - JACK dummy server
  - PortAudio callback and blocking modes through JACK
  - PulseAudio null sink
  - PipeWire null sink with WirePlumber
- Confirmed each backend shut down instead of hanging.
- Confirmed `csoundPerformOutputFade` is exported by the shared library.

Limitations:

- This machine has no physical audio device, so I could not listen for a pop on real hardware.
- PortAudio blocking mode crashed when opening ALSA's named `null` device. I reproduced the same crash with the exact pre-PR `rtpa.c`, so this does not appear to be caused by this PR. Both PortAudio modes passed through JACK.

From the Linux side, this looks good to me.

kunstmusik:
Thanks @hlolli for the check; probably still worth a smoke test with audio on some of those. 

@rorywalsh would you have time to check audio on Windows that it fades out gracefully instead of pops when ctrl-c is hit?

rorywalsh:
Yes @kunstmusik, but I won't be near a Windows machine for two weeks.

vlazzarini:
@kunstmusik should we extend support to WASM/Web Audio?

kunstmusik:
@vlazzarini: I've added an AI-assisted change for adding fadeout to webaudio. I'm going to do some further testing on various platforms to see how this performs.

kunstmusik:
I did some smoke tests on Linux (alsa, pa_bl, pa_cb) and Windows (portaudio, winmm, wasapi). Oddly on Windows, using trapped which was set to 44100 sr, using portaudio it wouldn't open the device and I had to use --sample-rate=48000. Not sure that's related to anything here. I did not hear pops when using these drivers.

I think we're okay to proceed here. We may need others to test elsewhere but I imagine users will raise issues quickly with further testing. 

@vlazzarini could I get an approval here?

vlazzarini (approved):

PR #2770, opened 2026-09-05
fix: keep logbtwo audio output indexing in range

hlolli:
`logbtwo` reused `n` for the lookup-table index inside the audio loop. That lookup index could then be used as the output buffer index, which can write far past the current audio block.

This keeps the sample index for `r[n]` and uses a separate index for `logbase2`.

### DIFF
--- a/OOps/aops.c
+++ b/OOps/aops.c
@@ -1538,10 +1538,10 @@ int32_t logbasetwoa(CSOUND *csound, EVAL *p)
   }
   for (n = offset; n < nsmps; n++) {
     MYFLT aa = a[n];
-    int32_t n = (int32_t)((aa -(FL(1.0)/INTERVAL)) / (INTERVAL-FL(1.0)/INTERVAL)
-                          *  STEPS + FL(0.5));
-    if (n<0 || n>STEPS) r[n] = LOG(aa)*ONEdLOG2;
-    else                r[n] = csound->logbase2[n];
+    int32_t indx = (int32_t)((aa -(FL(1.0)/INTERVAL)) / (INTERVAL-FL(1.0)/INTERVAL)
+                             *  STEPS + FL(0.5));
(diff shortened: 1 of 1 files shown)


OOps/aops.c
```diff
@@ -1538,10 +1538,10 @@ int32_t logbasetwoa(CSOUND *csound, EVAL *p)
   }
   for (n = offset; n < nsmps; n++) {
     MYFLT aa = a[n];
-    int32_t n = (int32_t)((aa -(FL(1.0)/INTERVAL)) / (INTERVAL-FL(1.0)/INTERVAL)
-                          *  STEPS + FL(0.5));
-    if (n<0 || n>STEPS) r[n] = LOG(aa)*ONEdLOG2;
-    else                r[n] = csound->logbase2[n];
+    int32_t indx = (int32_t)((aa -(FL(1.0)/INTERVAL)) / (INTERVAL-FL(1.0)/INTERVAL)
+                             *  STEPS + FL(0.5));
+    if (indx<0 || indx>STEPS) r[n] = LOG(aa)*ONEdLOG2;
```
vlazzarini, line 1543: I see your point but that changes the original code.  Can we make sure results are correct or at least they are the same as the original code?

vlazzarini: Ok. I found the very original code from 23 years ago.
```
    a = p->a;
    r = p->r;
    do {
      MYFLT aa = *a++;
      int n = (int) ((aa - (FL(1.0)/INTERVAL)) / (INTERVAL - FL(1.0)/INTERVAL)
                     *  STEPS + FL(0.5));
      if (n<0 || n>STEPS) *r++ = (MYFLT)(log((double)aa)*ONEdLOG2);
      else                *r++ = logbase2[n] ;
    } while (--nsmps);
```

so the bug is confirmed!

vlazzarini (approved):

PR #2782, opened 2026-09-07
fix: validate string getcol column bounds

hlolli:
`getcol` for string arrays checked the requested column against the row count. This rejected valid columns when a matrix had more columns than rows, and accepted invalid columns when it had more rows than columns, leading to out-of-bounds reads.

The check now uses the column dimension and rejects negative indexes. A command-line regression test covers the last valid column of a 2x3 string matrix.


hlolli:
Superseded by #2783, which uses the conventional fix/arrays-getcol-string-bounds branch.

PR #2798, opened 2026-09-08
Fix fof and fof2 edge-case arithmetic and overlap storage

hlolli:
`fof` and `fof2` can dereference missing or undersized overlap storage with a negative phase, leave zero-sample grains active indefinitely, and divide by zero when long rise times truncate the fixed-point increment to zero. The octave mask also left-shifts a signed negative value.

Allocate overlap storage when required, clamp unrepresentable grain and rise counts before integer conversion, and build the octave mask with unsigned shifts. Keep normal sample-count rounding and existing glissando behavior.


Opcodes/ugens7.c
```diff
@@ -249,17 +248,24 @@ static int32_t newpulse(CSOUND *csound,
...
+  if (ovp->timrem > p->durtogo &&
       (*p->iskip==FL(0.0))) /* ringtime */
     return(0);
   if ((oct = *p->koct) > FL(0.0)) {                   /* octaviation */
-    int64_t csnt = -1;
-    int64_t ioct = (int64_t) oct;
-    int64_t bitpat = ~(csnt << ioct);
-    if (bitpat & ++p->fofcount)
+    int64_t ioct = oct >= (MYFLT) INT64_MAX ? 64 : (int64_t) oct;
+    uint64_t bitpat = ioct < 64 ? (UINT64_C(1) << ioct) - 1 : UINT64_MAX;
```
vlazzarini, line 272: are we sure we get the same result here? We can't change behaviour, unless it leads to a crash.

Opcodes/ugens7.c
```diff
@@ -293,14 +299,21 @@ static int32_t newpulse(CSOUND *csound,
         
       }
       ovp->risincf = (CS_ONEDSR / *p->kris);
-      rismps = (int32_t) (1. / ovp->risincf);  
+      double rise_samples = 1. / ovp->risincf;
```
vlazzarini, line 310: why a double? this is compared below to an int and then copied into an int?

vlazzarini (changes requested):

Opcodes/ugens7.c
hlolli: I had GPT Luna check this against the old code. For the defined input range, with 0 <= floor(koct) < 64, the unsigned mask produces the same bits as the old expression, so the octave selection and fractional amplitude behavior stay unchanged. The change removes undefined signed shifts for larger values, which had no defined behavior before.

The unconditional overlap allocation only changes the negative iphs path. Previously, that path skipped allocation and then built the overlap free list through a null pointer, so it could crash during initialization. Negative phase is normalized with PHMOD1; allocation lets it use the same overlap path as other phases.

The rismps change also preserves values within range and only avoids overflow for excessively large rise times. The regression test covers these cases.

hlolli: I am re-reviewing this, after realising I had wrong model enabled... (should have been Astra not Luna)

hlolli: Amended in 139331093. You were right to question the behavior claim. The earlier duration cast could turn a 10 ms grain at 48 kHz into 479 samples instead of 480 in float builds. The calculation now retains the original MYFLT multiplication precision.

I removed the float-phase sampct reset from this PR because it changes glissando output on inputs that do not crash. I also narrowed negative-phase allocation to missing or undersized storage, so it leaves a sufficient existing buffer alone.

A correction to my earlier mask explanation: left-shifting signed -1 is undefined even for shift counts below 64. The unsigned expression preserves the intended mask for counts 0–63. The regression now asserts pulse timing and fractional-octave amplitudes for both table paths in fof and fof2, as well as grain duration and finite sample values. The previous fixture only checked successful completion, so my earlier assurance overstated what it proved.

Opcodes/ugens7.c
hlolli: The reciprocal was already evaluated as double in `(int32_t) (1. / ovp->risincf)`. Keeping it in `double rise_samples` lets us check its range before converting it. `INT32_MAX` is compared as a double too.

Values within range use the same reciprocal and integer truncation as before. Larger values clamp to INT32_MAX before the cast; converting to int first would make the range check too late.

vlazzarini: I would have thought `int32_t  rise_samples = (int32_t) (1./ ovp->risincf)` since the comparison is with INT32_MAX and the assignment is to an int32_t.

hlolli: The comparison is checking whether the sample count fits in `int32_t`. For example, a 50,000-second rise at 48 kHz gives roughly 2.4 billion samples, above `INT32_MAX`. With your declaration, that value would undergo an out-of-range conversion before reaching the check. The intermediate needs to hold that larger value so we can clamp it; the final `rismps` remains `int32_t`.

vlazzarini: Ok, so the check is in the wrong place. *p->kris should be clamped to the grain duration elsewhere, and the grain duration maybe needs to be limited to something, if it is not already.

hlolli: The check is here because this is where the floating-point sample count converts to `int32_t`. Earlier validation could make it unnecessary. The patch already caps grain duration at `INT32_MAX` samples.

My concern with also clamping `kris` to `kdur` is that it changes working cases: a 10 ms grain with a 20 ms rise currently ends partway through the rise. Clamping would change that envelope.

Is `kris > kdur` intended to be invalid? The current code allows it, so I kept the guard at the integer conversion to preserve those cases.

vlazzarini: Sure, but it seems like we should not permit ridiculous values like a 50k-second rise and maybe issue a warning or an error.

hlolli: Yes, an error for a rise time outside the implementation’s supported range seems reasonable. I used 50,000 seconds to demonstrate the overflow, rather than as an intended use case. Would you like me to change this to report an error before the unsafe conversion?

vlazzarini: good idea

hlolli: it's done

vlazzarini: sorry I can't see the changes. The latest commit [e406b6d](https://github.com/csound/csound/pull/2798/commits/e406b6d32420a2d4f7ae602759a3aa979297a7df) seems not be changed.

hlolli: my bad, I felt this conversation got resolved automatically because I already fixed it, I check what I'm forgetting..

hlolli: fixed

vlazzarini (approved):

PR #2799, opened 2026-09-08
fix: correct granule indexing and input validation

hlolli:
`granule` can divide by a zero source length, read outside its selected source region, and read before an envelope table. Zero or nonfinite pitch values can also reach invalid arithmetic and integer conversions. This change checks the affected inputs before using them and corrects the sample and envelope indices.

- Require a source region of at least one sample. A positive `ilength` can still truncate to zero samples, leaving a zero divisor in the wrapping code. Check the start and length for finite values and integer range before casting, then check the remaining table span and integer addition before computing the end. Keep the later bounds check because thresholding can shorten the table, and compare against its unsigned length without narrowing it to a signed integer.

- Preserve sample rounding while checking bounds. Compute sample counts with `MYFLT` rounding, then compare them in double precision. This keeps float builds from treating a rounded-up `INT32_MAX` as a valid integer limit, without shifting existing source regions. Initialize each voice from the checked `gstart` so the region and its voices use the same start. For example, at 8 kHz, multiplying the float value of `0.005` in double precision truncates to 39 samples, while the original float multiplication rounds to 40.

- Require positive, finite pitches only for the slots selected by `ipshift`. Zero pitch can reach a division by zero when a grain restarts; nonfinite values can reach undefined float-to-integer conversions. Unused pitch slots do not affect playback and should not cause initialization to fail.

- Use `[gstart, gend)` for both interpolation samples. The reverse path can produce `gend` as its primary index, so wrap it before the first read. Correct the neighboring sample's wrap in both directions as well. This prevents interpolation from including data after the selected region or wrapping to the wrong sample within it.

- Handle zero-length attack and decay stages without division. A zero attack, including one that truncates to zero samples, previously produced NaN at the start of a grain. Use full envelope level when no ramp is needed and divide only when the stage has a positive length. This also prevents that NaN from reaching the envelope index conversion.

- Map envelope levels onto valid table entries. The old `floor(level * length) - 1` expression produces index `-1` at level zero. Scale by `length - 1` and clamp the index, mapping zero to the first entry and one to the last. This intentionally changes some interior lookups in nonconstant envelopes to follow that normalized mapping.

The regression cases cover envelope start, end, and interior values; forward and reverse wrapping; unused pitch slots; and source-region rounding with a nonzero skip. Assertions reject NaN as well as incorrect finite output. Each invalid-input case runs separately so one rejected input cannot hide a missing check for another.


vlazzarini:
Just checking - the PR cites grain4 but this is about granule?

vlazzarini:
Just checking - the PR cites grain4 but this is about granule?

hlolli:
Yes, this fixes granule. The implementation lives in Opcodes/grain4.c, so I carried that name into the commit and tests. Sorry for the confusion. There's no opcode grain4?

vlazzarini (approved):

PR #2800, opened 2026-09-08
fix: stop diskgrain and syncloop grains at envelope end

hlolli:
Stop diskgrain grains when the envelope index reaches the table length. The previous `>` check allowed interpolation to read past the table's guard sample at the exact endpoint.

Also check the envelope index before interpolation: several overlapping grains can finish together, while cleanup removes only one per sample. Skip the finished grains still awaiting cleanup to prevent further out-of-bounds reads.

Apply the `>=` completion check to syncloop too. Its interpolation already checks the bounds, but grains at the exact endpoint stayed marked active, causing duplicate output when overlap slots were reused.

### DIFF
--- a/Opcodes/syncgrain.c
+++ b/Opcodes/syncgrain.c
@@ -338,7 +338,7 @@ static int32_t syncgrainloop_process(CSOUND *csound, syncgrainloop *p)
         /* if the envelope is finished */
         /* the grain is also finished */
 
-        if (UNLIKELY(envindex[j] > envtablesize))
+        if (UNLIKELY(envindex[j] >= envtablesize))
           streamon[j] = 0;
       }
 
--- a/Opcodes/syncgrain.c
+++ b/Opcodes/syncgrain.c
@@ -668,6 +668,11 @@ static int32_t filegrain_process(CSOUND *csound, filegrain *p)
         if (index[j]  < 0)
           index[j] += dataframes;
 
+        if (UNLIKELY(envindex[j] >= envtablesize)) {
+          streamon[j] = 0;
+          continue;
+        }
+
         /* sum all the grain streams */
         tndx = (int32_t)index[j]*chans;
         endx = (int32_t) envindex[j];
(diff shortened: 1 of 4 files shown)


Opcodes/syncgrain.c
```diff
@@ -668,6 +668,11 @@ static int32_t filegrain_process(CSOUND *csound, filegrain *p)
         if (index[j]  < 0)
           index[j] += dataframes;
 
+        if (UNLIKELY(envindex[j] >= envtablesize)) {
```
vlazzarini, line 671: if this is a real issue, then we ought to apply the fix to the other places where this code is used (eg. syncgrain). It's strange that we never had an issue with this code and it's over 20 years old.

vlazzarini (changes requested):

Opcodes/syncgrain.c
hlolli: Yes, I can reproduce this with `test_diskgrain_envelope_bounds.csd` in the PR. On develop (`c7aea5036`), ASan reports a heap-buffer-overflow in `filegrain_process`; the patched build passes.

The table has one guard sample at index `flen`. When the envelope index reaches `flen`, interpolation also reads `ftable[flen + 1]`, beyond that guard sample. The old `>` check allows this. There is also an overlap case: several grains can finish together, but cleanup removes only one per sample, so the remaining finished grains need the check before interpolation.

I checked the related code. `syncgrain` already checks `envindex < envtablesize` before interpolation and uses `>=` to finish grains. `syncloop` also guards interpolation, but still has a separate `>` completion issue that should be fixed.

I can’t say why this hasn’t been reported before. An out-of-bounds read does not necessarily cause a crash, so the lack of reports would not rule it out.

vlazzarini: I guess if syncgrain is good, then it hasn't been reported because diskgrain is not used as much. Also since this is all memory owned by Csound, it generally won't crash, we need asan or valgrind to find the bug.

Could you apply the syncloop
fix in this PR?

hlolli: on it..

hlolli: https://github.com/csound/csound/pull/2800/changes/c9b97e9423533920ba130193a47b5c76e9dae157

vlazzarini (approved):

PR #2801, opened 2026-09-08
fix: validate olabuffer overlap before integer arithmetic

hlolli:
Reject zero and negative olabuffer overlap factors before integer division or buffer allocation. Zero previously passed the integer check and caused division by zero.

Compare the overlap factor with the array size before converting it to `int32_t`. This rejects excessively large values and positive infinity without an out-of-range conversion, while preserving valid overlap factors.

### DIFF
--- a/Opcodes/framebuffer.c
+++ b/Opcodes/framebuffer.c
@@ -201,10 +201,10 @@ void OLABuffer_checkArgumentSanity(CSOUND *csound, OLABuffer *self)
 {
     MYFLT overlapCount = *self->overlapArgument;
 
-    if (UNLIKELY(floor(overlapCount) != overlapCount)) {
+    if (UNLIKELY(overlapCount <= FL(0.0) || floor(overlapCount) != overlapCount)) {
 
-      csound->Die(csound,
-                  "%s", Str("olabuffer: Error, overlap factor must be an integer"));
+      csound->Die(csound, "%s",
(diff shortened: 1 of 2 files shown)


Opcodes/framebuffer.c
```diff
@@ -201,10 +201,10 @@ void OLABuffer_checkArgumentSanity(CSOUND *csound, OLABuffer *self)
 {
     MYFLT overlapCount = *self->overlapArgument;
 
-    if (UNLIKELY(floor(overlapCount) != overlapCount)) {
+    if (UNLIKELY(overlapCount <= FL(0.0) || floor(overlapCount) != overlapCount)) {
 
-      csound->Die(csound,
-                  "%s", Str("olabuffer: Error, overlap factor must be an integer"));
+      csound->Die(csound, "%s",
```
vlazzarini: calling Die() here seems wrong, can we not give an error and carry on?

vlazzarini (changes requested):

Opcodes/framebuffer.c
hlolli: @vlazzarini Changed `olabuffer` validation to return `InitError()` and stop initialization of the affected instrument. Other instruments can continue performing; it no longer calls `Die()` for invalid arguments.

vlazzarini (approved):

PR #2803, opened 2026-09-08
fix: keep scanhammer table copies in bounds

hlolli:
`scanhammer` wrapped one point too late, allowing copies to overwrite the destination table's guard point. Wrap at the table length and check the starting position before converting it to an index, so invalid values cannot cause out-of-bounds writes or undefined floating-to-integer conversions. Fractional positions select the point below them.

Check table lookups before dereferencing them. When source and destination are the same table, preserve the source values before a nonzero-offset copy so writes cannot corrupt values that still need to be read.

### DIFF
--- a/Opcodes/wave-terrain.c
+++ b/Opcodes/wave-terrain.c
@@ -111,24 +111,48 @@ static int32_t wtPerf(CSOUND *csound, WAVETER *p)
 static int32_t scanhinit(CSOUND *csound, SCANHAMMER *p)
 {
   uint32_t srcpos = 0;
-  uint32_t dstpos = (uint32_t)MYFLT2LONG(*p->ipos);
+  uint32_t dstpos;
+  MYFLT *source;
+  MYFLT *sourceCopy = NULL;
 
   FUNC *fsrc = csound->FTFind(csound, p->isrc); /* Source table */
   FUNC *fdst = csound->FTFind(csound, p->idst); /* Destination table */
+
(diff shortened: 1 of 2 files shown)


Opcodes/wave-terrain.c
```diff
@@ -111,24 +111,48 @@ static int32_t wtPerf(CSOUND *csound, WAVETER *p)
...
 
+  /* Validate before converting the position or writing the first sample. */
+  if (UNLIKELY(!(*p->ipos >= FL(0.0) &&
+                 (double)*p->ipos < (double)fdst->flen)))
+    return csound->InitError(csound, "%s",
+                            Str("scanhammer: position must be within the "
+                                "destination table"));
+
+  /* A fractional position selects the point below it. */
+  dstpos = (uint32_t)floor((double)*p->ipos);
```
vlazzarini: instead of floor() we can just truncate the index. The index is always non-negative.

Also note we have macros. I think we have a FLOOR. These are preferable instead of the funcall.

vlazzarini (changes requested):

Opcodes/wave-terrain.c
hlolli: @vlazzarini Replaced `floor()` with a direct `uint32_t` cast. The preceding bounds check guarantees a non-negative index, so truncation preserves the same behavior.

vlazzarini (dismissed):

vlazzarini (approved):

PR #2806, opened 2026-09-08
ci: update workflow actions to latest stable releases

hlolli:
Update external actions in all three workflows and the local PR-label action to their latest stable release tags. Pin exact versions so each workflow states the release it uses, including replacing `jlumbroso/free-disk-space@main` with `v1.3.1`.

This includes checkout v7, artifact upload v7 and download v8, cache v6, GitHub Script v9, Docker action upgrades, and current releases of the Nix, CMake, Emscripten, tagging, release, and JUnit actions. `cachix/cachix-action@v17` remains current.

### DIFF
--- a/.github/actions/check-pr-label/action.yml
+++ b/.github/actions/check-pr-label/action.yml
@@ -15,7 +15,7 @@ runs:
   using: "composite"
   steps:
     - name: Checkout code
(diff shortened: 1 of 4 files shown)


kunstmusik (approved):
looks good to me

PR #2810, opened 2026-09-08
fix: handle zero stages across the ADSR family

hlolli:
Zero attack left ADSR envelopes silent or stale after reinit and could crash `xadsr`. The exponential variants also subtracted 1 ms from attack, breaking shorter attacks, and `mxadsr` ignored its release override.

Use one stage implementation for `adsr`, `madsr`, `xadsr`, and `mxadsr`. It handles zero stages immediately, honors delay and release overrides at both rates, and validates counts before conversion. Fixed stage storage removes invalid segment pointers.

Exponential curves retain their small positive floor, then finish at exact zero. Zero sustain stays silent during release. The hidden 1 ms timing adjustment is removed. The generic `linseg` and `expseg` families are unchanged.


OOps/ugens1.c
```diff
@@ -306,26 +306,29 @@ static int32_t adsrset1(CSOUND *csound, LINSEG *p, int32_t midip)
...
+      nsegs*sizeof(SEG) > p->auxch.size) {
     csound->AuxAlloc(csound, (size_t) nsegs * sizeof(SEG), &p->auxch);
     p->cursegp = segp = (SEG *) p->auxch.auxp;
     segp[nsegs-1].cnt = MAXPOS; /* set endcount for safety */
   }
-  else if (**argp > FL(0.0))
+  else if (**argp >= FL(0.0))
     memset(p->auxch.auxp, 0, (size_t)nsegs*sizeof(SEG));
-  if (**argp <= FL(0.0))  return OK;       /* if idur1 <= 0, skip init  */
+  if (**argp < FL(0.0))  return OK;        /* negative attack skips init */
```
vlazzarini: do you think we need this? It may still be problematic, for the same reasons of attack zero.

OOps/ugens1.c
```diff
@@ -210,15 +210,15 @@ int32_t klnseg(CSOUND *csound, LINSEG *p)
...
       if (UNLIKELY(!(--p->segsrem)))  {
         p->curval = segp->nxtpt;      /* advance the cur val  */
         return OK;
       }
       p->cursegp = ++segp;            /*   find the next      */
       if (UNLIKELY(!(p->curcnt = segp->cnt))) { /*   nonlen = discontin */
-        p->curval = segp->nxtpt;      /*   poslen = new slope */
-        /*          p->curval += p->curinc;  ??????? */
-        return OK;
+        *p->rslt = p->curval = segp->nxtpt;
```
vlazzarini: do we need these changes. Afaik linseg was ok with zero length segments. If we can avoid making changes beyond the adsr code, it would be better.

OOps/ugens1.c
```diff
@@ -433,7 +441,7 @@ int32_t klnsegr(CSOUND *csound, LINSEG *p)
       segp = ++p->cursegp;               /*   else find nextseg  */
     newi:
       if (!(p->curcnt = segp->cnt)) {    /*   nonlen = discontin */
-        p->curval = segp->nxtpt;         /*     reload & rechk   */
+        *p->rslt = p->curval = segp->nxtpt; /* apply the jump before output */
```
vlazzarini: same here, do we need this change for zero lengths to work?

vlazzarini (changes requested):

OOps/ugens1.c
```diff
@@ -289,108 +289,189 @@ int32_t linseg(CSOUND *csound, LINSEG *p)
...
+                         int32_t exponential)
+{
+  MYFLT **args = p->argums;
+  double rate = IS_ASIG_ARG(p->rslt) ? CS_ESR : CS_EKR;
+  double length = p->h.insdshead->p3.value;
+  int32_t counts[5], release, override = -1, length_count, i;
+  const int32_t inputs[4] = {4, 0, 1, 3};
+  const int32_t stages[4] = {ADSR_DELAY, ADSR_ATTACK, ADSR_DECAY, ADSR_RELEASE};
+
+  /* Preserve the historical negative-attack skip, but initialize zero. */
```
vlazzarini, line 319: ~~I don't think we should preserve that. It won't be very useful - we need a better "skip" initialisation strategy. Basically I think the approach should be to feed the last output of the previous envelope as the starting point of the new (instead of starting from zero), and then recalculate the attack re-increment and initialise the other segments too. We should add a dedicated skip parameter at the end so that we can have zero attack in this case too. It's not really skipping anything but doing a legato envelope. Does that make sense? Skipping all initialisation may be confusing - any changes to other parameters are then ignored.~~ Maybe I'm wrong - skipping bypasses the envelope altogether I guess - keeps to the same sustain level. That's how mono synths work in legato mode. Let's try it.


Btw I like your latest solution, separating the adsr family from linseg etc so any changes here do not have unintended side effects elsewhere.

vlazzarini (changes requested):

vlazzarini (approved):

PR #2813, opened 2026-09-09
Fix vdelay buffer sizing and indexing

hlolli:
`vdelay` and `vdelay3` could derive zero or mismatched ring-buffer sizes, use stale size metadata after skipped initialization, and convert unbounded delay values to integer indices. These cases could access memory outside the allocated buffer or cause undefined float-to-integer conversions.

This change validates the maximum delay before sizing the buffer and keeps a one-sample minimum. State is preserved only when a buffer already exists, without changing its size or write position. Delay values now wrap in floating point before index conversion, while non-finite values return an error. It also fixes the short-buffer, control-rate `vdelay3` path so it writes each input sample into the delay buffer.


OOps/vdelay.c
```diff
@@ -109,27 +116,14 @@ int32_t vdelay(CSOUND *csound, VDEL *p)               /*      vdelay  routine */
...
-          fv1 += (MYFLT)maxd;
-        while (UNLIKELY(fv1 >= (MYFLT)maxd))
-          fv1 -= (MYFLT)maxd;
-
-        if (LIKELY(fv1 < maxd - 1)) /* Find next sample for interpolation      */
-          fv2 = fv1 + FL(1.0);
-        else
-          fv2 = FL(0.0);
-
+        fv1 = vdelay_readpos(fdel, esr, maxd, indx);
```
vlazzarini: I don't like to see a function call in the processing loop. Can you replace by a macro or inline the code here?

OOps/vdelay.c
```diff
@@ -206,26 +195,23 @@ int32_t vdelay3(CSOUND *csound, VDEL *p)    /*  vdelay routine with cubic interp
...
-      fv1 = *del * -esr; v1 = (int32_t)fv1; fv1 -= (MYFLT) v1;
-      v1 += (int32_t)indx;
-      /* Make sure Inside the buffer      */
-      if ((v1 < 0L) || (fv1 < FL(0.0))) {
-        fv1++; v1--; while (UNLIKELY(v1 < 0L)) v1 += (int32_t)maxd;
-      }
-      else {
-        while (UNLIKELY(v1 >= (int32_t)maxd)) v1 -= (int32_t)maxd;
-      }
+      fv1 = vdelay_readpos(*del, esr, maxd, indx);
```
vlazzarini: Same here - replace by inlined code or macro. We can't have a funcall here.

vlazzarini (changes requested):

OOps/vdelay.c
hlolli: the function can be decorated as inline (seems more modern to me), I'll see what's best

vlazzarini: but decoration doesn't guarantee inlining in the C specs. So that's no good, compilers may just ignore it.

hlolli: That's true, but I felt in practice this wasn't ever the case, just that the compiler is free to decide according to the spec in a strict sense. The pro is readability and type/compiler warnings/errors. But turning this into a macro is np.

hlolli: fixed

OOps/vdelay.c
hlolli: fixed

OOps/vdelay.c
vlazzarini: thanks, we need to consider the whole range of platforms this code may be compiled for.

vlazzarini (approved):

PR #2814, opened 2026-09-09
Bound mirror folding for extreme values

hlolli:
`mirror` reflected out-of-range values in a loop. Extreme finite inputs could require many iterations, doubled-bound arithmetic could overflow, and infinite inputs could keep the loop running forever. Computing the midpoint for reversed or equal bounds could also overflow even when both bounds were finite.

This change gives the audio-rate and control-rate forms one bounded reduction based on floating-point remainders. It avoids forming an overflowing doubled range, preserves values already inside the range, and computes the fallback midpoint without overflowing the sum.

### DIFF
--- a/Opcodes/uggab.c
+++ b/Opcodes/uggab.c
@@ -80,24 +80,44 @@ static int32_t kwrap(CSOUND *csound, WRAP *p)
 
 /*---------------------------------------------------------------------*/
 
+static MYFLT mirror_value(MYFLT input, MYFLT lower, MYFLT upper)
+{
+    double value = input, low = lower, high = upper;
+    double width, remainder, lowrem;
+    int quotient, lowquot, odd;
+
+    if (low >= high) {
+      double sum = low + high;
+      return (MYFLT)(isfinite(sum) ? sum * 0.5 : low * 0.5 + high * 0.5);
+    }
+    if (value >= low && value <= high) return input;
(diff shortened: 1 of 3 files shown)


Opcodes/uggab.c
```diff
@@ -121,22 +141,15 @@ static int32_t mirror(CSOUND *csound, WRAP *p)
...
     for (n=offset;n<nsmps;n++) {
-      xsig = asig[n];
-      while ((xsig > xhigh) || ( xsig < xlow )) {
-        if (xsig > xhigh)
-          xsig = xhigh + xhigh - xsig;
-        else
-          xsig = xlow + xlow - xsig;
-      }
-      adest[n] = xsig;
+      adest[n] = mirror_value(asig[n], xlow, xhigh);
```
vlazzarini: same here as before - no function call in processing loop. Replace by macro or copy the code inline here

vlazzarini (changes requested):

Opcodes/uggab.c
hlolli: Fixed in f2dd2ff390f9494224fbe2c3d519a11edd1f8666. Replaced mirror_value with a MIRROR_VALUE macro, so the folding code expands directly in the audio loop. The i-rate and k-rate paths use the same macro.

vlazzarini (dismissed):

vlazzarini (approved):

PR #2815, opened 2026-09-09
Add stable chebyshevpoly2 evaluator

hlolli:
`chebyshevpoly` expands Chebyshev terms into ordinary polynomial coefficients. This keeps its short Horner loop for common low orders, but high-order intermediate coefficients can overflow even when the final series is finite. This change keeps the existing opcode unchanged and documents that known tradeoff.

`chebyshevpoly2` evaluates the series directly with Clenshaw's recurrence and double-precision intermediate values. It accepts the existing variadic coefficient form plus one-dimensional `i[]` and `k[]` coefficient arrays, rejects empty arrays, and has direct constant and linear fast paths. Variadic coefficients are copied into contiguous storage once per control block, and the recurrence stays in each processing loop so no helper function runs per sample.

### DIFF
--- a/Opcodes/shape.c
+++ b/Opcodes/shape.c
@@ -159,6 +159,12 @@ static int32_t ChebyshevPolyInit(CSOUND* csound, CHEBPOLY* p)
    Coefficients (k0, k1, k2, ... ) are k-rate multipliers of each Chebyshev
    polynomial starting with the zero-order polynomial T0(x) = 1, then
    T1(x) = x, T2(x) = 2x^2 - 1, etc. */
+/*
+ * Known limitation: expanding high-order Chebyshev terms can overflow even
+ * when the final series is finite. Keep this implementation unchanged for
+ * its established behavior and faster low-order sample loop. Use
+ * chebyshevpoly2 for direct evaluation; do not apply that fix here.
+ */
 static int32_t ChebyshevPolynomial(CSOUND* csound, CHEBPOLY* p)
 {
     int32_t     i, j;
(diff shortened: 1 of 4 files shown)


Opcodes/shape.c
```diff
@@ -139,29 +139,15 @@ static int32_t Polynomial(CSOUND* csound, POLYNOMIAL* p)
...
-    csound->AuxAlloc(csound, (2*ncoeff + 1)*sizeof(MYFLT), &(p->coeff));
-    p->chebn = ((MYFLT*)p->coeff.auxp) + ncoeff;
-    return OK;
-}
-
 /* Efficiently evaluates a sum of Chebyshev polynomials of arbitrary order.
    Coefficients (k0, k1, k2, ... ) are k-rate multipliers of each Chebyshev
    polynomial starting with the zero-order polynomial T0(x) = 1, then
    T1(x) = x, T2(x) = 2x^2 - 1, etc. */
 static int32_t ChebyshevPolynomial(CSOUND* csound, CHEBPOLY* p)
```
vlazzarini, line 168: Don't change this code. Add an alternative version using the new code - if possible to reuse the same opcode name with an extra optional parameter or otherwise a new name. There's a balance between overflow and efficiency at play here. We don't want to enforce any penalties.

vlazzarini (changes requested):

Opcodes/shape.c
hlolli: I’ll restore chebyshevpoly unchanged and add the direct-recurrence implementation as chebyshevpoly2. This keeps the current performance characteristics while providing a stable option for higher orders.

vlazzarini: excellent. That's in line with other opcodes we updated.

hlolli: done, and added comment on "legacy" chebyshevpoly about this limitation so others don't attempt to fix this, knowing we are aware and accept this status quoe.

vlazzarini (dismissed):

kunstmusik (approved):

PR #2816, opened 2026-09-09
Keep syncphasor phase within one cycle

hlolli:
`syncphasor` adjusted its phase by only one cycle after each sample. Frequencies at or above the sample rate, and equally large negative frequencies, could therefore leave the stored phase outside `[0, 1)`. Initial phases were reduced through an integer conversion that could become undefined for large values. The inactive part of the sync output could also retain stale samples.

This change normalizes finite phases with `floor()` after initialization and every advance, reports non-finite phases and frequencies as errors, and clears inactive samples in both outputs. Wrap detection still emits one sync trigger for any sample that crosses a cycle boundary in either direction.


Opcodes/shape.c
```diff
@@ -491,21 +500,17 @@ int32_t SyncPhasor(CSOUND *csound, SYNCPHASOR *p)
...
-          if (phase >= 1.0) {
-            phase -= 1.0;
-            syncout[n] = FL(1.0);        /* send sync when phase wraps */
-          }
-          else if (phase < 0.0) {
-            phase += 1.0;
-            syncout[n] = FL(1.0);
-          }
-          else syncout[n] = FL(0.0);
+          int32_t wrapped = syncphasor_advance(&phase, incr);
```
vlazzarini: function call in loop again.

Opcodes/shape.c
```diff
@@ -425,18 +425,29 @@ typedef struct {
...
     return OK;
 }
 
+static int32_t syncphasor_advance(double *phase, double increment)
+{
+    double next = *phase + increment;
+    int32_t wrapped;
+    if (UNLIKELY(!isfinite(next))) return NOTOK;
+    wrapped = next >= 1.0 || next < 0.0;
+    *phase = next - floor(next);
```
vlazzarini: use FLOOR() (include/sysdep.h) instead

vlazzarini (changes requested):

Opcodes/shape.c
hlolli: Fixed in 4325c9e8c. Removed syncphasor_advance and put the phase update directly in both processing loops.

Opcodes/shape.c
hlolli: Used FLOOR() for the initial phase in 4db9b3de8. The loops need to keep floor(): phase is double even in float builds, where FLOOR() maps to floorf(). That conversion can round next up to 1 before subtraction and leave a negative phase.

vlazzarini (dismissed):

kunstmusik (approved):

PR #2818, opened 2026-09-09
Fix distortion partial-block state and unsafe arithmetic

hlolli:
`distort` used the full control-block length when smoothing over a sample-accurate partial block. This changed its state and output based on inactive samples. It could also preserve uninitialized state when `istor` was set on first use. The opcode now processes only the active span, clears inactive output, and initializes or restores valid state as needed.

The change rejects non-finite parameters and state before they can reach table indexing, which avoids an undefined float-to-integer conversion. It also rewrites the equivalent `distort1` expression with scaled exponents so large pregain values do not produce an overflow through `inf / inf`.


Opcodes/compress.c
```diff
@@ -239,31 +244,45 @@ static int32_t distort(CSOUND *csound, DIST *p)
...
-      memset(&ar[nsmps], '\0', early*sizeof(MYFLT));
-    }
-    dinc = (dnew - dcur) / nsmps;
+    dinc = (dnew - dcur) / (nsmps - offset);
     for (n=offset; n<nsmps; n++) {
       MYFLT sig, phs, val;
       sig = asig[n] / dcur;             /* compress the sample  */
       phs = p->midphs * (FL(1.0) + sig); /* as index into table  */
-      if (UNLIKELY(phs <= FL(0.0)))
+      if (UNLIKELY(isnan(phs)))
```
vlazzarini: is there a way to avoid calling isnan() on every sample? Can we have a macro version (not sure if isnan() is actually a macro)

Opcodes/compress.c
```diff
@@ -239,31 +244,45 @@ static int32_t distort(CSOUND *csound, DIST *p)
...
+    }
+    if (UNLIKELY(offset >= nsmps))
+      return OK;
     q = p->prvq;
-    for (n=offset; n<nsmps-early; n++) {
+    for (n=offset; n<nsmps; n++) {
       q = p->c1 * asig[n] * asig[n] + p->c2 * q;
     }
-    p->prvq = q;
+    if (UNLIKELY(!isfinite(q)))
```
vlazzarini, line 259: do we really need to call isfinite() on every sample?

Opcodes/compress.c
```diff
@@ -239,31 +244,45 @@ static int32_t distort(CSOUND *csound, DIST *p)
...
+      return csound->PerfError(csound, &(p->h), "%s",
+                               Str("distort: input level is out of range"));
     rms = SQRT(q);    /* get running rms      */
     if (rms < p->min_rms)
       rms = p->min_rms;
-    if ((dist = *p->kdist) < FL(0.001))
+    dist = *p->kdist;
+    if (UNLIKELY(!isfinite(dist)))
```
vlazzarini, line 266: also here. The original code did not have these funcalls and no one reported issues.

Opcodes/compress.c
hlolli: Removed the separate isnan() check in 2d827c038. The table bounds comparisons now keep NaN out of the integer conversion.

Opcodes/compress.c
hlolli: An upstream NaN or overflow when squaring a large input can make q non-finite. Taking its square root does not recover a valid level, so we need to reject it before updating the gain and saved state. The later dnew check would also catch this, so this particular check is not strictly required for safety; keeping it reports an input-level error instead of blaming the distortion amount. It runs once per control block.

Opcodes/compress.c
hlolli: This also runs once per control block. A NaN kdist bypasses the lower clamp and can reach table indexing, so I kept the check.

vlazzarini (approved):

PR #2824, opened 2026-09-10
Fix partial-block audio input bounds

hlolli:
Audio input opcodes placed partial-block output at the right offset but still read from input frame zero. The array form also measured its offset in bytes, while `inx` and `in32` wrote each frame into sample zero and cleared the wrong end of the output.

Read each active sample from the matching input frame, keep offsets in samples, and clear the exact inactive prefix and tail across the scalar, array, fixed-channel, selected-channel, and generic input paths. Compute one end index for all `inch` outputs, and expand the shared 16/32-channel loop as a macro to avoid a helper call in the performance path.


OOps/aops.c
```diff
@@ -1767,42 +1768,40 @@ int32_t ino(CSOUND *csound, INO *p)
...
-      memset(ara[i], '\0', early*sizeof(MYFLT));
-  }
-  for (m = offset; m < nsmps; m++) {
-    for (i = 0; i < n; i++)
-      *ara[i] = *sp++;
-  }
-  CSOUND_SPIN_SPINUNLOCK
-    return OK;
-}
+#define INN(N)                                                          \
```
vlazzarini: here I think a function is OK, it's only called once per block. We do that in several places, I don't think we need a macro

OOps/aops.c
```diff
@@ -1767,42 +1768,40 @@ int32_t ino(CSOUND *csound, INO *p)
...
+      for (i = 0; i < (N); i++)                                        \
+        ara[i][m] = *sp++;                                              \
+    CSOUND_SPIN_SPINUNLOCK                                              \
+    return OK;                                                         \
+  } while (0)
 
 int32_t in16(CSOUND *csound, INALL *p)
 {
-  return inn(csound, p, 16u);
+  INN(16u);
```
vlazzarini: As mentioned above, one call per kcycle is fine. The compiler will optimise it away probably.

vlazzarini (changes requested):

OOps/aops.c
hlolli: Replaced INN with a shared static function in dc4677ff5.

OOps/aops.c
hlolli: Done in dc4677ff5. Both in16 and in32 now call the function once per block

vlazzarini (approved):

PR #2828, opened 2026-09-10
Fix track frequency bounds and trfilter Nyquist lookup

hlolli:
`trscale` and `trshift` now clamp output frequencies to zero through Nyquist. Downward shifts still work; results below zero become zero. NaN results also become zero.

`trfilter` keeps its own frequency bounds before indexing the filter table. Negative frequencies now become zero instead of being reflected with an absolute value.

At Nyquist, `trfilter` uses the table's guard-point value directly, without reading beyond it. This endpoint fix also applies to valid, non-negative track frequencies. Filtering within the valid range stays unchanged.

### DIFF
--- a/Opcodes/psynth.c
+++ b/Opcodes/psynth.c
@@ -641,6 +641,8 @@ static int32_t trscale_process(CSOUND *csound, _PTRANS *p)
         else
           frameout[i] = framein[i];
         outfr = framein[i + 1] * scale;
+        if (UNLIKELY(!(outfr >= FL(0.0))))
+          outfr = FL(0.0);
         frameout[i + 1] = (float) (outfr < nyq ? outfr : nyq);
         frameout[i + 2] = framein[i + 2];
         id = (int32_t) framein[i + 3];
@@ -670,6 +672,8 @@ static int32_t trshift_process(CSOUND *csound, _PTRANS *p)
         else
           frameout[i] = framein[i];
         outfr = framein[i + 1] + shift;
+        if (UNLIKELY(!(outfr >= FL(0.0))))
+          outfr = FL(0.0);
         frameout[i + 1] = (float) (outfr < nyq ? outfr : nyq);
         frameout[i + 2] = framein[i + 2];
         id = (int32_t) framein[i + 3];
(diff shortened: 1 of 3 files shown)


vlazzarini:
I think frequencies are strictly non-negative in pvs track analysis, so this extra check is not needed.

hlolli:
@vlazzarini The negative frequencies here come from track processing after analysis. Both `trscale` and `trshift` accept a TRACKS signal and clamp only the upper frequency bound:

- `trscale ftracks, -10` turns the 1 kHz track into **-10 kHz**.
- `trshift ftracks, -10000` turns it into **-9 kHz**.

This standalone CSD demonstrates both cases without calling `trfilter`:

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d -m0
</CsOptions>
<CsInstruments>
sr = 8000
ksmps = 8
nchnls = 1
0dbfs = 1
giSine ftgen 1, 0, 8192, 10, 1
instr 1
  asig oscili .5, 1000, giSine
  ffreq, fphase pvsifd asig, 64, 16, 1
  ftracks partials ffreq, fphase, .001, 1, 3, 10
  fscaled trscale ftracks, -10
  fshifted trshift ftracks, -10000
  fl1, kfreq1, kamp1 trlowest fscaled, 1
  fl2, kfreq2, kamp2 trlowest fshifted, 1
  kcycle timeinstk
  if kcycle == 90 then
    printks "scaled=%f shifted=%f amps=%f,%f\n", 0, kfreq1, kfreq2, kamp1, kamp2
    if kamp1 <= 0 || kamp2 <= 0 || kfreq1 >= -4000 || kfreq2 >= -4000 then
      exitnowk(-1)
    endif
  endif
endin
</CsInstruments>
<CsScore>
i 1 0 .1
</CsScore>
</CsoundSynthesizer>
```

Running it prints:

```text
scaled=-10000.000000 shifted=-9000.000000 amps=0.249990,0.249990
```

At this sample rate, Nyquist is 4 kHz. If either output feeds `trfilter`, its existing absolute-value operation turns the frequency into 10 or 9 kHz. That is why the upper clamp must come **after** the absolute value. The separate Nyquist endpoint guard also remains necessary for non-negative frequencies: at Nyquist, the interpolation index is `len`, so reading `fil[posi + 1]` goes past the guard point even though the interpolation weight is zero.

vlazzarini:
ok so the issue is trscale and trshift. They should limit their output to non-negative freqs. In pvs tracks format amps and freqs should strictly be non-negative.

hlolli:
@vlazzarini Agreed. I updated `trscale` and `trshift` to clamp output frequencies to zero through Nyquist. Downward shifts still work; results below zero become zero. `trfilter` keeps its own bounds check and the separate Nyquist endpoint fix, which is needed for valid non-negative frequencies too.

vlazzarini (approved):

PR #2829, opened 2026-09-10
Fix crossfm sample offsets and phase lookup bounds

hlolli:
Fix input alignment and table lookup bounds across `crossfm`, `crosspm`, `crossfmpm`, and their interpolating variants.

- When a note starts partway through a block, advance audio-rate frequency and modulation inputs to its first active sample. Control-rate inputs stay in place. Previously, the oscillators read the skipped prefix and produced the wrong onset.
- Keep wrapped phases below 1 before converting them to table indexes. Tiny negative phases can round to 1 after wrapping, causing interpolation to read past the guard point. The same comparison keeps NaN out of the integer conversion; empty tables are rejected at init.
- Reuse the integer table index to compute the interpolation fraction, removing redundant `floor` operations. Shared macros add no helper function calls to the sample loops.

### DIFF
--- a/Opcodes/crossfm.c
+++ b/Opcodes/crossfm.c
@@ -29,12 +29,32 @@
 #include "crossfm.h"
 #include <math.h>
 
+#define XFM_ADVANCE_INPUTS(offset_)                   \
+  do {                                                \
+    xfrq1 += (offset_) * p->frq1adv;                  \
+    xfrq2 += (offset_) * p->frq2adv;                  \
+    xndx1 += (offset_) * p->ndx1adv;                  \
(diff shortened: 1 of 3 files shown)


vlazzarini:
@fggp these are yours, are they not?

fggp:
This code is effectively mine (2009?); the point corrected here concerns the --sample-accurate case. This part of the code had been added later, maybe by John?
Anyway these corrections are subtle and well done. We can merge.

fggp (commented):
Ok for me

fggp (approved):

PR #2843, opened 2026-09-10
fix: update pitch conversion arrays every control cycle

hlolli:
Array `mtof` and `ftom` skip their first control cycle, leaving init-time results after the inputs change. For example, changing MIDI 69 to 81 still gives 440 Hz for the first block instead of 880 Hz. Remove the skip so both match their scalar versions from the first cycle.

Also check output capacity before either conversion loop, update output lengths to match the inputs, and return on failure. `ftom` lacked this check; `mtof` ignored its result. This prevents writes beyond reserved storage without allocating during performance or adding checks inside the element loops.

### DIFF
--- a/Opcodes/emugens/emugens.c
+++ b/Opcodes/emugens/emugens.c
@@ -358,7 +358,6 @@ typedef struct {
     MYFLT *irnd;
     MYFLT freqA4;
     int32_t rnd;
-    int32_t skip;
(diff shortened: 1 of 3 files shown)


vlazzarini:
@gesellkammer these are yours, I believe?

gesellkammer (approved):

PR #2845, opened 2026-09-10
fix: preserve lag state and partial audio blocks

hlolli:
Fix state and sample handling in `lag` and `lagud`:

- Save the new control-rate `lag` output when lag time changes. Keeping the old state made repeated changes prevent convergence to the input.
- Initialize audio state from the first active sample when no initial value is supplied. Reading sample zero gave notes starting within a block the wrong initial state.
- Remove duplicate tail handling in audio `lagud`, which dropped active samples at note ends.
- Ramp coefficients across the active samples of a partial block and leave state untouched in empty blocks.

The aliases share these fixes. The sample loops gain no calls or checks, and full-block audio processing keeps its existing formulas.

### DIFF
--- a/Opcodes/emugens/scugens.c
+++ b/Opcodes/emugens/scugens.c
@@ -25,7 +25,7 @@
 #include "emugens_common.h"
 
 #define LOG001 FL(-6.907755278982137)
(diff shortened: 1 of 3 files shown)


vlazzarini:

gesellkammer (approved):

PR #2846, opened 2026-09-10
fix: preserve trighold's first pulse after initialization

hlolli:
`trighold` called its control-rate performance routine during initialization. An already-positive input therefore started the hold before performance, shortening the first pulse by one control cycle and losing a one-cycle pulse entirely.

Set the initial output directly while leaving the counter and trigger state reset. This preserves the init-time output and gives the first pulse its full duration after initialization or reinitialization. The aliases share the fix; performance processing stays unchanged.


vlazzarini:
@gesellkammer another one of yours, could you review?

gesellkammer (approved):

PR #2847, opened 2026-09-10
fix: keep trigphasor starts and resets within range

hlolli:
Fix `trigphasor` and its aliases:

- Start from the first active `kstart` value. Previously, audio output began at zero even outside the requested range; empty blocks now leave state untouched.
- Use `kstart` when the reset position is omitted, while preserving an explicit zero. Control-rate triggers now reset the audio output directly, without an extra phase step or an unused division.
- Wrap audio state before output after a range change or trigger. Equal bounds produce the start value instead of NaN.

The audio loops use a wrap macro and retain the existing interpolation for audio-rate triggers. Ordinary wrap arithmetic stays unchanged.

### DIFF
--- a/Opcodes/emugens/scugens.c
+++ b/Opcodes/emugens/scugens.c
@@ -72,15 +72,17 @@ static inline MYFLT sc_wrap(MYFLT in, MYFLT lo, MYFLT hi) {
 }
 
 
-// Scalar branchless version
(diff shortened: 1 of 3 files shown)


vlazzarini:
@gesellkammer another one, please review.

gesellkammer (approved):

PR #2850, opened 2026-09-10
fix: restart triggered envelopes and correct segment timing

hlolli:
Restart `triglinseg` and `trigexpseg` once per trigger block, resetting both the segment and its sample count. Previously, retriggering during a later segment could drive the envelope past its specified peak, while the trigger block repeatedly reset each sample to the first step.

Share the timing code so both shapes advance between segments without an extra sample, reach each endpoint, and hold the starting value until triggered. Zero-length stages jump to their endpoint; negative exponential values of the same sign now work as documented.

Compute segment increments at initialization, removing division and `pow` from the sample loop. Reject incomplete segment lists and negative durations at initialization.


vlazzarini:
@rorywalsh are these yours? Can you review?

rorywalsh (approved):
Looks good.

PR #2861, opened 2026-09-10
Add opt-in corrected resony linear spacing

hlolli:
Add an optional trailing `icorrect` argument to select corrected `resony` spacing while preserving existing calls.

```csound
ares resony asig, kbf, kbw, inum, ksep [, isepmode, iscl, iskip, icorrect]
```

- `icorrect = 0` (default) keeps the original spacing formula and integer conversion of `isepmode`.
- `icorrect = 1` places the first filter at the base frequency and adds the spacing in hertz for each following filter when `isepmode` is nonzero. Zero `isepmode` keeps octave spacing. This mode also rejects a zero base frequency before relative bandwidth scaling divides by it.

For example, append `, 1, 1, 0, 1` after `ksep` to select corrected linear spacing, normalization mode 1, and cleared filter state.

### DIFF
--- a/Opcodes/uggab.c
+++ b/Opcodes/uggab.c
@@ -345,6 +345,11 @@ static int32_t product(CSOUND *csound, SUM *p)
 
 static int32_t rsnsety(CSOUND *csound, RESONY *p)
 {
+    /* icorrect opts into base-relative linear spacing; old calls keep
+       the original spacing and integer conversion of isepmode. */
+    if (UNLIKELY(*p->icorrect != FL(0.0) && *p->icorrect != FL(1.0)))
+      return csound->InitError(csound, "%s",
+                               Str("resony: icorrect must be 0 or 1"));
     double order = (double)*p->ord;
     double scale_value = (double)*p->iscl;
     size_t state_size;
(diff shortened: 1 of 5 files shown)


vlazzarini:
This modifies behaviour - even if it's not ideal. Can we add a mode for this fix instead of changing it?

vlazzarini (changes requested):
See my comment about changed behaviour

hlolli:
@vlazzarini Added a trailing optional `icorrect` argument, defaulting to `0`, so existing calls keep their spacing. Set it to `1` to enable the corrected base-relative linear spacing.

vlazzarini:
Thanks. We need to make sure these and other changes we're making are documented.

hlolli:
@vlazzarini Opened a [companion manual PR](https://github.com/csound/manual/pull/797) documenting `icorrect` and the default compatibility behavior, plus the related `pvsbin`, `gtadsr`, `unwrap`, and `pitchac` behavior. It includes examples for corrected `resony` spacing and `pitchac`.

vlazzarini (approved):

hlolli:
@vlazzarini The documentation for the merged changes is now in https://github.com/csound/manual/pull/798 as a regular PR. It covers `resony`’s default behavior and `icorrect` mode, with an example, plus `pvsbin` and `unwrap`.

PR #2884, opened 2026-09-11
Fix vosim non-power-of-two tables and add opt-in pulse corrections

hlolli:
Add optional `icorrect` (default `0`) after `iskip`. Power-of-two tables keep the legacy pulse timing and amplitude behavior unless `icorrect=1` selects the corrected path.

Non-power-of-two tables always use the corrected phase, pulse-boundary, and amplitude handling. In corrected mode, each event starts at the requested formant rate, and negative pulse factors reverse direction between pulses.


vlazzarini:
I'm wondering about behaviour change in this one. It's a longstanding opcode in the collection.

hlolli:
@vlazzarini These correctness fixes may slightly change the sound of scores that hit the affected paths. A one-sample timing difference would usually be barely noticeable, though some of these fixes can affect more than one sample. If the audible differences are common and substantial, we could introduce `vosim2` and document the correctness problems in `vosim`, preserving its existing behavior if you prefer.

vlazzarini:
@hlolli let's try adding a correctness parameter for the part that affects legacy power-of-two tables path. The fix for non-power-two can be applied without any further worries.

hlolli:
@vlazzarini Added optional `icorrect`, defaulting to `0`. Power-of-two tables retain the legacy behavior; `icorrect=1` opts into the pulse fixes. Non-power-of-two tables always use the corrected path.

vlazzarini (approved):

PR #2891, opened 2026-09-11
Fix spout_interleave dropping samples when ksmps > -b

ervoL:
Fixes a bug in `spout_interleave()` (`InOut/libsnd.c`) that truncates a k-cycle when `ksmps` is larger than the software buffer (`-b` / `outbufsamps`).

This shows up in **embedded / host** use (`-n`, `-+rtaudio=NULL`, `csoundSetHostAudioIO`), where `Engine/musmon.c` does **not** clamp `outbufsamps >= ksmps`. Standalone `-odac` is unaffected because that clamp is applied there. Default `-b` is 128 on macOS and 256 on Windows, so `ksmps = 200` or `512` is enough to hit it.

## Failure trace (`ksmps = 512`, `nchnls = 2`, `outbufsamps = 128`)

Pass 1: `start = 0`, `end = 128`, `outbufrem` hits 0, then `start = 128`, `end = 512`.

Pass 2: `spoutrem > outbufrem` is true, and `end` was recomputed as

```c
end = csound->libsndStatics.outbufrem / nchnls;  /* 128, ignores start */
```

so `end - start == 0`. The copy loop runs 0 times. `outbufrem` is still nonzero, so the reset / `goto nchk` path does not run. Frames 128..511 stay as the `kperf` memset zeros. Hosts then hear a periodic buzz at `sr/ksmps` (93.75 Hz at 48 kHz / 512).

## Fix

```c
end = start + csound->libsndStatics.outbufrem / nchnls;
```

One-line change. No API / behaviour change when `ksmps <= -b`.

## Test

`IOTests.testSpoutFilledWhenKsmpsExceedsSoftwareBuffer` in `tests/c/io_test.cpp`: host IO, `-n -d -b128`, `ksmps = 512`, always-on `poscil`, assert the full spout k-cycle is filled (not trailing zeros, no large adjacent steps).

## Context

Found while embedding Csound 7 in a VST3 host (papad.dev). With the unpatched library, `ksmps = 100` was clean on macOS (`-b 128`) and `ksmps = 200` / `512` buzzed at every DAW buffer size we tried. After this fix those cases are continuous.

### DIFF
--- a/Engine/musmon.c
+++ b/Engine/musmon.c
@@ -201,12 +201,22 @@ MYFLT initialise_io(CSOUND *csound) {
           * O->outbufsamps;
         if (O->oMaxLag <= O->outbufsamps && O->outbufsamps > 1)
           O->outbufsamps >>= 1;
-        if(O->outbufsamps < csound->ksmps) {
-          O->outbufsamps = csound->ksmps;
-          O->oMaxLag = csound->ksmps*2;
-        }
       }
       O->inbufsamps = O->outbufsamps;
+    }
+    /* Host/embedded IO skips the rtaudio block above, so default -b can
+       stay below ksmps. Raise -b and scale -B to keep the original ratio. */
+    if (O->outbufsamps > 0 && O->outbufsamps < (int32_t) csound->ksmps) {
+      int32_t old_b = (int32_t) O->outbufsamps;
+      int32_t new_b = (int32_t) csound->ksmps;
+      if (!O->oMaxLag)
+        O->oMaxLag = IODACSAMPS;
+      csound->Warning(csound, Str(
+        "-b (%d) < ksmps (%d); increasing -b to %d and scaling -B to preserve b:B\n"),
+                      old_b, new_b, new_b);
+      O->oMaxLag = (int32_t) (((int64_t) O->oMaxLag * new_b) / old_b);
+      O->outbufsamps = new_b;
+      O->inbufsamps = new_b;
     }
      csound->ErrorMsg(csound, Str("audio buffered in %d sample-frame blocks\n"),
                     (int32_t) O->outbufsamps);
--- a/InOut/libsnd.c
+++ b/InOut/libsnd.c
@@ -70,7 +70,7 @@ static inline void spout_interleave(CSOUND *csound, int32_t scal) {
    nchk:
     /* if nspout remaining > buf rem, prepare to send in parts */
    if (spoutrem > (int32_t) csound->libsndStatics.outbufrem) {
-     end = csound->libsndStatics.outbufrem/nchnls;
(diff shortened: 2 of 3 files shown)


vlazzarini:
@ervoL Thanks for spotting this. I think we should not allow ksmps to be larger than the software buffer, I don't think it makes much sense to do so. We need to issue an error on csoundStart()

ervoL:
thanks @vlazzarini  for fast review/feedback! 


I would rather not error on csoundStart() for this. i think standalone Csound does not error either: it raises -b. an error would break some orchestras (ksmps=512 for FFT / heavy k-rate work) unless every host also sets -b, and it would make host mode stricter than -odac.

ksmps and -b are also different layers. ksmps is an orchestra design choice; -b is an I/O chunk and its default is platform-dependent. spout_interleave() already copies one k-cycle in several -b pieces (the nchk continuation). that path is just computing `end` as an absolute index instead of start-relative. I think the one-line fix is the right repair for that path.

if you still want the -odac invariant in host mode as well, i think that clamping -b up to ksmps (same as DAC) would match existing policy better than a new error.

vlazzarini:
@ervoL let's try clamping -b (and by consequence -B) to ksmps then and put out a warning for the moment. I don't like to encourage the practice of ksmps being more than the buffer size as it can be misleading/confusing. 

NB: for -B we can preserve the b:B ratio when we increase -b to match ksmps.

ervoL:
@vlazzarini  sounds good. Unless you want to remove it, I'll keep the spout_interleave bounds fix (that path is still wrong if it ever runs) and add a follow-up that clamps -b up to ksmps on start, with a warning, including host/embedded IO.

when -b is raised, I'll scale -B by the same factor so the original b:B ratio is preserved.

I'll push that as a second commit on this PR.

vlazzarini:
sounds good, yes - keep the fix in case we change our minds later regarding this.

ervoL:
@vlazzarini  I clamped -b up to ksmps on start, including host/embedded IO, and scaled -B by the same factor so the original b:B ratio is preserved, with a warning. I left the spout_interleave bounds fix here as we said. I also checked this in a VST3 host with ksmps 32, 512, and 1024 across DAW buffers from 128 to 2048, and those combinations were running fine.

vlazzarini (approved):
Thanks @ervoL. I've invited to the development team. You will be able to merge this yourself when the checks are done.

ervoL:
Thanks Victor! I’ll merge it once the checks are green.

vlazzarini:
you can now create and manage branches in the Csound repo. So I will suggest to use that for future PRs, which allows collaboration by other team members.

ervoL:
Ok — I’ll put the next bugfix on a branch in the Csound repo and open the PR from there.

PR #2892, opened 2026-09-11
Fold unary +/- literals so functional init(-1) type-checks

ervoL:
`k_temp = init(-1)` failed with `opcode 'init' for expression with arg types c returns out-args != 1`. `init(1)` and classic `k_temp init -1` already worked.

`convert_statement_to_opcall()` treated unary minus as an expression and type-checked `init` in isolation, so `resolve_opcode_get_outarg` matched a multi-out `init` overload. `handle_negative_number()` already folded unary-minus literals, but only later in `expand_statement()`.

This folds unary `+/-` numeric literals on every `T_FUNCTION` argument before that type-check. `init(0-1)` (binary minus) is unchanged and still out of scope.

Test: `OrcCompileTests.testFunctionalInitUnaryNumericLiteral`.

### DIFF
--- a/Engine/csound_orc_expressions.c
+++ b/Engine/csound_orc_expressions.c
@@ -1104,6 +1104,13 @@ static TREE *create_synthetic_label(CSOUND *csound, int32 count)
 
 void handle_negative_number(CSOUND* csound, TREE* root)
 {
+  /* Fold unary +/- of a numeric literal into a constant token so
+     k = init(-1) is the same shape as k = init(1). Otherwise the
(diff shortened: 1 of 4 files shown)


ervoL:
@vlazzarini  — this is the init(-1) functional-syntax bugfix I mentioned. I’d appreciate a review when you find a moment.

vlazzarini:
@coderabbitai review

hlolli:
good catch

hlolli (approved):

PR #2893, opened 2026-09-11
Fix gtadsr stage transitions and retriggering

hlolli:
`gtadsr` can start decay before attack reaches its peak. With amplitude 1 and a four-step attack, it outputs `0.25, 0.5, 0.75, 0.625` instead of reaching `1`. This affects the control-rate and both audio-rate forms.

- Count down each stage so attack reaches its peak and decay reaches sustain at their endpoints. A new gate during release starts a full attack from the current level.
- Check attack and decay durations before converting them to integer counts, preventing undefined conversions. Separate countdowns also avoid overflow from adding stage lengths. Preserve the existing one-step minimum for zero and short stages.
- End nonpositive release times immediately, so negative release cannot make the envelope grow.
- Share the sample update through `process()`, defined inside the class, and compute the release coefficient once per control block, removing `pow` from the audio loop.

### DIFF
--- a/Opcodes/pvsops.cpp
+++ b/Opcodes/pvsops.cpp
@@ -307,90 +307,91 @@ struct TVConv : csnd::Plugin<1, 6> {
 };
 
 
+// Keep the shared sample update free of function calls in the audio loop.
+#define GTADSR_TICK()                                                       \
+  do {                                                                      \
+    if (gate) {                                                             \
+      if (a > 0) {                                                          \
+        e = --a == 0 ? MYFLT(1) : e + ainc;                                 \
+      } else if (d > 0) {                                                   \
+        e = --d == 0 ? s : e + (s - 1) * dfac;                              \
+        if (e < s) e = s;                                                   \
+      } else {                                                              \
+        e = s;                                                              \
+      }                                                                     \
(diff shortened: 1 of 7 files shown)


Opcodes/pvsops.cpp
```diff
@@ -307,91 +307,92 @@ struct TVConv : csnd::Plugin<1, 6> {
 };
 
 
+// Keep the shared sample update free of function calls in the audio loop.
+#define GTADSR_TICK()                                                       \
```
vlazzarini: I think we can put this inside the class definition and the compiler will inline it (I think the C++ spec will
almost always guarantee that, so we can lean on it with cpp code). Could you do that - call it process() instead of GTADSR_TICK, which I prefer in terms of naming.

vlazzarini (changes requested):

Opcodes/pvsops.cpp
hlolli: @vlazzarini Moved the shared update into `Gtadsr::process(MYFLT s)`, defined inside the class, and replaced the macro calls.

vlazzarini (dismissed):

vlazzarini (approved):

PR #2895, opened 2026-09-11
Fix lufs power scaling and measurement state

hlolli:
`lufs` uses raw sample power, so `0dBFS` values other than 1 shift its readings. Its short-term calculation also carries the previous average into the next power sum.

- Normalize mono and stereo power by `0dBFS` squared before calculating loudness and gates.
- Start each short-term power sum from zero, removing the carried average and stale state after reinitialization.
- Require the absolute gate before applying the relative gate, avoid division by empty counts, and use the same channel sum for stereo gating and loudness.
- Process only active samples in partial blocks. Replace per-sample filter calls with a macro that keeps the same arithmetic.

Relative gating still uses the existing running mean and does not revisit earlier blocks when the threshold changes.


vlazzarini:
@gleb812 this one is yours. Could you review it?

vlazzarini (approved):
approving this as it looks ok to me and author did not respond

PR #2899, opened 2026-09-11
Fix exact breakpoint handling in bpf and bpfcos

hlolli:
`bpfcos` and audio/array forms of `bpf` can reject an input exactly on an interior breakpoint: the binary search selects the preceding segment, but evaluation expects the segment starting at that point.

- Make binary and cached searches agree at exact breakpoints, including the final segment.
- Stop i-rate evaluation after initialization fails, and reject empty or invalid point-array shapes before reading their elements, including after runtime resizing.
- Save the last segment in the scalar array form of `bpfcos`, avoiding repeated searches for values in the same segment.

Point-array checks run once per control block, outside audio sample loops.

### DIFF
--- a/Opcodes/emugens/emugens.c
+++ b/Opcodes/emugens/emugens.c
@@ -497,7 +497,8 @@ static int32_t bpfx_k(CSOUND *csound, BPFX *p);
 
 
 static int32 bpfx_i(CSOUND *csound, BPFX *p) {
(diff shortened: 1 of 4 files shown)


vlazzarini:
@gesellkammer please review.

gesellkammer (approved):

PR #2902, opened 2026-09-11
fix: correct pitchac correlation bounds and window state

hlolli:
`pitchac` has no manual entry. Its signature is `kpitch pitchac asig, kfmin, kfmax, iflow`. It estimates pitch by correlating successive sample windows; `iflow` sets the lowest frequency used to size its buffers, while `kfmin` controls the window length and `kfmax` limits accepted pitch estimates.

Fix `pitchac` correlation indexing and window state:

- Wrap the circular index at the buffer length, preventing a read past its end, and process only active note samples.
- Reset the peak search and update pitch for every completed window. This preserves the latest accepted result when several windows finish in one control block. Keep correlation maxima in `MYFLT` to avoid narrowing double-precision values.
- Validate frequency-derived lengths before integer conversion and allocation. Clamp runtime window lengths to the allocated capacity and reject values that would produce an invalid length.
- Register the opcode with its actual `PITCHAF` instance size.

### DIFF
--- a/Opcodes/pitchtrack.c
+++ b/Opcodes/pitchtrack.c
@@ -473,7 +473,11 @@ typedef struct _pitchaf{
 } PITCHAF;
 
 int32_t pitchafset(CSOUND *csound, PITCHAF *p){
-    int32_t siz = (int32_t)(CS_ESR/ (*p->iflow));
+    double samples = CS_ESR / (double)*p->iflow;
+    if (UNLIKELY(!(samples >= 1.0 && samples <= INT32_MAX &&
+                   samples <= SIZE_MAX / sizeof(MYFLT))))
+      return csound->InitError(csound, "%s", Str("pitchac: invalid lowest frequency"));
+    int32_t siz = (int32_t)samples;
     if (p->buff1.auxp == NULL || p->buff1.size < siz*sizeof(MYFLT))
       csound->AuxAlloc(csound, siz*sizeof(MYFLT), &p->buff1);
     else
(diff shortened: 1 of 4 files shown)


vlazzarini:
I can't find a manual page for this opcode.

Opcodes/pitchtrack.c
```diff
@@ -496,23 +500,34 @@ int32_t pitchafset(CSOUND *csound, PITCHAF *p){
...
     MYFLT *cor = (MYFLT *)p->cor.auxp;
     MYFLT *s = p->asig, pitch;
-    //MYFLT ifmax = *p->kfmax;
 
-    for (n=0; n < ksmps; n++) {
+    for (n=offset; n < ksmps; n++) {
       for (i=0,j=lag; i < len; i++) {
         cor[lag] += buff1[i]*buff2[j];
-        j = j != len ? j+1 : 0;
+        if (++j == len) j = 0;
```
vlazzarini, line 524: why this change?

hlolli: @vlazzarini When `j == len - 1`, the old expression sets `j` to `len`, so the next iteration can read `buff2[len]`, past the end of the buffer. Incrementing first and wrapping at `len` keeps the index within `0..len-1`.

hlolli:
@vlazzarini There is no manual entry for `pitchac`. Its signature is `kpitch pitchac asig, kfmin, kfmax, iflow`; it estimates pitch by correlating successive sample windows. `iflow` sets the lowest frequency used to size its buffers.

Opcodes/pitchtrack.c
vlazzarini: ok, so the correct fix is to set the check to j != len - 1 or j < len - 1 instead of rewriting the whole line.

hlolli: @vlazzarini Changed it to `j = j != len - 1 ? j+1 : 0;`, keeping the original expression and correcting only the bound.

vlazzarini (approved):
Can we have a full man page with example for this too?

hlolli:
@vlazzarini Opened https://github.com/csound/manual/pull/798 with a full `pitchac` page, parameter descriptions, update behavior, and a standalone sine-wave example.

PR #2910, opened 2026-09-11
fix: correct pvsbin output rates and bin selection

hlolli:
Correct `pvsbin` output dispatch and bin selection:

- Register separate control-rate and audio-rate forms so audio outputs use the sample-processing function.
- Keep bin selection at the PVS frame rate for frame-based streams. Hold the selected amplitude and frequency between frames and fill audio outputs on every block.
- For sliding streams, read each active sample at audio rate and the first active sample at control rate. Clear inactive audio samples.
- Check bin bounds before integer conversion, and truncate the bin number before locating its amplitude/frequency pair. This prevents fractional indices from selecting mismatched values or reading past the frame.

### DIFF
--- a/Opcodes/pvsbasic.c
+++ b/Opcodes/pvsbasic.c
@@ -1165,8 +1165,21 @@ static int32_t pvsoscprocess(CSOUND *csound, PVSOSC *p)
 }
 
 
+static int32_t pvsbinset(CSOUND *csound, PVSBIN *p)
+{
+  IGN(csound);
+  p->lastframe = 0;
+  p->amp = p->freq = FL(0.0);
+  return OK;
+}
+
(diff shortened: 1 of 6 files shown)


vlazzarini (changes requested):
This changes behaviour - bin numbers should only be updated at the PVS rate (not the control rate). We can't have that. If the new behaviour is to be implemented, we need a mode parameter to select it.

hlolli:
@vlazzarini Restored bin selection at the PVS frame rate for frame-based streams. Both output forms hold the selected values between frames, so changing `kbin` no longer takes effect early.

vlazzarini (dismissed):

vlazzarini (approved):

PR #2916, opened 2026-09-11
fix: align modmatrix buffers and validate table dimensions

hlolli:
`modmatrix` placed floating-point buffers after integer maps, leaving doubles misaligned when the combined map length was odd. Group the floating-point buffers before the maps so each stays aligned.

- Check dimensions before integer conversion, require tables large enough for the matrix and vectors, and calculate allocation sizes without overflow.
- Index matrix rows directly so column traversal never forms pointers beyond the array.
- Avoid copying the parameter table onto itself when it also serves as the result table.
- Treat any nonzero update flag as a request to read the live matrix, as the [manual specifies](https://csound.com/docs/manual/modmatrix.html). Negative flags previously left the cached matrix in use.


vlazzarini:
@Oeyvind could you take a look?

Oeyvind:
I checked with Thom, and it looks good. I am unable to test immediately as I dont't have a build system set up at the time. Are there any automatic builds done somewhere?

rorywalsh:
If you log into github, you'll see beta development builds in the actions
page:
https://github.com/csound/csound/actions/workflows/csound_builds.yml

vlazzarini:
@Oeyvind specifically for this PR - https://github.com/csound/csound/actions/runs/34957551873?pr=2916#artifacts
but you need to be logged in to github as @rorywalsh  mentioned

PR #2920, opened 2026-09-11
fix: correct unwrap input and phase history

hlolli:
`unwrap` processed the output array without first reading its input, so a separate output could contain zeros or stale phases. Read the input directly while preserving in-place use.

- Allocate phase history as `MYFLT`, matching how it is accessed. The old float-sized allocation was too small in double builds.
- Unwrap each phase relative to its previous output with the correct correction direction, and reset history on initialization.
- Keep the arithmetic wrap/unwrap loops and the default `[-pi, pi)` interval, without a phase helper or remainder calls.
- Use a dedicated opcode struct, validate the mode, and reject array shape changes before accessing history.

### DIFF
--- a/Opcodes/arrays.c
+++ b/Opcodes/arrays.c
@@ -1077,36 +1077,60 @@ static int32_t shiftout_perf(CSOUND *csound, FFT *p) {
 }
 
 
-static int32_t unwrap_set(CSOUND *csound, FFT *p) {
+typedef struct {
+  OPDS h;
+  ARRAYDAT *out, *in;
+  MYFLT *mode;
+  AUXCH mem;
+  int32_t size, unwrap;
+} UNWRAP;
+
+static int32_t unwrap_set(CSOUND *csound, UNWRAP *p) {
   if(p->in->sizes == NULL)
     return csound->InitError(csound, "array not initialised\n");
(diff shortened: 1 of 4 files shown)


Opcodes/arrays.c
```diff
@@ -1077,36 +1077,60 @@ static int32_t shiftout_perf(CSOUND *csound, FFT *p) {
...
+  int32_t i;
+  MYFLT *in = p->in->data;
   MYFLT *phs = p->out->data;
-  if(mode == 0) { // wrap
-  for (i=0; i < siz; i++) {
-    while (phs[i] >= PI) phs[i] -= TWOPI;
-    while (phs[i] < -PI) phs[i] += TWOPI;
-  }
-  } else {  // unwrap
+  if (!p->unwrap) {
```
vlazzarini, line 1117: Leave the original code here without function calls. Don't make this replacement.

vlazzarini (changes requested):

Opcodes/arrays.c
hlolli: @vlazzarini Restored the arithmetic wrap/unwrap loops and removed the helper and `fmod` calls. The input and phase-history fixes remain.

vlazzarini (approved):

PR #2924, opened 2026-09-11
fix: honor ftconv impulse length and partition layout

hlolli:
`ftconv` used the requested impulse length to choose its partition count, but still read later table samples into the last partition. Zero-pad beyond the requested end so those samples no longer affect the output.

- Preserve state on skipped initialization only when channel count, partition size, and partition count match. Different layouts can need the same allocation size and must still rebuild their buffers and FFT setup.
- Check rounded parameters before integer conversion, calculate impulse ranges with wider integers, and bound allocation sizes and signed FFT indices before multiplication.
- Keep negative skip values as leading silence and handle multichannel impulse lengths in sample frames.

### DIFF
--- a/Opcodes/ftconv.c
+++ b/Opcodes/ftconv.c
@@ -112,19 +112,6 @@ static void multiply_fft_buffers(MYFLT *outBuf, MYFLT *ringBuf,
     } while (--nPartitions);
 }
 
-static inline int32_t buf_bytes_alloc(int32_t nChannels,
-                                      int32_t partSize, int32_t nPartitions)
-{
-    int32_t nSmps;
-
-    nSmps = (partSize << 1);                                /* tmpBuf     */
-    nSmps += ((partSize << 1) * nPartitions);               /* ringBuf    */
-    nSmps += ((partSize << 1) * nChannels * nPartitions);   /* IR_Data    */
-    nSmps += ((partSize << 1) * nChannels);                 /* outBuffers */
-
-    return ((int32_t) sizeof(MYFLT) * nSmps);
-}
-
 static void set_buf_pointers(FTCONV *p,
                              int32_t nChannels, int32_t partSize,
                              int32_t nPartitions)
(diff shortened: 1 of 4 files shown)


Opcodes/ftconv.c
```diff
@@ -149,99 +136,87 @@ static void set_buf_pointers(FTCONV *p,
...
         }
-        /* pad second half of IR to zero */
-        for (k = p->partSize; k < (p->partSize << 1); k++)
+        for (k = partSize; k < partSize * 2; k++)
           p->IR_Data[j][n + k] = FL(0.0);
-        /* calculate FFT */
-        csound->RealFFT(csound, p->fwdsetup, &(p->IR_Data[j][n]));
-        n -= (p->partSize << 1);
+        csound->RealFFT(csound, p->fwdsetup, &p->IR_Data[j][n]);
+        n -= partSize * 2;
```
vlazzarini: why change the left shift to multiplication by 2?

Opcodes/ftconv.c
```diff
@@ -149,99 +136,87 @@ static void set_buf_pointers(FTCONV *p,
...
+        for (k = 0; k < partSize; k++, frame++) {
+          if (frame >= 0 && frame < tableFrames && frame < endFrame)
+            p->IR_Data[j][n + k] = ftp->ftable[frame * nChannels + j];
           else
             p->IR_Data[j][n + k] = FL(0.0);
-          i += p->nChannels;
         }
-        /* pad second half of IR to zero */
-        for (k = p->partSize; k < (p->partSize << 1); k++)
+        for (k = partSize; k < partSize * 2; k++)
```
vlazzarini: why change left shift to multiplication by 2?

vlazzarini (changes requested):

Opcodes/ftconv.c
hlolli: @vlazzarini This rewrite wasn’t needed for the fix. I restored `<< 1`; the partition-size check keeps the shift within range.

Opcodes/ftconv.c
hlolli: @vlazzarini Restored `<< 1` here too. Multiplication wasn’t needed for the fix.

vlazzarini (approved):

PR #2925, opened 2026-09-11
fix: correct liveconv clear and reload state

hlolli:
Clearing `liveconv` resets the ring-buffer indices but leaves the input pointer on the old partition. Select that pointer after the reset so the next samples enter the correct partition.

Honor the documented flags: any nonzero `kclear` clears the buffers, while only `kupdate = 1` or `-1` starts a load or unload. Fractional values no longer get rounded into unintended commands.

Validate partition sizes before integer conversion and bound buffer sizes to prevent overflow. Stop processing if the impulse-response table is missing.


vlazzarini:
@Oeyvind we need you to have a look at this.

ssaue:
I will take a closer look: I just need to get my brain back into it

PR #2926, opened 2026-09-11
fix: correct PVS buffer spectrum and circular reads

hlolli:
`pvsbufread` treats bin numbers as float offsets, so it reads the wrong frequency range and omits the upper spectrum. Read complete amplitude/frequency pairs through the selected final bin, including Nyquist.

Both readers wrap over `frames - 1`, dropping the last stored frame and looping forever for a one-frame buffer. Wrap over the full buffer and interpolate from the last frame back to the first, with bounded handling of negative and large time positions.

Reject invalid buffer lengths and handles, and stop on missing or short delay tables before reading them. Reset reused output frames and cache handle changes to avoid repeated lookups.

### DIFF
--- a/Opcodes/pvsbuffer.c
+++ b/Opcodes/pvsbuffer.c
@@ -52,6 +52,18 @@ static int32_t pvsbufferset(CSOUND *csound, PVSBUFFER *p)
 
     if (UNLIKELY(p->fin->sliding))
       return csound->InitError(csound, "%s", Str("SDFT case not implemented yet"));
+    N = p->fin->N;
+    hop = p->fin->overlap;
+    if (UNLIKELY(N < 2 || N > INT32_MAX - 2 || (N & 1) || hop <= 0 ||
+                 (uint64_t) N + 2 > SIZE_MAX / sizeof(float)))
+      return csound->InitError(csound, "%s", Str("pvsbuffer: invalid frame size"));
+    double frames = (double) *p->len * CS_ESR / hop;
+    size_t frameBytes = ((size_t) N + 2) * sizeof(float);
+    if (UNLIKELY(!(frames >= 1 && frames <= UINT32_MAX &&
+                   frames <= SIZE_MAX / frameBytes)))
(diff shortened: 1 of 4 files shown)


vlazzarini (commented):
I think I will close this. There are too many changes and I am not convinced there is a problem here that requires such a lot of changes. I have used this extensively and have not identified any major problems.

The best thing is that if there is an issue that really needs addressing, a ticket is open and I can look at it.

hlolli:
@vlazzarini Understood. We could open up to three separate tickets, or fewer if you think some belong together:

1. **Frequency range in `pvsbufread`:** bin numbers are used as array positions, although each bin contains two values. At 48 kHz, the default read covers roughly 0–12 kHz rather than 0–24 kHz, and explicit frequency ranges are also shifted. How audible this is depends on the source.
2. **Circular reads in both readers:** wrapping excludes the last stored frame, shortening the loop by one spectral frame and changing interpolation at the boundary. This may be subtle with long buffers, but matters more with short ones. A one-frame buffer can hang.
3. **Invalid delay tables in `pvsbufread2`:** a missing table can be dereferenced, and a short table reports an error but continues reading beyond its end. These concern invalid inputs rather than normal use.

You may not consider these significant for your usual use, but we can report them separately so you can assess each without the broader changes in this PR.

vlazzarini:
please open one PR for the three. I'll deal with it then, thanks.

hlolli:
@vlazzarini Split this into three separate PRs: #2961 fixes frequency-bin selection, #2962 fixes circular frame reads, and #2963 stops on missing or short delay tables. Closed the combined PR #2959.

vlazzarini:
Sorry I meant open one issue, not PR. I want to do the fixes myself, as I know the code well.

hlolli:
@vlazzarini Understood. Opened https://github.com/csound/csound/issues/2966 with all three concerns and closed the three PRs. I’ll leave the fixes to you.

PR #2927, opened 2026-09-12
fix(pvoc): add opt-in analysis-rate file timing

hlolli:
Add a trailing optional `ifiletime` argument to `pvoc`, `pvadd`, `pvread`, `pvbufread`, `pvinterp`, and `pvcross`, for both string and numeric filenames.

The default, 0, preserves the output-rate time mapping used by existing scores. Set it to 1 to convert the time pointer using the analysis file’s sample rate, so it addresses seconds in the source recording when input and output rates differ.


vlazzarini (changes requested):
This changes behaviour, users may have already accounted for this. We need an optional mode parameter to introduce this.

hlolli:
@vlazzarini Added optional `ifiletime` to all six readers, defaulting to 0 so existing calls retain their time mapping. Setting it to 1 uses the analysis file’s sample rate. The companion manual PR documents both modes.

vlazzarini (approved):

PR #2928, opened 2026-09-12
fix(phasorbnk): preserve bank state across resizing

hlolli:
Preserve retained bank phases when negative `iphs` skips initialization during resizing, and initialize newly enabled banks to zero. Track the current bank count so removed indices cannot access retained storage.

Validate bank sizes and indices, and normalize an initial phase of 1 to 0. Frequency updates keep their existing behavior.

### DIFF
--- a/Opcodes/pitch.c
+++ b/Opcodes/pitch.c
@@ -1185,43 +1185,74 @@ int32_t phsbnkset(CSOUND *csound, PHSORBNK *p)
     int32_t    n, count;
     double  *curphs;
 
-    count = (int32_t)MYFLT2LONG(*p->icnt);
-    if (UNLIKELY(count < 2))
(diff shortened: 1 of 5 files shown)


vlazzarini (changes requested):
Same thing as in phasor: updates bigger than one cycle would indicate aliasing, wouldn't they?It's user error and I don't think we need a change unless there is a documented correct real use showing there is a problem.

hlolli:
@vlazzarini Removed the large-increment changes. The remaining issue occurs at ordinary rates: growing a bank with `iphs=-1` clears phases that should survive skipped initialization. The fix preserves those phases, starts new banks at zero, and checks indices against the current bank count.

vlazzarini (approved):

PR #2935, opened 2026-09-12
fix(metro2): add opt-in corrected timing

hlolli:
Keep the original metro2 timing by default, including startup offbeats and the offset between the two clocks.

Add a trailing optional `icorrect` argument. Set it to 1 to normalize startup phase from the first control-rate swing value, advance both clocks together, and wrap complete periods after large frequency changes. Existing calls use `icorrect=0` and retain their timing.

### DIFF
--- a/Opcodes/metro.c
+++ b/Opcodes/metro.c
@@ -34,9 +34,9 @@ typedef struct {
 // METRO2 ADDED BY GLEB ROGOZINSKY Oct 2019
 typedef struct {
         OPDS    h;
-        MYFLT   *sr, *xcps, *kswng, *iamp, *iphs;
+        MYFLT   *sr, *xcps, *kswng, *iamp, *iphs, *icorrect;
         double  amp2, curphs, curphs2, swng_init;
-        int32_t flag;
+        int32_t flag, flag2;
 } METRO2;
 //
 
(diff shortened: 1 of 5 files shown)


vlazzarini:
@hlolli maybe users are relying on this extra beat? This is behaviour change - I think.

hlolli:
maybe calls for metro3

hlolli:
@vlazzarini Yes, a score could depend on that first offbeat. I’ve kept the original timing as the default and added trailing `icorrect=1` for corrected timing. That lets us keep this in `metro2` without changing existing sequences. A separate `metro3` could focus on sample-accurate triggers, explicit reset/run controls, BPM subdivisions, and beat/phase outputs.

vlazzarini:
I agree on metro3 but only if it really has something extra to give, otherwise we may as well have an icorrect parameter or sonething.

vlazzarini (dismissed):

hlolli:
@vlazzarini Agreed. `metro2` now has a trailing `icorrect` parameter: `0` keeps the existing timing and `1` selects corrected timing. I’ll keep this PR focused on that.

vlazzarini:
thanks: could you get a PR for the manual too?

hlolli:
@vlazzarini Opened https://github.com/csound/manual/pull/799 with the `icorrect` syntax, default compatibility behavior, and an example showing corrected timing from a nonzero initial phase.

vlazzarini (approved):

PR #2939, opened 2026-09-12
fix(phasor): keep rounded float phase within range

hlolli:
In float builds, rounding a phase just below one before the wrap check can subtract a cycle too early, producing a small negative output at ordinary frequencies. Control-rate output can also round to one.

Keep the internal wrap comparisons in double precision and map output values that round to one to zero.

### DIFF
--- a/OOps/ugens2.c
+++ b/OOps/ugens2.c
@@ -41,15 +41,13 @@
 
 int32_t phsset(CSOUND *csound, PHSOR *p)
 {
-  double phs = (double)*p->iphs;
-  if (UNLIKELY(!isfinite(phs)))
-    return csound->InitError(csound, "%s", Str("phasor: initial phase must be finite"));
-  if (phs >= 0.0) {
-    if (UNLIKELY(phs >= 1.0)) {
+  MYFLT       phs;
+  int32_t  longphs;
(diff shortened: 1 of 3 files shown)


vlazzarini (changes requested):
Increments larger than one cycle would cause aliasing, wouldn't they?  I don't think we need this change to deal with user error. Unless there is a correct real use case where phasor can't do what it is supposed to.

I am mindful of preserving the code of longstanding opcodes like this one - it's been there for 40+ years.

hlolli:
@vlazzarini Agreed about aliasing; I’ve removed the large-increment and initialization changes. The PR now only fixes float rounding: at 440 Hz and 44.1 kHz, a valid initial phase can produce a tiny negative output (about −1.7e−8) at the wrap. The wrap check rounds the accumulator too early. The fix keeps that check in double precision and maps an output that rounds to 1 back to 0. This is a small phase-range correction, not a claim of a major audible problem.

vlazzarini (dismissed):

vlazzarini (dismissed):

vlazzarini (dismissed):

vlazzarini (approved):

PR #2957, opened 2026-09-12
Fix lphasor direction across multiple loop reflections

hlolli:
In forward/backward mode, `lphasor` loses the correct position and direction when a step crosses multiple loop boundaries. For a 0–4 loop and a step of 6, it produces `0, 2, 0, 2` instead of `0, 2, 4, 2`.

Reduce the excess distance over a full forward/backward cycle, then apply the remaining reflections. Also remove the integer casts used to count whole loops in the one-way modes, avoiding out-of-range conversions for large steps.

### DIFF
--- a/OOps/oscils.c
+++ b/OOps/oscils.c
@@ -168,15 +168,31 @@ int32_t lphasor(CSOUND *csound, LPHASOR *p)
       phs += (p->dir ? trns : -trns);
       if (loop_mode) {
         dir = (trns < 0.0 ? !(p->dir) : p->dir);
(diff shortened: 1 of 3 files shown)


vlazzarini (changes requested):
isn't this the same as in the phasor case? A step should not be more than one cycle, otherwise it aliases?

hlolli:
@vlazzarini Yes—the failing examples step farther than the loop span, so they already alias. I haven’t found a normal playback case that needs this change, so I’m closing the PR.

PR #2959, opened 2026-09-12
Fix PVS buffer bin ranges, circular reads and delay table errors

hlolli:
Addresses the three fixes requested in https://github.com/csound/csound/pull/2926#issuecomment-5644919454 with small changes to the existing routines.

- `pvsbufread` converts bin numbers to amplitude/frequency pair offsets, so the default range includes the full spectrum through Nyquist and explicit ranges select the intended bins.
- Both readers wrap across the full stored frame count, including the last frame and its interpolation back to the first.
- `pvsbufread2` stops when either delay table is missing or too short.


hlolli:
Split into three separate PRs: #2961 for frequency-bin selection, #2962 for circular frame reads, and #2963 for delay-table errors.

PR #2967, opened 2026-09-12
fix: correct mton octave and cent formatting

hlolli:
`mton` labels low MIDI notes an octave too high: MIDI 1 returns `0C#` instead of `-1C#`, and 11.75 returns `1C-25` instead of `0C-25`.

Derive the octave and pitch class from the whole note, including negative notes, and carry rounded cents into the next note without leaving a `-0` suffix. Format signed and multi-digit octaves correctly.

Grow short output strings before writing the name, allow room for the octave and cents, and reject values outside the integer conversion range. This prevents invalid pitch-class indexing and writes past the string buffer. Both init- and control-rate variants use the same conversion.


vlazzarini:
@gesellkammer one of yours, could you review?

gesellkammer (dismissed):

vlazzarini (dismissed):

vlazzarini (approved):

PR #2991, opened 2026-09-13
Preserve fractional skip times in ftsamplebank

hlolli:
`ftsamplebank` converts its skip time to an integer before loading files, so `.25` seconds becomes zero. Keep the skip time as `MYFLT` for both initial loads and triggered reloads.

Write GEN01 score events with enough precision to preserve the skip position, including values near sample-rounding boundaries. Use a fixed decimal-point locale so host locale settings cannot change the score syntax.


vlazzarini:
@rorywalsh one of yours, can you review?

rorywalsh (approved):
Looks good to me.

PR #3007, opened 2026-09-14
Fix specaddm multiplier and documented default

hlolli:
`specaddm` never initialized its cached multiplier, so it ignored the second spectrum regardless of `imul2`. Copy the multiplier at initialization so the output follows the [documented formula](https://csound.com/docs/manual/specaddm.html): `mag1 + mag2 * imul2`.

The manual specifies a default of zero, but the opcode registered a default of one. Use zero to match the manual and preserve the existing output when callers omit the argument.

Also refresh the output's source metadata on every initialization, even when its buffer size stays the same.


vlazzarini:
just to say this is marked as deprecated so I am
unsure we ought to put any effort on it.

hlolli:
@vlazzarini Understood. We could leave `specaddm` unchanged and add a code comment explaining its known behavior and why we preserve it, with a link to this discussion.

Would a successor be useful for the constant-Q analysis use case? Beyond correcting the addition, it could offer independent k-rate gains for both inputs, per-band weights, and clear rules for combining frames and handling magnitude versus dB values. I’d only propose it if it fills a gap in the existing spectral opcodes.

vlazzarini:
I don't think these opcodes ever fulfilled their promise - but maybe because they were just too awkward to use and were mostly designed to extract control information from inputs. Let's fix it and then kick the deprecation can down the road, maybe we can review if the set needs to be deprecated or just left alone.

vlazzarini (approved):

PR #3008, opened 2026-09-14
Fix spectrum sample processing and note history

hlolli:
`spectrum` decremented an unsigned sample offset after it reached zero, then replaced all remaining samples in each audio block with zeros. A 2048 Hz sine at 32768 Hz could therefore disappear from the analysis with `ksmps=16`.

Process every active audio sample and pad inactive samples with zeros. Keep audio sample offsets out of the control-rate path, which consumes one value per cycle. This restores the [documented support for both audio and control inputs](https://csound.com/docs/manual/spectrum.html).

Clear the analysis buffers at initialization so reused notes do not inherit samples from the previous note.


vlazzarini:
similarly - it's deprecated.

hlolli:
@vlazzarini Understood. This affects the same deprecated analysis chain as `specaddm`. Here, the audio path effectively keeps only one sample per block, so even a steady tone can disappear from the analysis; reused notes can also inherit the previous note’s history.

If we want to preserve that behavior, I can document these limitations and the reason for leaving the code unchanged. Any successor would need to address `spectrum` and its consumers together, so perhaps we should first establish whether a supported constant-Q replacement is wanted.

vlazzarini (approved):
let's merge it but then do not spend more time fixing other deprecated opcodes.

PR #3009, opened 2026-09-14
Fix hrtfer data loading and sample handling

hlolli:
`hrtfer` byte-swapped the shared `HRTFcompact` buffer on every initialization, so later notes or instances could corrupt the data used by earlier ones. Convert byte order once when loading the file, and reject missing or incomplete data.

Correct the elevation and azimuth record offsets so each direction reads the right measurements. Handle partial audio blocks consistently, clear inactive and unfilled output, and preserve unread input when an output reuses its buffer.

Preserve this deprecated opcode's legacy gain and block timing; the [manual](https://csound.com/docs/manual/hrtfer.html) recommends `hrtfstat` for new use.


vlazzarini:
also deprecated.

vlazzarini (approved):

PR #3011, opened 2026-09-14
Fix bob solver and filter state

hlolli:
`bob` starts its first RK4 derivative from uninitialized temporary data and uses a half step for the fourth derivative. Start from the saved filter state and use the full fourth step.

Return the fourth filter stage, matching [Pure Data’s correction](https://github.com/pure-data/pure-data/blob/master/extra/bob~/bob~.c). This changes the audible response from the first stage to the intended four-pole output; Csound’s existing `tanh` saturation curve stays intact.

Keep parameter inputs untouched, honor nonzero `istor` as [documented](https://csound.com/docs/manual/bob.html), and retain state at zero saturation without dividing by zero. Check the oversampling count before converting it to an integer. The derivative calculation uses a macro to avoid helper calls inside the sample loop.


vlazzarini:
@gleb812 another one of yours, could you review?

vlazzarini (approved):

PR #3021, opened 2026-09-14
Fix fmvoice shared state layout and modulation

hlolli:
`fmvoice` duplicates the shared FM state layout but omits its `opt` member, so shared routines access the wrong fields. Embed `FM4OP` directly and use that member throughout.

Initialize vibrato phase and rate, set formants from the first note’s controls, and apply the current vibrato factor to all four oscillators. Keep table-rate conversion, control reads, and output scaling outside the sample loop, with the sample helper inlined.

Bound vowel and gain-table indices and check frequency conversions at control rate. Preserve the existing four vowel banks (0–127), although the [manual currently lists 0–64](https://csound.com/docs/manual/fmvoice.html). Keep the existing spectral-tilt mapping.


vlazzarini (approved):

vlazzarini:
@hlolli manual needs adjustment.

hlolli:
@vlazzarini Opened https://github.com/csound/manual/pull/802 to document the full 0–127 vowel range and all four formant scales, correct the timbre-control range, and clarify vibrato depth and rate. The example now includes the same vowel from each bank.

PR #3027, opened 2026-09-14
Mark deprecated opcode policy at registrations and frozen functions

hlolli:
Csound marks deprecated opcodes in several ways, often without telling users what to use instead or telling developers which historical behavior must remain. Add replacement guidance and a maintenance policy for 209 deprecated names and renamed aliases.

Place `CSOUND_DEPRECATED_OPCODE(name, replacement, policy, note)` beside opcode registrations. Mark frozen initialization and processing functions with `CSOUND_PRESERVE_LEGACY_BEHAVIOR(name)` and an explicit comment requiring a maintainer decision before changing historical behavior. The policy distinguishes renamed aliases, legacy implementations, and known incorrect behavior retained for compatibility.

Generate `H/opcode_deprecations.def` and the reference catalog from those local descriptors. Reject missing local descriptors, conflicting policies, missing frozen-function markers, and stale generated files. Both annotation macros expand to nothing and leave opcode processing and the plugin ABI unchanged.

Compiler warnings name replacements when known. Both `_QQ` and the newer deprecation field honor `--error-deprecated` and warning suppression, and opcode listings report both consistently. Add missing source flags for five names already deprecated in the manual: `ptablew`, `sclag`, `sclagud`, `scphasor`, and `sctrig`. Also mark the audio-array form of `fin`, which the old manual includes in its deprecated list. The audio-to-array form of `array` remains supported. Where no direct replacement is documented, say so; FFT-based PVS opcodes do not directly replace the constant-Q `w`-signal chain.


vlazzarini:
Thanks. It may be worth looking at the old manual list of deprecated opcodes to make sure they are all properly marked in the sources.

hlolli:
@vlazzarini The old manual list exposed one more missing flag: `fin`'s audio-array form. I've marked it too. The other surviving entries are covered; the older names that no longer have source registrations need no new markers.

vlazzarini:
I'd like @kunstmusik and @rorywalsh to take a look

rorywalsh:
This looks good to me, and make the opcode registration a lot clearer, in terms of what has been deprecated.

kunstmusik:
Reviewed with AI and it raised a couple points:


must fix | Android ndk-build link breakage. The checked-in stale parser Android/CsoundAndroid/jni/csound_orcparse.c:2591 still calls the deleted query_deprecated_opcode(). CI's Android jobs use CMake (regenerated parser), so they stay green — anyone building via Android.mk hits an undefined symbol. Regenerate the parser or keep a stub; leftover declarations in H/csound_orc_compile.h:41 should go too.
-- | --
intended? | Stricter --error-deprecated. _QQ-only names (ptable, tabgen, ins/outs…) previously only warned; now they hard-error. And -m1024 now suppresses warnings it previously couldn't. Both documented in the PR — worth a release note since lint-style scores may newly fail.

Overall seems like a step forward.

hlolli:
@kunstmusik Fixed the Android path: `ndk-build` now regenerates the local parser when the grammar changes, and I removed both obsolete declarations.

The stricter `--error-deprecated` behavior is intended: opting into it should reject deprecated opcodes regardless of whether they use `_QQ` or the newer field. Without that option, they still compile. `-m1024` suppresses warnings but does not cancel an explicit request for errors. I added a release note for this change.

hlolli:
Btw, did we have release notes anywhere at all already? Good time to start collecting it before release?

vlazzarini:
No we don't. In the past we relied on going through commits (actually John did). It's such a big task, maybe we should use an LLM for it.

vlazzarini:
are we good? @kunstmusik can you approve and merge?

kunstmusik (approved):

PR #3028, opened 2026-09-14
Discover command-line tests and keep expectations in CSD files

hlolli:
Adding a command-line test currently means editing the shared list in `test.py`, which causes frequent merge conflicts. Discover `.csd` files recursively instead, with expectations and special arguments declared in each file.

TOML metadata goes at the top of the CSD, outside `<CsoundSynthesizer>`:

```csound
<CsTest>
description = "reject an invalid input"

[expect]
exit = "nonzero"
stderr = ["expected diagnostic"]
</CsTest>
<CsoundSynthesizer>
...
</CsoundSynthesizer>
```

Csound ignores this preamble; the runner reads it without an engine parser change. Files without metadata expect success. Output checks support substrings and regular expressions on stdout or stderr. Use `output` or `output_regex` for text that may appear on either stream, such as ordinary Csound messages whose routing differs by platform. Expected failures must match a stderr diagnostic; signals, Windows exception statuses, timeouts, and runner errors cannot pass.

This is intended to replace #2638. It includes that PR's separate managed-input/output failure cases and error-channel routing, while moving expectations out of Python. Descriptions, arguments, stack limits, and WebAssembly exceptions now live in each CSD. `--list` and `--test` support browsing and selecting tests. Python before 3.11 needs `tomli`; newer versions use the standard library.

Manual fixtures have visible skip reasons. The existing UDO recursion case also has an explicit skip because it exits by signal, which the old runner incorrectly counted as a pass. Runs use a temporary fixture copy so generated files do not change the checkout.


vlazzarini:
@hlolli can we add the metadata outside the CSD tags? Csound ignores everything that is not inside CsoundSynthesizer.

That is simpler and cleaner. We can just place the json at the the top of each file, no need for comment syntax. We don't modify the compiler, leaving tests independent, which I'd prefer.

If we must have it in the engine, then we should use an option section in the CSD, instead of placing it in a comment.

```
<CsTest>
</CsTest>
```

vlazzarini (approved):

PR #3033, opened 2026-09-15
Prevent circular buffer size overflow

hlolli:
`csoundCreateCircularBuffer()` calculates `(numelem + 1) * elemsize` with signed 32-bit arithmetic. For example, 65,536 elements of 65,536 bytes wrap to a 65,536-byte allocation; writing the second element then overflows the heap.

Validate the dimensions before allocation and use `size_t` for the allocation size and byte offsets. This preserves valid large buffers on platforms that can represent them while rejecting invalid dimensions and sizes that cannot be represented. Free the buffer header if allocation of its storage fails, and return zero for non-positive transfer counts as the API contract requires.


vlazzarini:
this is used extensively in the rt audio backends. How can we ensure the changes do not break them?

hlolli:
@vlazzarini Fair concern. I've kept this PR focused on size overflow and moved the synchronization changes into #3035. That keeps the existing locking and index updates intact here, while preserving buffer capacity, sample order and partial-read/write behavior for valid sizes. We can then assess the concurrency change with threaded tests and real-time backend runs before merging it.

vlazzarini (approved):

PR #3035, opened 2026-09-15
Synchronize circular buffer index reads and writes

hlolli:
The circular buffer writes its read and write indexes atomically, but reads them through plain accesses under a lock that the updates never take. A reader and writer can therefore access the same index without matching synchronization.

Use the existing atomic macros for index reads and writes, and remove the lock around the space check. Publishing the write index follows the sample copies; publishing the read index follows consumption. Buffer capacity, sample order, and partial-transfer behavior stay the same.

Document that each buffer supports one reader and one writer. Peeking and flushing share the reader's role, and destruction requires both sides to stop. Cross-thread use requires a build with atomic operations enabled.

This contains the synchronization changes separated from #3033 so they can receive their own review. The size-overflow fix from #3033 has merged into develop; this PR is rebased on it.


vlazzarini:
@hlolli is this superseded by #3238

PR #3036, opened 2026-09-15
Fix instance performance argument handling

hlolli:
Calling `perf()` with optional p-fields on a newly created instance reads the first value from `args[1]`, although the first optional value is stored in `args[0]`. A call such as `perf(voice, 440)` therefore dereferences a NULL pointer during automatic initialization and crashes. Start the copy at `args[0]`, then continue updating those p-fields on later control cycles.

The `play()`, `init()`, and automatic `perf()` initialization paths also allocate their event data as `PMAX + 1` bytes and treat it as an array of `MYFLT`. Use actual `MYFLT` arrays so larger argument lists cannot overwrite the stack, and fully initialize the event timing fields.

Return the performance error for a NULL instance and release an instance when `play()` cannot initialize it, instead of reporting success or retaining the failed instance until reset.


vlazzarini:
Can you make this into a PR so I can look at the issue and fix it manually?

PR #3056, opened 2026-09-16
Fix lastcycle timing without extending notes

hlolli:
`lastcycle` used to add a release cycle to notes that had none. It also cached timing before later init statements could change `p3` or `xtratim`, so it could report the last cycle too early.

Read the owning instrument’s scheduled end time without changing its duration or release. This follows later note and release extensions, early turnoff with release, and calls inside nested user-defined opcodes. Account for partial final blocks when the opcode uses a smaller local `ksmps`.

An indefinite note stopped without a release has no scheduled final cycle to report. `lastcycle` leaves its duration unchanged rather than adding a cycle to produce a pulse.


vlazzarini:

gesellkammer (changes requested):
The tests look too convoluted to really check if they are actually testing the different use cases.

hlolli:
@gesellkammer I split the test into separate cases for release extension, delayed starts, `p3` changes, finite and indefinite turnoff, and instance reuse. Each now states its expected final cycle and checks both `lastcycle` placements directly, without the array slots or flag-driven setup.

tests/commandline/test_lastcycle_delayed_start.csd
```diff
@@ -0,0 +1,48 @@
...
+gkFinalCycle init 0
+
+instr 1
+  kBefore lastcycle
+  ; No explicit release: lastcycle must keep its one-cycle minimum.
+  xtratim 0
+  kAfter lastcycle
+
+  ; Both placements must produce 0 on every cycle except the final one.
+  kCycle timeinstk
```
gesellkammer: timeinstk is (documented) always off by 1, starts at 1. eventcycles() shows the correct value

```csound

<CsoundSynthesizer>
<CsOptions>
--no-sound
</CsOptions>
<CsInstruments>

sr = 44100
ksmps = 64

instr 1
  kcounter init 0  
  kt = timeinstk()
  ke = eventcycles()
  ks = timeinsts()
  kx = eventtime()
  println "cycle=%2d  timeinstk=%2d  eventcycles=%2d  timeinsts=%.4f  eventtime=%.4f", kcounter, kt, ke, ks, kx
  kcounter += 1
endin

</CsInstruments>
<CsScore>
i1 0 0.1
e
</CsScore>
</CsoundSynthesizer>
```

gesellkammer (changes requested):

tests/commandline/test_lastcycle_delayed_start.csd
hlolli: @gesellkammer Switched the cases to `eventcycles()` and separated the zero-based cycle index from the total cycles rendered. Our assertions already accounted for `timeinstk`’s one-based count, so the expected timing is unchanged: five rendered cycles end at index 4. I also opened https://github.com/csound/manual/pull/803 to correct the `eventcycles` documentation.

tests/commandline/test_lastcycle_delayed_start.csd
```diff
@@ -0,0 +1,50 @@
...
+instr 1
+  kBefore lastcycle
+  ; No explicit release: lastcycle must keep its one-cycle minimum.
+  xtratim 0
+  kAfter lastcycle
+
+  ; Both placements must produce 0 on every cycle except the final one.
+  ; Cycle indices start at zero.
+  kCycleIndex eventcycles
+  kExpected = (kCycleIndex == 4 ? 1 : 0)
```
gesellkammer: How can this work?

0.0625 * 1024 / 16 = 4, so the instr runs for 4 cycles, which means that lastcycle should fire at kCycleIndex == 3 and the instrument should never reach 4

hlolli: @gesellkammer Your calculation gives the four cycles from `p3`. The extra cycle comes from `lastcycle` itself: its existing initializer reserves one release cycle when none is set, and `xtratim 0` does not undo it. That makes index 4 the final cycle here. This PR preserves that behavior. I opened https://github.com/csound/manual/pull/804 to explain the side effect and this example in the manual.

gesellkammer: There should not be an extra cycle. The previous implementation counted the cycles when there was no release phase (no extra cycles), now that logic is gone. lastcycle should have no side effects (so please do not document those), if it has this is a bug and should be fixed.

hlolli: I see ok, this behaviour predates this PR, I'll prepare a fix for this discovered bug.

hlolli: @gesellkammer The fix removes the extra cycle. Here are the cases and how they differ. Cycle indices start at **0**, as with `eventcycles()`.

| Case | Note length | When `lastcycle` returns 1 |
|---|---|---|
| Finite note without release, including a delayed start | 4 cycles | Index 3; no added cycle |
| Explicit release set after `lastcycle` initializes | 4 note + 4 release cycles | Index 7, regardless of opcode order |
| `p3` extended later during init, without release | 4 → 8 cycles | Index 7 |
| `p3` extended later during init, with release | 8 note + 4 release cycles | Index 11 |
| Finite note stopped early with release | 5 cycles before release + 4 release cycles | Index 8; follows the early stop |
| Indefinite note stopped with release | 5 cycles before release + 4 release cycles | Index 8; release gives it a known end |
| Indefinite note stopped without release | Ends immediately after 5 cycles | No pulse: it cannot predict the external stop or add a cycle to report it |
| Instance reused with shorter or no release | Successive notes last 8, 6, then 4 cycles | Indices 7, 5, then 3; no state carried over |
| Sample-accurate ending within a block | 72 samples with `ksmps = 16`: 5 cycles | Index 4, during the partial final block |
| Smaller local `ksmps` | 72 samples with local `ksmps = 4`: 18 cycles | Local index 17 |
| Nested UDOs, with later duration/release changes | 8 cycles without release, 12 with release | Index 7 or 11; follows the owning instrument’s end time |

In each case, `lastcycle` leaves the note’s duration and release unchanged.

gesellkammer (approved):

PR #3070, opened 2026-09-16
Fix midiarp held-note state and trigger timing

hlolli:
`midiarp` reads the note number as velocity from the current MIDI message buffer. A zero-velocity note-on therefore adds a held note instead of releasing it, and the arpeggio keeps playing after key release.

Read the four message fields correctly and process queued messages before generating a trigger. Keep a sorted list of held key/channel pairs within the existing ten-note capacity. This also fixes negative counts from unmatched releases, writes past a full note list, lost notes after removing and adding keys, and MIDI note 0 being treated as an empty slot.

Clear the trigger when idle and reset the clock on initialization. Up and bounce modes start at the lowest held note; down starts at the highest. Keep traversal within the held notes, and use Csound's seeded random generator instead of reseeding the process-wide C generator. Nonpositive rates pause triggering; rates above the control rate emit at most once per control cycle.

These changes restore the [documented behavior](https://csound.com/manual/opcodes/midiarp/): arpeggios follow currently held notes and the trigger is a pulse, rather than remaining high after the last release.


vlazzarini:
@rorywalsh one of yours?

vlazzarini (approved):

PR #3082, opened 2026-09-16
Fix lfsr seed conversion and random state

hlolli:
The documented `lfsr` seed of `-1` went through an undefined floating-point-to-unsigned conversion. Convert finite seeds modulo 2^32 so `-1` reliably sets all bits in both float and double builds. Reject unsupported register lengths and probabilities before narrowing them, and use unsigned arithmetic for bit shifts.

Give each instance its own Csound random state instead of reseeding the process-wide C random generator from the clock. Starting another instance no longer changes an existing instance’s random choices, and Csound’s `seed` controls repeatability. This changes randomized sequences while preserving the shift-register algorithm.


DaveSeidel:
Thanks Hlöðver, this looks good.

vlazzarini (dismissed):

vlazzarini (approved):

PR #3087, opened 2026-09-16
Fix variable-pool alignment in float builds

hlolli:
In a 64-bit float build, a scalar occupies four bytes but each variable's type header needs eight-byte pointer alignment. Packing slots at four-byte boundaries can misalign later headers, structured values, and the opcode state after the pool, causing undefined behaviour during ordinary instrument setup.

Pad each variable-pool slot to pointer alignment when adding variables or recalculating pool sizes. Keep each value's size unchanged, so scalar storage and array strides retain their current layout. This also fixes the same packing issue in the UGen API and adds no work to audio loops.


vlazzarini:
how do you propose to test it? Do we have a floats build on 64bit arch in the CI for this?

.github/workflows/csound_builds.yml
```diff
@@ -132,6 +132,53 @@ jobs:
             exit 1
           fi
 
+  linux_float64_ubsan:
```
vlazzarini: I think the name of the workflow is misleading. It's linux64_float_ubsan. The OS is 64 and the build is float, instead of linux with a float64 (that is, a double) build.

hlolli: @vlazzarini Agreed. Renamed the job to `linux64_float_ubsan`, the display name to “64-bit Linux / 32-bit float type-system tests (UBSan)”, and the build directory to `build-linux64-float` so the architecture and sample precision are clear.

vlazzarini (approved):

PR #3094, opened 2026-09-17
fix(zak): validate state and indices and preserve partial blocks

hlolli:
Zak's shared control-rate initializer dereferences missing global state before it can report that `zakinit` has not run. Check for that state first.

Validate allocation sizes and index conversions before using them, avoid signed overflow when reversing negative modulation indices, and use `size_t` for audio-buffer offsets. The index macro runs once per opcode call, outside the sample loops.

Also remove a second subtraction of the note-end offset in `zamod`, which skips active samples and can underflow the loop bound on short notes.

The [zakinit manual](https://csound.com/docs/manual/zakinit.html) says the size excludes the upper location, but the implementation and examples use that location. Keep the existing inclusive indexing and fractional-index truncation for compatibility.


vlazzarini (approved):
I wonder if we should deprecate the zak opcodes since we have more modern means of doing the same thing. Maybe a question for an LLM?

hlolli:
yeh I can check, my thinking would be because of LLM the time spend maintaining it is low. And there exists alternatives that also work well. I can ask llm what it thinks too

hlolli:
Astra says:

@vlazzarini I’d lean toward recommending channels and arrays for new code, while keeping Zak supported for existing orchestras. Before formally deprecating it, we should document replacements for indexed routing, mixing and clearing, and check for timing or performance differences. The fixes here remain useful either way. LLMs can reduce the work of editing code, but review and compatibility still need care.

vlazzarini:
We will sure keep zak - marking it as deprecated does not remove it. The plan for deprecated opcodes is to mark and document them as that but not remove them. Encourage users to take the modern alternatives. Backwards compatibility is a non-negotiable condition. This also allows the manual and similar reference docs to be more easily navigated.

How about that then, for zak?

PR #3104, opened 2026-09-17
fix(printsk): bound formatting and refresh control-time strings

hlolli:
`printsk` and `println` can write beyond their output buffer when printing long literals or several formatted fields. Their formatter also advances past truncated output without retrying after allocation, and caches a format length that can change at k-time.

Use the remaining buffer space for each field, grow to the required size, and retry before advancing. Read the current format, reuse working buffers, and free output storage when the note ends. Reject unsupported conversions and mismatched argument types before passing them to `snprintf`.

The [manual](https://csound.com/docs/manual/printsk.html) allows strings up to 8192 characters, but the old literal-copy path allocated only 2048 bytes. Its claim of full C formatting support is also broader than this formatter: widths and precisions must be fixed, and length modifiers and positional arguments are unsupported.


vlazzarini:
@gesellkammer review?

gesellkammer:
I cannot review this since this is a complete ai rewrite.

Opcodes/emugens/emugens.c
```diff
@@ -2760,234 +2760,186 @@ strstrip(CSOUND *csound, STR1_1 *p) {
...
+  for (j = 0; j < numVals; j++) {
+    if (UNLIKELY(IS_AUDIO_ARG(kvals[j])))
+      return PERFERR(Str("a-rate argument not allowed"));
+  }
+  j = 0;
+
+  /* Preserve literal copying when there are no format arguments. */
+  size_t initialSize = numVals == 0 ? strlen(fmt) + 1 : 64;
+  if (str->data == NULL || str->size < initialSize) {
+    char *data = csound->ReAlloc(csound, str->data, initialSize);
```
gesellkammer: This runs on performance also, it should not allocate/reallocate

gesellkammer (approved):

PR #3115, opened 2026-09-17
fix(strings): copy current values without timestamp checks

hlolli:
String assignment can keep stale text when a control-rate ternary selects a different input or `readf` returns a new line. Remove the timestamp check and copy-time stamping so assignments copy the current text on each control cycle, including empty strings at EOF and local `ksmps` blocks.

Compile init-time (`b`) string ternaries to init-only `strcpy`, so changing their source strings during performance does not change the selected text. Control-time (`B`) ternaries keep updating each cycle.

This includes the `readf` fix from #3119. The remaining timestamp writes and the `STRINGDAT` field are removed separately in #3126.


vlazzarini:
I suppose we need to make sure this does not affect init-time only string copy in ternary expressions when we have a b instead of a B expression.

I mean we don't want init time code to leak into perf time. Can you ensure that?

hlolli:
@vlazzarini This PR keeps `b` ternary copies at init time and lets `B` ternaries update at control rate. It now also removes the timestamp check from ordinary assignment and stops stamping copies, so neither path depends on timestamps.

hlolli:
@vlazzarini

> Can you ensure that?

Yes. For a `b` condition, the compiler now emits init-only `strcpy`, so those branch copies never enter the performance chain. The selected text stays fixed even if the source strings later change. Only `B` conditions use control-rate branch copies.

vlazzarini (approved):

PR #3116, opened 2026-09-17
fix(strings): copy strings on every local control cycle

hlolli:
Chained string assignments can keep stale text when an instrument uses a smaller local `ksmps`. The copy routine stamps the global cycle, while assignment checks the local cycle.


vlazzarini:
ok, timestamps were an idea that did not work, so while there is stamping here, I don't think the timestamp is read/used anywhere. We can fix this ok but I think will not have any impact.

hlolli:
@vlazzarini There was still a timestamp check in `strassign_k`. I've removed it and the copy-time stamping in this PR. Assignments now copy the current text each cycle, including local `ksmps` blocks. #3126 handles the remaining timestamp cleanup.

PR #3117, opened 2026-09-17
fix(channels): keep string assignments current

hlolli:
`chnget`, `chngetks`, and `chngets` can return new text while a following assignment keeps its old value. Remove the proposed timestamp writes from these readers and fix assignment itself, including copies of empty strings. Channel locking and the unchanged-text shortcut stay in place.


vlazzarini:
Same thing here - timestamps were tested and then abandoned, but the code was not removed.

Maybe the best thing is to ask the LLM to review and remove timestamping across the codebase.

hlolli:
@vlazzarini I've now removed the timestamp writes from this PR too, and removed the assignment check instead. Copies follow the channel text, including empty strings, without an update marker. #3126 handles the remaining timestamp cleanup.

PR #3118, opened 2026-09-17
fix(invalue): correct callback state and initialization

hlolli:
`invalue` can receive new callback text while a following assignment keeps its old value. Remove the proposed timestamp write and make assignment copy the current text each cycle, including when removing the callback clears the string. Remove the timestamp check and copy-time stamping. Keep init-time string ternary copies out of performance.

Reserve the same 256-byte callback buffer for numeric channel IDs as for named channels. The numeric form previously kept the default 64-byte string buffer.

The i-rate numeric forms also read the callback twice during initialization and discard the first result. Their setup routines already read the value, so remove the second call.


vlazzarini:
Same thing here with timestamps.

hlolli:
@vlazzarini I've removed the timestamp write from this PR and removed the assignment check instead. Copies now follow the callback text without an update marker. The callback-buffer and duplicate-init fixes remain; #3126 handles the remaining timestamp cleanup.

vlazzarini (approved):

PR #3119, opened 2026-09-17
fix(readf): keep string assignments current

hlolli:
`readf` can advance through a file while a following string assignment stays empty. Remove the proposed timestamp write from the reader and fix assignment itself, so copies receive each line and the empty string at EOF, including with local `ksmps`.


hlolli:
@vlazzarini This PR and #3115 now use the same fix: remove the timestamp check from string assignment while keeping init-time ternary copies out of performance. I've combined them in #3115 and am closing this PR as superseded.

PR #3120, opened 2026-09-17
Fix mton string output updates

hlolli:
`mton` updates its note-name string at control rate but leaves its timestamp unchanged. A downstream assignment such as `Scopy = Snote` therefore keeps the initial note name even as the MIDI input changes.

Stamp the output with the instrument's current control cycle after formatting it, so copies receive each update, including with a smaller local `ksmps`.


vlazzarini:
@gesellkammer can you review?

gesellkammer (approved):

PR #3121, opened 2026-09-17
Fix ftset slice bounds and table selection

hlolli:
`ftset` uses unchecked slice bounds and steps when writing to a table. Invalid ranges can write outside the table, and a step that does not advance can hang performance. Its zero-fill shortcut also mishandles reversed ranges.

- Validate bounds and steps before integer conversion, treat empty or reversed ranges as no-ops, and prevent the final loop increment from overflowing. Keep the fast zero-fill path and the [documented negative end indexes](https://csound.com/docs/manual/ftset.html).
- Match the table cache to `FTFind`'s rounding so changes in table number select the right table. Clear the cache at initialization so the built-in table alias works on the first call.


vlazzarini:
@gesellkammer review?

gesellkammer:
Sorry, same thing

hlolli:
Here I agree, the tests are bulky. From my side, the goal of the tests is to find regressions, and that goal is achieved. But in this case, I'll make a trade-off on that goal for even more simple tests.

hlolli:
@gesellkammer I think you'll like this test now

Opcodes/emugens/emugens.c
```diff
@@ -1800,61 +1799,67 @@ typedef struct {
...
-    FUNC *tab = p->tab;
-    MYFLT *data = tab->ftable;
-    int32_t tablen = tab->flen;
-    int32_t start = (int32_t)*p->kstart;
-    int32_t end = (int32_t)*p->kend;
-    int32_t step = (int32_t)*p->kstep;
+ftset_common(CSOUND *csound, FTSET *p, int32_t init) {
+    double number = (double)*p->tabnum;
+    if (UNLIKELY(!(number >= INT32_MIN && number <= INT32_MAX)))
+        return init ? INITERR(Str("ftset: table number out of range"))
```
gesellkammer: At least some macro to abstract this kind of pattern out, it is polluting everything

gesellkammer (changes requested):

gesellkammer (commented):
Also, I see no tests for ftset at init time and ftset with a table changing at perf time

gesellkammer (approved):

PR #3122, opened 2026-09-17
Fix ftslice source bounds and step handling

hlolli:
`ftslice` and `ftslicei` limit copies to the destination length but do not bound reads from the source. Invalid steps also reach division and integer conversion unchecked.

Validate table numbers, indexes and steps before conversion, cap the end at the source length, and leave empty or reversed ranges unchanged. Use an exact integer copy count and a wider source index to avoid rounding errors and overflow on the final step.

Keep the [documented exclusive end and destination truncation](https://csound.com/docs/manual/ftslice.html). Negative end values retain their existing meaning of copying to the end. Also correct the destination-table lookup diagnostic and report invalid slices through the appropriate init or performance error path.


vlazzarini:
@gesellkammer and this one?

gesellkammer:
Same as other PRs, this is a complete rewrite with unreadable tests, so cannot review.

hlolli:
it's 32 line removal and 42 addition in the c file, barely accounts as complete rewrite. I've made the tests little more clear, as always, feel free to reject the PR and I'll close it leaving a ticket for addressing it later.

hlolli:
simplified tests even more

gesellkammer (approved):

PR #3123, opened 2026-09-17
Fix tab2array capacity errors and slice bounds

hlolli:
`tab2array` ignores a failed output-capacity check and keeps copying. If a control-rate slice grows beyond the allocated array, it reports an error but can still write past the buffer.

Stop before copying when that check fails. Also validate table numbers and slice parameters before integer conversion, limit reads to the source table, use an exact integer copy count, and prevent the final index increment from overflowing. Empty or reversed ranges produce empty arrays; multidimensional outputs report an error.

The [manual describes allocation and resizing](https://csound.com/docs/manual/tab2array.html), but does not state the performance-time limit: allocation happens at init time, while later slices must fit the existing capacity. This keeps that limit and makes its error path safe. Preallocated arrays can still shrink and grow within their capacity.


vlazzarini:
@gesellkammer and another

gesellkammer:
Also cannot review. It is also impossible to check the tests since they are embedded into c++ and mangled.

hlolli:
@gesellkammer readability improved, test cases moved into more files, but c++ has a purpose there, but say the word and we can leave this as ticket and address this later, we don't need to solve this in a hurry, I have no problem with closing this PR.

hlolli:
because I had good result on another PR, I'm also simplifying here

hlolli:
even simpler now

gesellkammer (approved):

PR #3124, opened 2026-09-17
Fix linlin array error handling and in-place output

hlolli:
Both array forms of `linlin` ignore failed capacity checks and keep writing when their inputs grow beyond the output buffer. Their init-only wrappers also continue after initialization fails.

Propagate those errors and validate one-dimensional inputs and outputs before using their sizes. Report zero-width ranges through the shared `INITPERFERR` helper at the caller’s rate. Remove `restrict` from the mapping loop so using the input array as the output is valid C.

The interpolation formula and shorter-array blend behavior stay the same. Output arrays can still shrink and grow within their allocated capacity.


gesellkammer (changes requested):
* The pattern init ? INITERR(...) : PERFERR needs to be corrected
* Better tests, csd tests or at least tests where it is clear what is being tested
* Being a complete rewrite, the author needs to be aware that she/he will completely own this code

hlolli:
@gesellkammer I replaced the C++ harness with CSD tests that show the inputs and expected results directly, with comments for each case. Error handling now uses the shared `INITPERFERR` helper. The interpolation formula is unchanged; the code changes cover validation, error propagation, and in-place output.

gesellkammer (approved):

PR #3125, opened 2026-09-17
Fix bpf and bpfcos array capacity errors

hlolli:
The array forms of `bpf` and `bpfcos` report an error when the input grows beyond the output buffer, but then continue processing and write past its end. Return immediately when the capacity check fails.

Check array shapes and breakpoint counts before allocating the output, so invalid arguments leave an existing output unchanged. Remove the duplicate allocation call and correct the reported number of breakpoint pairs. Interpolation stays unchanged.


vlazzarini:

gesellkammer (dismissed):

hlolli:
@gesellkammer merge conflict, so I lost the approval

gesellkammer (approved):

PR #3126, opened 2026-09-18
Remove string update timestamps

hlolli:
Remove `STRINGDAT.timestamp`, its writes, and timestamp-gated string copies. String-array access uses `arrayMemberSize` so the smaller structure still works with alignment padding on 32-bit builds, including WASM.

Keep ordinary string `=` init-only, as in Csound 6. Use `strcpyk` when a copy must follow changes during performance. String expression assignments retain the expression’s rate: an init-time ternary stays fixed, while a control-rate ternary updates its result. The compiler now recognizes the current assignment token when applying that rule.

This also stops `SResult = ""` from clearing strings built during initialization on every control cycle.

The callback-buffer and duplicate-init fixes in #3118 remain separate. Removing the field changes the public `STRINGDAT` layout; native plugins need rebuilding.

Requested in https://github.com/csound/csound/pull/3117#issuecomment-5720782944. Addresses the assignment compatibility concern in https://github.com/csound/csound/pull/3126#issuecomment-5726665656 and the init-time concern in https://github.com/csound/csound/pull/3115#issuecomment-5720692885.


vlazzarini:
I think assignment should be init-time only. There's strcpyk for perf time. Actually it's best to check what the situation is with Csound 6. We maintain it - backwards compatibility.

hlolli:
@vlazzarini You’re right: ordinary string `=` is init-time only in Csound 6. I’ve restored that behavior; performance-time copies use `strcpyk`. Control-rate ternary expressions still update their result, as they do in Csound 6. This also fixes the print failures caused by `SResult = ""` clearing strings built at init.

vlazzarini (approved):

PR #3127, opened 2026-09-19
diskin2 crossfade and loop end

gesellkammer:
diskin2: adds optional crossfading for diskin2 when looping. It also adds a new iend param to set the end of the loop. Without this, using iwrap=1 will create discontinuities since this is a hard wrap. This extends iwrap to control crossfading between end and start of the loop.

iwrap=0 -> no loop (original behaviour)
iwrap=1 -> hard wrap (original behaviour)
iwrap > 1 (for example, 128) -> crossfade of this value as samples (128 sample crossfade)

NB: benchmarks show exact same performance for diskin2 with iwrap=0 and iwrap=1 before and after these changes. 

This makes it possible to use diskin2 as a streaming alternative to the looping table readers (loscilx, flooper2, etc.)


hlolli:
I think if you rebase with latest develop the CI should pass wrt prinkt errors, which are unrelated (obviously)

OOps/diskin2.c
```diff
@@ -80,18 +95,20 @@ static CS_NOINLINE void diskin2_read_buffer(CSOUND *csound,
 /* of opcode 'p', at sample index 'n' (0 <= n < ksmps), with amplitude  */
 /* scale 'scl'.                                                         */
 
-static inline void diskin2_get_sample(CSOUND *csound,
+static ALWAYS_INLINE void diskin2_get_sample(CSOUND *csound,
```
hlolli: I'm split on always inline vs macro. I like personally that we can toggle inlined functions on and off for debugging, but if we want to hard enforce it like that, maybe a macro is better?

What I wonder here, and I'm thinking, maybe we get neither, the ability to toggle it on and off, nor ensure all compilers actually do inline (besides mainstream ones).

vlazzarini:
@coderabbitai review

hlolli (commented):
Thanks for adding this. I have two suggestions for the crossfade behavior below.

OOps/diskin2.c
```diff
@@ -1256,25 +1353,128 @@ int32_t diskin2_async_deinit(CSOUND *csound, DISKIN2 *p)
...
+static inline void diskin2_file_pos_inc_xf(DISKIN2 *p, int32_t *ndx)
+{
+    p->pos_frac += p->pos_frac_inc;
+    *ndx = (int32_t) (p->pos_frac >> POS_FRAC_SHIFT);
+    /* Only reached by the crossfade readers (xf.len > 0 implies wrapMode), and
+       loopLength == loopEnd - loopStart, so one unsigned compare covers both
+       past-the-end and before-the-start */
+    if (UNLIKELY((uint32_t) (*ndx - p->loopStart) >= (uint32_t) p->loopLength)) {
+      if (p->xf.ready)
+        p->pos_frac = p->xf.headEnd;
```
hlolli: Could we account for playback speed when capping the crossfade and ensure the saved restart position stays inside the loop? I reproduced a case at 44.1 kHz with a loop from 0 to 0.04 seconds, `iwrap=1000`, and `kTranspose=2`: the cap gives 882 frames, but capturing 882 output frames advances through all 1,764 source frames, leaving `headEnd == loopEnd`. This assignment then returns to the same position on every sample, and both scalar and array playback become a constant sample after the first pass. It would be good to cover this case in a regression test too.

OOps/diskin2.c
```diff
@@ -1256,25 +1353,128 @@ int32_t diskin2_async_deinit(CSOUND *csound, DISKIN2 *p)
...
+
+    if (UNLIKELY(F <= 0 || inc == 0))
+      return;
+
+    /* span is the crossfade length in position units, positive in both
+       directions; dist below is the distance travelled from the relevant
+       loop edge, also positive in both directions. */
+    span = (int64_t)F * (inc > 0 ? inc : -inc);
+
+    if (x->dir != dir) {
```
hlolli, line 1494: Could we also refresh the captured head when playback speed changes without changing direction, or make the cached audio follow the current speed? Reading this path, a change from `kTranspose=1` to `2` keeps the head audio and `headEnd` captured at the old speed, while `span` uses the new increment. Later loop boundaries would therefore mix in audio at the old pitch and resume at the old source position. A test that changes positive `kTranspose` after the first capture would help cover this.

OOps/diskin2.c
```diff
@@ -1299,6 +1560,8 @@ int32_t diskin2_perf_synchronous(CSOUND *csound, DISKIN2 *p)
 #else
       p->pos_frac_inc = (int64_t)(f + (f < 0.0 ? -0.5 : 0.5));
 #endif
+      /* the captured head belongs to the previous speed */
+      diskin2_xf_reset(&p->xf);
```
hlolli: Could we keep crossfading active when `kTranspose` changes continuously? With `ksmps=64` and `iwrap=128`, this reset clears both `ready` and `count` every block, before the 128-frame capture can finish. A normal pitch ramp therefore keeps using hard wrapping, bringing back the boundary clicks. I confirmed the same behavior in the scalar and array versions.

The self-contained CSD below generates its own source file. The left output is the difference between hard wrap and crossfade at a fixed speed; the right is the same comparison during a ramp from 1 to 1.1. On `694814d67`, it prints `overall amps: 0.92075 0.00000`: the fixed-speed crossfade works, but the ramp output is identical to hard wrapping. Could we add a case like this alongside the single-step speed-change test?

<details>
<summary>Small CSD reproducer (no external sample needed)</summary>

Save as `diskin2-pitch-ramp.csd` and run `csound diskin2-pitch-ramp.csd` in a writable directory. It creates `diskin2-pitch-ramp-source.wav` there.

```csound
<CsoundSynthesizer>
<CsOptions>
-n -d -m0
</CsOptions>
<CsInstruments>
sr = 44100
ksmps = 64
nchnls = 2
0dbfs = 1

instr 1
  aSrc phasor 25
  fout "diskin2-pitch-ramp-source.wav", 14, aSrc
endin

instr 2
  kPitch linseg 1, p3, 1.1
  aFixedHard diskin2 "diskin2-pitch-ramp-source.wav", 1, 0, 1, 0, 2, 0, 0, 1, 0.04
  aFixedFade diskin2 "diskin2-pitch-ramp-source.wav", 1, 0, 128, 0, 2, 0, 0, 1, 0.04
  aRampHard diskin2 "diskin2-pitch-ramp-source.wav", kPitch, 0, 1, 0, 2, 0, 0, 1, 0.04
  aRampFade diskin2 "diskin2-pitch-ramp-source.wav", kPitch, 0, 128, 0, 2, 0, 0, 1, 0.04
  ; Left: fixed-speed difference. Right: changing-speed difference.
  outs aFixedHard - aFixedFade, aRampHard - aRampFade
endin
</CsInstruments>
<CsScore>
i1 0 0.08
i2 0.1 0.4
e
</CsScore>
</CsoundSynthesizer>
```

</details>

hlolli (commented):
Thanks for fixing the earlier cases. I found one remaining case with continuous pitch changes, with a small reproducer below.

OOps/diskin2.c
```diff
@@ -1487,7 +1804,11 @@ void diskin_file_read(CSOUND *csound, DISKIN2 *p)
 #else
       p->pos_frac_inc = (int64_t)(f + (f < 0.0 ? -0.5 : 0.5));
 #endif
+      /* a step invalidates the captured head; a ramp keeps it */
+      diskin2_xf_speed_change(&p->xf);
     }
+    else
+      p->xf.changing = 0;
```
hlolli: Could we track pitch changes by control period in the async readers, so an extra worker poll cannot clear `changing`? The worker can see the same pitch more than once within a control period, including when no frames need reading. Clearing the flag here makes the next ramp update look like a fresh step, which resets the captured head again. The array reader has the same case at lines 2303–2304.

I reproduced large loop-boundary jumps on `d59cd875d` in both async variants. Fixed-speed async playback and the same ramp with synchronous reads stay smooth. The current ramp test runs synchronously, so it misses this case. It would help to cover the async ramp too.

<details>
<summary>Self-contained CSD reproducer</summary>

Save this as `async-ramp.csd` and run `csound async-ramp.csd` in a writable directory. It takes about 6.5 seconds, generates its own source and measurement WAV files, and uses the null audio driver, so no audio device is needed.

One local run printed:

```text
Maximum sample jumps: fixed=0.000244141 async=0.986816 array=0.986816 sync=0.000366211
```

The exact async values vary with thread timing; the large jumps are the problem.

```csound
<CsoundSynthesizer>
<CsOptions>
-odac -+rtaudio=null --realtime -d -m128
</CsOptions>
<CsInstruments>
sr = 44100
ksmps = 64
nchnls = 1
0dbfs = 1

; Generate a source with a large discontinuity at the loop boundary.
instr 1
  aPhase phasor 1
  fout "async-ramp-source.wav", 14, aPhase - 0.5
endin

instr 2
  kPitch = 1 + 0.1*timeinsts()
  ;                           pitch   skip wrap fmt win buf  skip sync
  aFixed diskin2 "async-ramp-source.wav", 1,      0, 4096, 0, 2, 1024, 0, 0
  aAsync diskin2 "async-ramp-source.wav", kPitch, 0, 4096, 0, 2, 1024, 0, 0
  aArray[] diskin2 "async-ramp-source.wav", kPitch, 0, 4096, 0, 2, 1024, 0, 0
  aSync diskin2 "async-ramp-source.wav", kPitch, 0, 4096, 0, 2, 1024, 0, 1

  ; Measure adjacent-sample jumps, ignoring startup.
  kGate = (timeinsts() > 1 ? 1 : 0)
  aD1 = (aFixed - delay1(aFixed)) * kGate
  aD2 = (aAsync - delay1(aAsync)) * kGate
  aD3 = (aArray[0] - delay1(aArray[0])) * kGate
  aD4 = (aSync - delay1(aSync)) * kGate
  fout "async-ramp-jumps.wav", 14, aD1, aD2, aD3, aD4
  out aAsync
endin

instr 3
  prints "Maximum sample jumps: fixed=%g async=%g array=%g sync=%g\n", \
    filepeak("async-ramp-jumps.wav", 1), \
    filepeak("async-ramp-jumps.wav", 2), \
    filepeak("async-ramp-jumps.wav", 3), \
    filepeak("async-ramp-jumps.wav", 4)
endin
</CsInstruments>
<CsScore>
i1 0 1
i2 1.1 5
i3 6.2 0.1
f0 6.5
e
</CsScore>
</CsoundSynthesizer>
```

</details>

gesellkammer:
While debugging this last issues some crashes surfaced, which are due to longjmps across threads. I would post that as a separate PR once this is merged

hlolli:
Sorry, I merged #3221 before all tests had finished. The `SfloadTests.DuplicateLoadsReuseHandlesAcrossBankGrowth` failures here came from that merged change: parallel test processes can use the same temporary directory name, and one process can then remove the other's files.

The fix is in #3230.

OOps/diskin2.c
```diff
@@ -1488,6 +1840,11 @@ void diskin_file_read(CSOUND *csound, DISKIN2 *p)
       p->pos_frac_inc = (int64_t)(f + (f < 0.0 ? -0.5 : 0.5));
 #endif
     }
+    /* a step latched by the perf thread invalidates the captured head; a ramp
+       keeps it. Consumed here, outside the change test, so that no poll can be
+       missed even if the perf thread updates p->transpose right after we read
+       it. */
+    diskin2_xf_consume_step(&p->xf);
```
hlolli, line 1847: The ramp reproducer now passes. Could we also make the pitch and reset notification a consistent, synchronized handoff? This ordering still seems possible:

```text
worker: reads old transpose
perf:   publishes new transpose and increments resets
worker: consumes that reset, then captures the head at the old speed
worker: next poll sees the new speed, but the reset is already consumed
```

That can leave the old head span in use after a speed step, changing the loop period. Also, `resets` is shared through plain `volatile` reads/writes; it needs proper synchronization. Making only the counter atomic would still leave the ordering above.

I haven't reproduced this interleaving in audio, so this is a source-review finding. A focused test that pauses the worker after reading the pitch, publishes a step, then resumes capture would check it more reliably than a timing-dependent CSD. Could we consume a matching speed/reset snapshot in both async readers?

hlolli: I'm ok with this being in a follow-up PR. So I approve. You decide if you feel fixing it now

hlolli (approved):

PR #3130, opened 2026-09-20
Fix printarray formatting buffer bounds

hlolli:
`printarray` uses `snprintf`'s required output size as a buffer index, so long strings or wide fields can write past the stack buffer. Its `%d` replacement also copies into fixed buffers without checking their size.

Print elements through the message handler instead of accumulating fixed-size lines, and store the prepared format in Csound-managed memory. This prints long elements and all multidimensional columns without truncation. Read the current label pointer so growing a label cannot leave a dangling pointer.

Validate formats at initialization: each element supplies one argument, so reject mismatched conversions, extra conversions, and dynamic widths. Preserve literal `%%` and the existing bare `%d` alias to `%.0f`. That alias still rounds, although the [manual example](https://csound.com/docs/manual/printarray.html) describes integer truncation.


vlazzarini:
@gesellkammer one of yours.

gesellkammer (approved):
I have no problem approving this, but the rewrite addresses much more than the fixes.

hlolli:
The rewrite does go beyond the safety fixes. It shares the formatting code, allows the triggered form to print numeric arrays with more than two dimensions, and prints full matrix rows instead of dropping columns. The k-rate forms also accept single-character literal formats.

All formats are now copied at initialization, and message callbacks receive output in smaller pieces rather than complete lines. These are broader changes. The buffer bounds, format validation and stale label pointer are the bug fixes. I should have made that distinction clearer in the PR description.

cc @vlazzarini @gesellkammer

gesellkammer:
In ftprint the problem with the many calls to the message callback could be fixed. In general, when running csound under a host this poses a problem.

PR #3131, opened 2026-09-20
Fix slicearray source bounds and copy offsets

hlolli:
`slicearray` accepts negative start indices and can read before the input array. Large strides can also overflow the copy loop's final index increment. Validate bounds before integer conversion and calculate offsets only for the elements being copied.

Use separate input and output element strides, fetch storage after output preparation, and skip self-copies. This keeps in-place slices and local audio-array copies safe.

Separate initialization from performance so source-size changes get performance errors and output growth stays within reserved capacity. Preserve the inclusive end index; reversed ranges and empty inputs produce empty slices.


vlazzarini (approved):

hlolli:
@vlazzarini I have a feeling that windows will fail, it feels random if it prints error to stdout or stderr, lets see what happens maybe needs a fix

```
stderr = ["slicearray: invalid slice bounds", "slice larger than original size", "Array too small"]
```

hlolli:
all good here

PR #3133, opened 2026-09-20
fix(ftprint): bound table reads and remove the fixed row buffer

hlolli:
`ftprint` advances its row-buffer position by the length `snprintf` wanted to write, even when the text did not fit. A large column count can therefore write past the buffer.

Keep buffered output and check each append before advancing the write position. Rows that fit use one message callback. Split longer rows between values into chunks of at most 1,023 bytes so they also fit Csound's string-message callback buffer. Preserve every value, the existing row layout and line breaks without calling the host for each element.

Check table lookup, indices, step and column count before use. This prevents out-of-range table reads, zero-step loops and integer overflow on the final step. Fix the negative-index helper, which rejected valid indices after calculating them, and report errors at initialization or performance time as appropriate.


vlazzarini:
@hlolli could you assign a reviewer directly if the opcode author is in the team or contactable via github? I'll keep reviewing any others that have no well defined author.

hlolli:
@vlazzarini yes I can do that from now on

gesellkammer (commented):
The original implementation accumulated each row in a buffer , thus limiting the number of calls to the callback. This implementation fires the callback for each element, which has a clear impact in hosts like csoundqt, which set the message callback and implement highlighting , etc

hlolli:
You're right. Calling the message handler for every element adds unnecessary work for hosts that process each callback. I'll restore buffered output with checked appends, keeping the bounds and index fixes. Ordinary rows will use one callback, and longer rows will use safe chunks without changing the text or line breaks.

I could make the same change to `printarray` from #3130 (in a new PR) and check callback counts as well?

hlolli:
@gesellkammer I've restored buffered output in `ftprint`. Rows that fit go through one callback, and longer rows use safe chunks while keeping all values and line breaks. The bounds and index fixes remain. I've also updated the PR description to reflect this.

gesellkammer (approved):

PR #3138, opened 2026-09-20
fix(compiler): preserve at signs in string literals

hlolli:
The compiler treats every `@` in an argument as the start of a global annotation. For a string such as `"name@example.org"`, tokenization truncates the literal even though the suffix is not `@global`. The resulting string misses the constant pool, and the fallback inserts it after freeing its temporary buffer, causing a use-after-free.

Skip quoted strings with a one-line guard and keep the existing variable annotation code. Reuse the string-pool helper so lookup or insertion completes before the temporary buffer is freed. This also covers strings introduced by runtime compilation while preserving explicit global variables.


Engine/csound_orc_compile.c
```diff
@@ -2828,15 +2828,11 @@ static void build_const_pool(CSOUND *csound, INSTRTXT *ip, char *s,
...
-  if(strchr(varName, '@') != NULL) {
-    char* th;
-    char* baseType = strtok_r(varName, "@", &th);
-    char* global = strtok_r(NULL, "@", &th);
-    if(!strcmp(global, "global")) {
-      varName = baseType;
-    }
-  }
+  /* String contents are not variable annotations. */
+  if (*varName == '"') return;
```
vlazzarini, line 2831: This change only needs one line, does it not? Keep the original code for the rest, unless there is a problem we can't see.

vlazzarini (changes requested):

Engine/csound_orc_compile.c
hlolli: @vlazzarini Agreed. I restored the original annotation code and kept only the one-line guard for quoted strings. The separate string-pool fix remains because the fallback used a buffer after freeing it.

vlazzarini (dismissed):

vlazzarini (approved):

PR #3141, opened 2026-09-20
Fix ntom and ntof note-name validation

hlolli:
`ntom` and `ntof` read the second character before checking whether a note name is empty. They also turn malformed names into pitches: `ntom("4C?")` returns 59.5 and `ntom("xA")` returns 885.

Validate the length, octave digit, note letter, accidental and cents suffix before conversion, and report invalid names through the existing init/performance error macro. Valid names keep their current results, including the documented `+`/`-` quarter tones and signed cents.


gesellkammer (commented):
All messages indicating parsing problems are discarded

hlolli:
Errors are still reported through InitError and PerfError, which the tests explicitly check. However, you’re right that the more specific parsing messages were replaced by “invalid note name.” I’ll retain those details within the new error reporting so users can see why their input was rejected.

gesellkammer (approved):

PR #3143, opened 2026-09-20
Fix signed overflow in bexprnd sampling

hlolli:
`bexprnd` can receive `INT32_MIN` from its signed random draw. Negating that value as an integer overflows before conversion to floating point, causing undefined behavior and potentially a NaN sample. All opcode rates and GEN21's bilateral-exponential mode share this sampler.

Convert the draw to `MYFLT` before negating it. This makes the endpoint produce zero and preserves ordinary samples and the random sequence, with no added branches or function calls.

The manual describes `krange` as an output bound, but the existing code and example output use it as a scale factor. This fix keeps that behavior.


hlolli:
@psmaragdis, could you review this `bexprnd` fix? It handles the `INT32_MIN` random draw without changing ordinary samples or the random sequence.

vlazzarini (approved):

PR #3144, opened 2026-09-20
Fix unschedule string matching and storage

hlolli:
`unschedule` and `unscheduleall` can crash when a queued event has a string p-field but the cancellation uses a number in that position. The matcher decodes both values as strings without checking both types. Reject that mismatch before decoding.

Store cancellation strings with their NUL terminators and append after all existing strings. The old code omitted terminators and joined separate fields, causing failed matches and reads beyond the allocation. Also reject missing or excessive arguments before accessing the fixed argument storage.


vlazzarini (approved):

vlazzarini:
I am not sure if the spec written by John ffitch for string args is supposed to store terminators. This needs to be checked to be in line with the rest of the system. I am concerned about the changes in rdscor.c: this code may have many uses beyond this PR and we absolutely don't want to break the score string mechanism.

hlolli:
I checked the existing readers and writers, though I haven't located John's original spec. They store each string with a NUL terminator:

- The score parser in [`scanflt()`](https://github.com/csound/csound/blob/1105d6600b8fe53f77ef79eda61d1f295f9b572b/Engine/rdscor.c#L158-L188) skips existing strings with `strlen(sstrp) + 1` and writes `\0` after each new string.
- The [realtime score reader](https://github.com/csound/csound/blob/1105d6600b8fe53f77ef79eda61d1f295f9b572b/Engine/linevent.c#L362-L390) does the same. Both string lookup functions in [`rdscor.c`](https://github.com/csound/csound/blob/1105d6600b8fe53f77ef79eda61d1f295f9b572b/Engine/rdscor.c#L28-L98) also advance by `strlen(ss) + 1`.

The change in `rdscor.c` only touches `set_evt_strarg()`. Its only caller is `remove_event_op()`, shared by `unschedule` and `unscheduleall`; the score parser and lookup code are unchanged.

The old helper copied no terminator and appended at `strlen(e->strarg)`, overwriting the first string's terminator rather than adding a separate string. The fix copies the terminator and walks all `scnt` strings before appending. This makes the helper produce the same layout the existing score code already writes and expects, including empty strings. The p-field string-code encoding is unchanged.

PR #3149, opened 2026-09-20
Fix UDP text command parsing and orchestra bounds

hlolli:
The UDP text-command parser uses unbounded `%s` reads into 128-byte channel and address buffers. Long fields overwrite the stack, and missing fields leave uninitialized names in use. Bound each token, reject incomplete or overlong fields, and parse reply ports without overflowing `atoi`. Track the consumed text so extra whitespace does not shift the value or address.

Multi-packet orchestra input also keeps receiving full-size packets after advancing into its 1 MB buffer. Limit each receive to the remaining space and discard oversized input. Keep continuation packets in the orchestra parser and accept `{` and `}` in separate packets, as described in the [manual](https://csound.com/docs/manual/udpserver.html).

Reject invalid reply addresses and fix an unconditional return that prevents replies on Windows even when Winsock starts successfully.


Top/server.c
```diff
@@ -182,23 +193,48 @@ static uintptr_t udp_recv(void *pdata){
...
+      if(strlen(orchestra) < 2 && !cont && *orchestra != '{') continue;
       if (csound->oparms->echo)
         csound->Message(csound, "%s", orchestra);
       if (strncmp("!!close!!",orchestra,9)==0 ||
           strncmp("##close##",orchestra,9)==0) {
         csoundEventString(csound, "e 0 0", 1);
         break;
       }
-      if(*orchestra == '/') {
+      if(*orchestra == '{' || cont) {
```
vlazzarini, line 223: why is there a branch reordering? Does the fix require this?

hlolli: Yes, for orchestra messages split across packets. Once `cont` is set, the next packet belongs to the same code block. The old order could treat it as a new OSC or channel command before reaching the `cont` check. Moving this branch first keeps those packets together until the closing brace.

vlazzarini (approved):

PR #3150, opened 2026-09-20
Fix internal OSC packet and listener bounds

hlolli:
The internal UDP server queues OSC payloads without checking that they contain the data named by their type tags. `osclisten` then reads past short numeric payloads or scans beyond unterminated strings. Validate packet headers, string padding, and all supported arguments before queuing or dispatching a message, and discard malformed or unsupported packets.

Use aligned local copies for numeric decoding and read OSC characters as four bytes, as required by the [OSC specification](https://opensoundcontrol.stanford.edu/spec-1_0.html). Check scalar output counts and array sizes before decoding, truncate strings to the output buffer's existing capacity without allocating during playback, and reset the status output when no message arrives. Truncation still advances past the full input string and its padding so later arguments decode correctly.

Also use atomic reads of queue flags, retain the payload size, and stop multi-channel dispatch at the end of the channel names instead of repeatedly overwriting the last channel. Initialize the OSC queue lock when creating or resetting Csound so listeners can run before the UDP server starts. These changes affect the internal listener, not the separate liblo plugin.


OOps/compile_ops.c
```diff
@@ -157,15 +155,15 @@ void csoundClearOSCMessage(OSC_MESS *mess){
 */
 const char *csoundOSCMessageGetFloat(const char *buf, MYFLT *mf) {
   float f;
-  f = *((float *) buf);
+  memcpy(&f, buf, sizeof(f));
```
vlazzarini, line 159: this seems over convoluted when we can just use indirection

OOps/compile_ops.c
```diff
@@ -157,15 +155,15 @@ void csoundClearOSCMessage(OSC_MESS *mess){
...
   byteswap((char*)&f,4);
   *mf = (MYFLT) f;
   return buf + 4;
 }
 
 const char *csoundOSCMessageGetDouble(const char *buf, MYFLT *mf) {
   double f;
-  f = *((double *) buf);
+  memcpy(&f, buf, sizeof(f));
```
vlazzarini, line 167: same here.

OOps/compile_ops.c
```diff
@@ -177,7 +175,7 @@ const char *csoundOSCMessageGetDouble(const char *buf, MYFLT *mf) {
 */
 const char *csoundOSCMessageGetInt32(const char *buf, MYFLT *mf) {
   int32_t i;
-  i = *((int32_t *) buf);
+  memcpy(&i, buf, sizeof(i));
```
vlazzarini, line 179: and here.

OOps/compile_ops.c
```diff
@@ -188,7 +186,7 @@ const char *csoundOSCMessageGetInt32(const char *buf, MYFLT *mf) {
 */
 const char *csoundOSCMessageGetInt64(const char *buf, MYFLT *mf) {
   int64_t i;
-  i = *((int64_t *) buf);
+  memcpy(&i, buf, sizeof(i));
```
vlazzarini, line 190: and here

OOps/compile_ops.c
```diff
@@ -198,20 +196,21 @@ const char *csoundOSCMessageGetInt64(const char *buf, MYFLT *mf) {
...
 */
-const char *csoundOSCMessageGetString(const char *data, STRINGDAT *sdat) {
+const char *csoundOSCMessageGetString(CSOUND *csound, const char *data,
+                                     STRINGDAT *sdat) {
   size_t len = strlen(data)+1;
-  strncpy(sdat->data, data, sdat->size-1);
-  sdat->data[sdat->size-1] = '\0'; // safety
-  return data+((size_t) ceil(len/4.)*4);
+  if (len > (size_t)sdat->size) {
+    sdat->data = csound->ReAlloc(csound, sdat->data, len);
```
vlazzarini: realloc here is not good, this might be called at perf time. If the string does not have enough capacity, truncate.

vlazzarini (changes requested):

OOps/compile_ops.c
hlolli: I kept the fixed-size `memcpy` here so the reader can load from byte storage without relying on C type-aliasing assumptions. It does not allocate. I've added a shared comment explaining this and the alignment constraint on the 64-bit readers in c88fc2640.

OOps/compile_ops.c
hlolli: For `double`, direct indirection can also require stronger alignment than OSC guarantees. A double after a four-byte argument can start four bytes into the payload, so copying into a local double keeps the read safe on platforms that require eight-byte alignment. I've documented this and added a mixed-number decoding check in c88fc2640.

OOps/compile_ops.c
hlolli: I kept the fixed-size copy here for the same byte-storage and C aliasing reason as the float reader. It avoids a typed dereference of the packet buffer and keeps the four numeric readers consistent. The shared comment in c88fc2640 now explains the choice.

OOps/compile_ops.c
hlolli: The same alignment issue applies to `int64_t`: OSC arguments need only four-byte alignment, which may not satisfy an eight-byte typed read. I kept the local copy and documented why in c88fc2640.

OOps/compile_ops.c
hlolli: Done in c88fc2640. The reader now truncates to the existing capacity, leaves room for the null terminator, and does not allocate. It still advances past the full input string and its padding so the next argument decodes correctly. The added checks cover truncation, unchanged capacity, and reading a number after a truncated string.

OOps/compile_ops.c
vlazzarini: what you are saying is that the variable f in the stack can be 8-byte and that assigning it through indirection needs to be replaced by a byte copy using memcpy? Which means that the C indirection is not guaranteed to work?

vlazzarini (approved):

PR #3151, opened 2026-09-20
Fix atssinnoi partial and noise bounds

hlolli:
`atssinnoi` can read outside its buffers with valid ATS data: frequencies at or above 20 kHz leave its band lookup uninitialized, and selecting more than 25 partials reads past the noise tables.

- Remove the unused per-partial noise conversion and cache. Playback immediately overwrote those values with fixed-band energies.
- Limit noise reads and setup to the 25 stored bands, and skip noise reads for files without them. Decode the analysis window size for byte-swapped files.
- Validate partial selection before allocation, apply the increment once, and allow selections that end on the last partial. Reset oscillator phases when a note reuses its buffers.

The [manual](https://csound.com/docs/manual/ATSsinnoi.html) describes per-partial noise, while the current code uses fixed bands. This keeps that existing sound; changing the noise model would need a separate decision.


vlazzarini (approved):

vlazzarini (dismissed):

hlolli:
conflict, fixing... @vlazzarini

vlazzarini (approved):

PR #3153, opened 2026-09-20
Fix LPC frame bounds and analysis slot initialization

hlolli:
`lpread` can read before its frame data for single-frame files and long analyses: its signed 16.16 frame position becomes negative. Clamp a frame number and fraction before indexing, and hold the final frame without reading a nonexistent neighbour, matching the [documented playback behaviour](https://csound.com/docs/manual/lpread.html).

- Validate the header, frame layout and coefficient count before allocation or access. Keep headerless input and copy frame data only when legacy header padding leaves it unaligned.
- Keep the final text value without a trailing newline and handle text-reader failures without leaking the buffer.
- Initialize new LPC slots to empty and report missing analyses in dependent opcodes. Have `lpinterp` publish its coefficient fields directly instead of copying an unrelated opcode layout.

File and slot checks run at initialization; the audio loops stay unchanged.


vlazzarini (dismissed):

hlolli:
also conflict :( @vlazzarini

vlazzarini (dismissed):

vlazzarini (approved):

PR #3171, opened 2026-09-21
fix: correct LPC cepstrum recurrence

hlolli:
`csoundLPCeps` uses integer division in its recurrence, so the weighting terms become zero. It also discards every cepstral term beyond the filter order, although an all-pole filter generally has a longer cepstrum. A conversion round trip changes coefficients `[-1, 0.5]` into `[-1, 1]`.

Use floating-point weights and continue the recurrence for the full requested output length. Compute the final signs directly, removing the separate sign-flip pass. Also handle an order-zero filter in both conversions without accessing a second element, and put the two API descriptions above their matching declarations.


OOps/lpred.c
```diff
@@ -205,9 +201,7 @@ MYFLT *csoundCepsLP(CSOUND *csound, MYFLT *b, MYFLT *c,
                     int32_t M, int32_t N){
   int32_t n,m;
   MYFLT s;
-  b[0]  = 1;
```
vlazzarini, line 215: why this change? That's not in the original code. Why can't we start with the first two terms and iterate from the third?

vlazzarini (changes requested):

OOps/lpred.c
hlolli: @vlazzarini Yes, we can set the first terms explicitly and start the loop at `m = 2`. We just need to guard `b[1] = -c[1]` with `M > 0`, since an order-zero filter only has space for `b[0]`.

Starting at `m = 1` handles that case without a separate guard. On the first iteration the inner loop is empty, so `s = 0` and the result is exactly `b[1] = -c[1]`. For `M = 0`, the loop does nothing.

The removed `b[0] = 1` was unused: the inner loop starts at `n = 1`, and the function sets `b[0] = exp(c[0])` before returning. This is just a way to handle the boundary case; the recurrence does not require changing the original layout.

vlazzarini: zero-order filters are not practical in this context. Enforce minimum order 2.

hlolli: @vlazzarini Done in 7ce0c77b5. Both conversions now enforce `M >= 2` and initialize the first two terms explicitly before looping from index 2. Orders below two return `NULL` without writing to the output; I also added the check to `pvscfs` initialization so it reports an invalid order before calling the conversion.

I updated the API docs and replaced the one-pole example with a two-pole case. The focused conversion and opcode tests pass.

vlazzarini (approved):

PR #3186, opened 2026-09-21
build: select C11 per compiler and enable it on MSVC

gesellkammer:
The build only probed -std=gnu11/-std=gnu99, which are GCC/Clang options. MSVC sources therefore stayed in the legacy C89 mode and HAVE_C99 was never defined there.

Select C11 via CMAKE_C_STANDARD for the non-MSVC case, letting CMake emit the right switch for each toolchain and decay to gnu99 when C11 is unavailable. MSVC is handled separately by testing /std:c11 directly, because CMake only maps CMAKE_C_STANDARD to that switch from 3.18; probing the flag keeps the existing CMake 3.13.4 minimum.

With /std:c11 the MSVC compiler defines __STDC_VERSION__ as 201112L, so include/sysdep.h defines HAVE_C99 and the existing C99 paths compile.

No change for GCC/Clang, which still use -std=gnu11.


kunstmusik:
Did a review and there was a minor nit found:

* [P2] Use a fresh cache key for the MSVC probe" body="CMake caches `check_c_compiler_flag` results by variable name. An existing Windows build configured before this PR may already contain `HAS_C11=FALSE` from probing `-std=gnu11`; after upgrading, this check will not test `/std:c11`, so the flag remains disabled and the original problem persists. Use a new key such as `HAS_MSVC_C11` or explicitly invalidate the old cached result.

I think it's pretty minor, so I'm goign to go ahead and approve and merge. Thanks for finding this!

kunstmusik (approved):

PR #3214, opened 2026-09-21
fix: use condition variables for thread monitor notifications

hlolli:
The C11 thread monitor uses a mutex as a notification token. Normal cross-thread notification therefore unlocks a mutex owned by another thread, which is undefined behavior. This affects performance-thread pause/resume and message-queue flushing. Windows-pthread and Haiku use the same scheme.

Use a condition variable and stored signal state for C11, and the existing condition-variable implementation for all pthread platforms. This follows the API contract: monitors start notified, retain notifications until consumed, and allow another thread to notify them. A notification arriving at the timeout boundary now returns success instead of being consumed with a failure result.

Also translate C11 success codes to Csound’s zero-on-success convention for monitor waits and mutex operations. Rename the C11 barrier type to avoid its collision with the macOS pthread compatibility type.


vlazzarini (approved):
quite a big modification. Thread locks have always worked despite the UB, I don't think I ever saw a failure.

PR #3215, opened 2026-09-22
fix: release LPC workspaces and correct initial analysis

hlolli:
LPC opcodes allocate a new analysis workspace on every initialization but never release the old one, so repeated notes and reinitialization retain memory until engine reset. `csoundLPfree()` also leaves six buffers allocated.

Free every workspace buffer and add cleanup for `lpcfilter`, `lpcanal`, `pvslpc`, `pvscfs`, and `apoleparams`. Release old workspaces before reinitialization and release the i-rate analysis workspace as soon as its results have been copied.

Also correct i-rate `lpcanal` to accept the five inputs and optional window shown in the [manual](https://csound.com/docs/manual/lpcanal.html); its four-input registration leaves required pointers unset. Bound the shared pitch search so it cannot read past the autocorrelation array, return zero when no peak exists, and let `pvslpc` use bin-centre frequencies when the detected pitch is zero or below one bin, avoiding integer division by zero.


vlazzarini (changes requested):
Fix for lpc freeing function is ok. Leave the rest as is, to be deleted at reset time. This was the original design.

OOps/lpred.c
```diff
@@ -537,8 +544,43 @@ MYFLT *csoundStabiliseAllpole(CSOUND *csound, void *parm, MYFLT *c, int32_t mode
...
+  return lp_free_setup(csound, &p->setup);
+}
+
+int32_t coef2parm_deinit(CSOUND *csound, CF2P *p) {
+  return lp_free_setup(csound, &p->setup);
+}
+
 /* lpcfilter - take lpred input from table */
 int32_t lpfil_init(CSOUND *csound, LPCFIL *p) {
+  lpfil_deinit(csound, p);
```
vlazzarini, line 583: what happens if deinit is called on an instance, then we init the same instance again and deinit is called again to free the memory that has already been freed?

hlolli: @vlazzarini The first deinit frees the workspace and sets `p->setup = NULL` through `lp_free_setup()`. `csoundLPfree()` now returns immediately for `NULL`, so the cleanup call at the start of the next init does nothing.

That init then stores a new workspace from `LPsetup()` in `p->setup`. The next deinit frees that new allocation and clears the pointer again. Calling deinit twice without an intervening allocation is also safe: the second call receives `NULL`.

The added lifetime test covers reinitialization and reuse across successive notes.

vlazzarini (changes requested):
ok, just checking this again. It's fine to add the deinit functions. I don't think they should be called directly on every init.

vlazzarini (approved):

PR #3216, opened 2026-09-22
feat: add typed JSON conversion for UDTs

hlolli:
## Summary

Add initialization-time JSON conversion using declared UDT member names and types:

- `jsonunmarshal` reads JSON text into a UDT or typed array.
- `jsonunmarshalfile` reads through Csound's file handling and search paths.
- `jsonmarshal` returns compact or indented JSON.

Strict JSON is the default. Separate options enable comments and trailing commas. The converter handles nested UDTs, one-dimensional typed arrays, numeric and Boolean fields, and UTF-8 strings. It checks the full input's nesting depth and reports field paths for schema errors. Missing, unknown, duplicate and wrongly typed fields fail without partially updating the destination.

Use temporary values during decoding, then commit into existing member storage so compiled member references and shared array copies remain valid. Cache field lookup tables and reuse binding buffers across array entries. Keep parsing and file I/O outside real-time processing.

Vendor yyjson under its MIT license, keep its symbols private, and disable compiler fast-math for JSON conversion so DSP build flags do not change numeric round trips.

Add a dedicated `json_udt_tests` executable and `json-udt` CTest label, plus command-line cases used by the WASI test runner. Document the opcodes, options, supported types and limits in `docs/json-opcodes.md`.


vlazzarini:
just wondering about the JSON parsing code - are we OK to pull it into the code. Are the licenses compatible between each other. Does it need to keep up to date with upstream sources? Would a git submodule be preferred (and are we happy to introduce the dependency in that case)?

